PR #9, opened 2017-02-06
Add a stupid commit

qgp:
Another test commit


qgp:
Why am I asked to approve this? If I have the rights, this should just go through ...

qgp:
I think, this is a stupid commit that should be discarded (remember 1+1=2) ...

qgp:
This was mean, but things like this should not be taken as an approval ...

PR #26, opened 2017-02-23
This will break

ktf:


ktf:
+test

ktf:
@dberzano I think known people should have test run immediately, without waiting for an actual +test, regardless of wether they can commit or not to the given package.

PR #30, opened 2017-02-24
Update AliEveBeamsInfoEditor.cxx

ktf:


ktf:
+test

ktf:
Ok... I managed to pick something which is not actually compiled, but tests
are down to 8 minutes even on the initial build.

dberzano:
+1

PR #53, opened 2017-03-01
Commit 1

dberzano:


dberzano:
OK @arossi81, same story as #52: in case of an empty pull request (I have two commits, and the second reverts the first), which obviously makes no sense, the checker gets crazy. Fixing it!

PR #54, opened 2017-03-01
This is a hell of a large commit

dberzano:


dberzano:
This is to test what goes on with an extremely large pull request.

dberzano:
Very well. Now if someone accidentally/deliberately opens a PR with too many changed files, individual permissions are not checked (we don't want to exhaust the API calls, and it's likely to be a mistake!), and admins are contacted.

PR #58, opened 2017-03-02
Stress Test 2

dberzano:


dberzano:
(test in progress)

dberzano:
@rbertens you have commented on the _commit_, not on the _pull request_. Your commit is getting ignored. It happened the same to @arossi81 earlier this morning. @jgrosseo I think we should clarify it in the documentation.

jgrosseo:
@dberzano I fear the documentation is not enough - it will happen again and lead to frustration. Can we drop the link to the commit in the alibuild comment? This would avoid the mistake?

dberzano:
Hi @jgrosseo, I am solving the problem by adding a clear link to the pull request. We cannot drop the commit hash as it is used by the checker to determine what pull request iteration permissions are requested for (if we lose it, we lose the ability to react when pull requests are updated), and the link is added by GitHub automatically, we cannot prevent this from happening. As said, let me put a clear link there which tells you exactly what to do, it's even better than fixing the documentation.

PR #139, opened 2017-03-13
bug fix: use correct filename for OADB access

mazimm:


dberzano:
+1

Please also have a look at what I just wrote to [ALIROOT-7127](https://alice.its.cern.ch/jira/browse/ALIROOT-7127).

PR #173, opened 2017-03-17
Update of the AliAnalysisMuMu Framework

benjaminaudurier:
Implement event mixing to the mumu framework and other rather small updates


ktf:
+1

PR #184, opened 2017-03-23
fix cluster id assignment

jdmulligan:
A difference was identified between "tender" clusterizer and new correction framework clusterizer (based on ClusterizeFast): the centralized clusterizer seems to set cluster ID always to 0 -- this commit fixes it, setting the ID as done in the tender.


ktf:
+test

mfasDa:
+1

PR #189, opened 2017-03-23
Add .settings in .gitignore

aiola:
Same as done for AliRoot. Useful for Eclipse users.


dberzano:
+1, see also AliRoot#71.

PR #212, opened 2017-03-25
Adding a couple of TH3D and the option to perform Isolation exclusively with clusters

dlodato:
A couple of QA TH3D added plus implementation of the possibility to choose from Isolation/UE computation with Only Clusters or Clusters + Tracks (which is the default method)


gconesab:
+1

PR #219, opened 2017-03-27
New AddTask macro for single-track efficiency task in pp + updates in…

fprino:
This commit has:
- New AddTask macro for single-track efficiency task in pp
- corresponding updates the Pb-Pb macro


dberzano:
+1 (I think that @fprino should be added to `alice-svn-pwgpp` though.)

PR #224, opened 2017-03-27
TPC PID studies and Mult Update

hzanoli:
Hi,

I added the histograms to check the TPC Splines as function of momentum (p), the invariant mass plots as function of pT bins and changed the default multiplicity to use the ZNA. It is still possible to use V0A in case it is needed.


fprino:
+test

arossi81:
+1

PR #228, opened 2017-03-27
EMCal CorrFW: Compare configs, remove run period

raymondEhlers:
Implement suggestion from the EMCal Question and Answer meeting on 23
March 2017 to provide a comparison function for YAML configuration
files. Thanks Constantin for the suggestion!

Also removed the run period setter function announced two weeks ago.


mfasDa:
+1

PR #236, opened 2017-03-27
Added histos for jet tagging efficiency

haandrew:
Added histograms to compare the detector level jets that are matched and in total to measure jet matching efficiency for matching hybrid subtracted jets to detector level


adriansev:
+test

adriansev:
+1

PR #241, opened 2017-03-28
changed a cut

sashingo:
changed E/p cut (higher) for the systematic study at high pT


sashingo:
+test

ktf:
@sashingo no need to say +test if your PR is already approved. @deberzano
we should probably rephrase the message so that the testing is ongoing.

PR #244, opened 2017-03-28
add more histograms on col-row location of exotics and cluster cell e…

gconesab:
…nergy distribution; comment out other histograms not useful for the moment


gconesab:
+1

PR #240, opened 2017-03-28
Changes in ThreePart code

pbatzing:
Incremental changes, forces warnings as errors in PWGCF/Correlations/ThreePart

PR #248, opened 2017-03-28
Fix auto_ptr deprecation errors in PWGCF

dberzano:


dberzano:
+test

#240 broke compilation on some machines because of the deprecated `auto_ptr` (also notifying @pbatzing). Can some expert check and merge if it works?

pbatzing:
As I am the sole user and maintainer of libPWGCFThreePart at the moment, nobody except me will be able to confirm that this works and merge the code. At my end this fine and can be merged.

I am however puzzled that this issue first appears in #240 , as this particular lines of code have been present since I first introduced the class in AliRoot.

rbertens:
Hi Paul, 
Since you're the only user I've approved the commit - otherwise it's stuck forever.

dberzano:
Hi @rbertens, you need to comment on the PR, not on the commit :-)

dberzano:
Hi @pbatzing, in #240 you turn all warnings into errors in the `CMakeLists.txt`, see [here](https://github.com/alisw/AliPhysics/pull/240/files#diff-69500f5234320e6f5ea8c74ade0d8309). This is good (very good actually), but had the side effect that the compiler on my machine (GCC v5.4.0) as _more_ warnings than the production one used for tests (GCC v4.9.3). One of these additional warnings (turned into errors) was the deprecation of `auto_ptr`.

rbertens:
+1

pbatzing:
Hi @dberzano I see. Then at least it should not happen more in the future. I think also I will update my local compiler, so that I catch such warnings (errors) on my end. Thanks for the fix.

dberzano:
No worries, it's not your fault at all!

PR #249, opened 2017-03-28
fixing filling of a TH2D and removing Printf

dlodato:
I fixed a bug about the filling of a TH2D which was initialised within a flag condition but not filled within the same flag. 
CTest didn't spot it...
Also i took out lots of print out which were slowing down the running time and making the code ugly to read


gconesab:
+1

PR #253, opened 2017-03-29
Revert errors as warnings (for now)

pbatzing:
Reverts errors as warnings, as it seems to be unstable wrt compiler versions atm. Will investigate locally.

### DIFF
--- a/PWGCF/Correlations/ThreePart/CMakeLists.txt
+++ b/PWGCF/Correlations/ThreePart/CMakeLists.txt
@@ -18,12 +18,6 @@ set(MODULE PWGCFCorrelationsThreePart)
 add_definitions(-D_MODULE_="${MODULE}")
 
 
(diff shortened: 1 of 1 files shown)


pbatzing:
As I don't want to break anyones compile, I will quick fix this for now. Could @rbertens or @jgrosseo please +1 this? It doesn't change any actual code, only the makefile.

dberzano:
Can you _delete_ the part as opposed to commenting it?

pbatzing:
I deleted it now, and ammended the commit. Can you confirm that this looks ok?

dberzano:
+1 (thanks!)

In general the message is that we have the Git history for recovering old code, so at least for important stuff like `CMakeLists.txt` it's better to keep it clean and not using too many comments for old code.

pbatzing:
Sure, I will keep that in mind :)

PR #258, opened 2017-03-29
Updating task: RequestITScls, nsigmaTPC for species

strogolo:
Adding:
- possibility to require and cut on a minimum number of ITS clusters
- selection of the pT range for proton and deuteron
- separating the nsigma_TPC cut for the different species (pion, proton, deuteron)
- new histograms to check the sigma on the DCA between tracks and secondary vertex


mpuccio:
+1

PR #259, opened 2017-03-29
Implement event counter in AliBasicParticle to allow using of all-event pool for mixed event

rhaake:
This commit only (directly) affects the usage of custom particles in AliAnalysisTaskPhiCorrelations.
It is to allow the task to use an imported event pool correctly. Without, the tracks from an imported pool are mixed with tracks from the same events.

### DIFF
--- a/PWGCF/Correlations/Base/AliUEHistograms.cxx
+++ b/PWGCF/Correlations/Base/AliUEHistograms.cxx
@@ -657,7 +657,7 @@ void AliUEHistograms::FillCorrelations(Double_t centrality, Float_t zVtx, AliUEH
 	    AliBasicParticle* particleBasic        = dynamic_cast<AliBasicParticle*>(particle);
 	    if(!triggerParticleBasic || !particleBasic)
 	    {
-	      AliError("If fCheckEventNumberInCorrelation is set, particle must be derived from AliBasicParticle");
+	      AliFatal("If fCheckEventNumberInCorrelation is set, particle must be derived from AliBasicParticle");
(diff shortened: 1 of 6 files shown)


PWGCF/Correlations/Base/AliUEHistograms.cxx
```diff
@@ -711,6 +727,16 @@ void AliUEHistograms::FillCorrelations(Double_t centrality, Float_t zVtx, AliUEH
         // check if both particles point to the same element (does not occur for mixed events, but if subsets are mixed within the same event)
         if (mixed && triggerParticle->IsEqual(particle))
           continue;
+        if (fCheckEventNumberInCorrelation)
+        {
+          AliBasicParticle* triggerParticleBasic = dynamic_cast<AliBasicParticle*>(triggerParticle);
+          AliBasicParticle* particleBasic        = dynamic_cast<AliBasicParticle*>(particle);
+          if(!triggerParticleBasic || !particleBasic)
+            AliError("If fCheckEventNumberInCorrelation is set, particle must be derived from AliBasicParticle");
```
jgrosseo: This will crash two lines below. Better put AliFatal then.

jgrosseo:
Hi Ruediger,
Please fix the one thing which I commented. Then it is fine with me.
https://github.com/alisw/AliPhysics/pull/259/files
Cheers,
Jan

rhaake:
Thanks, I submitted a patch for the two occurrences.

jgrosseo:
+1

PR #265, opened 2017-03-29
Fix pointer deletion in class destructor

jhamoncern:
Fix crash of class deletion when using >3rd order polynomials ('fBkgFuncSb' has to be deleted before 'fHistoInvMass')


ebruna:
+test

ebruna:
+1

PR #270, opened 2017-03-29
Fixed jet matching bug in user task

haandrew:
Fixed jet matching bug for matching subtracted jet container to detector level jet container


lcunquei:
+1

PR #274, opened 2017-03-29
PWGGA/GammaConv: excluded if for gamma from elec. for TrueSec, check TrueSec is primary for AOD loop

lleardin:
The if in the last else of the TrueSecGamma histo filling loop was excluding the cases in which the condition was not filled and the was no appropriate else for these circumstance -> removed for both ESD and AOD loops.
In the same AOD loop as above, the condition that the photon grandmother is a primary has been added, to be consistent with the ESD loop.


FriederikeBock:
+1

PR #282, opened 2017-03-30
EMCal starts at 80 degrees not 100 deg; start isolation analysis at 5…

gconesab:
… GeV like correlation; do correlation with any selected cluster, not event leading


gconesab:
+1

PR #288, opened 2017-03-30
Code grouping into a new Method to avoid repetition (both tasks) and option for Container 

dlodato:
Code for LeadingParticle analysis updated/included; 
Code to select the candidate particle grouped into *new* SelectCandidate method ;
Added Option for Output Container to specify whether an Analysis is done with Leading Particle option ON or OFF


gconesab:
+test

gconesab:
+1

gconesab:
+1

PR #289, opened 2017-03-30
New direct photon analysis

csergilo:
New direct photon analysis, work in progress. One of the aims is to learn the analysis of photon conversion processes and for this, developing a new task "from scratch". Has also a pedagogical purpose (I am an undergrad student). When the understanding of all the processes is mature and the analysis more meaningful, I will revert to using existing tools.

### DIFF
--- a/PWGGA/GammaConv/AliAnalysisTaskDirectPhotons.cxx
+++ b/PWGGA/GammaConv/AliAnalysisTaskDirectPhotons.cxx
@@ -61,22 +61,13 @@ ClassImp( AliAnalysisTaskDirectPhotons );
 AliAnalysisTaskDirectPhotons::AliAnalysisTaskDirectPhotons():
     AliAnalysisTaskSE(),
     fOutputList(NULL),
-    /*fNegativeParticles(NULL),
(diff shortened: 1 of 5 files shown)


FriederikeBock:
Hi Carolina, 
if you must you can develop a class completely from scratch, however the conversion group strongly discourages it in particular aiming at direct photons. 
The direct photon analysis has a complexity which requires not only a simple analysis task but also an extensive "after-burner" code first to extract and correct the mesons, parametrize them, create the photon cocktail and only then can one even think about aiming at the direct photons. The existing framework has grown over the last 5-8 years and I am sure you don't wanna spend this much time in writing approximately 1Mio lines of code. We are happy to teach you how the analysis works and use our framework and certainly it is a good exercise to write a task from scratch, but I am not sure it is wise to include it in the central packages.
All that said I will approve the merge, however not with the current class name. Thus please rename the class as it suggest a maturity, which it will simply not reach and might be accidentally executed by others.  
Regards,
Friederike

dberzano:
I have closed the PR upon request.

PR #291, opened 2017-03-30
Fixing a double colon which should have been a single colon 

dlodato:
With the previous commit i introduced a small bug,I added two colons instead of one as a binary operand in the composition of the output container for the AliAnalysisTaskEMCALPhotonIsolation.. 
I am really sorry but not even the CTest spotted it.


gconesab:
+1

PR #301, opened 2017-03-31
AliAnalysisTaskHypertriton3: fixing issue for invariant mass

strogolo:
Use tracks propagate for invariant mass (trkD, trkP, trkPi)


mpuccio:
+1

PR #309, opened 2017-03-31
JCORRAN: Change MC truth particle selection

jusaviin:
To study the difference seen in the first try to compare MC truth to MC reconstruction, I have made a small change to my selection of MC truth particles.


aohlson:
+1

PR #311, opened 2017-03-31
Improvements for Leading particle Analysis on MC truth

dlodato:
I have perfected the Leading Cluster Analysis on MCtruth and implemented the Leading Track analysis plus added a couple of feature that were missing..
I also rearranged the definition of the histograms in output, according to the flow of the AliAnalysisTaskEMCALTrackIsolation


gconesab:
+1)

PR #319, opened 2017-04-02
Fix division by zero crash in AliDielectronVarManager::FillAODTrack()

iarsene:


iarsene:
+test

iarsene:
Hi, i see there was a compiling error for the test of my commit but this seems to be completely unrelated to the small change i made.

ktf:
Indeed the errors seems unrelated and it's probably coming from some bad interaction with alisw/AliRoot#94. I am checking what is going on.

PR #325, opened 2017-04-03
Fix ROOT6 build

dberzano:


dberzano:
+1

dberzano:
For the record this should fix our ROOT 6 test builds, no functional change is done in those files.

PR #330, opened 2017-04-03
added different algorithm reclustering softdrop

haandrew:
Modified user task to be able to recluster jets using different algorithms for softdrop


mfasDa:
+1

PR #343, opened 2017-04-04
Unique event ID correctly evaluated and propagated to tracks

rhaake:
This is a bug fix. Before, the unique event ID was only unique on a per-task level.
With this commit, the event ID is taken from the header. In addition, the ID is now correctly set for cloned tracks.

### DIFF
--- a/PWGCF/Correlations/DPhi/AliAnalysisTaskPhiCorrelations.cxx
+++ b/PWGCF/Correlations/DPhi/AliAnalysisTaskPhiCorrelations.cxx
@@ -1610,6 +1610,12 @@ TObjArray* AliAnalysisTaskPhiCorrelations::CloneAndReduceTrackList(TObjArray* tr
     else
       copy = new AliBasicParticle(particle->Eta(), particle->Phi(), particle->Pt(), particle->Charge());
     copy->SetUniqueID(particle->GetUniqueID());
+
+    // Set unique event index if input tracks are AliBasicParticles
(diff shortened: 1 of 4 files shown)


PWGCF/Correlations/DPhi/AliAnalysisTaskPhiCorrelations.cxx
```diff
@@ -2051,6 +2058,17 @@ Bool_t AliAnalysisTaskPhiCorrelations::InitiateEventPlane(Double_t& evtPlanePhi,
 }
 
 //____________________________________________________________________
+Long64_t AliAnalysisTaskPhiCorrelations::GetUniqueEventID(AliVEvent* inputEvent)
+{
+  // Get event ID from header
+  AliVHeader* eventIDHeader = inputEvent->GetHeader();
+  if(eventIDHeader)
+    return eventIDHeader->GetEventIdAsLong();
+  else
```
jgrosseo: Isn't this restarting from 0 for each new raw data file?

rhaake: I thought since the ID is calculated using BC, period, and orbit, it is kind of global.
http://aliroot-docs.web.cern.ch/aliroot-docs/src/AliVHeader.cxx.html#49

Maybe I am wrong and you know the correct code to get a global event ID?

jgrosseo:
+1

PR #352, opened 2017-04-05
New trigger argument + code aesthetics

e-masson:
The possibility to pass a **L1 trigger name as argument** is added (only "EG1" and "EG2" for now).
**Spaces and general indentation** are corrected, some **obsolete comments** are deleted.


gconesab:
+1

PR #368, opened 2017-04-05
updated pPb BDT pKpi systematics

jaimenorman:
Lc pPb BDT pKpi systematic uncertainty function updated with final values

### DIFF
--- a/PWGHF/vertexingHF/AliHFSystErr.cxx
+++ b/PWGHF/vertexingHF/AliHFSystErr.cxx
@@ -5120,14 +5120,13 @@ void AliHFSystErr::InitLctopKpi2013pPb() {
 	for(Int_t i=1; i<=12; i++) fBR->SetBinContent(i,0.05);
 }
 
-
 //--------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


arossi81:
+test

arossi81:
+1
p.s. is it correct that the MC pt shape uncertainty "oscillates" with pt (0,3,1,3)?

jaimenorman:
Hi Andrea,

Thanks for the +1.
This is how the systematic uncertainty has been evaluated, reflecting in principle the extent to which the different pt shape changes the efficiency in each pt bin, which has been accepted. Does this sound ok?

Cheers,
Jaime
________________________________

arossi81:
Hi Jaime,
if those are the right numbers it is ok, at least for the commit. Then, whether or not it is expectable to have this trend has to be discussed in D2H (you did it in the past, right?).
Cheers,
Andrea

jaimenorman:
Hi Andrea,

Yes, it had been discussed in the D2H. The tighter BDT cut in 4-6 with respect to 6-8 GeV/c, and the wider pt bit 8-12 I think are the two reasons why this trend would be expected.

Cheers,
Jaime
________________________________

PR #374, opened 2017-04-05
Updates for the HF multi-trial class and the D-tagged jet raw yield uncertainty

aiola:
These PR contains a number of commits:

- Fix an issue with invalid memory access when the results of the mass fit are drawn in the canvas and the mass fitter is deleted
- Fix various potential (and actual) memory leaks
- Fix indentation

### DIFF
--- a/PWGHF/jetsHF/AliDJetRawYieldUncertainty.cxx
+++ b/PWGHF/jetsHF/AliDJetRawYieldUncertainty.cxx
@@ -406,42 +406,48 @@ Bool_t AliDJetRawYieldUncertainty::RunMultiTrial()
   cname = TString::Format("MassFit_%s_%s_%1.1fto%1.1f", fDmesonLabel.Data(), fMethodLabel.Data(), fpTmin, fpTmax);
   TCanvas* c0 = new TCanvas(cname,cname);
 
-  if(fFitRefl) { //reflection treatment
+  if (fFitRefl) { //reflection treatment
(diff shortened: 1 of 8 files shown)


arossi81:
+test

arossi81:
Hi @aiola,
you should increase the ClassDef for AliDJetRawYieldUncertainty.
Concerning the modifications to the MultiTrial and MassFitter classes, your proposals looks reasonable to me. However, before committing the changes, could you ask someone from D2H to make an explicit test to cross check that he gets what expected? 
Ciao,
Andrea

aiola:
Ciao @arossi81,
thanks for starting the test. Regarding AliDJetRawYieldUncertainty I think that it is not needed to increase the ClassDef version because the two fields that I added are transient (i.e. defined with //!<!).

For D2H: @fprino? Someone else?

Best,
Salvatore

fprino:
I looked in the code modifications in the MassFitter and MultiTrial classes and in principle the functionalities should be the same.
Andrea, do you think one needs a test running it for some cases?

arossi81:
Ciao Salvatore,
yes, I think you're right, it should not be needed to increase the ClassDef version for the addition of transient data members.
Francesco: well, for me also the changes are ok, I was proposing a test just to be sure that I did not overlook something because the diff was large, but it's not strictly needed. We can commit it.
Cheers,
Andrea

arossi81:
+1

PR #390, opened 2017-04-06
Config. files for CF_PbPb_MC_AOD Lego Train

KonstantinMikhaylov:
+1 Configuration files for Lego Train do not need to be compiled


KonstantinMikhaylov:
+1 just in case

PR #396, opened 2017-04-06
EMCal Embed: Bug fixes, docs, and debug info

raymondEhlers:
Includes:
- Small bug fixes related to the interface for selecting files to embed.
- Improve documentation and debugging information to simplify comparison between the new and old frameworks, including an option to disable some randomness in the old embedding framework.
- Small changes to the Sample Task also serve to simplify comparison between the two frameworks


mfasDa:
+1

PR #401, opened 2017-04-07
Code aesthetics

e-masson:
Comments and general indentation.


dlodato:
+1

e-masson:
Hi Davide, thanks! I am using macOS as main interface but my AliPhysics installation is on the same SL server at Subatech. Anyway I did not modify anything concerning md5, do you think that the code cleaning I made could cause an issue?

dlodato:
@e-masson I remember there have been some problems of this sort, but not specifically which.
It could also have been due to something completely unrelated.
Maybe (hopefully) a bit of rumour for nothing, but I thought it was worth mentioning the possibility for this to happen..

e-masson:
Of course, you did well!

PR #405, opened 2017-04-07
Monitor SPD primary vertex before and after event selection

jhamoncern:
- Fix missing initialisation of 2 data members
- Add an additional histogram monitoring SPD primary vertex after event selection


arossi81:
+test

ebruna:
+1

PR #407, opened 2017-04-07
class name added for newly committed HM task

nirbhaygit:
The class name of the newly added Net-particle higher moment task is added

### DIFF
--- a/PWGCF/EBYE/PWGCFebyeLinkDef.h
+++ b/PWGCF/EBYE/PWGCFebyeLinkDef.h
@@ -40,6 +40,8 @@
 #pragma link C++ class AliEbyEMultFluctuationTask+;
 #pragma link C++ class AliEbyEHigherMomentsTaskPID+;
 
+#pragma link C++ class AliEbyEPidEfficiencyContamination+;
+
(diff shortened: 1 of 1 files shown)


aohlson:
+test

nirbhaygit:
Hi @aohlson
 Is the error related to my task? I saw the log with some "warning" related to my task. Does it need to be fixed before the final merge or it is OK for today tag ?

aohlson:
Hi @nirbhaygit 
I am not sure about the origin of the error, since I am still learning the workflow of github myself (maybe @jgrosseo can offer some advice...).  It may be because the updates to CMakeLists and the linkdef file were done separately?  Perhaps you could add the change to CMakeLists to this commit (and close the other commit) so that they are done together.  If that does not work, then it indicates there is a compilation error in the class, and we can look in the log files for any clues where it could come from.

rbertens:
You have to make the changes to the CmakeList and pwglinkdef file in one pull request. Now the test falils, because you want to link a class that is not actually compiled (no Cmake instruction)

nirbhaygit:
@rbertens 
OK, I am closing both the pull requests and will make both the changes on one pull request.

PR #408, opened 2017-04-07
Class name added for the newly commited HM task

nirbhaygit:
The class name of the newly committed net-particle higher moments analysis task is added


aohlson:
+test

PR #421, opened 2017-04-07
Added normalized Impact Parameter to AliDielectronVarManager

hscheid:
Added Impact Parameter normalized to resolution to the AliDilectronVarManager for ESD and AOD tracks


miweberSMI:
+1

PR #430, opened 2017-04-08
PWG/Cocktail parametrization file updates

nschmidtALICE:
PWG/Cocktail replaced individual pp params with combined file and updated 0.9, 7 and 8 TeV params


FriederikeBock:
+1

PR #432, opened 2017-04-09
fixed issue with streamer

kreisl:
Fix for issue with streamer causing a memory leak.


zwound40:
+1

kreisl:
The Issue seems to be within the `AliDielectronEventCuts.h`
```
TH1* fCorrCutMin[5];       //parametrization of lower limit correlation cut
TH1* fCorrCutMax[5];       //parametrization of upper limit correlation cut
```
are non-transient members of the class and are initiated as null pointers. This seems to cause the problems when running the train.

PR #440, opened 2017-04-10
Fixed issue of DCA in AOD analysis.

kreisl:
fgEvent in `AliDielectronVarManager` was not set.
Therefore DCA wasn't calculated for all AOD filter bits in the ReducedTreemaker.
Fixed by adding `AliDielectronVarManager::SetEvent(event)`.


zwound40:
+1

PR #451, opened 2017-04-10
Fixed AliReducedEventPlaneInfo not being written to tree.

kreisl:
Fixed eventplaneinfo not being correctly written to file after previous commit #440.
`AliDielectronVarManager::SetEvent(event);`fills values into `fgData`
`AliDielectronVarManager::GetValue(...)` gets the values from `fgData`


zwound40:
+1

PR #462, opened 2017-04-11
Fix for fastjet dependence introduced in 06cd0021f1e6c4eeaa7c0d648472704246b465d9

dstocco:
Compile classes depending on fastjet only if fastjet is found


adriansev:
+test

adriansev:
+1

PR #469, opened 2017-04-11
Files get a name, that depend on the used trigger

jluehder:
The TRDPIDTree Output gets a name which depends one the chosen Trigger. So multiple Triggers can be used in one train.


ktf:
+test

tdietel:
+1

PR #483, opened 2017-04-12
PWG/Cocktail: updated params and option for dyn. adaption of pt rang

laltenka:
update of pp 7TeV parametrizations and implementation of option for dynamical adaption of pt range for generation depending on particle mass, also the number of generated particles is adapted to the increased range in pt if the option is used


FriederikeBock:
+1

PR #493, opened 2017-04-12
Added centrality cuts

aaroncapon:
-Minor updates to histograms
-Definition of TPC variables update in TreeMaker class


miweberSMI:
+1

PR #497, opened 2017-04-12
New class for EMCal alignment

hzanoli:
Hi,

This new class redo the propagation for electrons. The residual is recalculated. The default method propagates only until the EMCal Surface and it is also saved for comparison. It is necessary to check in details the alignment of EMCal and DCal.


gconesab:
+test

gconesab:
+1

PR #501, opened 2017-04-13
Fix in log message (does not affect processing)

shahor02:
@chiarazampolli, @fprino, could you please transfer updated PWGPP/TPC/CalibMacros/ProcessOutputCheb.C to ALIDPG, the processing does not change but the log message was wrong (if the distortion object was missing, it was complaining about missing correction object)


shahor02:
+1

PR #502, opened 2017-04-13
Fixed AddTaskZDCGainEq for multiple wagon

rihanphys:
Removed  'mgr->InitAnalysis()' from AddTask macro which was prohibiting from running multiple wagons in same train. 
Also changed return type from 'void' to class pointer so that "__R_ADDTASK__" option is available for sub-wagon configuration.


rbertens:
+1

PR #511, opened 2017-04-13
Bug Fix in AliAnalysisTaskZDCGainEq.cxx and Few updates

rihanphys:
A bug fixed related to PileUp rejection. Also added histograms to Store average ZDC Energy RunbyRun.


rbertens:
+1

PR #530, opened 2017-04-17
tot momentum cut applied to compare results with Identity method

nirbhaygit:
additional total momentum cut applied for comparison of C1 and C2 with identity method.


aohlson:
+1

PR #547, opened 2017-04-18
Added fraction shared pt function to user task

haandrew:
Moved function for shared pt fraction from jet container class to my class for matching to subtracted jets. Need to have a less stringent check for same particle due to modified particle pt.


aiola:
+1

PR #550, opened 2017-04-18
EMCal Embed: Fix jet constituent index

raymondEhlers:
AliEmcalJet::Track(i) calls AliEmcalJet::TrackAt(i), so calling
AliEmcalJet::Track(AliEmcalJet::TrackAt(i)) is redundant and
causes an index out of bounds. This occurred here a few times
since the calls to the two functions were separated.


aiola:
+1

PR #552, opened 2017-04-19
Adaptation to 'usedefault' container naming convention

e-masson:
Modification of hard-coded track container name + adaptation to the "usedefault" naming convention already used in other tasks. Minor corrections/comments.


dlodato:
+1

e-masson:
Thanks @dlodato!
I indeed tried running the modified AddTask on a data sample but not on MC. Everything is fine with data using (for instance) either ntracks="usedefault" of ntracks=<another name> as parameter of the AddTask.
Best, Erwann

mfasDa:
Hi @e-masson,

The functionality to handle "usedefault" for default containers was added to the EMCAL containers themselves. You maybe want to consider this!

Cheers

Markus

e-masson:
Hi @mfasDa,
Indeed I saw that a few weeks ago. Maybe we can see my today implementation as a first step to a bigger modification taking the EMCal default containers into account, thanks for your advice!
Best, Erwann

PR #558, opened 2017-04-19
Task: standardisation of acceptance values + fix of the UE area calculation. AddTask: removal of two comments previously written.

e-masson:
Phi acceptance values are still to be fixed: sometimes the EMCal phi range (minPhi, maxPhi) is (1.398, pi) or (1.4, 3.15), whereas in a more accurate way it is set as (5/9)*pi. Those three cases give slightly different maxPhi-minPhi values.
Some minor code aesthetics is also done in this commit.

### DIFF
--- a/PWGGA/EMCALTasks/AliAnalysisTaskEMCALPhotonIsolation.cxx
+++ b/PWGGA/EMCALTasks/AliAnalysisTaskEMCALPhotonIsolation.cxx
@@ -1115,7 +1115,8 @@ Bool_t AliAnalysisTaskEMCALPhotonIsolation::Run()
   
   fVevent = dynamic_cast<AliVEvent*>(InputEvent());
   
-  if(fVertex[2]>10. || fVertex[2]<-10.) return kFALSE;
+  if(fVertex[2]>10. || fVertex[2]<-10.)
+    return kFALSE;
   
   AliClusterContainer* clusters = GetClusterContainer(0);
   Int_t nbTracksEvent;
@@ -1138,7 +1139,8 @@ Bool_t AliAnalysisTaskEMCALPhotonIsolation::Run()
       }
     }
     // AliError(Form("EMCAL L1 trigger for MC simulation anchored to LHC13 data"));
-    if(!MCSimTrigger(fVevent,fTriggerLevel1) && fAnalysispPb) return kFALSE;
+    if(!MCSimTrigger(fVevent,fTriggerLevel1) && fAnalysispPb)
+      return kFALSE;
   }
   
   fVz->Fill(fVertex[2]); // Fill Vertex Z histogram
@@ -1988,7 +1990,7 @@ void AliAnalysisTaskEMCALPhotonIsolation::EtIsoClusEtaBand(TLorentzVector c, Dou
   
   Float_t sumEnergyEtaBandClus =0., sumEnergyConeClus=0., sumpTConeCharged=0, sumpTEtaBandTracks=0.;
   Double_t clustTOF=0;
-  Double_t minEta = -0.67, maxEta = 0.67, minPhi=1.398,maxPhi=TMath::Pi();
+  Double_t minEta = -0.67, maxEta = 0.67, minPhi=1.398,maxPhi=TMath::Pi(); // maxPhi-minPhi != (5/9)*pi !! Anyway, minPhi = 1.4 elsewhere in the task
   
   // AliParticleContainer *clusters = static_cast<AliParticleContainer*>(fParticleCollArray.At(1));
   AliClusterContainer *clusters = GetClusterContainer(0);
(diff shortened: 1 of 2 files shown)


dlodato:
+1

e-masson:
Hi @dlodato,

Following my commit of the last week, I would like to standardise the phi acceptance values set in the code. Have you motivations in favour of particular phi bounds, i.e. for you which one of the values already set is the most accurate/true regarding the EMCal geometry?

Thanks in advance, best,
Erwann

dlodato:
Hi @e-masson , 
If you want to generalise the EMCAL Acceptance, I think you should also consider the different years.
You should be able to make use of the EMCALGeometry class to retrieve the lower and upper bound of EMCAL in phi, and from there start computing the boundaries in phi for the Analysis (depending on one wants the UE/Eiso with charged +neutral or if you want to use only the charged component (in the latter case one would have a bigger acceptance for the candidates).
As long as the implementation is correct I don't have any preference, the difference should be (at most) on the 3rd digit in phi.

Best 
Davide

e-masson:
Hi @dlodato,

I am setting up the implementation of EMCal acceptance using the AliEMCALEMCGeometry class, and for instance GetArm1EtaMax() gives me fArm1EtaMax = 0.7 (whatever the EMCal geometry). This makes me doubt (again) the accuracy of the eta bounds set as 0.67 in the task (and maybe also for the TPC max eta as 0.87). Maybe @gconesab could enlighten us on this point?
On the other hand the AliEMCALEMCGeometry GetArm1PhiMin() and GetArm1PhiMax() methods give as expected 80° and 180° (i.e. ~1.396 rad and pi rad) for the EMCal geometry "EMCAL_COMPLETEv1" with 10 SMs, but if I get an instance of the geometry from e.g. a 2013 run number, it will be "EMCAL_COMPLETE12SMV1" with phi from 80° to 120°, even if the 2 last added SMs were not activated at the time...
So for the current implementation I can go around and set "EMCAL_COMPLETEv1" for 2013 runs, or I also can set the real geometry and removing 20° from the max phi value (i.e. 120-20 = 100 as desired). Anyway this does not concern pp data that you Davide use, but in the future we might want to consider all the geometries throughout the years as you suggested (especially with DCal).

Please feel free to give your preferences concerning this implementation, the result will be better :).

Best,
Erwann

gconesab:
Ciao, the purely geometrical limit is 0,7 of the last/first cell border, but the cluster will not have never a center at 0.7 or at 80 deg in phi, etc, take also into account the cell border cut, etc. Anyway, it does not harm to have those as limits of the histograms

dlodato:
Hi @e-masson,
the boundary at 0.67 was put to take out from the Analysis Clusters whose eta/phi are closer than 0.03 from the border of EMCAL (which, for example, covers eta [-0.7;0.7]). For these clusters the recorded reconstructed energy might be smaller than that of the particle that produced it (because of border leakage), so this cut is a QA cut. 
About 2013, whatever the geometry is, if the two 2/3 SMs were not active, you won't have any entries in the Eta-Phi regions covered by them, so you will have (in those eta/phi ranges) blank histograms. 
Maybe leaving the associated geometry (even if the specific SM were not switched on) is best for coding purposes.
Best
Davide

e-masson:
Hi, thanks for your answers,

In that case the phi limits should be reduced by 0.03 as are the eta limits. Taking the exact geometry boundaries, the EMCal phi acceptance would be (minPhi = (4./9.)*pi+0.03, maxPhi = pi-0.03) instead of (minPhi = 1.4, maxPhi = 3.15), leading to a phi range from 1.75 to 1.685 (–3.7 %).
Is it ok for you @dlodato?

Best,
Erwann

dlodato:
Hi @e-masson, 
I get slightly different values... From the TDR I remember EMCal covering from 80 to 180 degrees in phi.. bringing it to radians it yields 4/9*pi (+0.03 to exclude the borders)... This leads to 1.39626 (+0.03)= 1.39926 (hence the 1.4 for the minPhi, it is strange that maxPhi is bigger than Pi, though). 
If you got those values from the AliEMCALGeometry class, then please change them using that definition (namely invoking AliEMCALGeometry::GetPhiMin() or something like it), this way we would be sure that our task definitions change accordingly with the definitions agreed on by the whole collaboration.
Best, 
Davide

e-masson:
Hi @dlodato,
Ok, thanks. Just one more comment: in the CheckBoundaries() method, the phi bounds depend on the fAnalysispPb flag. I guess the !fAnalysispPb case is dedicated to Run1 pp analysis and then should be equivalent to set the geometry accordingly. In any case, maxPhiBound = 2.740 (!fAnalysispPb) is almost equal to maxPhiBound = 3.15-fIsoConeRadius (fAnalysispPb).
So do you think I can replace the if(!fAnalysispPb){...} by geometry-based conditions, or were there motivations for the !fAnalysispPb values?
Best,
Erwann

dlodato:
Hi @e-masson, 
the only reason I remember was to account for the 2 1/3 SMs (implemented very early in the code)
Yes you can change it accordingly and as you see fit. This is mainly a condition exclusive to Run1.
Cheers,
Davide

PR #560, opened 2017-04-19
Use EMCAL standard names 

hzanoli:
Hi,

I renamed the tasks to match the default EMCAL (instead of Emcal). I also added a few comments on the data members.


gconesab:
+test

gconesab:
+1

PR #569, opened 2017-04-20
Add: macro to produce performance report

miranov25:
*  PWGPP-272 - Use AliESDs.root in Pass2 filtering
*  PWGPP-272, Performance maps support


mkrzewic:
+1, as commented above

PR #603, opened 2017-04-21
Fix to Sub->Det matching for embedding

haandrew:
1. Cleaned up tree by removing second call of SoftDrop function in user task
2. Added AcceptJet condition for matched detector level jets
3. Modified AddTask to add correct particle container types


aiola:
+1

aiola:
Something went wrong in the build check... no log seems to be available.
@dberzano do you know what happened?

PR #618, opened 2017-04-24
increasing the number of histograms available to work with more pt bins

cbedda:
increasing the number of histograms available to work with more pt bins in the D* mass spectra analysis


ebruna:
+test

ebruna:
+1

PR #643, opened 2017-04-25
Tagging efficiency for HFe-h correlations

hzanoli:
The changes split Run 1 and Run 2 tagging efficiency calculations. Run 1 is kept exactly how it was (if was just moved to a function) in case any future check is needed. The Run 2 one uses only the electrons from the enhanced pi0 and eta. 

The function to compare the spectra with and without enhanced signal is implemented too. The weights are stored in a TH1F that can be defined in the AddTask.


sashingo:
+test

sashingo:
+1

PR #648, opened 2017-04-25
Remove efficiency calculation when running in data

hzanoli:
I forgot to add the protection in case it is running in data. So the task was crashing in data.


arossi81:
+1

PR #651, opened 2017-04-25
Cleanup of AliAnalysisTaskCEP

pbuehler:
. cleaned directory by packing old files into tar file
. cleaned some printouts
. corrected bug in track selection AliCEPUtils::GetCEPTracks


dberzano:
Hello @pbuehler, your commit 7453dff is truly ugly. You do not pack old code, you have the Git history for that. In addition, Git does not perform very well with non-text files. I am going to remove your file in a separate pull request (unfortunately as per our policy we do not remove it from the history).

PR #653, opened 2017-04-25
Setting SetTMClusterRejection() according to fTMClusterRejected (instead of fixed value), outside and inside the isolation cone

e-masson:
This was previously set in the run macro (or in a wagon) on the current instance of the task. Now the two parameters bTMClusterRejection and bTMClusterRejectionInCone are used directly inside the AddTask.
Calling SetTMClusterRejection()/InCone() in the run macro (or wagon) overwrites this new inside-AddTask setting so that nothing changes for a user who does not change his/her run macro (or wagon).


dlodato:
+1

PR #707, opened 2017-04-28
Allow daughters of V0s to pass track cuts

pbuehler:
With the old initialisation of GetStandardITSTPCTrackCuts2010 the daughter of V0s did not pass the track selection. With this modification V0 daughters pass the cut.


dberzano:
This build problem is unrelated to your commit, please be patient and it will be fixed automatically (see alisw/alidist#696).

PR #724, opened 2017-04-30
pT weight for D meson Raa for 5.02TeV at 0-10%

xinyepeng:


arossi81:
+test

arossi81:
+1
Hi Xinye,
we can commit it now. I find it a bit unpractical the implementation of the weights defining 500 points, but it's ok (and not really differing from storing an histo with 500 bins).
Cheers,
Andrea

PR #736, opened 2017-05-01
Added PHOS nonlinearity, bad channel map and tender task to GammaConv group directory

nschmidtALICE:
Added PHOS nonlinearity, bad channel map and tender task to GammaConv group directory
- includes 7 TeV LHC10b-f bad channel map and non linearity correction
- adds trainconfig which uses the nonlinearity 
- includes an AddTask made from the PHOStender to load custom bad channel maps


FriederikeBock:
+1

PR #756, opened 2017-05-02
Implementation of geometry-based (eta,phi) boundaries

e-masson:
The EMCal (eta,phi) limits used for selection of clusters, isolation and UE studies are now based on the detector geometry corresponding to the analysed run (using the geometry instance fGeom defined in AliAnalysisTaskEmcal).
A new data member (fPeriod) and a new function member (SetPeriod()) are defined in order to retrieve the LHC period passed as the first argument of AddTaskEMCALPhotonIsolation.C. This allows to deal with the EMCal max phi for 2012-2013 runs, 20° higher in fGeom than in reality (the two last SMs at the time were already integrated in EMCal but not activated).


dlodato:
+1

PR #770, opened 2017-05-03
Initialised TString variables in the member initialization list

rihanphys:
Initialised TString variables in the member initialization list.


rbertens:
+1

PR #795, opened 2017-05-04
New codes to analyze MCEvent and produce charmed baryon spectra

yowatana:
This code is to analyze the MCEvent at generator level (directly from galice.root and Kinematics.root). I am using these codes to analyze Herwig7 events to produce Lc/D0 ratio using PWGMM train.


ebruna:
+test

ebruna:
+1

PR #809, opened 2017-05-05
Weight MC background to data

hzanoli:
Added the possibility to weight the background to the pi0 and eta spectra available in data. The weighting to Hijing is still available for comparison.


sashingo:
+test

sashingo:
+1

PR #817, opened 2017-05-05
Update of class used in the LMRDimuon analysis.

bteyssie:
Adding the cut in pileup and the parametrization of this cut in function of collision system
    Correctif in physics selections
    Adding new multiplicity estimator


dstocco:
+test

aphecetche:
+1

PR #827, opened 2017-05-06
Use histograms for weights

hzanoli:
When the Hijing + enhanced pi0 and eta spectra are merged, the total shape is harder to fit. It is easier to use histograms with very fine binning, since the statistics at high momentum is good due to the enhancement.


sashingo:
+test

sashingo:
+1

PR #835, opened 2017-05-08
Z coordinate values taken from CPass1

dcaffarr:
Due to the HM trigger selection a bias in the evaluation of the Z coordinate in CPass0 was found. These values are now taken from CPass1 and the correspondent OCDB objects are updated (both GRP/Calib/MeanVertex and GRP/Calib/MeanVertexSPD).


fprino:
+test

fprino:
+1

PR #848, opened 2017-05-09
fixed memory allocation bug

bschaefe:
unnecessary histograms are removed that had not been initialized with null pointer


aohlson:
+1

PR #862, opened 2017-05-10
add some QA histos

jmargutt:
add some QA histos:
store separately real and imaginary components of scalar product
store Q-vector cross terms (<QAxQCy>, ...)
store NUAs, both of D mesons and reference plane (<QAx>, ...)


fprino:
Dear Jacopo,

the code of this task is very similar to the one of AliAnalysisTaskSEHFvn.
Is there a reason why you do not use AliAnalysisTaskSEHFvn, which can be fonfigured to do the first harmonic?
If this is possible it would avoid code duplications.

Cheers, Francesco

jmargutt:
Hi Francesco,
we want to use the reference plane from the ZDCs, which require some tricky corrections not easily implementable in the event plane framework. I currently use my own class (AliAnalysisZDCEP) for that and I needed to change the corresponding methods; we therefore preferred to commit a separate class for testing purposes, without affecting the existing one.
I would propose the following: if the results of this first exploratory study are promising, and if other people are interested in pursuing these kind of studies as well, we can extend the original class (AliAnalysisTaskSEHFvn) to include the ZDCs, of course. It was in the back of our minds already :)

ebruna:
+test

ebruna:
+1

PR #866, opened 2017-05-10
new HFE task

adubla:


ebruna:
+test

ebruna:
+1

adubla:
Last Spam :)

code has been officially committed and it will be available in today tag.

Mattia, I will try to set up some wagons as soon as the tag will be out. In
the worse scenario after dinner.
But you can start to send me which setting and on which list you would like
to run.
I guess we can start with the Run2 data and later on with the anchored MC.

You can reply to me privately so we do not spam everybody ;)

Ciao,
Andrea

PS: i just asked to become operator of the train in pp as well, should not
take long :)


*************************************************************
Andrea Dubla
ALICE Group - CERN
GSI Helmholtz Centre for Heavy Ion Research
Planckstraße 1 - 64291 Darmstadt-Germany
*************************************************************

PR #867, opened 2017-05-10
Fixed usage of histograms for runwise TPC PID correction

carstooon:
Old code recreated histogram in new scope and was therefore not able to use histogram outside this scope


pdillens:
+1

Could you commit the same change for line 1769 as well. Should be the same problem as far as I see it.

PR #869, opened 2017-05-10
Adding pileUp option to container name

dlodato:
Adding piece of name for output container in case one does pileUp rejection analysis.


dlodato:
Testing watching for @rqwa

PR #870, opened 2017-05-10
small change in the tree maker

iarsene:


iarsene:
+1

kreisl:
I think this pullrequest reintroduces the problems fixed in pull request #432.

iarsene:
Hi,
Did you check that the memory leak is there with this pull request? The things is that without the fix, the event selection on the tree maker is not applied when running on the grid. As you can see in the tree maker there are a few more filters which are also streamed since longer time ago. To me it seems that the problem is in the AliDielectronEventCuts, not in the event filter of the tree maker.

PR #871, opened 2017-05-10
EMCal Embed: Fix pythia scaling

raymondEhlers:
The xSec hist is a TProfile, so we should _NOT_ scale it by the
number of entries - it is already the mean xSec/event!

The corresponding entries for the xSec and nTrials were renamed to
clarify. Note that the nTrials is _still_ scaled by the number of
entries because it is _not_ a TProfile!


raymondEhlers:
+1 Is this stuck?

PR #876, opened 2017-05-11
PWGGA/GammaConv: added additional nonlinearity wagon configurations for LHC15n

nschmidtALICE:
- Additional trainconfigs for Calo and ConvCalo including EMC triggers and all nonlinearity variations


FriederikeBock:
+1

PR #888, opened 2017-05-11
Disable root history when performing ctest

dstocco:
The ctest runs a series of load-libraries commands that are stored in the root history.
Since there are lots of libraries, the user history basically goes outside the buffer.
The patch disable the root history before loading the libraries.


majanik:
+1

majanik:
This is a very useful update, lack of the user history after aliroot restart was very inconvenient.
Thanks for this!

dberzano:
+1 this is the same as alisw/AliRoot#182, thanks @dstocco!

PR #896, opened 2017-05-12
changed TOF timing and momentum bining

bschaefe:
minor changes over ranges and granularity of both timing and momentum, used for systematic studies


aohlson:
+1

PR #897, opened 2017-05-12
PWGGA/GammaConv: Added empty PHOS bad channel map for QA

nschmidtALICE:
- commit adds an empty bad channel map that can be force loaded by the PHOS tender


FriederikeBock:
+1

PR #901, opened 2017-05-12
Weight in MB

hzanoli:
I included histograms to check the tagging efficiency using only the non-enhanced part of the sample, but weighting it to data.


sashingo:
+test

sashingo:
+1

PR #913, opened 2017-05-13
modified two codes for kink analysis pp 5 TeV

nurgu:
Few lines are deleted due to previous error during compiling


bdoenigu:
+test

bdoenigu:
+1

PR #932, opened 2017-05-15
PWGGA/GammaConv: Added switch to use external NonLinearity for PHOS

nschmidtALICE:
- Added boolean to AddTask_PHOSTender_PCMconfig.C to disable the nonlinearity correction from the PHOS group
- This bool must be set to kFALSE if a NL correction from AliCaloPhotonCuts.cxx should be applied


dmuhlhei:
+1

PR #945, opened 2017-05-16
store separately even and odd components

jmargutt:
store directly even (uQA+uQC) and odd (uQA-uQC) components of v1:
higher signal expected --> better fits to inv.mass. distribution


ebruna:
+test

ebruna:
+1

PR #948, opened 2017-05-16
Only one output file created (compatible with LEGO train)

SHornung1:
Previous version was not compatible with LEGO train (multiple outputfiles)


dberzano:
It appears there is something wrong. Did you rebase before pushing? In order to avoid confusion and accidental merging, I am closing this PR now.

PR #956, opened 2017-05-16
PWGGA/GammaConv and TENDER/TenderSupplies: Added feature to use custom OADB bad channel maps for PHOS

nschmidtALICE:
This pull request adds support for loading custom bad channel maps (OADB or standard file) in AliPHOSTenderSupply. In addition a function is added that allows to obtain the current bad channel maps from the PHOSTender. This function is then used in AliCaloPhotonCuts for the extended QA.

-  PWGGA/GammaConv/AliCaloPhotonCuts.cxx: 
    -- Changed loading procedure of PHOS bad channel maps. 
    -- It will now always load the current bad channel maps from the AliPHOSTenderSupply (tested on ESDs)

- PWGGA/GammaConv/CMakeLists.txt:
    -- Included PHOSTasks in GammaConv CMakeLists to get access on AliPHOSTenderTask

- PWGGA/GammaConv/macros/AddTask_PHOSTender_PCMconfig.C:
    -- Updated custom PHOS tender AddTask for GammaConv-Group to allow loading of custom bad channel maps (0: no forced map, 1: forced OADB map, 2: forced single map file)

- TENDER/TenderSupplies/AliPHOSTenderSupply.cxx:
    -- added fPrivateBadMapIsOADBfile and fBadChannelMapFile as necessary variables
    -- adjusted ForceUsingBadMap() to differentiate between single bad map file and OADB file (without changing the second argument of this function, it will work as before and only load the bad maps from the single bad map file) 
    -- Added the possibility to use a force loaded OADB object in InitTender(). Without changing the boolean fPrivateBadMapIsOADBfile (done in ForceUsingBadMap()), this part will work as before

- TENDER/TenderSupplies/AliPHOSTenderSupply.h:
    -- Added boolean to ForceUsingBadMap which indicates weather an OADB object or a single bad map file should be loaded
    -- Added function GetPHOSBadChannelStatusMap() to get access to current bad channel map (this is now used in PWGGA/GammaConv/AliCaloPhotonCuts.cxx)


FriederikeBock:
+1

peressounko:
Hi Nicolas,

I agree with adding
+ TH2I * GetPHOSBadChannelStatusMap(Int_t iModule) const { return
(TH2I*)fPHOSBadMap[iModule] ; }
method (but you should  keep in mind that histograms will be available
only after processing of the first event.

Concerning private OADB I am not sure I understand you correctly. Do you
want to produce a full OADB object with PHOS bad maps and then use it in
analysis? Why not provided approach?
Then it will be easier to make method like
SetPrivateOADBBadMap(char * filename){fPrivateOADBBadMap = filename;}

and in the InitTender method to test if(fPrivateOADBBadMap.Length()!=0) then use it, otherwise - default path:

   AliOADBContainer badmapContainer(Form("phosBadMap"));
   if(fPrivateOADBBadMap.Length()!=0)
badmapContainer.InitFromFile(fPrivateOADBBadMap.Data(),"phosBadMap");
   else
badmapContainer.InitFromFile("$ALICE_PHYSICS/OADB/PHOS/PHOSBadMaps.root","phosBadMap");

   Best regards, Dmitri

FriederikeBock:
Hi Dmitri, 

our idea was to just change the path for the BadChannel maps OADB-object, for tests for instance, while keeping the rest the same. In this case one wouldn't need to change anything in the structure of your tender supply except to add a string for the path. For EMCal we do exactly the same to change for instance between the OCDB, OADB and CVMFS if necessary.
I would prefer that to externally initializing an object which might be problematic using the lego trains. 

Regards,
Friederike

peressounko:
OK, then indeed implement change of the file path string, for example as I suggested. Proposed changes can be dramatically simplified.
          Dmitri

FriederikeBock:
We will adjust the necessary things. 
Thanks for reviewing.
Friederike

mfasDa:
Hello Nico,

Please remove "using namespace std" from the header file. Using directives in header files create problems.

As this class is rather unrelated to the other classes in PWG/Tools and more dealing with common calorimeter tools I suggest putting this into an extra library PWG/CALO. This helps keeping track of the intended functionality, and prevents that the library PWG/Tools becomes a collector of code where the authors don't really know where to put it.

Best Regards

Markus

nschmidtALICE:
@peressounko : I just changed how the OADB object would be loaded. This follows your suggestion to use the path.
@mfasDa : I removed the namespace std from the header.

FriederikeBock:
+1

FriederikeBock:
@peressounko I think you directly need to comment on the pull request (you gave the approval of the commit directly), otherwise it won't merge the commit. 
Regards,
Friederike

nschmidtALICE:
@mfasDa : I just moved the new class to a new class PWGCALO in PWG/CALO.

I hope that this is now okay.

peressounko:
+1

FriederikeBock:
+test

mfasDa:
@nschmidtALICE Looks good! Suggesting approval!

FriederikeBock:
+1

FriederikeBock:
+1

PR #968, opened 2017-05-17
fix to allow correct energy and centrality selection 

jaimenorman:
The bracket placement meant that the selection of the energy and centrality for p-Pb collisions when creating the AliHFSystErr object did not work correctly. It should be working now. I also added the option to select the BDT analysis.


arossi81:
+test

arossi81:
+1

PR #969, opened 2017-05-17
Fix in the analysis tracks type for OrthCones and FullTPC methods

e-masson:
The two isolation/UE methods PtIsoTrackOrthCones() and PtIsoTrackFullTPC() were using the TPC-only track container instead of the Hybrid one dedicated to analysis.
In addition, the track containers are now called by their name and no longer by their index (as recommended in the EMCal framework documentation).


dlodato:
+1

PR #976, opened 2017-05-17
Update: HFE AddTask and Config for pp analysis

SHornung1:


ebruna:
+test

dberzano:
Hi @ebruna, I see you `+test` a lot - FYI, if you have actually reviewed the code already, it's more convenient for you to use `+1`: this will imply an automatic merging without further intervention, but only in case our automatic tests pass. In case you have `+1`d a pull request which does not compile, worry not: it won't be merged.

The reason why you see the admins (as myself) doing `+test` is to speed up the process in case the real mentioned persons haven't replied yet. Note that this is not going to be necessary in the future as we plan to automatically start tests for all pull requests.

ebruna:
Ciao @dberzano , thanks for the tip!

ebruna:
+1

SHornung1:
Sorry, there was some delayed commit which has not taken part in the test.
(The AddTask for 5 TeV was up-tp-date.)

PR #979, opened 2017-05-17
Fixing ConfigHFEnpepp.C and cleaning up AddTask-macros

SHornung1:
Cleaned-up the AddTasks and fixed some bug in the 5 TeV version

The general Config-macro had a wrong name for the main function.

(Sorry, I had some problems with the pull request I started yesterday, which was not rebased correctly.)


ebruna:
+1

PR #981, opened 2017-05-17
Add method AliAnalysisMuMuEventCutter::IsSPDPileUp

benjaminaudurier:
Add a new data member fAnalysisUtils object and a method in order to call the AliAnalysisUtils::IsPileUpSPD in AliAnalysisMuMuEventCutter


aphecetche:
+test

aphecetche:
For the record, instead of adding a method to AliAnalysisMuMuEventCutter (and risking to introduce a bug in it) another (slightly better, IMHO) option would have been to provide a completely independent new class (implementing the event cutting policy) with the new method IsSPDPileUp. That was the original idea behind (event) cutters : to keep them independent.  The MuMuEventCutter was originally meant as an example implementation only.

aphecetche:
+1

PR #987, opened 2017-05-18
Add possibility for Cocktail to work with V2 parameterizations stored in file

peressounko:
Added functionality is following:
If you do nothing, code behaves as before.
If you set non-zero last parameter in AddMCEMCocktailV2.C - dir name "v2_20-40"  //or v2_00-20, v2_40-80 etc. , then code will take v2 parametrizations from file which you select in configuration (paramFile) with pT spectra and v2 parametrizations

This file should contain directory with v2 parametrizations of at least pi0 and optionally of other hadrons.
Then these parametrizations are used in simulations. If V2 for a given hadron is provided, it is used, if not, it uses K0s parametrization with EKt scaling, if not - pi0.  This is approach which I used in cocktail for direct photon v2/V3

Best regards, Dmitri


FriederikeBock:
Hi Dmitri,

a) If we now have the functionality to load v2 parametrisations from the input file I would completely take out the hard-coded ones, to keep it clean and simple (PWG/Cocktail/AliGenEMlibV2.*).
b) Does this change already include that its not just v2 but vN (with one N at a time in general). If so the functions should be renamed accordingly.
c) If no v2 parametrisation file is given the function shouldn't be executed (meaning check the string before executing in the addtask, or put a bool in addition)
d) Could you please also create the pull request for our cocktail git-lab to create those input files, as @lleardin and you need to make sure you're not interfering with each other when pushing new PbPb.root files.

Regards,
Friederike

peressounko:
Hi Friederike, 
a) Presently they are used with Lucia's spectra parametrizations since her files do not contain V2 parametrizations. In addition I am not sure that nobody uses these parametrizations in his analysis. 
b) No, my changed concern only V2. There is no possibility to simulate Vn, n>2 in AliRoot, wee need dedicated commit to be able to do this. However, proposed scheme can easily extended to Vn by adding corresponding dirs into file. So I propose to start with V2 and then extend it to Vn when proper modifications to AliRoot will be ported.
c) Most of cocktail users do not need Vn parametrizations and therefore should not care if it exists and if it is correct or not. I would prefer not to modify standard behaviour for them. 
d) OK. So far we use different file names, so this should not be a problem. BTW, why do we need several files for different centralities in PbPb and not one file with different dirs in it?
  Best regards, Dmitri.

PWG/Cocktail/macros/AddMCEMCocktailV2.C
```diff
@@ -51,6 +52,7 @@ AliGenerator* AddMCEMCocktailV2(  Int_t collisionsSystem      = 200,
   gener->SelectMotherParticles(selectedMothers);
   gener->SetCollisionSystem(collisionsSystem);                //pp 7 TeV
   gener->SetCentrality(centrality);                           // kpp
+  gener->SetParametrizationFileV2Directory(paramV2FileDir);
   (AliPythia::Instance())->SetMSTU(22, pythiaErrorTolerance);   // tolerance for error due to rhos
```
FriederikeBock: I would add here:
if (paramV2FileDir.CompareTo("")! = 0 ) gener->SetParametrizationFileV2Directory(paramV2FileDir);

instead of 
gener->SetParametrizationFileV2Directory(paramV2FileDir);

FriederikeBock:
Hi Dmitri,

concerning a) Lucia might run them but she's not analysing it versus v2 so it shouldn't influence her. For anybody who still want's to do it the old way there is still the older version of the classes available, or one can produce a file with exactly the same settings for reading. Let's remove the legacy code now.

concerning b) too bad ... but lets prepare the changes for aliroot then as well 

concerning c) See my comment in the Addtask- code, I just wanted a kind of protection

concerning d) Ideally we shouldn't have different files or make them clearly different in their names. Ie. rename yours to Flow_PbPb.root. Furthermore I would really like to avoid to commit here files which we cannot reproduce using the cocktail git lab, so I urge you to commit your modifications there as well.

Best regards,
Friederike

peressounko:
a) one may not analize it, but some V2 parametrization will be called in PbPb simulations. We have to provide some default V2 parametrization to handle case when special one is not present in the root file.
d) here we need some format of root files. I propose to use same as for pp case: different parametrizations in dedicated dirs in one PbPb_2.76.root file. Dirs should have (unique) self-explanatory names and detailed descriptions in twiki page. Compared to set of separate files this this approach is more vulnerable to corruption of existing parametrisations while adding new ones.  Lucia, what do you think?

lleardin:
Unless I am misunderstanding what you mean, we already have this kind of structure in the PbPb.root file. Directory are per centrality and per variation, with the name as explanatory as I could while keeping them short. A description on the twiki can for sure help too.

peressounko:
I see set of files for different centralities (including yesterday commit) like this one
 TFile*         parametrisations/PbPb_2.76TeV_0010_default.root
  KEY: TH1D     histoMtScaleFactor;1
  KEY: TF1      111_pt;1        [0]*TMath::Power(x,-1*([1]+[2]/(TMath::Power(x,[3])+[4])))
  KEY: TF1      221_pt;1        x*[0]*pow(exp(-[1]*x-abs([2])*x*x)+x/[3],-[4])
  ...
these are something special?

lleardin:
No, as i said in the meeting, this is an obsolete file. I do not think it even works with the code we have now and it should be removed. The one I committed recently, called only PbPb.root, is the one to consider.

FriederikeBock:
Hi Dmitri,

in case no v2 is specified it should be set to a flat phi distribution. Could you add this feature please. 

Regards,
Friederike

peressounko:
With latest commit I added default zero flow if no specific parametrization provided, added PHOS spectra and v2 parametrizations to PbPb.root file.

FriederikeBock:
Hi Dmitri, 

thanks so much! I have 2 remaining remarks though. 

a) Instead of calculating on the fly the pt value for a certain particle in case of Et-scaling could we simply create the correct functions at the beginning according to Et scaling and afterwards just draw from them.
b) Could we include a Get function in the AliGenEMlibV2 & AliGenEMCocktailV2 which returns us those functions used for the v2 generation, mainly for bookeeping purposes later in the tasks (we do something similar for the pt params), this kinda requires part a) of my request.

Furthermore from a conceptual point of view it should be better to use the v2 parametrization of the protons for all baryons, if we need to do mt scaling, could this still be integrated. 

Regards,
Friederike

peressounko:
Hi Friederike, 
a)anyway one has to re-calculate pt on-fly. In your code you implement this re-calculation into TF1 formula, while I use compiled function. Normally, compiled function is 100-1000 times faster. So I think my approach is more efficient. This also is related to your second request: In my approach I can not construct and return TF1* from static functions of a class for all my parametrizations. Could you please provide example where this functionality is required?
Concerning proton parametrization,  yes it would be more consistent. But so far baryons are not used (contributions 100 times smaller than sys. uncertainties should be neglected). So I propose to concentrate on necessary part and introduce complications only if necessary.
 Best regards, Dmitri

laltenka:
Hi Dimitri,
I would also think that your approach is more efficient than using a TF1 for the scaled parametrizations. However, we chose to use the TF1s to provide also for an easier bookkeeping also for the scaled parametrizations. In our cocktail analysis task (PWGGA/GammaConv/AliAnalysisTaskGammaCocktail) we pipe the parametrizations, both the original and scaled ones, that were used in the generation through to our analysis output. I think this helps to ensure that no information is lost as well as it makes it easier to debug possible problems. Would it not be possible to use are approach for the generation but calculate the scaled TF1s just once to pipe them through?
On a side note this would also be possible for the pt parametrizations and should improve the efficiency of the code then.
Best regards, Lucas

dberzano:
Hi, this PR has been opened quite some time ago, and now it has conflicts. What is its status? Can we close it so that you can open a new one when ready?

peressounko:
Hi, 
I do not see how to fulfil requirements for bookkeeping. Lets close the pull request.
         Dmitri

PR #989, opened 2017-05-18
added transverse kinetic energy scaling

bschaefe:
KET scaling is added so that comparisons of deuteron correlations can better be made with hadron-hadron and proton-hadron correlations


aohlson:
+1

PR #995, opened 2017-05-18
Fix installation of Doxygen documentation

dberzano:


dberzano:
Same as alisw/AliRoot#193.

PR #996, opened 2017-05-18
Migrate TOF calibration AddTasks from AliRoot to AliPhysics

jezwilkinson:
Tasks AliTOFAnalysisTaskCalibTree & AliTOFAnalysisTaskCalibPass0 had been migrated to AliPhysics after AliRoot/AliPhysics split, but AddTask macros were orphaned in AliRoot (& not added to any CMakeLists, so inaccessible from LEGO train system). This patch migrates the missing macros to where they should be.

### DIFF
--- a/PWGPP/CMakeLists.txt
+++ b/PWGPP/CMakeLists.txt
@@ -326,6 +326,7 @@ install(FILES ITS/QA/MakeTrendingITSQA.C
 install(FILES TOF/AddTaskTOFQA.C
               TOF/AddTaskTOFqaID.C
               TOF/AddTOFAnalysisTaskCalibTree.C
+              TOF/AddTOFAnalysistaskCalibPass0.C
               DESTINATION PWGPP/TOF)
 install(FILES TOF/trending/MakeTrendingTOFQA.C
 	      TOF/trending/DrawTrendingTOFQA.C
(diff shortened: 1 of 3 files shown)


preghenella:
+1

preghenella:
+test

chiarazampolli:
Hello,

What are these macros? Those for CPass are in:

$ALICE_PHYSICS/PWGPP/CalibMacros/CPass0
$ALICE_PHYSICS/PWGPP/CalibMacros/CPass1

Cheers,

Chiara

jezwilkinson:
@chiarazampolli Ah. Sorry, my mistake then. Looks like those are indeed the same macros (give or take a couple of debug lines). Thought I'd searched through but must have missed this folder, not sure how that happened. I'll revert the pull request, apologies for the confusion.
In any case, those two macros seem to still exist in $ALICE_ROOT/TOF, which is what led to the confusion in the first place (as it looked like the tasks had moved but the macros not...).

chiarazampolli:
Ciao Jeremy,
If they are the same macros, then we should remove them from TOF (not from PWGPP/CalibMacros, since we use them from there). Could you take care?

Chiara

jezwilkinson:
Ciao Chiara,
The revert request (#997, https://github.com/alisw/AliPhysics/pull/997) should remove from the PWGPP/TOF folder and not touch the PWGPP/CalibMacros ones. Or do you mean to also remove from AliRoot?
--Jeremy

chiarazampolli:
Ciao Jeremy,
Sorry for the late reply. I don't know if someone uses the macros from AliRoot, so I cannot really answer, but I would guess the macros are not needed in 2 places, so I would remove them from AliRoot.

Cheers,

Chiara

PR #1006, opened 2017-05-19
Added period LHC16l for correction of the number of accepted tracklets…

crisjahnke:
Added period LHC16l for correction of the number of accepted tracklets given its vertex dependence. Also added the profile of the dependence on file estimators.root for the LHC16l


zwound40:
+1

zwound40:
+1

PR #1018, opened 2017-05-20
Few modification in Kink pp 5 TeV codes

nurgu:
Modifications are done in analysis codes and macros both for data and MC


bdoenigu:
+test

bdoenigu:
+1

PR #1023, opened 2017-05-22
Increase mixed event stats

yowatana:
Change to increase the mixed event statistics


fprino:
Hi Yosuke,
are you sure this change will increase your statistics of mixed events?
From the code modification it seems that the events that you reject with the current code have either no tracks ot no V0 passing the selections, so adding these events in you analysis may not add any candidate.
But maybe I overlooked something.
Cheers, Francesco

yowatana:
Dear Francesco,

Yes, it does. (I checked) I was storing only the events having both electron and Xi.
The number of such events is very scarce, and when we split the job,
we most likely finish running before we accumulate enough events.
After the modification, events with single e and single Xi are also stored, so
same event pairs are not affected, but mixed-event pairs benefit from it.

Actually, I am reverting the code to the code at the very beginning. This line is added
to check the effect in PbPb analysis and I forgot to revert. For PbPb analysis, it was
easy to accumulated mixed-event statistics even with the condition. 

Best regards,
Yosuke

fprino:
+1

fprino:
Hi Yosuke,
thanks for the explanations.
Cheers, Francesco

PR #1030, opened 2017-05-22
added TOF channel TProfile, extensive house-cleaning

bschaefe:
this edit is made largely for organizational purposes, but additionally has KET scaling histograms as well as a simple TOF channel timing study


aohlson:
+1

PR #1050, opened 2017-05-23
Added jet mass to jet extractor task

rhaake:
This commit changes the AliBasicJet definition and the extractor task to extract the jet mass which is calculated by the jet finder


ebruna:
+1

PR #1054, opened 2017-05-23
Additional info to do improved vertexing with KFParticle

llayer:
In order to use KFParticle offline it is necessary to store the 
track parameters and covariance matrix for each electron track,
therefore these parameters are included in AliReducedTrackInfo,
besides the covariance matrix of the event vertex is needed to 
use the topological constraint of the KFParticle


dberzano:
+test

iarsene:
+1

PR #1062, opened 2017-05-24
Added pthard extraction/parameter

rhaake:
pT hard parameter for extracted jets is necessary for correct MC scaling.


ebruna:
+1

PR #1067, opened 2017-05-24
update on correction for multiplicity dependence on vertex for LHC16l.

crisjahnke:
estimators.root: fixing name of TProfile for LHC16l period

AliDielectronHelper: update on reference number for LHC16l to apply correction for multiplicity dependence on vertex.


otonvd:
+1

PR #1072, opened 2017-05-24
Compute acceptance with toy MC on the basis of pt-shape histogram

jhamoncern:
Compute the acceptance correction factor of D mesons with a toy MC (13 TeV). The macro is very similar to ComputeAcceptance.C but:
- the pt shape (FONLL or PYTHIA) is taken from histogram (TH1D) instead of function (TF1)
This is motivated by the fact that no efficient fit functions has been found to match both the low- and high-pt part of the FONLL and PYTHIA D-meson spectra.

### DIFF
--- a/PWGHF/vertexingHF/macros/ComputeAcceptanceFromHistogram.C
+++ b/PWGHF/vertexingHF/macros/ComputeAcceptanceFromHistogram.C
@@ -180,7 +180,7 @@ void ComputeAcceptanceFromHistogram()
     } else {
       funcPt = LoadPYTHIA13TeV_D0();
     }
-    outFileName.Append("promptFONLL13ptshape.root");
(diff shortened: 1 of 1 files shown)


fprino:
Hi Julien,

wouldn't it be better to set the usage of histogram as an option in the existing macro?
The risk of duplicating the full macro is that we will have to maintain two macros.

Cheers, Francesco

jhamoncern:
Hi Francesco,
I created a new macro to avoid an if/else inside the long loop. But being smarter it can be avoided I think. I will update the original ComputeAcceptance.C macro and let you know.
Cheers, Julien

jhamoncern:
As Francesco suggested, it might be better to update the actual ComputeAcceptance.C macro. I close the request.

PR #1076, opened 2017-05-24
matching inefficiency tune on data set to 0 for LHC15o

pantonioli:
Modified TOF OADB for LHC15o: matching inefficiency tune on data set to 0 for LHC15o


chiarazampolli:
+1

PR #1077, opened 2017-05-24
MLtreemaker update

sebaleh:
- MLtreemaker now uses AliDielectron cut classes to be consistent with nanoAOD cuts
- vertex positions are now also written to tree for no MC events
- chi2Global was replaced by chi2GlobalPerNDF

treemaker cuts and nanoAOD filter cuts were checked to be consistent


miweberSMI:
+1

PR #1080, opened 2017-05-24
Add 13 TeV pt-shape histograms to the toy MC

jhamoncern:
- Add D meson pt-shapes as obtained from FONLL and PYTHIA at 13 TeV. They have been added as histograms (TH1D) instead of function (TF1) since no fit functions have been found to efficiently match the low- and high-pt part of the spectra.

- Increase the number of point of TF1 to make sure that the sharp peak region is well described (as advised in TF1::GetRandom).

- Avoid new/delete of the TLorentzVector inside the loop to speed-up the task


fprino:
+1
Maybe we need finer pt binning for the histograms, you can consider this for a possible future update

PR #1082, opened 2017-05-24
PWGGA/GammaConv: adapted cut string initializers for use of alphanumeric

laltenka:
initializers are now able to handle alphanumeric characters, i.e. usual digits 0-9 and characters a-z, which are translated into digits 10-35


FriederikeBock:
+1

PR #1088, opened 2017-05-25
total momentum cut imposed, rapidity cut removed

nirbhaygit:
Total momentum cut imposed, rapidity cut removed. This change (temporary) is made to compare results with Identity method.


altsybee:
+1

PR #1107, opened 2017-05-26
Add PYTHIA and FONLL pt shapes for feed-down D mesons

jhamoncern:


arossi81:
+1
Hi Julien,
thanks for the work. I did not go through the changes in the code in detail, but I assume that you tested the new functionalities you added to the macro, so it's fine for me to commit it. Notify the others about the fact that they can try to use the weights also for the acceptance factor.
Cheers,
Andrea

jhamoncern:
Hi Andrea,
Yes, I tested the changes. I will send a mail to all analysers.
Cheers, Julien

PR #1126, opened 2017-05-29
Added FONLL scale error in fprompt uncertainty

fgrosa:


fprino:
+1
Did you test if the treatment is OK also for fc?
In case, we could consider to take the FONLL uncertainty from the TGraphErrors in the output of the Raa macro.

fgrosa:
Ciao Francesco,
yes I checked and it is ok also for fc: the error on the FONLL scale as it is computed now it gives exactly the result that is in the TGraphErrors in the Raa macro, so if you prefer we can take the uncertainty from the TGraphErrors.

fprino:
Hi Fabrizio,

I think it is OK to take the uncertainties from the ntuple as you did in the current code, since you verified that it gives the correct result..
Thanks!

Ciao, Francesco

PR #1133, opened 2017-05-29
PWGGA/GammaConv: bugfix to fill the TOF histograms

hmurakam:
this is a small fix to fill TOF histograms.


FriederikeBock:
A similar fix is then needed for the AliConversionPhotonCuts.cxx could you provide it please.

FriederikeBock:
+1 
Thanks for the fix!

PR #1134, opened 2017-05-29
Update default AliPhysics path in the toy MC macro

jhamoncern:
The AliPhysics path was referring to the old path, before AliBuild.


zconesa:
+1

PR #1135, opened 2017-05-29
Adding the possibility to read B polarity dependent histograms

afestant:
Possibility to read templates for the impact parameter in the transverse plane vs pt depending on the phi of the track and on the B field polarity of the event.


ebruna:
+1

zconesa:
+1

PR #1144, opened 2017-05-30
PWGGA/GammaConv: Added run dependent PHOS bad channel map container

nschmidtALICE:
This pull request add an OADB container with run dependent bad channel maps for PHOS in the LHC10[b-f] dataset
- this is a follow-up to the previous update that allows loading of custom OADB objects as bad channel maps in the PHOS tender


FriederikeBock:
+1

PR #1149, opened 2017-05-30
Fix directory naming to run multiple wagons in the same train

afestant:
Fix directory naming to run multiple wagons in the same train


zconesa:
+1

PR #1163, opened 2017-05-31
Add possibility to disable ITS and TPC Refit

hzanoli:
Hi,

In order to use the new cuts in the p-Pb analysis I have made a small flag to disable ITS and TPC refit. I need this commit to be able to run trains for systematics. The change doesn't modify the default behaviour of the task (fRequireITSAndTPCRefit = kTRUE).  So unless you explicit request, no other user is affected. Since this is a common task, I would suggest that someone else that also use it review the changes. I am tagging some people that I know use the task @crisjahnke @deepathoms.


sashingo:
+test

ebruna:
+1

PR #1173, opened 2017-06-01
Remove strong tracks cuts on the partner

hzanoli:
The cuts are now only TPCOnlyTracks and |TPC Nsgima| < 3. There is also a small change in the AddTask to check the bin size effect.


sashingo:
+test

ebruna:
+1

PR #1176, opened 2017-06-01
Fixes for TF1 in Root6

pzhristov:
Root6 expects reference to function f(double *, double*) that returns double.


pzhristov:
+test

pzhristov:
+1

PR #1183, opened 2017-06-01
possibility to use generated Pt for MC match.eff.

abarbano:


fprino:
Hi,

you need to increase the class revision in the header file since you added some data members

Ciao, Francesco

fprino:
+1

PR #1192, opened 2017-06-02
fix in the check on SDD presence for MC

afestant:
Adding the possibility to run on MC without SDD included in the reconstruction


fprino:
+1

PR #1196, opened 2017-06-02
PWGGA/GammaConv: bugfix for updated cutstrings

laltenka:
the return of the cut strings was adapted (previously returned integer values for each cut and not letters of the original cut string)


dmuhlhei:
+1

PR #1207, opened 2017-06-04
Update of class used in dimuon LMR analysis

bteyssie:
Update of LMR Analysis code for the normalization factor


aphecetche:
+test

aphecetche:
Hi,

What's the status here ? Are the test failing  because of that PR or because of the github issues of yesterday (June 5th,  21:10) ?

dberzano:
Hi, if the system asks you to `+1` then the PR is fine for the system - this is why the red test is being ignored (it's optional). In any case we have disabled it from now on (until it turns green) to avoid confusion like in this case.

Note that our CI will never let you `+1` a truly failed PR. Even if you accidentally approve a PR whose tests fail, our system will ignore your approval.

aphecetche:
+1

PR #1208, opened 2017-06-04
PWGGA/GammaConv: adapted hadronic cocktail analysis task

laltenka:
analysis task can now deal with charged pions


FriederikeBock:
+1

PR #1212, opened 2017-06-05
fixed bug in deuteron KET scaling

bschaefe:
A small edit with about 10 lines of code altered.  Fixes a bug in the way deuteron KET was
calculated which allows easy and immediate comparison to other species.


aohlson:
+1

PR #1220, opened 2017-06-06
updates to trigger study and calo clusters study

jdmulligan:
updates to my personal user task


adriansev:
+test

jdmulligan:
Hi All,
My commit passes the test for the release build, however there seems to be some problem loading the libraries for the root6 build in jet directories. This is not caused by this commit (the same problem occurs for several other pull requests). Can one of you please advise how to proceed? My local root6 build is having some problem so I cannot investigate at the moment -- but would like to get my commit into the build asap. 
Thanks,
James

mfasDa:
@jdmulligan Looks like there are problems with linking against fastjet. From what I saw other pull requests were merged even though the PR-check for ROOT6 failed. So probably with full approval the code will be merged as well. In my opinion there is nothing that should prevent the merge.

jdmulligan:
@mfasDa I see, thanks. Can you please approve the commit then? (so far only test was approved). Thanks!

aiola:
+1

PR #1223, opened 2017-06-06
Adding THnSparses and recovertex option for MC

afestant:
Adding 2 THnSparse for d0 pulls (in rphi and z) vs pt, phi, B, charge. Adding also the possibility to use reconstructed primary vertex when running with MC option enabled.


fprino:
+1

PR #1252, opened 2017-06-07
PWGLF: install required macro

dberzano:


dberzano:
@dcolell I think this is what you wanted to do. You can find additional information on the `install` command from CMake at [this location](https://cmake.org/cmake/help/v3.0/command/install.html). Can you triple-check if this works for you before approving?

dcolell:
@dberzano It should be fine. Approved. Thank you.

bdoenigu:
+test

bdoenigu:
+1

PR #1254, opened 2017-06-07
Add the possibility to have different size of band in eta and phi in …

asvauthi:
…which we estimate the UE

Same as previous commit : no error while building but strange behaviour as I was used to.


efranzis:
+test

gconesab:
Hi Eliane, we are discussing the commit, do not approve yet.

efranzis:
Understood. I just wanted to start the test to save time and then would have actually waited for you anyway to approve.

PR #1272, opened 2017-06-08
PWGDQ: functions for TOF PID post-calibration in dielectron framework

ivorobye:
Implemented in the same way as it is already done for TPC and ITS


tdahms:
+1

PR #1286, opened 2017-06-10
Adding trackcuts for systematics

vvislavi:
Added ITSTPC2011Sys cuts -- loose 2011 cuts for systematics estimation. Also, reduced gamma inv mass range even further (15 MeV) to reduce the size (1st commit) and added the missing ; (2nd commit)


ktf:
@dberzano: tests here were successful, however bot is not merging the PR. Doing it by hand.

ktf:
@dberzano: me commenting on the issue forced a merge. I think you have some race condition in the script...

PR #1288, opened 2017-06-10
adding AliAnalysisTaskEMCALPi0GammaCorr

miguelignacio:
adding  AliAnalysisTaskEMCALPi0GammaCorr. This code is based on previous EMCAL-based pi0 and gamma correlation analysis codes but it does uses THnSparse instead of multiple TH1s.

### DIFF
--- a/PWGGA/EMCALTasks/AliAnalysisTaskEMCALPi0GammaCorr.cxx
+++ b/PWGGA/EMCALTasks/AliAnalysisTaskEMCALPi0GammaCorr.cxx
@@ -1,0 +1,758 @@
+//
+#include <Riostream.h>
+#include <TClonesArray.h>
+#include <TH1F.h>
+#include <TH2F.h>
+#include <TH3F.h>
+#include <THnSparse.h>
+#include <TList.h>
+#include <TLorentzVector.h>
+
+#include "AliAnalysisTaskEMCALPi0GammaCorr.h"
+
+#include "AliParticleContainer.h"
+#include "AliClusterContainer.h"
+#include "AliAnalysisManager.h"
+#include "AliCentrality.h"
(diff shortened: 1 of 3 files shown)


dlodato:
+test

dlodato:
Hi @miguelignacio , 
before approving this new task, did you already discuss with the interested developers
 the possibility of implementing a THnSparse output in the already existing tasks? 
This could avoid code-repetition (as one understands from the description of this commit).
Let me know.
Cheers, 
Davide

miguelignacio:
Hi @dlodato ,

I do not agree that this is a code repetition. With that reasoning one could argue that any code that inherits from _AliAnalysisTaskEmcal_ and loops over clusters to save some histograms is a repetition.
Moreover, we want a loose selection on events and clusters; you cannot do that with existing codes without disrupting them a lot.  Finally, given that I got rid of all TH1 outputs, this code is rather orthogonal. 

best regards,
Miguel

efranzis:
Hi Davide,

I think his code is mainly based on my task AliAnalysisTaskGammaHadron. IN PRINCIPLE there is nothing against merging both of our tasks and using one. However, I also moved to THnSparses and I have large issues with memory. So adding additional stuff is not an option, at the moment. I think we should approve this. 
Regards,
Eliane

gconesab:
+1

EMCALTasks is for user code, if it does not disturbe the compilation, it
can be added.



2017-06-12 19:54 GMT+02:00 Eliane Epple <notifications@github.com>:

dlodato:
Hi @miguelignacio, 
in the description I read "code based on...". I didn't mean to say it is repeated...I just asked..
Having said so, as Gustavo pointed out, no problem...
Davide

PR #1313, opened 2017-06-12
Make task a bit more readable

mfasDa:
- Acceptance check just return true or false depending on whether
  particle is in the acceptance
- Ploton acceptance stored in bitmaps, comparisons replaced by TESTBITs
- PDG codes defied within enum, use constants defined there everywhere
  in the code instead of PDG code numbers


FriederikeBock:
+1 
Thanks for the improvements!

mfasDa:
Latest commit fixing case-sensitivity issue. Code was tested on OS X for which this was not caught in local test.

FriederikeBock:
+1

PR #1322, opened 2017-06-13
PWGGA/GammaConv:Modified weights for JetJetMC LHC16h3 anc LHC15n pass…

hmurakam:
…4 reproduction


FriederikeBock:
+test
Thanks for the update, for my understanding some questions:
a) Did they now remove the other JJ productions for LHC15n completely?
b) How did you obtain the new weights? Did you run locally on a few files - meaning that after a full grid run they might still change slightly?

hmurakam:
a) Did they now remove the other JJ productions for LHC15n completely? -> yes.
b) How did you obtain the new weights? Did you run locally on a few files - meaning that after a full grid run they might still change slightly? ->  new weights obtained by train 2969  output (which requested in the last weekend)

FriederikeBock:
+1
very good, thanks for the clarification.

PR #1325, opened 2017-06-13
PWGGA/GammaConv: Omit deletion of the fListHistos in AliCaloTrackMatc…

pgonzale83:
…her as the analysis fails in the GSI train if list is stored in the global output

### DIFF
--- a/PWGGA/GammaConv/AliCaloTrackMatcher.cxx
+++ b/PWGGA/GammaConv/AliCaloTrackMatcher.cxx
@@ -51,6 +51,7 @@ ClassImp(AliCaloTrackMatcher)
 AliCaloTrackMatcher::AliCaloTrackMatcher(const char *name, Int_t clusterType) : AliAnalysisTaskSE(name),
   fClusterType(clusterType),
   fV0ReaderName(""),
+  fAnalysisTrainMode("Grid"),
   fMatchingWindow(200),
   fMatchingResidual(0.2),
   fRunNumber(-1),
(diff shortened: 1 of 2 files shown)


FriederikeBock:
Hi Pedro,

thinking about it a little I am wondering whether we don't wanna introduce a flag for GSI running, because its so specific to GSI and in general we should free this memory again. 
Wouldn't this make more sense and then implement it as well for all the other classes, where we removedn the deletion to not make it crash at GSI?

Regards,
Friederike

pgonzale83:
Hi Friederike,
I agree with you, it can be dangerous if we don't free the memory in the Grid.
I am going to implement a flag to check if we are running on the grid or  on the GSI.
Best regards,
Pedro G.

FriederikeBock:
Hi Pedro,
very good, thanks so much!
Regards,
Friederike

FriederikeBock:
Hi Pedro,

could you increase the ClassDef please.

Regards,
Friederike

FriederikeBock:
+1
Thanks!

pgonzale83:
Hi Friederike, 
I have increased the classDef from 2 to 3.
Another thing, I wanted to implement something similar in the classes: AliConversionMesonCuts and AliConversionPhotonCuts but I was not sure if the deletion of fHistograms were commented in those classes because they crashed at the GSI or they were commented because they also crashed in the Grid.  do you know the reason?

Best regards,

Pedro G.

PR #1326, opened 2017-06-13
Updates for the multi-trial code for the D-tagged jets

aiola:
Two updates to make the extraction of the raw yield in the SB method more accurate


ebruna:
+1

PR #1329, opened 2017-06-13
-m Improving visualization of D* fit parameters

gluparel:
Improve the visualization of the gaussian mean and sigma in case of D* fit


arossi81:
+1

PR #1344, opened 2017-06-14
Fix T0 calibration macros (from Alla)

dberzano:


dberzano:
+1 as agreed with @chiarazampolli and @shahor02

PR #1345, opened 2017-06-14
PWGGA/GammaConv: Added new cut variations and configs for pp + configs for material task

nschmidtALICE:
This pull request adds two new cutvariations in pseudorapidity as well as new configs in the GammaConv AddTasks.
A second part of this request is the addition of two configs for the Material Budget tasks that will be used for pileup crosschecks.


dmuhlhei:
+1

PR #1397, opened 2017-06-18
Compatibility for parton type from stack for other generators than PYTHIA

rhaake:
For other generators than PYTHIA, it is possible to have less particles on the stacks than 8. This is now taken into account.


ebruna:
+1

PR #1407, opened 2017-06-19
Homogenising the isolation/UE methods + fixing missing initialisations

e-masson:
Each isolation/UE method now follows the same scheme for clusters and tracks while testing the position of the current cluster/track with respect to the cluster of interest.
Two different radii are introduced (radiusClust and radiusTrack) instead of the single radius variable.
A condition is added on the cluster position in order to skip clusters outside the considered acceptance.
Several missing initialisations are added and reordered for clarity.

### DIFF
--- a/PWGGA/EMCALTasks/AliAnalysisTaskEMCALPhotonIsolation.cxx
+++ b/PWGGA/EMCALTasks/AliAnalysisTaskEMCALPhotonIsolation.cxx
@@ -1920,7 +1920,7 @@ void AliAnalysisTaskEMCALPhotonIsolation::EtIsoClusPhiBand(TLorentzVector c, Dou
     // AliParticleContainer *clusters = static_cast<AliParticleContainer*>(fParticleCollArray.At(1));
   AliClusterContainer *clusters = GetClusterContainer(0);
   Int_t localIndex = 0;
-  TLorentzVector nClust = 0x0;
+  TLorentzVector nClust(0x0);
   AliVCluster *candidate = 0x0;
   
   for(auto it : clusters->accepted()){ // Check the position of other clusters with respect to the trigger cluster
@@ -2103,8 +2103,8 @@ void AliAnalysisTaskEMCALPhotonIsolation::EtIsoClusEtaBand(TLorentzVector c, Dou
     // AliParticleContainer *clusters = static_cast<AliParticleContainer*>(fParticleCollArray.At(1));
   AliClusterContainer *clusters = GetClusterContainer(0);
   Int_t localIndex = 0;
+  TLorentzVector nClust(0x0);
   AliVCluster *candidate = 0x0;
-  TLorentzVector nClust = 0x0;
   
   for(auto it : clusters->accepted()){ // Check the position of other clusters with respect to the trigger cluster
     
@@ -2930,7 +2930,7 @@ void AliAnalysisTaskEMCALPhotonIsolation::FillInvMassHistograms(Bool_t iso, Doub
     else{
       localIndex++;
       
-      TLorentzVector nClust = 0x0;
+      TLorentzVector nClust(0x0);
       coi->GetMomentum(nClust,fVertex);
       
         // Float_t phiClust =nClust.Phi();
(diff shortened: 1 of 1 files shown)


dlodato:
Hi @e-masson, 
If you want to keep the changes at (now) lines 2082,2083,  please add (protection) initialisation commands (to zero) for etaClust and phiClust at line 2110 (or 2118). 
Same goes for all other places where similar changes to the code where made. 

Ultimately, why do you need 2 radiuses for tracks and clusters? 
I would prefer you change this back to how it was. 
Cheers, 
Davide

e-masson:
Hi @dlodato,

I don't understand why initialising etaClust and phiClust twice is necessary (because those variables are set to zero at (now) line 2083: phiClust = 0., etaClust = 0.). There were no such double initialisation before in EtIsoClusEtaBand(), so am I missing something?
Otherwise I defined two different radii for clarity and in order to prevent any mix between clusters and tracks. Anyway, if you think this is not useful I can restore the previous convention.

Best,
Erwann

dlodato:
Hi @e-masson, 
before the variables etaClust and phiClust were created and destroyed (local variables in the cycle) at each cycle, so no initialisation was needed. Now you are defining them outside of the cycle (local variables **in the method**, this means that they keep the value from the previous cycle. 

If (by any chance) there's some mis-propagation of information for some clusters, with this commit you would be affected (because you would not see it), with the previous version these situations can be better spotted. 
That's why I called them **protection** initialisations..
Alternatively, you can revert the whole commit. 

The part about the radius introduces the understanding (from the reader) of two different variables which do the same as the single previous one (radius), so why introducing them in the first place?
Cheers, 
Davide

e-masson:
Hi again @dlodato,

Ok I see now. I am going to fix this point and the radius part. I don't want to revert the whole commit because I reordered some parts to make the various isolation/UE methods coherent (concerning the cluster/track belonging or not to the cone/UE).
BTW, could you tell me why the cut on the cluster energy in each cluster loop in the isolation/UE methods is written with nClust.E() and not nClust.Et() or nClust.Pt()?

Thanks and best,
Erwann

dlodato:
+test

dlodato:
hi @e-masson , 

for the UE (as already said) I left Lucile taking care of it, so I wouldn't know.

For the Isolation , in a version of the code prior to your commit, I see (in EtIsoClusPhi/EtaBand)

`else if(radius<fIsoConeRadius && radius != 0.){ // The cluster is inside the isolation cone -> add the cluster pT to pT_iso
                                                    // Printf("Cluster Not Matched, Inside the Cone, with Energy %.4lf",eTcluster);
      sumEnergyConeClus +=nClust.Pt();`

I remember before being `nClust.E() * TMath::Sin(nClust.Theta())`  which was giving E_T..

Where did you see this?

Cheers, 
Davide

PS i am +1ing your PR.

dlodato:
+1

e-masson:
Hi @dlodato,

Ok, but I think this `nClust.E()` cut also affects isolation, and as we cut on p_T for tracks, I would be in favour of replacing `if(nClust.E()<0.3)` by `if(nClust.Et()<0.3)` for consistency (BTW each cut applied on cluster/track energy at the correction task(s) level involves E_T or p_T, but rarely E).

For the code snippet you quote here, do you mean that `nClust.Pt()` was `nClust.E() * TMath::Sin(nClust.Theta())` in the past? As far as I remember it already was `nClust.Pt()` when I began to use the task... However, do we agree on the fact that, for a cluster, the `Et()` and `Pt()` functions give the same results?

Thanks for the +1!
Best,
Erwann

dlodato:
Hi @e-masson, 
I see what you mean, now...
this cut was placed in this way to compare **as much as possible** 1 to 1 with the results obtained with CaloXXX Framework (so I think that also in the Calo Framework nClust.E()<0.3 is used). 
This is not a bad choice (IMO) as we measure the clusters energy, and we reject those clusters with a reconstructed Energy smaller than 0.3 GeV/c (I think this is because the Energy non-linearity correction stops around this value).. 

your statement is a bit too generic. 
let's move it to a private exchange ... 
Send me an email with your points, if you want ... 
Cheers, 
Davide

dlodato:
+1

PR #1444, opened 2017-06-20
PWGGA/GammaConv: Fix constant-correctness

mfasDa:
- Make arguments in constructor which are not changed by the constructor const
- Make functions which just return values const


FriederikeBock:
+test

FriederikeBock:
+1

PR #1458, opened 2017-06-21
Implementation of hit coordinates in TOF calibration task

jezwilkinson:
Adds option to save the delta-X and delta-Z coordinates in a lighter version of the output tree (in discussion with @preghenella), for studies to check calibration as function of position of hit in the pad. Also fixes handling of streamed pointers/arrays in header/destructor, which will potentially help protect against possible memory leaks.
Mode to save coordinates (and therefore tree with some removed branches for optimisation) is handled by an added flag in the AddTask macro; default behaviour is unaffected.


preghenella:
+1

PR #1473, opened 2017-06-21
Adding pt weights for pp at 13 TeV (Susanna)

afestant:
Adding method to compute pt weights for pp at 13 TeV (Susanna)


ebruna:
+1

PR #1486, opened 2017-06-22
correspondence analysis separated for identified deuteron, proton, and un-id hadron

bschaefe:
For improved efficiency, separate wagons will be used for different species.  At present these include deuterons, protons, and hadrons.

### DIFF
--- a/PWGCF/Correlations/DPhi/AliAnalysisTaskCorPIDTOFQA.cxx
+++ b/PWGCF/Correlations/DPhi/AliAnalysisTaskCorPIDTOFQA.cxx
@@ -75,7 +75,7 @@
 
 using namespace std;            // std namespace: so you can do things like 'cout'
 
-using namespace BSchaefer_devel;
+//using namespace BSchaefer_devel;
 
 //ofstream fout;
 Int_t NEvents            = 0;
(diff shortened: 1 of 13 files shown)


aohlson:
Hi @bschaefe , 
Before approving the commit, I wanted to ask if it is necessary to have separate analysis tasks for the different particle species (I don't know the details of your code or analysis, so maybe the answer is yes).  Repeating code is more likely to introduce bugs (for example, it's easy to change something in one class but forget to change it in the other nearly-identical class), and so it's good coding practice to have the functionality change with a switch, rather than having large amounts of copied-and-pasted code in different classes or functions.  Also, if possible, we would also like to try to avoid unnecessary proliferation of classes with very similar functionality.  
Thanks, 
Alice

bschaefe:
splitting the code 3-ways (at present) is necessary because I intend to run the three simultaneously.  The same tasks could be combined into a single file, however it would add the complexity of multiple track loops.  Also, I intend to soon retire the previously used code from the repository.  Of course the trade-off is 3 classes versus 1 long class.  Future use will be optimal as only one species will typically be run at a time.

aohlson:
+1

PR #1492, opened 2017-06-22
Perform event selection in an independent task

cbourjau:
The event selection task runs befor the analysis task and exposes
methods that yield the selected tracks. All QA histograms etc are
created in the event selection task keeping the actual analysis clean.


aohlson:
+test

aohlson:
+1

PR #1501, opened 2017-06-23
changed conflicting output names to prevent simultaneous run conflict

bschaefe:


aohlson:
+1

akubera:
I will re-test the trains after this, thanks!

PR #1511, opened 2017-06-23
allow user to configure exotic cluster options

jdmulligan:
Tested locally to verify that default behavior is unchanged.


jdmulligan:
@raymondEhlers

PR #1523, opened 2017-06-25
D0 systematics for 010 and 60100 in pPb 2016 added, HFPtSpectrum, HFPtSpectrumRAA updated with TpA

cterrevo:


zconesa:
Ciao Cristina,
On AliHFSystErr, when you want to call the function for the 0-10 and 60-100% systematics (line 159 onwards), wouldn't it better to define from the start an option for the 2016 data (if clause?) ? As you left it now, it may be initializing the histograms and numbers twice, which I would avoid. Also, later you might want to study the other centralities, so it might be better for the long term?
My two cents,
Zaida

zconesa:
+test

zconesa:
Maybe you can try something like this on line 159 ? 
I am just unsure the 60-100 histos will be filled well as you implemented them. Did you look at them?

if (RunNumber==16 || fRunNumber==2016){
	if (fCentralityClass=="010ZNA") InitD0toKpi2016pPb010ZNA();
	else if (fCentralityClass=="60100ZNA") InitD0toKpi2016pPb60100ZNA();
} else {
… // previous 2013 functions vs centrality 
}

zconesa:
+1

PR #1535, opened 2017-06-26
Changeback systunc Dstar

amveen:
Dstar systematic now includes only effect d0 improver compared to old improver (all bins but 16-24 are set back to old cutvariation value)


arossi81:
+1

PR #1536, opened 2017-06-26
Hotfix namespacestd

mfasDa:
2 commits:
- [Trivial] Remove "using namespace std" from several header files: The using directive is not needed within the header files and might lead to unwanted effects
- Add stl iterator for AliV0ReaderV1: This allows range-based iterations and stl algorithms (in case operator< and operator== will be defined). This commit just adds functions and does not modify existing ones.
- Make getter functions which do not modify objects const in AliV0ReaderV1: This is needed for the iterator commit as the iterator works on a const AliV0ReaderV1


FriederikeBock:
+1 
Thanks for the implementation.

Issue #1507, opened 2017-06-23
~25-30 % of time spend in AliPhysicsSelection::EvaluateTriggerLogic when analyzing LHC15o AODs


cbourjau:
Hello,

I noticed today that my analysis spends a lot of time in AliPhysicsSelection::EvaluateTriggerLogic. I am running on the LHC15o AODs and have the AliPhysicsSelection task added. 
To be more precise, the vast majority of the time is spend in TFormula::Compile(). Now I don't know exactly what this piece of code is doing, but that surely sounds like a lot of CPU power to me. Even if the logic were to be unaltered, I think we can cut the time here in half. Namely, TFormuar::Compile is called in the constructor, but its later called again to check the result [(link)](https://github.com/alisw/AliPhysics/blob/da28d5c678c8d50e1381b23ad4b13b74c297162f/OADB/AliPhysicsSelection.cxx#L302-L305)

As far as I can see (on first glance), the second call to Compile only returns not 0 if ndim ==0 [(link)](https://root.cern.ch/doc/master/TFormula__v5_8cxx_source.html#l02289). So one could just call GetNdims() directly, saving the costly compilation.

Might this be worth a consideration?

Attached is a "FlameGraph" showing the issue

[8224.pdf](https://github.com/alisw/AliPhysics/files/1098206/8224.pdf)

ekryshen:
Hi Christian,

Thanks for reporting this issue. I am not sure if this piece of code can be easily optimized because it is called for every event and the logic of the formula may be different event by event. This part was actually introduced by Michele/Jan Fiete. I will have a look but probably not earlier than in two weeks from now.

Cheers,
Evgeny

cbourjau:
Hi Evgeny,

I created a PR fixing this issue to the best of my knowledge. It certainly needs an expert's sets of eyes to take a look: #1537

Cheers,
Christian

PR #1537, opened 2017-06-26
Trigger logic optimization

cbourjau:
*This PR needs more vetting before it may be merged!*
============================================

Background to this PR
----------------------------
I noticed last week that  AliPhysicsSelection::EvaluateTriggerLogic was one of the main consumers of CPU cycles when running analyses (see issue #1507). This PR is an alternative implementation for this function. The PR commits are structured as follows: 

1st: Adds comments to the old code; This is the understanding I had of the previous code when writing the new one

2nd: The new implementation is in place, but the old one was not touched. The results of the two implementations are compared and AliFatal is called should they be different. I tested the validity of the new implementation with this commit on LHC15o and LHC16k AODs. The behaviour of the new and old implementation was identical.

3rd: Removal of the old code and a bit of clean up. I ran the profiling on this commit. The Speedup was probably several orders of magnitude. The probabilistic profile can no longer pick up this function; compared to ~35% of the time spend within this function using the old implementation under otherwise identical conditions.

Summary of new implementation
------------------------------------------
The new Logic does not rely on the _very_ slow TFormula. Instead it
does the job by creating a vector of booleans; one bool for each token
in the trigger logic (e.g. "V0A"). At the end, the final trigger
decission is true if all elements in that vector were true (taking
possible negations ("!") into account. If one or more are false, this
logic yields false.

I tried this new implemenation on LHC15o and LHC16k AODs without
issues and the results were identical to the old implementation. The
Efficiency gain of this new implemntation is several orders of
magnitude. This code is called many times per event and seems to
be responible for most of the CPU consumption. See issue #1507

### DIFF
--- a/OADB/AliPhysicsSelection.cxx
+++ b/OADB/AliPhysicsSelection.cxx
@@ -68,6 +68,7 @@
 #include <algorithm>
 #include <vector>
 #include <iterator>
+#include <regex>
 
 #include <Riostream.h>
 #include <TH1F.h>
@@ -131,18 +132,20 @@ fPSOADB(0),
 fFillOADB(0),
 fTriggerOADB(0),
 fRegexp(new TPRegexp("([[:alpha:]]\\w*)")),
-fCashedTokens(NULL)
+fCashedTokens(NULL),
+fTriggerToFormula()
 {
   // constructor
   fCollTrigClasses.SetOwner(1);
   fBGTrigClasses.SetOwner(1);
   fTriggerAnalysis.SetOwner(1);
   fHistList.SetOwner(1);
+  fTriggerToFormula = new StringToTFormula();
   
   AliLog::SetClassDebugLevel("AliPhysicsSelection", AliLog::kWarning);
 }
 
- AliPhysicsSelection::AliPhysicsSelection(const char *name) :
+AliPhysicsSelection::AliPhysicsSelection(const char *name) :
  AliAnalysisCuts("AliPhysicsSelection", "AliPhysicsSelection"),
  fPassName(""),
  fCurrentRun(-1),
@@ -161,14 +164,15 @@ fCashedTokens(NULL)
  fFillOADB(0),
  fTriggerOADB(0),
  fRegexp(new TPRegexp("([[:alpha:]]\\w*)")),
- fCashedTokens(NULL)
+ fCashedTokens(NULL),
+ fTriggerToFormula()
  {
    // constructor
    fCollTrigClasses.SetOwner(1);
    fBGTrigClasses.SetOwner(1);
    fTriggerAnalysis.SetOwner(1);
    fHistList.SetOwner(1);
-
+   fTriggerToFormula = new StringToTFormula();
    AliLog::SetClassDebugLevel("AliPhysicsSelection", AliLog::kWarning);
  }
 
@@ -178,6 +182,7 @@ AliPhysicsSelection::~AliPhysicsSelection(){
   if (fTriggerOADB)  delete fTriggerOADB;
   delete fRegexp;
   delete fCashedTokens;
+  delete fTriggerToFormula;
 }
 
 UInt_t AliPhysicsSelection::CheckTriggerClass(const AliVEvent* event, const char* trigger, Int_t& triggerLogic) const {
@@ -254,63 +259,84 @@ UInt_t AliPhysicsSelection::CheckTriggerClass(const AliVEvent* event, const char
   return returnCode;
 }
 
-//______________________________________________________________________________
-Bool_t AliPhysicsSelection::EvaluateTriggerLogic(const AliVEvent* event, AliTriggerAnalysis* triggerAnalysis, const char* triggerLogic, Bool_t offline){
-  // evaluates trigger logic. If called with no event pointer/triggerAnalysis pointer, it just caches the tokens
-  // Fills the statistics histogram, if booked at row i
+/// Evaluate if the given event fulfills a given trigger logic
+///
+/// \param event Pointer to the current event
+/// \param triggerAnalysis Pointer to the TriggerAnlysis class
+/// \param triggerLogic Describing trigger logic; e.g. "V0A && V0C && ZDCTime && !TPCHVdip"
+/// \param offline Offline analysis(?)
+///
+/// \return True if the given event matches the trigger logic
+Bool_t AliPhysicsSelection::EvaluateTriggerLogic(const AliVEvent* event,
+						 AliTriggerAnalysis* triggerAnalysis,
+						 const char* triggerLogic, Bool_t offline){
+  using namespace std;
+  // Helper function to create a TFormula from the current trigger string
+  auto make_formula =
+    [this, triggerLogic](const char* name, const char* trigger) {
+#if ROOT_VERSION_CODE >= ROOT_VERSION(6,3,0)
+      // In case of ROOT6 it is necessary to stay for the moment with the v5 version of TFormula
+      // as the v6 version produces a large amount of warnings at runtime.
+      ROOT::v5::TFormula formula(name, trigger);
+#else
+      TFormula formula(name, trigger);
+#endif
+      if (formula.Compile() > 0)
+	AliFatal(Form("Could not evaluate trigger logic %s (evaluated to %s)", triggerLogic, trigger));
+      return formula;
+    };
+
+  // split the trigger logic on the individual trigger names and replace them
+  // with TFormula parameters
+  string trg_logic_orig(triggerLogic);
+  // Do we have a TFormula for this trigger already?
+  // Regex identifying the trigger (>=1 alphanumeric and _)
+  regex trg_re("\\w+");
+  if (fTriggerToFormula->count(trg_logic_orig) == 0) {
+    string trg_logic_formated;
+    // Create an iterator on all the _not_ mached things in the trigger string
+    sregex_token_iterator iter(trg_logic_orig.begin(), trg_logic_orig.end(), trg_re, -1);
+    sregex_token_iterator end;
+    Int_t nparam = 0;
+    while (iter != end) {
+      trg_logic_formated.append(*iter);
+      trg_logic_formated.append(Form("int([%i])", nparam));
+      iter++;
+      nparam++;
+    }
+    // Insert this new TFormula to the hashmap
+    fTriggerToFormula->insert({trg_logic_orig,
+			       make_formula(Form("dummy_name_%zu", fTriggerToFormula->size()),
+					    trg_logic_formated.c_str())});
+  }
+  // Get the values for each individual trigger in the trigger logic string;
+  // These values are the parameters of the TFormula
+  regex_iterator<string::iterator> rit (trg_logic_orig.begin(), trg_logic_orig.end(), trg_re);
+  regex_iterator<string::iterator> rend;
+  vector<Double_t> paras;
+  while (rit!=rend) {
+    auto bit = triggerAnalysis->FindTriggerBit(rit->str().c_str());
+    if (offline) {
+      typedef AliTriggerAnalysis::Trigger Trigger;
+      bit = static_cast<Trigger>(bit | AliTriggerAnalysis::kOfflineFlag);
+    }
+    paras.push_back(triggerAnalysis->EvaluateTrigger(event, bit));
+    ++rit;
+  }
+  // Get the TFormula for this trigger logic, set the parameters and
+  // evaluate it std::map::at throws an exception if the trigger
+  // string is not found, since it should be present at this point in
+  // the function
+  auto trg_formula = fTriggerToFormula->at(trg_logic_orig);
+  trg_formula.SetParameters(&paras[0]);
+  Bool_t new_result = trg_formula.Eval(0);
+
+  // Old code
+  // evaluates trigger logic. If called with no event
+  // pointer/triggerAnalysis pointer, it just caches the tokens Fills
+  // the statistics histogram, if booked at row i
   TString trigger(triggerLogic);
 
-  // add space after each token (to use ReplaceAll later); AFAIKT, the
-  // regex matches on 0 or more \w, though?
-  // Eg: Before: "V0A && V0C && ZDCTime && !TPCHVdip"
-  //      After: "V0A && V0C && ZDCTime && !TPCHVdip "
-  fRegexp->Substitute(trigger, "$1 ", "g");
-  // Split the trigger string at the `&`; Note: we still have to strip whitespaces
-  TObjArray* dirty_tokens = trigger.Tokenize("&&");
-  // collect all the trigger results in this vector which we can later
-  // fold to see what the final decission is
-  std::vector<bool> trigger_results;
-  for (auto s: *dirty_tokens) {
-    // Strip whitespace on both sides
-    TString cleaned_token = TString(static_cast<TObjString*>(s)->GetString().Strip(TString::kBoth));
-
-    // Strip the (possibly leading "!" to check if its a valid token
-    TString pure_token_name = TString(cleaned_token.Strip(TString::kLeading, '!'));
-
-    // Get the `bit` value for this triggera / check if this token is valid
-    // Maybe this would be better solved with a proper hashmap?
-    TParameter<Int_t>* param =
-      dynamic_cast<TParameter<Int_t> *>(fCashedTokens->FindObject(pure_token_name));
-    if (!param) {
-      TInterpreter::EErrorCode error;
-      Int_t bit = gInterpreter->
-	ProcessLine(Form("AliTriggerAnalysis::k%s;", pure_token_name.Data()), &error);
-
-      if (error > 0) {
-	AliFatal(Form("Trigger token `%s` unknown", pure_token_name.Data()));
-      }
-      // This might leak memory!
-      param = new TParameter<Int_t>(pure_token_name, bit);
-      fCashedTokens->Add(param);
-      AliDebug(AliLog::kDebug, "Added token");
-    }
-    // Why the implicit conversion from Int_t to Long64_t here?! This
-    // is copy/pasted from the previous version. Bits are scarry...
-    Long64_t bit = param->GetVal();
-    if (offline) {
-      bit |= AliTriggerAnalysis::kOfflineFlag;
-    }
-    if(event && triggerAnalysis) {
-      // Replace the current trigger name with 0 or 1, base on AliTriggerAnalysis::EvaluateTrigger
-      Bool_t this_trig_res = triggerAnalysis->EvaluateTrigger(event, (AliTriggerAnalysis::Trigger) bit);
-      Bool_t was_negated = cleaned_token.BeginsWith('!');
-      if (was_negated) {
-	trigger_results.push_back(!this_trig_res);
-      } else {
-	trigger_results.push_back(this_trig_res);
-      }
-    }
-  }
   while (1) {
     AliDebug(AliLog::kDebug, trigger.Data());
     
@@ -346,27 +372,13 @@ Bool_t AliPhysicsSelection::EvaluateTriggerLogic(const AliVEvent* event, AliTrig
       trigger.ReplaceAll(token, Form("%d", triggerAnalysis->EvaluateTrigger(event, (AliTriggerAnalysis::Trigger) bit)));
     }
   }
-
-  // The final modified trigger string now might look something like: "1 && 1 && 1 && !0"
-
-#if ROOT_VERSION_CODE >= ROOT_VERSION(6,3,0)
-  // In case of ROOT6 it is necessary to stay for the moment with the v5 version of TFormula
-  // as the v6 version produces a large amount of warnings at runtime.
-  ROOT::v5::TFormula formula("formula",trigger);
-#else
-  TFormula formula("formula", trigger);
-#endif
-  if (formula.Compile() > 0)
-    AliFatal(Form("Could not evaluate trigger logic %s (evaluated to %s)", triggerLogic, trigger.Data()));
   // Check if all constituents of the trigger string evaluate to
   // `true`; the "function" is constant, so the value in `Eval` does
   // not matter
-  Bool_t result = formula.Eval(0);
-  // Fold the vector with the individual trigger results by requiring all to be true
-  Bool_t new_result
-    = std::all_of(trigger_results.begin(), trigger_results.end(), [](Bool_t t){return t == true;});
+  Bool_t result = make_formula("name", trigger).Eval(0);
+
   if (result != new_result) {
-    AliFatal("Old and new method do not aggree!");
+    AliFatal(Form("Old and new method do not aggree! Trigger logic: %s", triggerLogic));
   }
   AliDebug(AliLog::kDebug, Form("%s --> %d", trigger.Data(), result));
   return result;
--- a/OADB/AliPhysicsSelection.cxx
+++ b/OADB/AliPhysicsSelection.cxx
@@ -81,10 +81,15 @@
 #include <TParameter.h>
 #include <TInterpreter.h>
 
+// In case of ROOT6 it is necessary to stay for the moment with the v5
+// version of TFormula as the v6 version produces a large amount of
+// warnings at runtime.
 #if ROOT_VERSION_CODE >= ROOT_VERSION(6,3,0)
 #include <v5/TFormula.h>
+typedef ROOT::v5::TFormula R5TFormula;
 #else
 #include <TFormula.h>
+typedef TFormula R5TFormula;
 #endif
 
 #include "AliPhysicsSelection.h"
@@ -111,6 +116,7 @@
 #include "AliVZEROTriggerData.h"
 #include "AliITSOnlineCalibrationSPDhandler.h"
 #include "AliITSTriggerConditions.h"
+
 ClassImp(AliPhysicsSelection)
 
 AliPhysicsSelection::AliPhysicsSelection() :
@@ -131,8 +137,6 @@ fHistStat(0),
 fPSOADB(0),
 fFillOADB(0),
 fTriggerOADB(0),
-fRegexp(new TPRegexp("([[:alpha:]]\\w*)")),
-fCashedTokens(NULL),
 fTriggerToFormula()
 {
   // constructor
@@ -163,8 +167,6 @@ AliPhysicsSelection::AliPhysicsSelection(const char *name) :
  fPSOADB(0),
  fFillOADB(0),
  fTriggerOADB(0),
- fRegexp(new TPRegexp("([[:alpha:]]\\w*)")),
- fCashedTokens(NULL),
  fTriggerToFormula()
  {
    // constructor
@@ -180,8 +182,6 @@ AliPhysicsSelection::~AliPhysicsSelection(){
   if (fPSOADB)       delete fPSOADB;
   if (fFillOADB)     delete fFillOADB;
   if (fTriggerOADB)  delete fTriggerOADB;
-  delete fRegexp;
-  delete fCashedTokens;
   delete fTriggerToFormula;
 }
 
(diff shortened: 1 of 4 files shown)


cbourjau:
Profiling flamegraph of previous implementation:

cbourjau:
Profiling flamegraph of this new implementation:

BerserkerTroll:
There are expressions, more complex than negation and conjunction e.g. `"(SPDGFO >= 1 || V0A || V0C) && !V0ABG && !V0CBG && ZDCTime && !TPCHVdip"` [1]

[1] https://github.com/alisw/AliPhysics/blob/86fe0202e97546eec734c022c0a42f606aece353/OADB/macros/BrowseAndFillPhysicsSelectionOADB.C#L209

cbourjau:
@BerserkerTroll Would have been too easy otherwise ;)

The reason I'm getting into this, is that I am running out of CPU quota when I run on the LHC15o PbPb sample. In my quest to optimize my analysis a bit I bumped into this issue. 

Given that this function seems to be such a hotspot I still wonder how one might optimize this in the light of the more complicated trigger strings. Please correct me if I am wrong, but it seems that such trigger strings are quite rare at least in the LHC16l and LHC15o sample? If so, it might be worth having the TFormular base solution as a slow fall-back whenever the string is more than just a chain of conjunctions?

This PR will currently call AliFatal whenever it encounters such complicated tirgger strings (It splits the string on the "&&" but will then fail to fined whatever chunks it split it into within AliTriggerAnalysis::Trigger.

Am I missing something obvious, maybe?

ekryshen:
Hi Christian, 

Many thanks for your efforts! Your solution looks natural indeed but I think we should try to cover the cases with complicated trigger strings, like (V0A | V0C). I thought about the following solution, still based on TFormula. 
* At the initialization step (constructor/init) create one TFormula per trigger alias such that those tokens are replaced with parameters, e.g. "(V0A || V0C) && !SPDClsVsTrkBG" would become "([0] || [1]) && ![2]", etc. This formula can be precompiled once in the init function. No need to compile it for every event, no need to convert TString with tokens into the string of "0"s and "1"s. 
* Evaluate those formulas in EvaluateTriggerLogic by changing those [0]...[n] parameters event by event. This should significantly speed up the function, probably similar to your implementation. The only doubt I have is that those parameters should be interpreted in TFormula as integers, not floats, otherwise this implementation will fail. May be foresee on-the-fly conversion in the formula string, like "Int_t([0]) || Int_t([1]) && !Int_t([2])".

Since you already went deep into this subject, do you think you could try this approach?

Thanks a lot, cheers,
Evgeny

BerserkerTroll:
I suggest to turn `triggerLogic` string into ordinary C++ function, compile, load and call it.
Of course, `triggerLogic <-> function pointer` relation should be cached in the hash map. See here https://gist.github.com/BerserkerTroll/124693b614743fb25f11bbf81957e484

cbourjau:
I guess both solutions are very similar in their core. To me, though, the concept of generating code to be compiled by a script sounds quite scary (even thought the gist is kind of cool). I am out of the office myself for the next week, though, so I don't know if I will have time to take a stab on Evgeny's solution in the next few days

hcab14:
Hi Christian, Evgeny, all,

what about memoizing the return value of the TFormula? This could be a
non-invasive and flexible approach.

Cheers
Christoph

BerserkerTroll:
> generating code to be compiled by a script

Code compiled by a compiler, not a script.

cbourjau:
@hcab14 I think this might be what Evgeny was suggesting? Or I might be misunderstanding one of you two. I understood that we have a `std::map` member variable to cache the `TFormula`s. This is my understanding without checking the syntax:

```cpp
// Have this member in AliPhysicsSelection
std::map<std::string, TFormula> fTriggerFormulaCache;
//....
// in EvaluateTriggerLogic
std::string trigger("(V0A || V0C) && !SPDClsVsTrkBG");
TFormula cacheMe("SomeSillyName", "Int_t([0]) || Int_t([1]) && !Int_t([2])");
fTriggerFormulaCache.insert({trigger, cacheMe});
// Use the formular again later if we see the same trigger again
```
I won't get around to take a stab at this before next week, though.

@BerserkerTroll Yes, sorry for the hasty and sloppy language. Point and case is that its automatically generated code and will have to be called through the interpreter at run-time since we won't have the header files at compile time. Or am I getting this wrong?

BerserkerTroll:
@cbourjau the code doesn't go through the interpreter. The code is written into the `.cxx` file and compiled into shared object — machine code, in contrast to `TFormula`, which is compiled into bytecode and interpreted [1].
Then the `.so` is (automatically) loaded and the only call to `gInterpreter` is necessary to get the address of the function from the shared object. It is a cross-platform version of `dlsym` [2].
The header files `AliVEvent.h` and `AliTriggerAnalysis.h` are available in binary `AliRoot`/`AliPhysics` distribution.

The only unpleasant thing here is a couple of temporary `.so`-files. Unfortunately, in ROOT 5 it is impossible to generate machine code in memory, without external files. It will be possible when `AliRoot` completely ported to ROOT 6.

[1] https://github.com/alisw/root/blob/a089153cd7e0d5c56f7ffa52fbed6841ce83e178/hist/hist/src/TFormula_v5.cxx#L2700
[2] http://pubs.opengroup.org/onlinepubs/009695399/functions/dlsym.html

cbourjau:
Thanks to @BerserkerTroll for the clarification. I now updated this PR with a TFormula based solution, though. If performance issues remain, I think your solution would be the one to go for.

The current approach which I implemented now is what Evgeny recommended. The trigger logic string is changed such that the triggers are parameters in a TFormula. The TFormula is only created once per trigger logic and than cached in a hasmap. I suppose that this should solve all issues around the call to Compile(), which was the entire reason for this exercise. I did not profile the new solution yet, though. I can do that Thursdays or Fridays. Also, commit 58032fd contains the original code and the new solution and crashes if the two disagree. I tested only on LHC16l pass1 AODs, yet, but there all was good.

OADB/AliPhysicsSelection.cxx
```diff
@@ -157,24 +167,22 @@ fCashedTokens(NULL)
...
    fBGTrigClasses.SetOwner(1);
    fTriggerAnalysis.SetOwner(1);
    fHistList.SetOwner(1);
-
+   fTriggerToFormula = new StringToTFormula();
    AliLog::SetClassDebugLevel("AliPhysicsSelection", AliLog::kWarning);
  }
 
 AliPhysicsSelection::~AliPhysicsSelection(){
   if (fPSOADB)       delete fPSOADB;
```
BerserkerTroll: Since you already modifying code around here, I think this might be a great opportunity to remove these unnecessary null-pointer checks.

BerserkerTroll:
@cbourjau I think I'll never understand this. I bet my solution will take like 30-40 LOC and it is obviously faster than what you've implemented. But you've chosen to write 160 lines of more complex code instead.
Even if the difference between performances these two solutions will be small against the background of `TTree::GenEntry`, I don't understand why use more complex and slow code instead of simple and fast one.

cbourjau:
@BerserkerTroll I am not claiming any authority in this case. Clearly both proposed solutions have their downside. About half of the LOC in this PR are the set up of the hashmap between trigger-strings and their respective enum, in order to be able to do this mapping fast and without the need of dropping down to the interpreter; obviously not necessary in your solution but not exactly complicated either.

To me, the advantage of using TFormula is that it is well tested and rather future proof. I lack the insight into the inner workings of ROOT to be sure about those things wrt to compiling and loading functions at runtime. Is it possible to write tmp files on all platforms? Are there potential ABI compatibility issues on the grid? Which header files are chosen if one were to use `par` files for AliPhysicsSelection? Could we run into a race condition where two nodes write to a tmp file of the same name? Maybe all these questions are very obvious to some people, but I preferred to take the route of least surprise (to me). I hope that cleared up my reasoning in favour of TFormula on my side a bit!

Cheers,
Christian

BerserkerTroll:
@cbourjau please don't treat the following as a promotion of my solution, I just want to answer to some questions.

>  Are there potential ABI compatibility issues on the grid? 

No doubt you know that when you are preparing your task to run in grid, your macro file contains something like `plugin->SetAnalysisSource("AliAnalysisYourTaskName.cxx");`. Then you get a bunch of .sh, .jdl and .C files generated and named after your taskname; These files are uploaded to the grid. In the file named `<your_taskname>.sh` there is a line like `root -b -q -x <your_taskname>.C` and in that macro there is a line `gROOT->ProcessLine(".L AliAnalysisYourTaskName.cxx+g");`. I'm not ready to bet that `CompileMacro("file.cxx")` and `ProcessLine(".L file.cxx+")` has absolutely the same effect, but I think there is no difference. Both are compiling (if necessary) and loading the shared object into the `root`. If you think that these functions might have different behaviour, we may just use `ProcessLine` instead of `CompileMacro` in my solution.
So, why are you worrying that there might be ABI compatibility issues with compiling and loading generated files and not worrying about the issuses with the task source? They are both compiled on every node, in the same environment: same versions of AliPhysics, AliRoot, ROOT, the same compiler etc. I don't see how it is possible to step into ABI issues with generated files and not to step with the user task source code at the same time.

> Is it possible to write tmp files on all platforms?

I'm not writing files to some special destination for tmp files like `/temp`, I'm writing files to the current working directory. If it was possible to compile user task (i.e. create and write to `.o` and `.so` and other ACLiC-generated `.h` and `.cxx` files) I don't see why it won't be possible to create other files.

> Which header files are chosen if one were to use par files for AliPhysicsSelection?

I think the answer is "the same as for user analysis task source code".

> Could we run into a race condition where two nodes write to a tmp file of the same name?

No doubt we may run into a race if we write to a file with the same name. So, just don't create files with the same name! Simple as that. As you might have noticed, in my macro I've used a counter to name the temporary files. It will be enough to make it `atomic` to safely use from multiple threads and get unique file names.

cbourjau:
Ok, this PR really turned out to be more work than I anticipated. For the TFormula based solution, it is necessary to tokenize the trigger logic by regexes. Since I did not find this functionality within TPRegex, I went for the cxx11 regex implementation. However, it turns out that gcc < 4.9 has a very much incomplete regex library, which is a show stopper. I therefore when ahead and used @BerserkerTroll approach of compiling things on the fly, as proper cpp functions. This really seems to be the way to go and I have to say that it indeed seems to be the better solution. Thanks for the further explanation. The current head of this PR remove all the old code, and is what I used for the below flame graph. Note that it is base on a stocastical profiler of 100 samples. The code still spends 12% of the time compiling the trigger logic, but this number is only so high because I profiled from the very start. All those hits in `EvaluateTriggerLogic` were indeed created in the very beginning of the execution and I am sure that this implementation does solve the initial problem of this PR.

The second to last commit is again the original code and the new implementation side by side. The code crashes if the old and new results differ. All was find for LHC15o AODs.

I think the code should now be ready for prime time, but given its crucial nature, a second set of eyes should really check it. I think it would be really good if we could get this code online as soon as possible, given how much CPU power the current implementation wastes. This PR got a bit messy in the progress. Is this a problem?

[4402.pdf](https://github.com/alisw/AliPhysics/files/1127605/4402.pdf)

BerserkerTroll:
@cbourjau I think we need more than a second sets of eyes to check it, especially from one who definitely knows how files are distributed across nodes in the grid to verify that I've not lied much when explained how it works. =)

How many events you are processing when you profile?

cbourjau:
@BerserkerTroll About 2300 events. Are you worried that the profile is inaccurate or if there are some exotic trigger strings hiding behind the low statistics? Who would be the right person to ask? @dberzano , just because he seems to always be the person to ask ;)?

BerserkerTroll:
@cbourjau Yes, I think the profile is inaccurete because in the grid, if I understand correctly, the minimal amount of work for a node is a run and a node processes all files of a run without recompilation. A run usually contains much more than 2300 events, so I expect that the fraction of time spent in `CompileMacro` will be much less than 12%.

If you have time, could you please add another 1000 or so of events to your local analysis to see if there is a trend in how much time spent inside `CompileMacro`? I think it would be enough to just duplicate the AODs you already have under another names.

OADB/AliPhysicsSelection.cxx
```diff
@@ -851,3 +800,40 @@ void AliPhysicsSelection::DetectPassName(){
...
+
+  TString trigger(triggerLogic);
+
+  fRegexp->Substitute(trigger, "triggerAnalysis->EvaluateTrigger(event, (AliTriggerAnalysis::Trigger) (AliTriggerAnalysis::k$1 | (offline ? AliTriggerAnalysis::kOfflineFlag : 0)) )", "g");
+
+  TString tmpfilenam = "__trig_tmpfile";
+  tmpfilenam += counter++;
+  TString fnam = tmpfilenam;
+  tmpfilenam += ".cxx";
+  FILE* tmpfile = fopen(tmpfilenam, "w");
```
BerserkerTroll: Could you add a check for `tmpfile` value?

BerserkerTroll:
@cbourjau on the other hand, each job is also split into subjobs. Each subjob is running on its own node and I think the user task gets compiled per subjob. It seems that split into subjobs is performed basing on the number of files in the run
```
# Maximum number of input files to be processed per subjob
SplitMaxInputFileNumber = "100";
```
(from `<taskname>.jdl`)
If the number of events per subjob is small, than the overhead of `CompileMacro` will be high, so it would be better to write own parsing function (should not be difficult, because the regex is pretty simple) and go back to `TFormula`, because `TFormula::Compile` + bytecode interpretation might be faster than compilation into machine code in this situation.

cbourjau:
@BerserkerTroll Thanks for the thorough review! I'll comment on the code-comments directly. Just about your thoughts on how often your code is compiled. Yes, my understanding is that each job (i.e., slave-job, not master job) is pretty much completely independent from each other. So each of these jobs would need to compile each trigger-logic individually. Each of these jobs probably has between 1 and 50 files. All the bench-marks I showed above were on one local file. So this new solution won't be slower than the current one, even in that case. Depending on what is being done, this might be different for ESDs where we have considerably fewer events per file and job. Then again, its ESDs and the IO related part will be much larger as well - premature optimization is the root of all evil ;). Never the less, the same reasoning applies to TFormula as well: Each node would need to create its own triggerLogic-TFormula map as well. But I suppose you are right that TFormula is faster to compile.

Note that the TFormula solution as done in one of the previous commits used the c++ regex engine which turned out to be a dead end in terms of compiler compatibility. So unless I am missing a way to tokenize the trigger logic  directly with TPRegex, the regex solution will turn more ugly if chosen.

About profiling this with more events. Maybe I'll have time to do that tomorrow, but I was checking the sats while they came in for the latest flame-graph. All hits in the `EvaluateTriggerLogic` function were at the very beginning. I am certain that those 12% are an upper limit.

OADB/AliPhysicsSelection.cxx
```diff
@@ -253,61 +254,13 @@ UInt_t AliPhysicsSelection::CheckTriggerClass(const AliVEvent* event, const char
...
-    AliDebug(AliLog::kDebug, Form("Tok %d %d %s %lld", pos[0], pos[1], token.Data(), bit));
-    
-    if (offline) 
-      bit |= AliTriggerAnalysis::kOfflineFlag;
-    
-    if(event && triggerAnalysis) {
-      trigger.ReplaceAll(token, Form("%d", triggerAnalysis->EvaluateTrigger(event, (AliTriggerAnalysis::Trigger) bit)));
-    }
+  // Do we already have a function for this trigger logic? Add to hashmap if not
+  if (fTriggerToFunction->count(triggerLogic) == 0) {
```
cbourjau: Thats a neat way! I'll fix this if we go this route! I just relied on hashmap look-ups to be dirt cheap anyways.

BerserkerTroll: Please also look at outdated comments. There I'm also pointing on that double-querying is suboptimal, so this route is applicable to the previous set of commits.

And `std::map` is not a hashmap, it is an ordered container. The `std::map` relies on `opetator<`, not `std::hash`.

OADB/AliPhysicsSelection.cxx
```diff
@@ -251,63 +259,66 @@ UInt_t AliPhysicsSelection::CheckTriggerClass(const AliVEvent* event, const char
...
+/// \return True if the given event matches the trigger logic
+Bool_t AliPhysicsSelection::EvaluateTriggerLogic(const AliVEvent* event,
+						 AliTriggerAnalysis* triggerAnalysis,
+						 const char* triggerLogic, Bool_t offline){
+  using namespace std;
+  // split the trigger logic on the individual trigger names and replace them
+  // with TFormula parameters
+  string trg_logic_orig(triggerLogic);
+  // Regex identifying the trigger (>=1; alphanumeric and _)
+  regex trg_re("\\w+");
```
cbourjau: aha! Man I wish we'd have unit tests... Anyways, given that the cxx11 regex is out of the game, I reverted back to the TPRegex one with the old pattern

OADB/AliPhysicsSelection.cxx
cbourjau: Yes, will do if we go this route!

OADB/AliPhysicsSelection.cxx
```diff
@@ -251,63 +259,66 @@ UInt_t AliPhysicsSelection::CheckTriggerClass(const AliVEvent* event, const char
...
-  if (formula.Compile() > 0)
-    AliFatal(Form("Could not evaluate trigger logic %s (evaluated to %s)", triggerLogic, trigger.Data()));
-  Bool_t result = formula.Eval(0);
-  
-  AliDebug(AliLog::kDebug, Form("%s --> %d", trigger.Data(), result));
-  
-  return result;
+  // Get the values for each individual trigger in the trigger logic string;
+  // These values are the parameters of the TFormula
+  regex_iterator<string::iterator> rit (trg_logic_orig.begin(), trg_logic_orig.end(), trg_re);
```
cbourjau: again, I think the cxx11 regex is out of the game anyways

BerserkerTroll:
@cbourjau 
https://gist.github.com/BerserkerTroll/a1967831bd6e507e6e2c22e1ef5bc91f
`TFormula`-based solution, but without C++11 regexs. Differs from your `TFormula`-based solution mainly in that it also caches the vector of bits together with the `TFormula`.
Minor differences:
1. Queries the map only once.
2. Splits the `triggerLogic` into tokens only once and only when there is a cache miss.
3. `if (offline)` check is moved out of the loop

In case there will be problems with `pfn`-based solution.

BerserkerTroll:
@cbourjau 
> Since I did not find this functionality within TPRegex

How about this, mon?

```
#include <iostream>

#include <TArrayI.h>
#include <TPRegexp.h>


void tpregextest(const char* triggerLogic = "(SPDGFO >= 1 || V0A || V0C) && !V0ABG && !V0CBG && ZDCTime && !TPCHVdip")
{
	TPRegexp fRegexp("[[:alpha:]]\\w*");
	TString trigger(triggerLogic);

	TArrayI pos;
	Int_t nrMatch;
	Int_t b = 0;
	Int_t e = 0;

	while ((nrMatch = fRegexp.Match(trigger, "", e, 1, &pos)) != 0) {
		b = e;
		e = pos[0];

		std::cout << trigger(b, e - b) << '\n';

		b = e;
		e = pos[1];

		std::cout << trigger(b, e - b) << '\n';
	}
}
```

BerserkerTroll:
https://gist.github.com/BerserkerTroll/1293f36715a17efb785180615c6829b8 based on `TPRegexp`

cbourjau:
@BerserkerTroll I didn't get around to do more profiling today or to further look into this nor will I probably until next week. I am certainly happy with either solution so please feel free to modify this PR at will if you have time before that! 

Did anybody with more knowledge about compiling temp files on the grid take a look at this, yet?

dberzano:
Hi all, I have just quickly read the whole conversation.

* The presence of a compiler on the Grid environment is expected. The presence of a **working** compiler is not, and will never be. This is to say, if you put this in production and then it does not work on some Grid nodes, we will not move a finger to make it work, you have been warned. Maybe, a fallback on `TFormula` in case the compilation fails is what you are looking for.

* The official stance is that we will never move AliRoot to ROOT 6, so do not count on it. O2 uses ROOT 6 and it's the only place where we support it officially.

cbourjau:
Ok, here we go again. I used @BerserkerTroll's TPRegex based tokenizer together with a cached TFormula. No funky compiling of temporary files. I also noted that the data we are running over makes a big difference on how big the change from the old to the new implementation is. As you can see in the attached files, the speedup is substantial with this new implementation (for LHC16k)

Old implementation on LHC16k AODs: [21165.pdf](https://github.com/alisw/AliPhysics/files/1145140/21165.pdf)
New implementation on LHC16k AODs: [27599.pdf](https://github.com/alisw/AliPhysics/files/1145141/27599.pdf)

I did not observe a regression on LHC16k nor LHC15o. Maybe we are getting close to closing this PR? @ekryshen

OADB/AliPhysicsSelection.cxx
```diff
@@ -251,63 +253,31 @@ UInt_t AliPhysicsSelection::CheckTriggerClass(const AliVEvent* event, const char
...
+/// \param triggerLogic Describing trigger logic; e.g. "V0A && V0C && ZDCTime && !TPCHVdip"
+/// \param offline Offline analysis(?)
+///
+/// \return True if the given event matches the trigger logic
+Bool_t AliPhysicsSelection::EvaluateTriggerLogic(const AliVEvent* event,
+						 AliTriggerAnalysis* triggerAnalysis,
+						 const char* triggerLogic, Bool_t offline){
+  auto formula_and_bits = FindForumla(triggerLogic);
+  auto trg_formula = formula_and_bits.first;
+  auto bits = formula_and_bits.second;
```
BerserkerTroll: Well, you are again copying `TFormula` and also a `std::vector`.
So, you are calling this https://root.cern.ch/doc/master/TFormula__v5_8cxx_source.html#l02458 ton of `new`s for every trigger logic evaluation.
It sould be `auto&`, as in my code. And `FindFormula` should return by reference.

OADB/AliPhysicsSelection.cxx
```diff
@@ -851,3 +817,55 @@ void AliPhysicsSelection::DetectPassName(){
   AliInfo(Form("pass name: %s\n",passName.Data()));
   fPassName = passName;
 }
+
+FormulaAndBits AliPhysicsSelection::FindForumla(const char* triggerLogic) {
```
BerserkerTroll: As I mentioned earlier, the return type should be `FormulaAndBits&`.

OADB/AliPhysicsSelection.cxx
```diff
@@ -251,63 +253,31 @@ UInt_t AliPhysicsSelection::CheckTriggerClass(const AliVEvent* event, const char
...
-#endif
-  if (formula.Compile() > 0)
-    AliFatal(Form("Could not evaluate trigger logic %s (evaluated to %s)", triggerLogic, trigger.Data()));
-  Bool_t result = formula.Eval(0);
-  
-  AliDebug(AliLog::kDebug, Form("%s --> %d", trigger.Data(), result));
-  
-  return result;
+  trg_formula.SetParameters(paras.data());
+  return trg_formula.Eval(0);
```
BerserkerTroll: `TFormula::SetParameters` copies the parameters into the `TFormula`'s member `fParams`. I don't think this copying is necessary, because they are used only once and will be reset when the `EvaluateTriggerLogic` is called next time.
In addition to `TFormula::Eval`, there is a `TFormula::EvalPar` method: https://root.cern.ch/doc/master/classROOT_1_1v5_1_1TFormula.html#a1f08c006a60991cc790d5c9410605787
It is defined as
```
virtual Double_t EvalPar(const Double_t *x, const Double_t *params=0){return ((*this).*fOptimal)(x,params);};
```
It uses pointer-to-member call syntax, but I think it is not difficult to guess that it points to one of the `TFormula`'s protected methods with suitable signature:
```
Double_t        EvalParFast(const Double_t *x, const Double_t *params);
Double_t        EvalPrimitive(const Double_t *x, const Double_t *params);
Double_t        EvalPrimitive0(const Double_t *x, const Double_t *params);
Double_t        EvalPrimitive1(const Double_t *x, const Double_t *params);
Double_t        EvalPrimitive2(const Double_t *x, const Double_t *params);
Double_t        EvalPrimitive3(const Double_t *x, const Double_t *params);
Double_t        EvalPrimitive4(const Double_t *x, const Double_t *params);
```
For example, the definition of the `EvalParFast` starts with the following lines
```
Double_t TFormula::EvalParFast(const Double_t *x, const Double_t *uparams)
 {
     const Double_t  *pdata[3] = {x,(uparams!=0)?uparams:fParams, fConst};
```
It uses the argument as an array of parameters if it is not null pointer.

So the calls to `SetParameters` and `Eval` can be merged into call to the `EvalPar`. It will not only remove unnecessary copying, but also take one line instead of two. It is not a premature optimization, it is removal of a premature pessimization.

BerserkerTroll:
@cbourjau could you upload flamegraphs in zoomable svg format, please?

OADB/AliPhysicsSelection.cxx
```diff
@@ -277,7 +277,8 @@ Bool_t AliPhysicsSelection::EvaluateTriggerLogic(const AliVEvent* event,
     paras[i] = triggerAnalysis->EvaluateTrigger(event, bit);
   }
   trg_formula.SetParameters(paras.data());
```
BerserkerTroll: You forgot to remove.

cbourjau: m-/ fixed now

BerserkerTroll: Could you profile the current state of this PR, please?

cbourjau:
@BerserkerTroll Github does not support svgs which is why I converted them to pdfs. You can find them on my dropbox:

https://www.dropbox.com/s/t0fwiv482ctyoyq/21165.svg?dl=0
https://www.dropbox.com/s/4cauxi843sd5u7o/27599.svg?dl=0

If you are interested, I added a small utility in my deployment tool [nittygriddy](https://github.com/cbourjau/nittygriddy) which makes it really easy to create those graphs if you have gdb set up. That feature is independent of the rest, so you can start your local execution any way you want and you just pass the pid to `nittygriddy profile`

BerserkerTroll:
@cbourjau 
> If you are interested, I added a small utility in my deployment tool nittygriddy which makes it really easy to create those graphs if you have gdb set up.

Yeah, thanks. Sorry for that I'm doing nothing and only command here. I was going to profile some things myself, but, unfortunately, the `aliBuild` is kinda broken (you can find a couple of my issue reports in the `aliBuild` repository). I fixed a couple of problems in it, but never got to the state when I had a working `AliPhysics`. Well, I didn't try hard to build it, but, TBH, the position "it works for us, but if it doesn't work for you we will not move a finger to make it work" doesn't put me in the mood to try harder.

But if I build the `AliPhysics`, I'll definitely try to measure some things which seem to work suspiciously slow and try to optimize it.

cbourjau:
Ok, latest profile of this PR did not change with the minor optimizations. Not a big surprise, in hindsight, given that the previous profile showed that we only spend ~1.5% of the time in EvaluateTriggerLogic anymore [link to dropbox](https://www.dropbox.com/s/kb669h1w31oj94o/31429.svg?dl=0). 

I very well understand the position, thought, that we don't want to support a new use case on the grid if it so  easily avoidable.

From my point of view, I consider this PR done, unless some major issue shows up.  I suspect that we could easily juice out major speed-ups from those 70% of the time spend in `TTree::GetEntry()`, if one could identify some branches which don't need to be read in. But that is clearly unrelated to this PR.

I hope we didn't drive away all the people to sign of on this PR by all the chatter ;)

BerserkerTroll:
@cbourjau thanks for the profile!

>  the previous profile showed that we only spend ~1.5% of the time in EvaluateTriggerLogic 

Well, but it spent the 1/3 of the time copying the `TFormula` and another 1/3 — destroying this copy. So it was doing nothing useful for the 2/3 of the time. It is not great.

Anyway, I also think that the work is finished. Well done.

ekryshen:
Dear @cbourjau and @BerserkerTroll,

Thanks a lot for your efforts. This is a very important contribution to speed up execution of user analyses. I suggest to approve it.

Thanks, cheers,
Evgeny

ekryshen:
+1

PR #1539, opened 2017-06-26
Fix for efficiency and response matrix evaluation from MC

btrzecia:
Correction for decay D channel at the MC level for efficiency calculation and correction for daughter removal for the Response Matrix


aiola:
+1

PR #1546, opened 2017-06-26
update D0 sys.unc about improver task

xinyepeng:
combine it with cut variation sys.unc, however, at 0-10%, only few bins has larger uncertainty on it(maximum 7%), at 60-80%, all within the uncertainty about cut variation


fprino:
+1

PR #1548, opened 2017-06-26
added method for setting OCDB path for FineSlewing calib

jezwilkinson:
Adds method for setting "FineSlewing" path from OCDB (same implementation as for ParOffline and RunParams)


preghenella:
+1

PR #1549, opened 2017-06-26
PWGGA/GammaConv: Factorize out CutHandlerConv to separate class

mfasDa:
Factorize out the cut handler from the macro AddTask_GammaConvV1_pp.C
to a separate class so that it can be shared among various AddTasks.


FriederikeBock:
+1

PR #1551, opened 2017-06-26
calculate back the sum of cell energy removed in previous commit

gconesab:
bug reported by @e-masson


gconesab:
@e-masson, please check it is ok

e-masson:
@gconesab It seems ok to me, we will check at the next QA output. Thank you!

PR #1566, opened 2017-06-27
PWGGA/GammaConv: Simple doxygenize

mfasDa:
Add doxygen class header with description of cut observables.
This will help the users to decode the cut string, allthough description
of the value to be set for a given cut is still missing.

In addition:
- Create doxygen group for GammaConv
- Move existing member declarations to doxygen


FriederikeBock:
+1

PR #1569, opened 2017-06-27
Update Ds systematics coming from ImproverTask

jhamoncern:
ImproverTask systematics estimated from cross-section ratio (Improver_Old / Improver_d0mean) instead of efficiency ratio.


fprino:
+1

PR #1597, opened 2017-06-29
NUA studies

zkhabano:
Macro used for non-uniform acceptance studies together with some basis QA distributions


rbertens:
Hi @zkhabano , I assume that your goal is that this code is compiled into a library so that it will be part of AliPhysics. If so,  I think that you also have to change the CMakeLists.txt to make sure that this new folder NUA will picked up by the compiler.

dberzano:
Hi all, can we close this PR if there is no more progress?

zkhabano:
Hello Redmer, Dario,

please close the request as I will come back to it later. 

Thank you

PR #1598, opened 2017-06-29
Add missing std::

wiechula:
root6 does not globally import namespace std.
std:: needs to be added in order to compile.


adriansev:
+1

wiechula:
Thanks for the fast reaction Adrian.
Cheers,
Jens

PR #1605, opened 2017-06-29
Option to set flavour tag property in AliEmcalJets

rhaake:
The idea is to set the flavour tags according to our definition in the JetExtractor and then to use those tagged AliEmcalJets in a response matrix maker.


ebruna:
+1

PR #1612, opened 2017-06-30
PWGGA/GammaConv: Added more pT cut variations and trainconfigs for PCM 7 and 8 TeV

nschmidtALICE:
This pull request adds several new pT cut variations on the electron and photon level in AliConversionPhotonCuts.cxx.
For running and testing of these variations, several trainconfigs have been added.


FriederikeBock:
+1

PR #1616, opened 2017-06-30
PWGGA/GammaConv: taken out negation from if for filling TOF/ITS histos

lleardin:
The negation (==0) was supposed to fix a problem in the filling the ITS and TOF histograms. As far as I can see, it has instead the effect of rejecting almost everything (in PbPb). The filling problem will be further investigated.


FriederikeBock:
+1

PR #1625, opened 2017-06-30
Update Ds systematics (PID + Cut variations)

jhamoncern:
New PID systematics:
   * 2% between 2 and 12 GeV/c (StrongPID)
   * 0% between 12 and 24 GeV/c (ConservativePID)

Smoothening of [8-12] cut variation systematics

Note: the normalisation syst. is set to 3.7% as for D* and D0 w/o vertexing, while it set to 2% for D+ and D0 standard.


fprino:
+1
(the normalization systematics is not used in the processing macros)

PR #1626, opened 2017-06-30
Adding more plots to identify anomalous peak in dphi correlation

CalPolyPat:
Added invariant mass plots and dphi without secondaries.


adriansev:
+test

adriansev:
+1

PR #1631, opened 2017-07-01
added folders in PWGPP/EMCAL

efranzis:


dberzano:
+1 (but next time please stick to adding new items instead of changing the order of the existing ones!)

PR #1632, opened 2017-07-01
PWGGA/GammaConv: Fix operator!=

mfasDa:
operator!= was implemented as operator==, leading
to a crash in case no photon was reconstructed, and
to a misbehaving otherwise.


FriederikeBock:
+1

PR #1645, opened 2017-07-03
PWGGA/GammaConv: Add task for K0 to two Pi0s.

hjbossi:
-Doing conversion and calo photon selection.
-Building Pi0 candidates for all three methods separately.
-K0 candidates still disabled.
-Task configured with loose cuts for the V0 reader and default event gamma conv and calo cuts for 2012 pp.
-Adding new class in the build.


FriederikeBock:
+1

PR #1684, opened 2017-07-06
reverse correspondance process with high-pt trig and IDparticle associate

bschaefe:
this code compares events with both a high-pt track and deuteron, to events with a high-pt track and proton, and high-pt track and unidentified hadron.  Two thresholds are used for the high-pt trigger: 5, 10 GeV/c.


aohlson:
+1

PR #1685, opened 2017-07-06
PWGGA/GammaConv: changed case 4 chi2 cut for PbPb, added trainconfigs…

lleardin:
… fro chi2 syst studies


FriederikeBock:
Hi lucia,

why did you change the cut and not simply added case 10 ("a"  in the cutstring)?

This would make it more backward compatible.
Regards,
Fredi

lleardin:
Hi, 

because case 4 was chi2=7, and if chi2=10 is already too much, this is an excessive cut. Between the two, I preferred to have the chi2=25 as "mainstream" cut, that's why i switched them. It not likely I will ever use the chi2=7, thus i prefer this to be with the letter.
This said, if it could be  that somebody else needs the chi2=25 variation not in HI  I can switch them back.
I can modify it here on the fly, can't I?
Cheers,
Lucia

FriederikeBock:
well in principle it would be good to have the 25 available for general use and yes you can edit here directly, but keep in mind that the indentations might not be the same as in your editor.

lleardin:
Changes done, just a character so the indentations are hopefully safe...

FriederikeBock:
+1

PR #1697, opened 2017-07-07
PWGGA/GammaConv: Run K0 candidate selection.

hjbossi:
-Restricting Pi0 candidates to Pi0 mass window.
-Enable K0 candidate selection.


dmuhlhei:
+1

PR #1699, opened 2017-07-07
PWGGA/GammaConv: Adding parameter for trigger configuration in order …

hjbossi:
…to execute different trigger set-ups at the same time.


FriederikeBock:
+1

PR #1700, opened 2017-07-07
Increase Event Mixing statistics

hzanoli:
I am increasing the event mixing statistics to reduce the statistical errors at large delta eta bins.


ebruna:
+1

PR #1706, opened 2017-07-07
PWGPP-350: Add ctest for $AliPhysics/scripts/utilities

miranov25:


miranov25:
FYI

@ktf @mkrzewic @sawenzel 
I added to the make test test for the $AliPhysics/PWGPP/scripts/utilities.sh
On my laptop the test takes about 0.15 s.
```
96/97 Test #96: histmgr_fill_grouped .......................   Passed    0.44 sec
      Start 97: utilitiesTest
97/97 Test #97: utilitiesTest ..............................   Passed    0.15 sec
100% tests passed, 0 tests failed out of 97
```
I would like to proceed with my other tests. I have following questions:
* What can I consider as upper time limit for other tests
* Is it possible to access logs from ctest?
** as I obtained  invoking ctest --verbose

PR #1717, opened 2017-07-09
PWGGA/GammaConf: Fix in iterator over second leg

mfasDa:
Copy/Paste error in iteration over second leg: Instead
of calling operator ++ and != on the iterator over the
second leg it was done on the first leg, leading to
undefined behaviour.


dmuhlhei:
+1

PR #1718, opened 2017-07-09
Revert "Adding a histogram for inclusive track pT spectrum"

e-masson:
This reverts commit 09077e43c600e18e7e05ba832ead08e36b4b77ce which was faulty for multiple histograms. Looking for another way to implement the inclusive track pT spectrum.


gconesab:
+1

PR #1737, opened 2017-07-10
Applying new cuts to associated particles

CalPolyPat:
Adding some standard loose cuts to associated particles for a dphi correlation analysis


adriansev:
+1

PR #1751, opened 2017-07-11
PWGGA/GammaConv: Added QA before meson selection.

hjbossi:
-In addition fixed negation for meson is selected.

### DIFF
--- a/PWGGA/GammaConv/AliAnalysisTaskK0toPi0Pi0.cxx
+++ b/PWGGA/GammaConv/AliAnalysisTaskK0toPi0Pi0.cxx
@@ -40,6 +40,7 @@
 #include "AliLog.h"
 #include "AliVCluster.h"
 #include "AliV0ReaderV1.h"
+#include "AliGammaConversionAODBGHandler.h"
 
 /// \cond CLASSIMP
 ClassImp(AliAnalysisTaskK0toPi0Pi0)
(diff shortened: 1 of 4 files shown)


dmuhlhei:
Could you please help me: I don't understand the changes in the BGHandler. Where is "GetNBackgroundEventsInBuffer" defined, for example?

hjbossi:
Hi Daniel,

I believe this is a result of conflicts with different versions of the code
that Markus and I worked on separately. We will prepare another commit
today with the changes.

Thank you for asking! I appreciate the help!

Hannah

mfasDa:
Hi @dmuhlhei,

The new function is intended to return the number of background events for a given bin in (z, multiplicity) currently stored in the buffer (not the maximum number of events / buffer size!). I don't see such a function implemented up to now.

We access this information via the function size of std::vector, which returns the number of events currently stored in the vector.

It would be helpful to have such a member function available, i.e. in a loop over background events one wants to loop only over the events which are actually stored in the buffer, not the maximum possible amount.

Best Regards

Markus

dmuhlhei:
thanks a lot :)

What is the purpose of 'AliGammaConversionMotherBGVector1'? The typedef is added, but I do not see where it is used.

mfasDa:
ah, this is a leftover remanent which was there just for illustration. Thanks for spotting it! While it does not harm it certainly should not end up in the repo. @hjbossi could you please remove the typedef and add a new patch for this?

hjbossi:
This change should be fixed now!

dmuhlhei:
thx! +1

dmuhlhei:
+1

PR #1758, opened 2017-07-12
PWGGA/GammaConv: Adding interface function adding background mesons f…

mfasDa:
…rom std::vector

In addition removing "using namespace std" from header file and
replacing instances with "std::" in order to avoid implicit
inclusion of the standard namespace in recursive include chains.


mfasDa:
New interface function is used by the K0task (internally using type safe std::vector as container for mesons). Functionality is only added, no existing functions are modified.

dmuhlhei:
+1

PR #1771, opened 2017-07-13
Fill by index and v0

cbourjau:
This reduces the number of times the bin indices have to be calculated for the pair histogram, which is a very costly operation. Also, this PR enables the usage of the V0 for cross checks


aohlson:
+1

PR #1772, opened 2017-07-13
PWGGA/GammaConv: More QA histograms

mfasDa:
Figure out where and why mesons are lost:
- Histograms with the amount of meson candidates per event before / after meson cuts
- Histogram with the selection code for Pi0 and meson selection


dmuhlhei:
+1

PR #1788, opened 2017-07-14
PWG/CALO: changed GetCorrectedEnergy last argument to period enum

nschmidtALICE:
In order to reduce the computation time of the energy correction, the string comparison to obtain the period enum is removed. Instead the GetCorrectedEnergy function now directly gets the enum as input parameter.


FriederikeBock:
+test

FriederikeBock:
+1

PR #1803, opened 2017-07-16
Fix array lengths 

marcovanleeuwen:
Code seems not to be in active use, but this silences a warning and may be useful if it is used again.
I checked with Panos (the original author) and he agrees.


bdoenigu:
+1

PR #1815, opened 2017-07-17
DeltaPT and centrality added

rvertesi:
1) Replaced the dummy kCFmult field in the tagged jet container with kCFcent, and fill this from the BJet tagging class with ZNA multiplicy percentile as centrality estimator.
2) Added a simple random cone method to estimate the background fluctuation (deltaPt). Created corresponding histo to fill. Incluced a switch into the class definition and also a switch 
Bool_t doRndmCone =  kFALSE into the macros/AddTaskEmcalJetBtagSV.C.


ebruna:
+1

PR #1816, opened 2017-07-17
PWGGA/GammaConv: Setting cluster energy cut to 0

mfasDa:
Setting energy cut to 0 in order to run with strongly relaxed tender
settings (300 MeV/50 MeV).


dmuhlhei:
+1

PR #1825, opened 2017-07-17
New variable for AliDilelectronVarManager

jjungIKF:
Added ITS shared cluster map
Added personal electron efficiency task


pdillens:
+1

PR #1829, opened 2017-07-17
PWGGA/GammaConv: Added cocktail task subwagon feature via division of rapidity

nschmidtALICE:
The rapidity string that is given as argument to the AddTasks must now be for example "80" instead of "0.8" as the rapidity gets divided by 100 in the task itself. This allows us to use the subwagon feature on the grid.


FriederikeBock:
+1

PR #1834, opened 2017-07-18
PWGDQ: Changed handling for AOD MC

carstooon:
The handling of AOD MC in the dielectron framework is tricky. Up to now the tracks are compared via their label on the stack. This is not working for AODs therefore a few changes are made. The handling for ESDs is unaffected by this patch.


tdahms:
+test

tdahms:
+1

PR #1841, opened 2017-07-18
"flag added to reduce the output size"

cterrevo:
flag to reduce the number of directories in the output file. Default value is set to kFALSE, in order to have all the directories


fcolamar:
+1

PR #1849, opened 2017-07-19
Lc2V0 bachelor code updates for PbPb

yowatana:
Dear all,

I made the following modifications to the codes to analyze Lc->pK0s in PbPb. 
I contacted the original authors (Annalisa,Elisa) of the codes and they agreed that I make modifications to the codes.  

AliRDHFCutsLctoV0.cxx(h)
- Added 4 variables to the array of cut variables
  - If the cut values are not in reasonable ranges, the cuts are not applied at all. This is intended to avoid any impacts on the past analyses.
- Added functions to calculate the variables
- Added one more option for PID

makeTFile4CutsLctoV0bachelor.C
- Added the 4 variables accordingly

AliAnalysisTaskSELc2V0bachelor.cxx(h)
- Added AOD protection against AOD-dAOD mismatch
- Added histograms for the variables

ConfigVertexingHF_****.C
- The number of cut variables and the cut arrays are updated accordingly. 

Best regards,
Yosuke


ebruna:
+1

PR #1850, opened 2017-07-19
Adding a histogram to study the cluster shower shape per SM

e-masson:
This histogram will be used to study the cluster shower shape per supermodule, integrated over the whole cluster energy range for now (could be detailed per energy bin later).


e-masson:
+test

e-masson:
+test

PR #1860, opened 2017-07-19
PWG/Cocktail: Updated Kaon cocktail parametrization for pp 8 TeV

nschmidtALICE:
Noticed a strong drop in the Kaon parametrization at high pT. Therefore changed the fit function from doHagPt to oHagPt which continues nicely towards high pT.


FriederikeBock:
+1

PR #1862, opened 2017-07-19
updates to QA/Calibration archive

efranzis:
A. added renamed folders to doxygen/Doxyfile.in
B. added files and folders to QA/Calibration archive in PWGPP/EMCAL/


dberzano:
Hmmmmmm... what is the purpose of `PleaseCopy...` files? Cannot you just copy those files there? We'd better off without placeholders.

I also see notes saying who should/should not delete a certain folder. If you don't want people to press a button, don't make a button. Let's change, centrally, the permissions for that folder, adding only the persons with permissions to change files there.

efranzis:
I know we'd be better off without placeholders. We are reorganizing our EMCal Calibration and I asked everybody to mover their files to specific folders. I don'T have these files and we named the folders according to a specific scheme - that's why.

efranzis:
That should be a very temporary measure. I hope this is done in 2 week but I just don't always want to rename folders and then rename them in the doxygen as well

efranzis:
@dberzano, well I don't consider this as mean but it would have saved me some extra work. I have no time for this and will close this now.

PR #1876, opened 2017-07-20
Adding the relative calibration and channel recovery

dhruvdixit:
Added single channel calibration for the EMCAL. Calibrations are relative and use good channels to try to calibrate and recover the bad channels.


efranzis:
+test

efranzis:
+1

PR #1883, opened 2017-07-21
Updates in the naming of containers for the train

pchrista:
Updates in the naming of containers for the train + plotting macro


rbertens:
+1

PR #1887, opened 2017-07-21
PWGGA/GammaConv: Changed parameter to the add macro to selectConfig i…

hjbossi:
…n order to select a train configuration.

-Added train config for the 7 TeV case.
-Added a separate cut for the mixedPi0 case.
-Fixed histogram not being filled in GammaPureMC.


FriederikeBock:
+1

PR #1897, opened 2017-07-22
Optimize AliPhysicsSelection::CheckTriggerClass

BerserkerTroll:
After @cbourjau optimized the `AliPhysicsSelection::EvaluateTriggerLogic`, `AliPhysicsSelection::CheckTriggerClass` became the most expensive method in the `AliPhysicsSelection` class:

master
(This flamegraph is for LHC16k data)

I've decided to take a closer look at it:
master_close
and found that it is **very** inefficient.

Currenlty trigger string "or" patters are manually expanded by the `AliOADBPhysicsSelection::AddCollisionTriggerClass` and `AliOADBPhysicsSelection::AddBGTriggerClass` methods, i.e. `+CINT1-[I|B]` is turned into `+CINT1-I,CINT1-B`.

When the analysis is performed, `AliPhysicsSelection::CheckTriggerClass` receivs the trigger string in the form `+TRIGGER1,TRIGGER1b,TRIGGER1c -TRIGGER2 [#XXX] [&YY] [*ZZ]`, splits it using space as a separator and then splits requred and rejected trigger substrings (those with `+` or `-` in the front), if any, using comma as a separator. Then it tries to find required or rejected triggers in the strings of the fired triggers of the current event.

As you can see, `TString::Tokenize` eats a lot of CPU time.

### My version

First of all, I've decided to keep "or" patterns unexpanded. One of the advantages of this approach is that `event_stat.root` directory structure became much more readable. Just compare old and new files:
luka

The minor difference is that I parse the trigger string manually.

The major improvement is that now required and rejected trigger strings are turned into the regular expressions and cached in the `map`. The way of how they turned into regular expressions is pretty straightforward:

- for one trigger, the "or" subpattern of the form `[this|or that]` is unsurprisingly transformed into the `(?:this|or that)` (`(?:...)` is a non-capturing group. It should be a bit faster than the capturing one `(...)`)

- the group of triggers `TRIGGER1,TRIGGER1b,TRIGGER1c` is also turned into non-capturing group of the "or"-subpatterns: `(?:TRIGGER1|TRIGGER1b|TRIGGER1c)`.

- and `\b` in the beginning and in the end checks for word boundaries. E.g. `\b123\b` does match `0 123 45` but does not match `012345`.

That is all. Ah, yes. My version of the `AliPhysicsSelection::CheckTriggerClass` does not touch its third argument (passed by reference) if the matching fails. The current version assigns `0` to it.

Lets take a look at flamegraphs for my implementation:
opt
(Leftmost box with the `A..` text is the `AliPhysicsSelection::CheckTriggerClass`)
Zoomed:
opt_close

Now the `AliPhysicsSelection::CheckTriggerClass` takes about 17-18% of the time and its caller — the `AliPhysicsSelection::IsCollisionCandidate` — 21.5-22%.
With my modifications, `AliPhysicsSelection::CheckTriggerClass` takes 2.0±0.5 % of the time and `AliPhysicsSelection::IsCollisionCandidate` around 6.5-7%.

It seems that PhysicsSelection became as cheap as MultSelection.

I can't _prove_ that my version behaves the same way as the original one, but I can, comparing `fHistStat`s from `event_stat.root`, try to show that there is no evident discrepancies. So the histograms for the original and my version:
0master 0opt

The zip-archive with flamegraphs in svg format [checktriggerclass.zip](https://github.com/alisw/AliPhysics/files/1167621/checktriggerclass.zip)

BTW. @pgonzale83 I think [the lines 16—20 here](https://github.com/alisw/AliPhysics/blob/628f7aace0145ccaa855ff2393e2b8577e913edc/PWGGA/GammaConv/macros/AddTaskCustomPhysicsSelection.C) should be fixed. The syntax is `+TRIGGER1,TRIGGER2` and not `+TRIGGER1,+TRIGGER2`, so the plus signs after the commas should be removed.


cbourjau:
I guess not too many people care about such considerable efficiency gains in AliPhysics ;) Never the less, I much appreciate your efforts! Just a few comments:

1. Is it intentional that you changed `OADB/COMMON/PHYSICSSELECTION/data/physicsSelection.root`?

If so, I suppose that it is because of the expansion of the trigger strings. While that certainly makes things look cleaner, I could also imagine that these fine grained statistics are actually useful/needed elsewhere. Is this a necessity for this PR? I suppose we could get the same efficiency gains without changing how we expand the trigger strings, could we not? If so, maybe we should rather leave the expansion as it is now to be on the safe side?

2. What do you mean that you cannot "proof" that the old and the new version are identical?

Life would be so easy with unit tests... Never the less, might it be possible to have a commit where the old and the new implementation are available in parallel and where we just `Fatal` if the results diverge for an event?

BerserkerTroll:
2 @cbourjau:

> Is it intentional that you changed OADB/COMMON/PHYSICSSELECTION/data/physicsSelection.root?
> If so, I suppose that it is because of the expansion of the trigger strings.

Yes, you are right. `OADB/COMMON/PHYSICSSELECTION/data/physicsSelection.root` is generated by the `OADB/macros/BrowseAndFillPhysicsSelectionOADB.C`, which calls `AddCollisionTriggerClass`, which previously expanded trigger string and now does not.

> I suppose we could get the same efficiency gains without changing how we expand the trigger strings, could we not?

I haven't measured, but I think the gain will be lower.

> While that certainly makes things look cleaner, I could also imagine that these fine grained statistics are actually useful/needed elsewhere.

If someone needs expanded names, it is not a problem to expand them at file creation time. Or better, keep them unexpaded and write a macro which will expand them on request.
Anyway, it is easier to get performance boost by not expanding patterns and expand them later (if it is ever needed) than do the opposite.

> What do you mean that you cannot "proof" that the old and the new version are identical?

I mean mathematically.

> might it be possible to have a commit where the old and the new implementation are available in parallel and where we just Fatal if the results diverge for an event?

It is not that easy, because the old implementation can't handle unexpanded string.  My implementation doesn't care about expansion and will work for expanded triggers, but comparing implementations for expanded strings doesn't make much sense.

BTW, I'm preparing another PR with sligtly different implementation than this one. I've managed to reduce the time taken by the `CheckTriggerClass` to 0.8-1.4%. Mind looking at it or I should not cast you?

ekryshen:
+1

dberzano:
Hello all, it was correctly pointed out that PCRE is not required by aliBuild, leading to build problems for some users (see @bdoenigu) and potential troubles on larger scale deployments.

Adding dependencies to AliRoot and AliPhysics, as already discussed for boost, and ROOT 6 (see what Predrag remembered yesterday at the plenary), is forbidden as a general rule - but it can be granted under some conditions, and coordination with us.

Now the job done here is good job and certainly above average, we're wondering if it can be done without using this external library (ROOT has regular expressions). Or maybe, it can be done by not using regular expressions at all.

Anyway the point is, this will have to be reverted without some action from you, but it's a pity to waste this work.

BerserkerTroll:
@dberzano fixed in #1972

PR #1900, opened 2017-07-23
Updates of CosThetaStar for Lc->pK0s analysis

yowatana:
Dear all,

The calculation of the proton emission angle was performed in the Lc rest frame, but it is changed to the pK0s pair rest frame. The old method assumes the mass of Lc even when the pair has an invariant mass far away from the Lc peak. 

Best regards,
Yosuke


ebruna:
+1

PR #1921, opened 2017-07-24
Fix: Don't reject B to JPsi in MC

dweiser90:
Bugfix: Keep all JPsi on the reconstructed MC level to be consistent with generated level


iarsene:
+1

PR #1936, opened 2017-07-25
Enable/disable the update of the cov matrix also for PbPb (F. Prino)

fgrosa:
FillRecoCand called before the smearing of the track parameters. Possibility enable/disable the update of the cov matrix also for PbPb (F. Prino)


fprino:
I think it is not needed to increase the class revision, since you did not change the data memebrs, if I didn't overlook anything.

fgrosa:
You are right, I just took the files that you modified, but I forgot about the partial modification done by Anastasia. Maybe she already updated the class revision, since it was added the flag for enable/disable the update of the cov matrix (which is a data member).

fprino:
+1

PR #1944, opened 2017-07-25
Another way of optimizing AliPhysicsSelection::CheckTriggerClass

BerserkerTroll:
This PR does the same thing as #1897, but more optimally. I did want to get rid of temporary `std::string` created and destroyed every time the method is called and of `TPRegexp`, which also do allocate and free memory every time the `TPRegexp::Match` method is called.

To be able to query the hashmap (in contrast to the previous PR, this one uses hashmap and not ordered `std::map`), which stores `std::string` using `const char*` I had to use `multi_index_container` from boost.
If the use of boost scares you, I hasten to point out how the boost is used: it is only a compile time dependency, user don't need to have boost installed to use compiled `AliPhysics` (although one already has it, because boost is one of the AliPhysics indirect dependecies, see https://dberzano.github.io/alice/alibuild/aliphysics_deps.png). boost headers are included only in the `AliPhysicsSelection.cxx` file and not in `AliPhysicsSelection.h` file, so they do not poison anyone who includes `AliPhysicsSelection.h`. And I don't link with any boost shared libraries.

The flamegraph for current implementation (zoomed):
opt2

Now, the method takes something like 1.1±0.3% of time instead of 2.0±0.5% in the previous PR.

@cbourjau what do you think about it?


dberzano:
Hi, thanks for your commit, however it has been decided (in quite a number of computing boards) not to add additional dependencies to AliPhysics, so even if I am not in principle against it, the CB decision still stands. You'll have to live without boost (and you don't know how I regret to tell this to you) :-(

BerserkerTroll:
@dberzano it is not possible to add even buildtime (not runtime!) dependencies?

dberzano:
Well, boost and yaml-cpp are quite different in size. No boost in AliRoot/AliPhysics was the decision, and adding it as a build-time dependency will break all the other cases using "proper boost". I regret to say you'll have to find a solution which does not depend on boost.

PR #1945, opened 2017-07-25
Changed the default of fAODProtection from 0 to 1

yowatana:
Dear all,

The default value is changed to 1 because 1 is used as default in D meson codes. 

Best regards,
Yosuke


ebruna:
+1

PR #1951, opened 2017-07-26
Update PID BF code

panjinjin:
added pile-up cuts

### DIFF
--- a/PWGCF/Correlations/DPhi/AliAnalysisTaskPIDBFDptDpt.cxx
+++ b/PWGCF/Correlations/DPhi/AliAnalysisTaskPIDBFDptDpt.cxx
@@ -72,6 +72,7 @@
 #include "AliMultInput.h"
 #include "AliAnalysisUtils.h"
 #include "TVector2.h"
+#include "AliEventCuts.h"
 
 
 using namespace std;
@@ -225,6 +226,7 @@ AliAnalysisTaskPIDBFDptDpt::AliAnalysisTaskPIDBFDptDpt()
   _weight_1      ( 0    ),
   _weight_2      ( 0    ),
   _eventAccounting ( 0),
+  fEventCut ( 0),
   _m0 ( 0),
   _m1 ( 0),
   _m2 ( 0),
@@ -575,6 +577,7 @@ AliAnalysisTaskPIDBFDptDpt::AliAnalysisTaskPIDBFDptDpt(const TString & name)
   _weight_1        ( 0    ),
   _weight_2        ( 0    ),
   _eventAccounting ( 0),
+  fEventCut ( 0),
   _m0 ( 0),
   _m1 ( 0),
   _m2 ( 0),
@@ -793,6 +796,8 @@ void AliAnalysisTaskPIDBFDptDpt::UserCreateOutputObjects()
   OpenFile(0);
   _outputHistoList = new TList();
   _outputHistoList->SetOwner();
+
+  fEventCut.AddQAplotsToList(_outputHistoList);
     
   //if ( _singlesOnly )   _outputHistoList -> Add( fHelperPID -> GetOutputList() ); // add AliHelperPIDBFDptDpt object output list to task output list only for singles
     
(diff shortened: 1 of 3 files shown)


majanik:
+1

panjinjin:
Hi! I don't quite understand why the error happened for the test. It seems that the reason for the error is that I included "AliEventCuts.h" in the header file. However, this should not give us any error, since I tested the code locally very well. Could you please help on this?

rbertens:
hi @panjinjin ,
probably you need to add ${AliPhysics_SOURCE_DIR}/OADB as include directory in the CMakeLists.txt . 
cheers,
redmer

aohlson:
Hi @panjinjin 
Do you really mean to create the object in the header file?  (i.e. "AliEventCuts fEventCut" instead of "AliEventCuts *fEventCut")  If you do the latter, then you can just declare "class AliEventCuts" instead of including AliEventCuts.h in your .h file and I think it will compile.  Not sure if this gives you the functionality you need, otherwise you can do as Redmer suggests.

panjinjin:
Hi, @rbertens!
Thank you for your reply! I never changed anything in CMakeLists.txt, so I am not quite familiar with it. Could you please tell me where exactly CMakeLists.txt to add ${AliPhysics_SOURCE_DIR}/OADB? 
Thanks!
Au-Au

panjinjin:
Hi, @aohlson !
Thank you for your reply! I create the object in the header file because I want to use the class AliEventCuts to do the standard pile-up cuts for run2 data suggested by Alex. Then I followed his instructions to create the object in the header file. I also tried to declare "class AliEventCuts" instead of including AliEventCuts.h in the .h file as you suggested, but segmentation violation happened... Do you know any other code which also does pile-up cuts using class AliEventCuts? I want to learn from them.
Thanks,
Au-Au

rbertens:
Hi @panjinjin ,

Perhaps this clarifies:
- you probably want to create a pointer to the AliEventCuts object as a member of your class. So, in header of your analysis class, declare 'AliEventCuts *fEventCut'  (as Alice suggests). If you do so, you do not need to include the header AliEventCuts.h in your own analysis class header: making a forward declaration suffices (i.e. doing 'class AliEventCuts'). 
This forward declaration basically tells the compiler that the  type 'AliEventCuts' exists, but since the header of your analysis class only includes a pointer to this class, it does not need to know what the specifics (i.e. memory layout) of AliEventCuts actually are (so you do not need to include the header AliEventCuts.h in your analysis header). 

In the implementation of your class, however, you want to use the functionality of AliEventCuts, so there you *do* need to include the header AliEventCuts.h, and create an instance of the object ( something like fEventCuts = new AliEventCuts(), or you can pass the instance via a setter function) . 
This means that the compiler also needs to know where to find the actual file AliEventCuts.h. This can be specified by including an include path o the CMakeLists.txt file, that you can find under $ALICE_PHYSICS/PWGCF/Correlations/DPhi/CMakeLists.txt . 

This file contains a list of include folders (around line 23)

# Additional includes - alphabetical order except ROOT
include_directories(${ROOT_INCLUDE_DIRS}
                    ${AliPhysics_SOURCE_DIR}/CORRFW
                    ${AliPhysics_SOURCE_DIR}/PWG/muon
                    ${AliPhysics_SOURCE_DIR}/PWG/Tools
                    ${AliPhysics_SOURCE_DIR}/PWG/JETFW
                    ${AliPhysics_SOURCE_DIR}/PWGCF/Correlations/Base
                    ${AliPhysics_SOURCE_DIR}/OADB/COMMON/MULTIPLICITY
)
 
what is missing is the folder in which AliEventCuts.h lives, which is ${AliPhysics_SOURCE_DIR}/OADB/, this would probably need to be added if you want to use AliEventCuts in your analysis class.

panjinjin:
Hi, @rbertens !
Thank you so much for your detailed information! Now the logic is clear to me. I am now closing this pull and opening up another one with another set of codes which work.
Cheers! Have a nice day!
Au-Au

mpuccio:
Hi!
Actually as @panjinjin said, in the twiki the example code uses AliEventCuts as simple object.
This is necessary if you want to pass easily the user settings to the final task without using the comment syntax for the ROOT streamer and to avoid unnecessary dynamical allocation of memory.
The only issue with this pr is related, as @rbertens said, to the lack of the OADB dependence in the `CMakeLists.txt` file.
Please remember to add, on top of what @rbertens said, `OADB` among the Aliroot dependences of your library in the `CmakeLists.txt` file. The line should look like the following:
```CMake
set(ALIROOT_DEPENDENCIES ANALYSISalice CORRFW PWGmuon PWGCFCorrelationsBase PWGJETFW OADB)
```
Cheers,
Max

rbertens:
Hi @mpuccio , thanks for the clarification, I missed that AliEventCuts is used as a simple object in the Twiki, good to know!
my best,
redmer

panjinjin:
Hi, @rbertens !
I've changed the code according the discussion we have here, and I started a new pull request #1966. Could you please approve it for test? Thanks!
Regards,
Au-Au

panjinjin:
Hi, @mpuccio !
Thank you so much for your detailed information! I really appreciate it!
Cheers,
Au-Au

PR #1959, opened 2017-07-26
Update weight calculation

hzanoli:
I found an inconsistency on the weight calculation and weight use and this fix it.


sashingo:
+1

PR #1971, opened 2017-07-27
fixed filling of fBGProbability in AliGammaConversionAODBGHandler

fjonasALICE:
The array fBGProbability in AliGammaConversionAODBGHandler was filled without considering the range of the array. A lookup array was created to avoid segfaults.


dmuhlhei:
+1 :)

PR #1972, opened 2017-07-27
TPRegexp instead of PCREWrap in AliPhysicsSelection

BerserkerTroll:
Fixes issues with #1897.

No changes in logic, just standard TPRegexp for regular expressions instead of custom pcre wrapper.

### DIFF
--- a/OADB/AliPhysicsSelection.cxx
+++ b/OADB/AliPhysicsSelection.cxx
@@ -65,7 +65,6 @@
 //   Origin: Jan Fiete Grosse-Oetringhaus, CERN 
 //           Michele Floris, CERN
 //-------------------------------------------------------------------------
-#include <memory>
 #include <vector>
 
 #include <Riostream.h>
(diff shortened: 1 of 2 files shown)


dberzano:
+1 and thanks!

Just out of curiosity, did you measure the performances here? Is there anything we lose by the ROOT wrapper?

BerserkerTroll:
@dberzano performance difference is around 1%. Yes, it is negligible, but it is just hard for me to accept that it will allocate and free memory in a (event-)loop: `TPRegexp::MatchInternal` (called by `TPRegexp::Match`) uses `new` and `delete` every time it is called.
Also, `TPRegexp::IsValid` method is broken (see [1]). It requires user to know that `TPRegexp` does not compile regular expression after construction and this method always return `false` untill you trigger regular expression compilation by `TPRegexp::Match`.
I just wanted to write a tiny wrapper for this particular task I solved which doesn't have such issues.

[1] https://sft.its.cern.ch/jira/browse/ROOT-8923

PR #1977, opened 2017-07-27
Use same event selection as AliMultiSelection

hzanoli:
This solves the discrepancy in the event selection in the task. The details of the event selection used for calibration are here (https://twiki.cern.ch/twiki/pub/ALICE/PACentStudiesRun2/pA_centralityRun2_19042017.pdf).


sashingo:
+1

PR #1978, opened 2017-07-27
task pp 13 TeV update

cdeconti:
removing old task and updating the new addtask


sashingo:
+test

sashingo:
+1

ebruna:
Hi Camila,

is there a reason why the class is called "AliAnalysisHFETPCTOF_2"? Since AliAnalysisHFETPCTOF is not there anymore, wouldn't have been better to simply change the name to AliAnalysisHFETPCTOF?
Thanks
Elena

cdeconti:
Hi, Elena

You are right, I can change the name, removing the "_2". But about the
addtask, there is already one AddTaskHFEtpctof.C and I'm not sure if
keeping my addtask name like AddTaskHFETPCTOF.C is ok, since the only
difference is the capital letters. So I think for the addtask I'll mantain
the "_new". So soon I update these things, thanks!
Cheers,

*Camila*

2017-07-27 17:44 GMT+02:00 ebruna <notifications@github.com>:

PR #1984, opened 2017-07-28
PWGDQ: Added TOFmismatch calculation in efficiency task + ITS related variables in VarManager

carstooon:
There is now the option to reject TOF mismatch when computing the efficiency with the AliAnalysisTaskElectronEfficiency task.

Addition of ITS dEdx and shared cluster map for every ITS layer in VarManager.h

Cheers
Carsten


pdillens:
+1

PR #1986, opened 2017-07-28
PWGDQ: Added centrality switch in Addtask

carstooon:
Added the possibility to change centralities via the Addtask


pdillens:
+1

PR #2004, opened 2017-07-30
Add files via upload

totonalam:


totonalam:
For calculating BF

altsybee:
+test

rbertens:
Hi @totonalam , I have two questions to your pull request

1) I assume that your goal is that this code is compiled into a library so that it will be part of AliPhysics. If so, I think that you also have to change the CMakeLists.txt (in PWGCF/EBYE/CMakeLists.txt) to make sure that this class will picked up by the compiler.
2) you include quite some headers that do not seem to be necessary (e.g. TCanvas, TAxis, etc). To ensure efficiency in compiling, it would be good to include only the headers that are actually needed for your code

Thanks,
Redmer

totonalam:
Dear Redmer

You are right . I have included  AliPidBFBase.cxx in CMakeLists.txt.  I have removed header TAxis and TCanvas , since It is not required. I have attached all things  with this mail .


regards
Noor, VECC
________________________________________

rbertens:
Hi @totonalam  ,
Unfortunately the code cannot be approved like this, you need to commit the changes to your local git repository ,and then make a new pull request. This pull request will then be 'merged' with the request that's currently open. As soon as you've done that, we can approve the commit. 
Thanks,
Redmer

rbertens:
Hi @totonalam ,
Thanks for the update, but one more thing: you've uploaded theCMakeLists.txt into the wrong folder. It should go in

PWGCF/EBYE/CMakeLists.txt

but you've put it in

PWGCF/EBYE/BalanceFunctions/CMakeLists.txt

Could you move the file, and commit it again? 

My best,
Redmer

aohlson:
Hello @totonalam , 
Your task also needs to be added to PWGCF/EBYE/PWGCFebyeLinkDef.h before it can be compiled and linked as part of AliPhysics.
Regards, 
Alice

rbertens:
@aohlson Thanks for adding this, I completely forgot!

rbertens:
+test

rbertens:
Hi @totonalam ,
The code now doesn't compile, because there are global variables that are defined multiple times. In general, the usage of global variables is discouraged. Maybe you could discuss with Panos how to remove/change these global variables (because he seems to be the original author of the template classes)? 
Thanks,
Redmer

totonalam:
Dear Redmer

ok.  I have fixed it . I have committed now . Please have a look 

regards
Noor, VECC
________________________________________

aohlson:
+test

rbertens:
Hi @totonalam , 

Sorry to come back to this, but is there a reason that you use global variables for defining some binning things ( kTrackVariablesPair1 )? I'm a bit reluctant to approve the commit with global variables, for two reasons

1) global variables are defined globally (not within the class scope), so can easily cause troubles when someone wants to declare a variable of the same name
2) this also means that they can be changed outside of the class scope (ok, they are 'const' here, but in general, this is a risk)

I would recommend to just add these Int_t's as class members, then you can also set/change them using a simple setter function, which is more flexible. Since you do not use them to allocate memory in your .h, this should not be problematic (as far as I see, but please correct me if I'm wrong). 

Thanks,
Redmer

totonalam:
Dear Redmer

It is not  a problem .  Any way I have put two setter function to initialize AliTHn dimension .  Thank you so much . Please  commit the code . 

I have  made a pull request


regards
Noor, VECC
________________________________________

rbertens:
+1 thanks!

PR #2026, opened 2017-08-01
update on small bug in MC D-meson daughter removal

btrzecia:
Fix for the daughter removal of D meson on the MC level when there is more than 2 daughters.


adriansev:
+1

aiola:
Hello @btrzecia,
sorry, I overlooked this PR.
Just for my understanding, could you explain what was the issue and how this resolves it?
At first glance it is not so obvious te me that this is correct. It seems to assume that the daughters are always stored contiguously in the MC particle array. This is probably almost always true, but I am not sure that it is guaranteed.
Thanks,
Salvatore

btrzecia:
Hi Salvatore,
I changed this since GetDaughter() returns indices of up to 2 daughter - the first and the last ones, as I realized recently, and in some cases I have a decay with more of them. I also discussed this with D2H people, and the daughters should be stored continuously. 
Cheers,
Barbara

PR #2028, opened 2017-08-01
Suffix for add task and histogram for event mixing

hzanoli:
Added the possibility to use a suffix in the add task and increase the control histogram used to keep track of event mixing statistics.


sashingo:
+1

PR #2044, opened 2017-08-02
Fixing the compilation error from Noor and adding sphericity in the t…

pchrista:
…ask (more to come)


aohlson:
+1

aohlson:
Hi @pchrista , 
Thanks very much for the quick action, but unfortunately there are compilation errors (see above).  I have also been in touch with @totonalam to get a fix for this bug as soon as possible, and I think he is working on it too.  
Cheers, 
Alice

pchrista:
Hi @aohlson 

sorry, I did not pay attention to how Noor is using these variables.
I let him work this through then since you are in contact with him.
I thought that he was offline (e.g. due to time diff).
Greetings,
Panos

PR #2056, opened 2017-08-03
Bug fixes and updates to AliJJetJtAnalysis

TWSman:
-Fix bug that caused crashes on pp trains, missing AliAnalysisUtils
-Fix bug where dijet mass response matrix would be tried to access for data events and not only for MC
-Leading track jT now finds the correct leading track of jet
-Disable histograms with log(jT) binning by default and add option to enable them


aohlson:
+test

aohlson:
+1

PR #2061, opened 2017-08-03
PWGGA/GammaConv added nonlin corrections and train configs for PbPb5T…

mikesas:
…eV and pPb5TeV


FriederikeBock:
+1
But I really don't like how you did it now for the PbPb configurations (should be 81,82, 91,92) furthermore I would only use the basic conv calo function if you square it

PR #2069, opened 2017-08-04
PWGGA/GammaConv: Fixed the case of empty periodName strings in the Ad…

hjbossi:
…dTask.

-Removed unnecessary print statements from the task.


dmuhlhei:
+1

PR #2071, opened 2017-08-04
Added option to pass flag in case the EMCal framework is used, togeth…

crisjahnke:
Added option to pass flag in case the EMCal framework is used, together with filtered tracks (in a TCLones array). Nothing changes if the flag is not passed.
This change is necessary in order to use the index of the tracks to check if they are same track or not. Before we were comparing the momentum to check if the tracks are same or not. However, it start to fail.


sashingo:
+1

mfasDa:
Raising a compiler error due to missing forward declaration of TClonesArray in the header file of AliSelectNonHFE.h.

DEBUG:AliPhysics:ali-work: /home/markus/alice/sw/SOURCES/AliPhysics/ali-work/0/PWGHF/hfe/AliSelectNonHFE.h:39:77: error: 'TClonesArray' has not been declared
DEBUG:AliPhysics:ali-work:    void FindNonHFE(Int_t iTrack1, AliVParticle *Vtrack1, AliVEvent *fVevent, TClonesArray  *fTracks_tender=0, Bool_t fUseTender=kFALSE);
DEBUG:AliPhysics:ali-work:                                                                              ^~~~~~~~~~~~

PR #2098, opened 2017-08-07
new histos for trigger efficiency added

ReneChristianSchach:
new histos for EMC and EGA ADC amplitudes and trigger efficiency for photon isolation added
also histo to connect clusters to FastOrs added


e-masson:
+test

dlodato:
+1

PR #2122, opened 2017-08-08
removed TOF requirement for high-pt trigger

bschaefe:
high-pt trigger tracks have their TOF requirement removed for more uniform phi acceptance.


aohlson:
+1

PR #2128, opened 2017-08-09
Added post shift-corrected recentering method. 

rihanphys:
post shift-corrected recentering method added to AliAnalysisTaskZDCGainEq class


rbertens:
+1

PR #2157, opened 2017-08-11
PWGGA/GammaConv: fix in neutral meson analysis AddTask

fjonasALICE:
- AddTasks of NeutralMesonToPiPlPiMiPiZero now hand the periodName to V0Reader to avoid crashes while running on MC on the grid


dmuhlhei:
+1

PR #2158, opened 2017-08-11
Wall emcal2

aaniin:
Hi EMCal, hi @mfasDa , hi @raymondEhlers ,

Here's a pull request with a few more fixes of EMCal warnings. I restricted myself to the very simple, obvious, and safe changes. Can you have a look? The compilation log before the fixes is here: www.physi.uni-heidelberg.de/~hbeck/EMCal2.txt 

There are two more compilation warnings which prevent AliPhysics from being compiled with -Wall. I refrained from touching them since the fix is not super-obvious and I don't want to interfere too much with EMCal code. The warnings are however still easy to fix. Can you EMCal guys take of them? http://www.physi.uni-heidelberg.de/~hbeck/EMCal3.txt

Cheers,
Hans


aaniin:
Hi EMCal,

After fixing the warnings/errors of EMCal3.txt locally, more show up:
http://www.physi.uni-heidelberg.de/~hbeck/EMCal4.txt

Cheers,
Hans

mfasDa:
+test

Hi @aaniin,

Thanks for spotting! I think all are safe, however for what concerns the embedding part it would be good to hear the opinion from @raymondEhlers. Therefore I will just launch the test.

For the recent ones I will have a look.

For what concerns a general enabling of Wall I would like to add a point of warning: As compilers get more and more clever, Wall will cover more cases for newer compilers compared to older compilers. So the compilation result will suddenly depend on the compiler version used. With different compilers supported for different platforms switching on Wall & Werror might end up in a nightmare (we had such a case a few months ago). In this case standards would need to be set to the most accurate compiler, otherwise this might end up in different compiler errors for different users. Better first get the code warning-free, then enable Wall & Werror.

Cheers

Markus

raymondEhlers:
+1

Hi @aaniin,

Thanks for checking this out and providing the fixes! These changes look good to me. (It seems that clang did not to throw these warnings when I was working on it because I would have fixed them).

I'll take a look at the other warnings that you mentioned when I next get a chance.

Best,
Raymond

PR #2159, opened 2017-08-11
Use static function to set up task (instead of AddTask)

cbourjau:
A static function on this task replaces the AddTask macro. This help
keeping everything in one plasce and ensures a tight integration and
most importantly compile time checks! The latter is what one misses
out when using the ungodly .C macros.


aohlson:
+1

PR #2163, opened 2017-08-11
Apply Shift Corr for combined EP and Code cleanup

rihanphys:
Apply Shift Correction for combined ZDC EP and Code cleanup


jmargutt:
+1

PR #2164, opened 2017-08-11
/PWGGA/GammaConv: changed Int_t to UInt_t in NeutralMeson task

fjonasALICE:
- changed Int_t to UInt_t when comparing with .size() in NeutralMesonPiPlPiMiPiZero task to avoid compilation warnings


FriederikeBock:
+1

PR #2184, opened 2017-08-14
task for Initialize base class AliPidBFBase and update version of code for PID BF

totonalam:


aohlson:
Hi @totonalam , 
Like I warned you about in my email, in this pull request you have reverted the fix that we committed last week, so this pull request will break the compilation for some users.  Please remove commit bd69564d94cfe4c5fc969d4c6a5f1fe7fa443fac from your pull request.  
Thanks, 
Alice

aohlson:
Hi @totonalam , 
My mistake, I see that you did not revert the fix, but rather that you're pushing the fix again.  I still think you can remove this commit.  
Thanks, 
Alice

totonalam:
Dear Alice 

I can not get your point . I want to include two new task macro and my analysis class.  


regards
Noor, VECC
________________________________________

aohlson:
Hi @totonalam 
Your pull request consists of three commits.  The first one, from 10 days ago, is the same as the fix I committed then (as I recall you had network issues which prevented your fix from going through).  The other two are new and comprise your analysis task.  I will try testing things, but it may not work because the first commit is a copy of what was done last week...
Thanks, 
Alice

aohlson:
+test

aohlson:
+1

aohlson:
It looks like the testing and merging worked, so everything is fine, sorry for the confusion!

PR #2185, opened 2017-08-14
Fix parentheses AliAnalysisTaskSEDs

aaniin:
Fixed three logical mass window comparisons. The mathematical 0.1 < 0.2 < 0.3 does not translate into a c++ 0.1 < 0.2 < 0.3, which in a boolean usage like if(0.1<0.2<0.3) evaluates to false: 0.1<0.2 evaluates to a boolean true which turns into a 1. The final comparison 1<0.3 is not true. Triggered by compiling AliPhysics with -Wall.


aaniin:
From Anastasia:

> Ciao Hans,
> 
> yes thanks, even if it is a part of code I do not use, it is sure wrong what I wrote. Pleas go ahead with your pull request,
> and thanks!
> 
> cheers, Anastasia

fprino:
+1

PR #2197, opened 2017-08-15
Added functionality to use 2D pair efficiency map

dweiser90:
Added few functions/variables to use a 2D pair efficiency map


iarsene:
+1

PR #2203, opened 2017-08-15
Added 3particle correlator histograms to AliAnalysisTaskZDCGainEq class

rihanphys:
Added 3particle correlator histograms to AliAnalysisTaskZDCGainEq class


jmargutt:
+1

PR #2213, opened 2017-08-16
Add -Wall to AliPhysics

aaniin:
Added -Wall -Wno-missing-braces warning to AliPhysics. Missing braces is
included in Wall in clang but not in gcc>=4.8.

Hello @pzhristov @dberzano @rbertens @mfasDa @raymondEhlers @loizides @aiola @ddobrigk @cbourjau !

This PR is meant as a place to discuss the possibility of adding the -Wall warning to AliPhysics. The compiler warnings can help in catching some unintended behavior in our physics analyses. Looking at the warnings of -Wall, most warnings are harmless reordering but a substantial amount of warnings points out code which does not do what it is intended to do. 

A counterargument is the additional work load on coders, fixing harmless reorder warnings. I'd like to argue that making coders aware of such issues is a benefit on its own and the few harmful cases we catch outweigh the effort in easily&quickly fixing harmless cases. 

I added -Wno-missing-braces to have uniform behavior between clang and gcc. 
@dberzano the build system check only runs gcc, right? In general, we might run into problems when clang is more sensitive than gcc. It could cause non-compilation with clang since some parts of AliPhysics have a -Werror flag. The vast majority of code in AliPhysics is compiled without the -Werror flag, so different behavior with various compilers will mostly only cause different warning messages. 

Raymond Ehlers, Markus Fasel, and I successfully exercised fixing all -Wall warnings in the EMCal part of AliPhysics. While it was some work, I think all of the fixes were straightforward and did not cause too much extra work. 

I think this would also good to be discussed on Monday in the Computing Board.

Cheers,
Hans

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,6 +146,16 @@ else()
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror=delete-non-virtual-dtor")
   endif()
 
+  # Turn on "all" warnings, but skip missing braces warning
+  check_cxx_compiler_flag(-Wall CXX_COMPILER_HAS_WALL)
+  if(CXX_COMPILER_HAS_WALL)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
+  endif()
+  check_cxx_compiler_flag(-Wno-missing-braces CXX_COMPILER_HAS_WNO_MISSING_BRACES)
+  if(CXX_COMPILER_HAS_WNO_MISSING_BRACES)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-missing-braces")
+  endif()
+  
   # Issue a warning if AliRoot Core does not have AliEn support (this is wrong in most cases)
   if(NOT AliRoot_HASALIEN)
       message(WARNING "AliRoot Core has been built without with AliEn support: some features might be unavailable!")


ktf:
+test

hcab14:
Dear all,

I strongly support adding -Wall (and eventually -Werror) to AliPhysics.

Cheers
Christoph

dberzano:
Hi all, the problem is the following. Though I always check all warnings and I make sure my code always compiles without any, it is sufficient to move to a new compiler to make new warnings pop up.

The common use case is: user X updates Fedora to version 99 with GCC 10. User X is unable to compile _just because of the `-Werror`. `-Wall` is, quite simply, not comprehensive enough, and never will be.

Definitions of warnings depend so much on the system that making them fatal would just be annoying and counter-productive for most people. And it's true that people seeing warnings should fix them, but I can imagine the number of complaints if a Mac user just wants to produce a plot for an important conference, and suddenly has to fix something in STEER during the weekend, and people with the right to approve there are not available.

What we should maybe do is enabling the `-Werror` or another mean of enforcing warnings _only on PR checks_. That way we will not prevent people using compilers different from our production's to work if somebody has hit a corner case.

aaniin:
Hello @dberzano ,

Thanks for joining this discussion and sharing your insight. I appreciate how you always think of the analyzer having to, e.g., produce a plot.
I read your comment as disfavoring `-Werror` at least for the user. What do you think of this specific PR, namely adding `-Wall` for everybody? From looking at the code highlighted by the `-Wall` warnings, I think it will be quite useful. Most of the warnings are harmless re-orderings. Others like pointing to precedence of logical operators (look for -Wlogical-op-parentheses in http://www.physi.uni-heidelberg.de/~hbeck/AliPhysics-Wall-clang-Aug17.txt ) exhibit clearly wrong code. I believe that a warning for the analyzer as a first step will be very helpful. What do you think?

Cheers,
Hans

dberzano:
I do think it's OK but I would like to discuss it with a couple of usual suspects first (@sawenzel, @ktf).

sawenzel:
I fully support `-Wall` and would personally even impose `-Werror` (**on the offical build system**) once we have a green status. I have very good experience with this in other projects (Geant4, VecGeom,...) After all, we do not change our compiler that often and when we change the compiler some dedicated clean up can be done to regain a green state.

rbertens:
Hi all,
Since I was mentioned in the original PR, I would just like to add that I absolutely support -Wall . The first step of having less mistakes in code is making people aware of the fact that they exist. 
Thanks,
Redmer

dberzano:
+1 and for what concerns `-Werror` support, yes, we should have it on on the continuous integration platform. I have added support for enabling it optionally in #2377.

PR #2216, opened 2017-08-16
PWGGA/GammaConv/ Adding eventmixing method for CaloCalo

mikesas:


FriederikeBock:
@mikesas would you mind explaining what this changes are meant to do? As I think one could implement them in a more effective way. From what I see you might end up with no events mixed at all like this.

Regards,
Friederike

mikesas:
Previously the "pT max method" was mixing events which had the maximum pt cluster with delta eta delta phi of 0.2. This didn't improve the quality of the mixed events. The change of the code is to have the weighted average of all clusters in the same region between the mixed events.

FriederikeBock:
shouldn't we better actually build the pools properly so that we don't have to reject so many candidates from the start

FriederikeBock:
+1

FriederikeBock:
+test

FriederikeBock:
+1

PR #2220, opened 2017-08-16
Fix abs -> TMath::Abs in AliHMPIDTaskQA

aaniin:
This changes the behavior of the task! Clearly, the code intends to compare doubles (explicit type) but the integer abs() is used.


shahor02:
+1

PR #2223, opened 2017-08-16
update NanoAOD feature

lhusova:
add new variables for NanoAODs
remove the test classes from NanoAOD development
introduce header index variables


mazimm:
This update should be included into AliPhysics at the same time as this update to AliRoot:
https://github.com/alisw/AliRoot/pull/367

This patch into AliPhysics will not work without the AliRoot change. 
@dberzano maybe you can trigger this when a new AliRoot tag is created and include both updates at the same time?

dberzano:
Yes, let's make sure of that. Can you open a porting request in the AliRoot JIRA targeting v5-09-14?

dberzano:
+1

PR #2230, opened 2017-08-16
PWGGA/GammaConv: added light output to NeutralMeson task

fjonasALICE:
- added light output mode for NeutralMesonPiPlPiMiPiZero task
- cut histos and task histos can be disabled seperately by the integer


FriederikeBock:
+1 
Thanks!

PR #2231, opened 2017-08-16
Actualization of reference trigger classes

rolavick:
Adding reference trigger class names and corresponding cross sections.


ekryshen:
+1

ekryshen:
+1

PR #2232, opened 2017-08-16
Actualization of EVS fils

rolavick:
Actualization of EVS files to match EOS version, specifically triggerInfo.C and periodLevelQA.C. These should complete previous actualization.


ekryshen:
+1

PR #2234, opened 2017-08-16
add PHOS_Run2 macros

dsekihat:
Macros to analyze Run2 PHOS data are uploaded. These macros support pp/PbPb/pPb data.
An array which contains PHOS clusters is created by PHOSObjectCreator at first.
Then, this array is analyzed by Pi0EtaToGammaGamma.
Event selection is done by AliPHOSEventCuts, and clusters are selected by AliPHOSClusterCuts.


kharlov:
+1

PR #2241, opened 2017-08-17
Fix AliCFDataGrid::Copy

aaniin:
Function was calling itself infinitely

Revealed by a compiler warning.


aaniin:
Here's the warning:
```c++
DEBUG:AliPhysics:0: /Users/hbeck/alice/alibuild/sw/SOURCES/AliPhysics/0/0/CORRFW/AliCFDataGrid.cxx:136:1: warning: all paths through this function will call itself [-Winfinite-recursion]
DEBUG:AliPhysics:0: {
DEBUG:AliPhysics:0: ^
```
AliCFDataGrid inherits from AliCFGridSparse which properly implements AliCFGridSparse::Copy

dberzano:
+1 makes sense to me, I am wondering if the error was ever seen in production.

PR #2244, opened 2017-08-17
Removed unused value AliPhysicsSelection

aaniin:
Removed unused value in AliPhysicsSelection. The function used to obtain this value is const, thus the function call itself has also no effect.

@BerserkerTroll @cbourjau @ekryshen please look. This commit does not change behavior of the task, but the warning that some value is unused might also mean that one forgot to use it.

Revealed by compiler warning:
```c++
DEBUG:AliPhysics:0: /Users/hbeck/alice/alibuild/sw/SOURCES/AliPhysics/0/0/OADB/AliPhysicsSelection.cxx:466:15: warning: unused variable 'thresholdInner' [-Wunused-variable]
DEBUG:AliPhysics:0:         Int_t thresholdInner = tri->GetAlgoParamValueLI("0SH1",0); // algorithm name and param index 0
DEBUG:AliPhysics:0:               ^
```

Cheers,
Hans

### DIFF
--- a/OADB/AliPhysicsSelection.cxx
+++ b/OADB/AliPhysicsSelection.cxx
@@ -463,7 +463,6 @@ Bool_t AliPhysicsSelection::Initialize(Int_t runNumber){
       else h.ReadPITConditionsFromDB(runNumber,man->GetDefaultStorage()->GetURI().Data());
       AliITSTriggerConditions* tri = h.GetTriggerConditions();
       if (tri) {
-        Int_t thresholdInner = tri->GetAlgoParamValueLI("0SH1",0); // algorithm name and param index 0
         if (fTriggerOADB->GetSH1OuterThreshold()<0) {
           Int_t val = tri->GetAlgoParamValueLI("0SH1",1);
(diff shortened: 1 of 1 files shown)


cbourjau:
I noticed that as well when I was working on the AliPhysicsSelection. It was unrelated to what I was doing then, though, so I did not touch it. Frankly, I do not know what this code is supposed to do, so I cannot comment if its a bug or not. But good on ya for this cleaning efforts! If we want to do more of this bug-squashing (warning-fixing), maybe it would be a good idea to have an automated way of checking an analysis result before and after the modification? Something like:

```bash
$ compare AnalysisResults_old.root AnalysisResults_new.root
"Result: All histograms were identical"
```

 We had some ideas in this regards at the last Juniors Hackathon, but didn't get to a point where anything was releasable.

aaniin:
@cbourjau Great idea to compare the results. Feels like we do need a PWG-junior which gets s**t done. :D Cheers!

ekryshen:
Thanks for the fix. This part of the code is not used in the standard physics selection so it should have no consequences.

ekryshen:
+1

PR #2268, opened 2017-08-17
PWG/Cocktail: added pPb params based on latest spectra & reduction of size of pp.root file

FriederikeBock:
- reduce data volume of parametrisation file for pp
- addition of parametrizations for pPb according to latest pi0 spectra


FriederikeBock:
Hi @dberzano ,
I would like to keep the file in general here for now, but I cut the size down to 1/5 do you think this is alright? 
Regards,
Friederike

dberzano:
Hi, the problem is that every file will inevitably increase the whole Git clone - so we'll never get those 11 MB back. Git repositories monotonically increase in size, and if you upload a 2.2 MB file, but change it 20 times per month, you'll have increased the whole Git clone by 44 MB by the end of the month.

My point being:

* our policy, as agreed within the Git working group, is to limit PRs to 20 MB in size...
* ...my opinion, as computer scientist, is that binary files should stay away as much as possible from a Git repository.

If the 11 MB file will never/rarely change, then leave it in place. If you intend to change it frequently (say > 10 times in total), I suggest you store it on EOS as described before.

FriederikeBock:
Hi @dberzano,
these files usually didn't change that often (on average 1/month), while currently its a little more frequent as we are including more collision energies. I am a bit hesitant to move them because they are needed during the MC event generation when running the cocktail. We tried to load them from alien (for about two month) but that failed miserably, as there is something with the streamers of AliGenParam which is not as straight forward as we would have like it. I know you solution is different, but as I am currently on a hard time limit I'd like to avoid introducing yet another potential source of errors.
We'll try to limit the updates of these files to a minimum though.

Regards,
Friederike

dberzano:
@FriederikeBock, I am not suggesting to store the files on AliEn. During the Git migration we've come up with a solution that makes them available from CVMFS. And, time-wise, it's easy: just copy the file to a proper directory on lxplus, and it will be *instantly* available for your analysis. At 4pm we take daily snapshots and store those files on CVMFS.

Before proceeding with this PR, I really really suggest you have a look [here](http://alisw.github.io/git-advanced/#how-to-use-large-data-files-for-analysis).

nschmidtALICE:
+1

PR #2284, opened 2017-08-18
Apply recentering after CombinedShift Corr and Fill ZDC-QA vs Qnx,Qny

rihanphys:
Apply recentering after Combined Shift Correction and Fill ZDC-QA vs Qnx,Qny


jmargutt:
+1

PR #2292, opened 2017-08-18
Disable Weffc for gcc for AliV0ReaderV1

mfasDa:
gcc implements the class std::iterator to have a non-virtual
destructor. Consequently classes inheriting from std::iterator
will raise the Weffc warning. Inheriting from std::iterator
however is a popular and easy way to make a class std::iterable,
and the way chosen in this approach. As the std::iterator doesn't
allocate memory, the warning is still vaid but harmless, and
as it cannot be handled by any implementation it should be treated
as a false positive. Therefore Weffc will be disabled for gcc
for AliVOReaderV1 in a similar way is it is done for the iterator
of ROOT's TMap.

### DIFF
--- a/PWGGA/GammaConv/AliV0ReaderV1.h
+++ b/PWGGA/GammaConv/AliV0ReaderV1.h
@@ -29,7 +29,17 @@ class TH1F;
 class TH2F;
 class AliAODConversionPhoton;
 
-using namespace std;
+#if (__GNUC__ >= 3) && !defined(__INTEL_COMPILER)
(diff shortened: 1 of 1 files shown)


FriederikeBock:
+1
Thanks for the explanation!

mfasDa:
Another change coming with the pull request is that I removed using namespace std from the header file. You should be careful with using statements in header files: I saw them in many header files from the gamma package. As header are included in other files, this can lead to unintended consequences. Rather put using ... when really needed in the source file.

PR #2300, opened 2017-08-20
option to use 2012 trigger bad map added

ReneChristianSchach:
Option that only analyses events with with L1 recalculated trigger patches (from TriggerMakerTask) over 2012 threshold added.
Needed to use trigger bad map.
triggername must contain EGArecalc or L1recalc, and periodstr must contain 12, to use this option


dlodato:
+test

dlodato:
+1

PR #2307, opened 2017-08-21
Fixes for train definitions in the TrainSetup framework

cholmcc:
- When unpacking Zip files, do not overwrite existing files
- Updated logic for finding local files.  Had some problems.


cholmcc:
+test

PR #2325, opened 2017-08-22
Added some track QA plots

rhaake:
I added some track QA histograms that might be useful.
In addition, I fixed a warning on a comparison of a signed to an unsigned integer.


ebruna:
+1

PR #2334, opened 2017-08-23
Update on EVS

rolavick:
RunLevelStatistic: Special treatment for T-mask classes - different reference trigger used.
Lumi plot: Projection plot function sensitive to starting day


fprino:
Just to be sure before giving a plus one: @ekryshen are these modifications fine with you?

ekryshen:
+1

PR #2341, opened 2017-08-23
PWGGA/GammaConv: addition of new TTree in NeutralMesonToPiPlPiMiPiZer…

fjonasALICE:
…o task

In ProcessTrueMesonCandidates()
- added TTree to get information about pi+pi-pi0 that have same mother, but it is not an omega or eta
- wrongly identified pi0 are now also filled into contamination histo


FriederikeBock:
+1

PR #2342, opened 2017-08-23
Added base for MCLEGO trains with Therminator2

maciekniewielki:
I added a default configuration file and a gen_therm2.sh script which should enable running MCLEGO trains with Therminator2 and parsing them to HepMC.


rbertens:
+1 Hi @maciekniewielki , nice work. If the behaviour of the generator is properly validated, it would be good if you could also provide some reference material for this (presentation at a meeting or a JIRA ticket or so). 
My best,
Redmer

PR #2357, opened 2017-08-24
JES systematics procedure

btrzecia:
MC studies: procedure to reject tracks from an event according to a provided track inefficiency, for JES systematics studies.


aiola:
+1

PR #2362, opened 2017-08-24
Tracklet cuts and basic support for tracks

cbourjau:
- changing the default eta bin with
- cutting on dphi of the tracklets
- disable additional pileup cuts


aohlson:
+1

PR #2372, opened 2017-08-25
Consider Trigger CCUP25 for LHC17 data

pbuehler:
. CCUP25 is counted as function of run number
. information about TOFMaxipad hits is stored


pbuehler:
+test

PR #2376, opened 2017-08-25
Added alpha parameter to constituent subtraction

haandrew:
Added ability to invoke the alpha parameter in constituent subtraction to apply pT weighting to subtraction.

### DIFF
--- a/PWGJE/EMCALJetTasks/AliFJWrapper.h
+++ b/PWGJE/EMCALJetTasks/AliFJWrapper.h
@@ -142,6 +142,7 @@ class AliFJWrapper
   void SetMinJetPt(Double_t MinPt) {fMinJetPt=MinPt;}
   void SetEventSub(Bool_t b) {fEventSub = b;}
   void SetMaxDelR(Double_t r)  {fUseMaxDelR = kTRUE; fMaxDelR = r;}
+  void SetAlpha(Double_t a)  {fUseAlpha = kTRUE; fAlpha = a;}
 
  protected:
   TString                                fName;               //!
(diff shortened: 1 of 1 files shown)


aiola:
Hello @haandrew,
I would suggest to simplify the implementation.
If I understand correctly both the alpha and deltaR parameters can be set in the constructor of fj::contrib::ConstituentSubtractor which is called in CreateConstituentSub and CreateEventConstituentSub. I would remove the fields fUseAlpha and fUseDeltaR and simply assign default values to fAlpha=0 and fDeltaR=-1 (defaults in the constructor of fj::contrib::ConstituentSubtractor).
Then use the values fAlpha and fDeltaR in CreateConstituentSub and CreateEventConstituentSub.
It would be good if the other people involved in the jet shape analysis could comment and/or approve @lcunquei, @nzardosh 
Best,
Salvatore

nzardosh:
Hi Salvatore and Harry,

Yes I agree with your comment Salvatore, it seems a good way of doing it. My only question is with the lines: 

else                 { fConstituentSubtractor = new fj::contrib::ConstituentSubtractor(fBkrdEstimator); }

in both CreateConstituentSub() and CreateEventConstituentSub(). If the fAlpha and fDeltaR are added here directly into the constructor then does a mass version of fBkrdEstimator need to also be added or defined? I'm asking becasue in the fastjet page (http://fastjet.hepforge.org/svn/contrib/contribs/ConstituentSubtractor/tags/1.2.0/ConstituentSubtractor.hh) it says that if a fj::JetMedianBackgroundEstimator option is used both the pt and mass background are calculated together but it doesnt say whether the mass part of the constructor can be skipped or not (I assume this is not allowed in C++?). This is my only question that needs to be looked into otherwise I agree with your comment.

Thanks,
Nima

haandrew:
Hi @aiola, I have made the changes you recommend I agree that this is a neater way of doing it. Is it okay to merge the pull request like this or do you need a new pull request? 

RE Nimas comment I have left this call of the constructor unmodified as this functionality is only required for cases when we provide external background. 

Cheers,

Harry

lcunquei:
Hello, looks fine to me, I will approve it

lcunquei:
+1

PR #2382, opened 2017-08-26
Update TRD classes for Run-2 analyses

zwound40:
In Run-2, the wake-up trigger was not used anymore for the TRD triggers, so the trigger string changed and does not contain "WU" anymore, but in this class it was still used, so it could not be used for Run-2 data.


tdietel:
+1

iselyuzh:
+1

PR #2402, opened 2017-08-29
PWGGA/PHOSTasks/UserTasks: Update on AliAnalysisTaskPHOSCluster

akraiker:
- Updated initialization order of variables
- Updated initialization parameters of several variables to -1 instead of 0
- Added an if condition to remove clusters with no energy
- Updated class version


kharlov:
+1

PR #2408, opened 2017-08-29
PWGGA/GammaConv: Added additional histos for PhotonQA

hmurakam:
- 2D histograms were added for photo QA. The one is R vs pt and the other one is dcc_z vs pt


dmuhlhei:
Why do we need the additional histograms in the GammaConvV1? We do have a dedicated PhotonQA task, which produces all necessary output for QA.

FriederikeBock:
Hi Hikari,
I agree with Daniel, that there is not need to introduce these histograms, in particular as the variable should already be contained in the DCA tree and can the histograms can be build from that.
Adding these histograms will only increase the memory consumption of the task.
Regards,
Friederike

hmurakam:
Hi Daniel, FriederikeBock,

thank you for your comments. as long as I see, tree of QA task doesn't contain dca_z and DCAtree of pi0 task doesn't contain R distribution.  I wanted to look at correlations between R,dca_z and pt at the same time. Then do you mind adding dca_z distribution to the tree of QA task?

hmurakam:
Hi, Daniel, Friederike,
I wanted to close this request according to your comments, but I did another commit (add train config) and open the pull request which is nothing to do with this pull request. super sorry and anyway once I close this pull request.

PR #2414, opened 2017-08-29
Added alpha setting to ConstSub utility

haandrew:
- Added ability to set alpha parameter in constituent subtractor utility 
- Protected against potential variable initialisation error in EventSubUtil

### DIFF
--- a/PWGJE/EMCALJetTasks/AliEmcalJetUtilityConstSubtractor.h
+++ b/PWGJE/EMCALJetTasks/AliEmcalJetUtilityConstSubtractor.h
@@ -46,14 +46,14 @@ class AliEmcalJetUtilityConstSubtractor : public AliEmcalJetUtility
   TString                fRhomName;                           // name of rhom
   Double_t               fRho;                                // pT background density
   Double_t               fRhom;                               // mT background density
+  Double_t               fAlpha;                              // pT weight exponent applied in const sub
+  Double_t               fMaxDelR;                            // Max distance between ghost and constituent pair in subtraction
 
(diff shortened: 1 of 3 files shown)


PWGJE/EMCALJetTasks/AliEmcalJetUtilityConstSubtractor.h
```diff
@@ -49,6 +51,8 @@ class AliEmcalJetUtilityConstSubtractor : public AliEmcalJetUtility
   TClonesArray          *fParticlesSub;                       //!subtracted particle collection
   AliRhoParameter       *fRhoParam;                           //!event rho
   AliRhoParameter       *fRhomParam;                          //!event rhom
+  Double_t              fAlpha;                              //!pT weight exponent applied in const sub
```
aiola: Hello,
are you sure that you want these fields as transient members (//!) ?
This means that the content of these fields won't be kept when an instance of this class is stored in a root file (which is necessary when the task is used on the grid).
If you change them into non-transient members (i.e. remove the ! after //) remember to increase the class version (last line of this file, ClassDef).
Best,
Salvatore

haandrew:
Hi @aiola ,

I will change these to non-transient members as this was indeed a mistake. What exactly should I increase the class version to in this case? It is currently 1..

Cheers,

Harry

aiola:
Hello @haandrew,
it's enough to increase it to 2. The class version must be increased by one unit every time any of the non-transient member is added, deleted or changed.
Best,
Salvatore

haandrew:
Okay thanks I have fixed the issue.

aiola:
+1

PR #2421, opened 2017-08-30
update D* QpPb systematics

cbedda:
update D* QpPb systematics (multiplicity weights not implemented in the previous version and PID updated after the approval)


zconesa:
+1

PR #2423, opened 2017-08-30
Removal of daughter tracks from q2 computation on the fly

fgrosa:
Removal of daughter tracks from q2 computation on the fly


fprino:
+1

PR #2424, opened 2017-08-30
Fix bash warnings

dstocco:


dstocco:
Dear @dberzano and @ktf ,
this is a rather trivial commit (just modifies a .sh). So the compilation is bound to succeed.
However, I get an aliDoctor error.
What's the issue exactly?

Thanks in advance,
ciao,
Diego

dberzano:
Hi, the problem you see is unrelated to your commit. It will clear out automatically, don't worry. Sorry for the inconvenience!

dstocco:
Ciao Dario,
ok, thanks for the info!
Ciao,
Diego

PR #2426, opened 2017-08-30
example for custom NanoAODSetter

lhusova:


jgrosseo:
+1

jgrosseo:
@dberzano 

Is the "aliDoctor error" (see below) transient? If not, where one can see the output?

image

dberzano:
It is transient and does not depend on the PR, please ignore it.

jgrosseo:
Hi @dberzano 
Why didn't this make it in yesterday's tag? It was approved at about 3pm, although the merge happened much later...

dberzano:
Because the test was marked as failed (red) at 4pm, and we do not wait for failed PRs to be green before merging them. Only yellow/green tests before 4pm are considered.

This is valid even when the problem is temporary and not due to the PR itself, as for this one.

jgrosseo:
So, it this right:
transient aliDoctor error -> PR red -> ignored at 4pm

Then, this problem with aliDoctor is severe. Does this affect all PR? Can we disable this, until the issue is fixed.

We should accept all PRs which are approved before 4pm and not yet processed...

dberzano:
Hi @jgrosseo, the catch is that is was impossible for the system in this case to tell between an error due to the PR itself, or to something else.

Let me point out, for completeness, that I believe that the problem was not severe:

* the figures tell that it affected 3 PRs since we started the system, so out of > 2400 PRs
* the system self-recovered, in spite of a problem not caused by our faults (this was a temporary connectivity problem on some central IT service)
* 3 PRs will be included today instead of yesterday

Your proposal would unfortunately make anybody opening a truly faulty PR to delay - indefinitely - the daily tag (and faulty PRs happen all the time, see #2429 for instance, and most of the time they are not fixed immediately).

We have now put in place a proper solution, instead of a workaround, that can better tell between temporary errors and true PR errors, so no need to worry for the next time, this problem will not reoccur, which is I guess what you want :-)

Thanks for your patience and apologies for the inconvenience!

jgrosseo:
Hi @dberzano. Exactly this was my worry. Thanks that it is fixed!

PR #2432, opened 2017-08-31
PWGGA/GammaConv: added pi0 mass window cut variations

fjonasALICE:
- added pi0 mass window cut varations for NeutralMesonPiPlPiMiPiZero task


nschmidtALICE:
+1

PR #2439, opened 2017-08-31
Eta jet and D rapdity added to the output THnSpars

btrzecia:
Eta jet and D rapdity added to the output THnSpars, for Monte Carlo and data studies.


aiola:
+1

PR #2442, opened 2017-08-31
PWGDQ new class for Event vs Track histograms

pdillens:
New class in the dq-framework, to create histograms plotting event variables versus track variables. Currently only the ITS-TPC matching efficiency is implemented but further options can now more or less easily be added. Although one always has to implement the recalculation of the event variable.


pdillens:
+test

dberzano:
Ignore the error. Temporary problem.

PR #2474, opened 2017-09-04
PWGGA/GammaConv:Modified weights for LHC16h3 due to transcription error

hmurakam:
modify scaling factor of JetJet LHC16h3 due to transcription error.


dmuhlhei:
+1

PR #2487, opened 2017-09-05
Use Abs(label) to take the MC track

rbailhac:
Like to check the pdg of the associated MC track in the function AliDielectronMC::IsMCTruth take the abs(label) of the reconstructed track to get its corresponding MC track.


tdahms:
+1

PR #2489, opened 2017-09-05
PWGGA/GammaConv: introduced likesign and pi0 sideband mixing to MesonCuts

fjonasALICE:
- likesign and pi0 sideband mixing is now set in meson cutstring rather than in MixMode option


nschmidtALICE:
+test

nschmidtALICE:
+1

PR #2509, opened 2017-09-06
major revision of initial task

FlHerrmann:
change of analysis flow, new anlaysis objects (correlations hadron-hadron, correlations with missing partner), MC implementation, etc.


fcolamar:
+1

PR #2511, opened 2017-09-06
New calibration OADBs for LHC16k, LHC16o and LHC16p

rderradi:
Added new multiplicity selection calibration OADBs for LHC16k, LHC16o and LHC16p.
Also updated the run ranges for the corresponding periods in AliMultSelectionTask.cxx.


ddobrigk:
+1 (thanks, Rafael!)

PR #2512, opened 2017-09-06
Missing protection in the IsJPsiPrimary function 

rbailhac:
Missing protection in the AliDielectronMC::IsJpsiPrimary function.
This did not show up until now since the track were anyway rejected by the HaveSameMother condition rejecting track with negative label. After the change in HaveSameMother this is crashing.
Add the protection on the two daughters for the AliDielectronPair in IsJpsiPrimary. In this way, this function does not change at all, even after the change in HaveSameMother.
(Sorry)


tdahms:
+1

PR #2515, opened 2017-09-07
Jet-hadron: Fix warnings and rename

raymondEhlers:
Updates to my personal task to fix warnings and to rename the task to clarify it's purpose (with the previous names symlinked since they are referenced in analysis notes, etc).


lcunquei:
+1

PR #2521, opened 2017-09-07
PWGDQ: Updated efficiency addtask

carstooon:
Small update to allow several wagons of the same addtask within one LEGO train


pdillens:
+1

PR #2544, opened 2017-09-08
Remove return value to fix root 6 compilation

raymondEhlers:
Function was given a return type of float, but doesn't return a value.
This changes the return to void.

@lcunquei - Can you please confirm that this is the desired way to change this?


aiola:
+1, looks good to me. Certainly it does not change the behaviour in a negative way (avoids random number returned by the method).

PR #2525, opened 2017-09-08
PWGCF: AliFemtoConfigObject - Introduce new class for "dynamic" data storage

akubera:
AliFemtoConfigObject is a TObject subclass that stores dynamic data in a JSON-like builtin types.

The goal of the type is to facilitate setting analysis configuration parameters and provide a means for storing these parameters in ROOT file (in a file-mergeable way).

PR #2545, opened 2017-09-08
Fix OS X compilation

raymondEhlers:
To compile on OS X, needs <numeric> included for accumulate.

Some fixes for root 6 are also needed, but the implementation is not yet clear to me - see [here](https://github.com/alisw/AliPhysics/pull/2525#issuecomment-328223744)

### DIFF
--- a/PWGCF/FEMTOSCOPY/AliFemtoUser/AliFemtoConfigObject.cxx
+++ b/PWGCF/FEMTOSCOPY/AliFemtoUser/AliFemtoConfigObject.cxx
@@ -12,6 +12,7 @@
 #include <iostream>
 #include <exception>
 #include <functional>
+#include <numeric>
 
 // #include <TIter.h>
 #include <TList.h>


akubera:
Oh, I thought I checked with both `master` and `IB/master/root6` branches of alidist (on ubuntu) before pushing, but the streamer stuff was added late so I might not have.

From your [other comment](https://github.com/alisw/AliPhysics/pull/2525#issuecomment-328223744) the `incomplete type 'TBuffer'` error implies that `#include <TBuffer.h>` should be included at the top of the cxx file. Could you add that for me and see if it fixes the issue?

rbertens:
+1

raymondEhlers:
@akubera, Thanks for the suggestion on TBuffer - that fixed it! (I should have looked a bit more closely, but I was distracted by other things at the time). I created a pull request to resolve it ( #2552 ).

PR #2551, opened 2017-09-09
Update to get deltaPhi for Spherocity axis

mbjadhav:
Adding a 3D histogram to analyse deltaPhi between phi angle of spherocity axis and D-meson direction

### DIFF
--- a/PWGHF/vertexingHF/AliAnalysisTaskSEDvsEventShapes.cxx
+++ b/PWGHF/vertexingHF/AliAnalysisTaskSEDvsEventShapes.cxx
@@ -84,6 +84,7 @@ fHistnTrackvsEtavsPhi(0),
 fHistnTrackvsEtavsPhiEvWithCand(0),
 fHistTrueSovsMeasSo(0),
 fHistTrueSovsMeasSoEvWithCand(0),
+fHistSpheroAxisDeltaPhi(0),
 fSparseEvtShape(0),
 fSparseEvtShapewithNoPid(0),
 fSparseEvtShapePrompt(0),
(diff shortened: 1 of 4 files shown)


fprino:
Hi Manoj,

I had a look at the modified code.
There are two points which need to be modfied:
In the task when you compute
Double_t deltaPhiRef = (phiRef-phiCand);
you should then modify it to have it in the range from -pi to pi or equivalently from 0 to 2*pi to match the periodicity of the azimuthal angle.
The same range you should then use to bin the axis of the 3D histogram.

Cheers, Francesco

mbjadhav:
Ciao Francesco, 

To get it in 0 to 2*pi, I can take the absolute of (phiRef-phiCand). But When I checked the distribution of deltaPhi locally, it shows three peaks (it is with very low statistics) at 0 (D0 along sphero axis), pi and -pi (D0 opposite to sphero axis). I emailed you the plot. In such a case we cut any one of the peak distributions? Won't it affect much?

Thank you.

fprino:
Hi Manoj,

what you have to do to apply the periodicity in phi is something like:
if(deltaPhiRef<0) deltaPhiRef+=2*TMath::Pi();
if(deltaPhiRef>2*TMath::Pi()) deltaPhiRef-=2*TMath::Pi();

which is different from taking the absolute value.

Cheers, Francesco

mbjadhav:
Ciao Francesco, 

I have updated the task to implement above suggestions. Thank you.

fprino:
Hi Manoj,

there is still one minor change that is needed on the upper edge of the phi axis in the 2D histos.
What you have now as upper range 6.28 is a bit less than 2pi, so there will be unwanted overflows.
You should use 0. and 2*TMath::Pi() as the axis limits.

Cheers. Francesco

mbjadhav:
Ciao Francesco, 

Yes, Sure. Updated it. 

Thank you.
Manoj

fprino:
+1

mbjadhav:
Thank you, Francesco!

PR #2552, opened 2017-09-09
Fix ROOT 6 compilation

raymondEhlers:
Explicitly add TBuffer include. Thanks @akubera for the suggestion! I should have looked more closely at the includes and the TBuffer header to see what the problem was.

Note that you also need #2544 to compile with ROOT 6 at the moment.


akubera:
+1 - Thanks for checking this, I'm glad it was an easy fix!

PR #2570, opened 2017-09-11
add TRU bad maps in LHC15no

dsekihat:
Update TRU bad map in LHC15n and put empty TRU bad map in LHC15o for QA purpose.


kharlov:
+1

PR #2573, opened 2017-09-12
allow user to include cells from rejected clusters, and a couple binn…

jdmulligan:
…ing improvements (updates to my user task).


aiola:
+1

PR #2596, opened 2017-09-13
Add the possibility to select INEL>0 event

mpuccio:
In addition to this I had to add a missing include in AliMultSelectionTask to fix the compilation when including its header.


fprino:
+1

PR #2616, opened 2017-09-14
Update improver adding tuning of the pulls

afestant:
Update improver adding tuning of the pulls and flags to switch it on/off (same for the covariance matrix)


arossi81:
Ciao Andrea,
increase the ClassDef. 
I guess that the modification (0.85 hard coded) is still a temporary one to check the impact on the D efficiency running on train, right?
ciao
Andrea

afestant:
Ciao Andrea,
I increased the ClassDef.
Yes, the 0.85 factor is not final, I committed this version to be able to run on trains but once we will have decided the final strategy we can change it.
Cheers,
Andrea

arossi81:
+1
Thanks!

PR #2622, opened 2017-09-15
Added low-pT extraction method

rhaake:
This introduces an "underflow" mechanism to the jet extractor to allow extraction of lowest-pT jets with reasonable statistics.

The commit also fixes some warnings.


fcolamar:
+1

PR #2628, opened 2017-09-15
Fix warnings in the jet finder

raymondEhlers:
All trivial modifications to fix warnings from -Wall, including:
- Initialization order
- Unused variables


aiola:
+1, thanks for fixing!

PR #2634, opened 2017-09-16
Systematic unc. for Ds in MB pp 13 TeV (LHC16k,l)

jhamoncern:
Systematic uncertainties for Ds in MB pp 13 TeV (LHC16k,l), approved for Initial Stages.


fprino:
+1

PR #2636, opened 2017-09-17
PID optimization

panjinjin:
1. added PID veto cuts
2. modified pT upper limit for use_p_cut
3. added centrality groups


aohlson:
+1

PR #2639, opened 2017-09-18
Add possibility to fill THnSparse at AccStep level

jhamoncern:
- New method, FillMCAccHistos, to MC-related THnSparse at AcceptanceStep level. The method is similar to FillMCGenAccHistos, which fill MC-related THnSparse at Gen+Acc level.
- Remove 'Double_t massPhi' variable, already defined above in the code.
- Comment unused variable 'Double_t normdl'

### DIFF
--- a/PWGHF/vertexingHF/AliAnalysisTaskSEDs.cxx
+++ b/PWGHF/vertexingHF/AliAnalysisTaskSEDs.cxx
@@ -917,13 +917,7 @@ void AliAnalysisTaskSEDs::UserExec(Option_t */*option*/)
     Double_t zMCVertex = mcHeader->GetVtxZ();
     if (TMath::Abs(zMCVertex) > fAnalysisCuts->GetMaxVtxZ())
       return;
-
-    if (fFillAcceptanceLevel)
-      // Fill MC histos for cuts study at AccStep
-      FillMCAccHistos(arrayMC, mcHeader, nTracklets);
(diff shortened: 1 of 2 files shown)


fprino:
Hi Julien,

can't we avoid the huge code duplication in the new method FillMCAccHistos, which is almost identical to the existing FillMCGenAccHistos? 
For example, one could consider to pass some options to the existing FillMCGenAccHistos to treat the different selections.
In general, it is better to avoid code duplications whenever possible to avoid extra work and possible mistakes in the code maintenance.

Thanks!

Cheers, Francesco

jhamoncern:
Hi Francesco,

Indeed it is possible to modify directly FillMCGenAccHistos, with the help of a flag. However the running time of this function would consequently increase: the test of the flag would necessarily be performed inside the loop over AODMCParticles.
I chose the option of saving the running time by including a new function. Would you prefer to be safer (in terms of maintenance) but slower? I don't have a strong opinion.

Cheers, Julien

jhamoncern:
Hi Francesco, all,

The new function (FillMCAccHistos) has been merged to the standard one (FillMCGenAccHistos), as you suggested. A flag allow to select candidates at AccStep or Gen+AccStep.
I tested the change locally and everything works.

Cheers, Julien

fprino:
+1 and thanks Julien for considering my suggestion

PR #2644, opened 2017-09-18
Simplify normalisation histogram code

mpuccio:
The normalisation histogram code has been simplified by putting some
commonly used bitmask as general enum of AliEventCuts.
This modification allows the user to extract these information using the
new `bool AliEventCuts::CheckNormalisationMask(AliEventCuts::NormMask mask)`
method.


fprino:
+1

PR #2647, opened 2017-09-18
PWGDQ EvtVsTrk now only evts aft evt cuts are used

pdillens:
The placement of the filling of the EvtVsTrk object in the AliDielectron::Process function has been revisited, so that now the filling happens after the event cuts passed to the Dielectron object by the user.


pdillens:
+1

PR #2653, opened 2017-09-19
PWGDQ EvtVsTrk now only evts aft evt cuts are used

pdillens:
Repositioned the filling of the EvtVsTrk object in the AliDielectron class. Now it is filled after the current event had to pass the event cuts, setted by the user.


pdillens:
+1

PR #2659, opened 2017-09-19
deuteron dataskim, removed fiducial cut, added std TOF/TPC 4-sig PIDresponse cut

bschaefe:
This skim-data creator is 4-sigma standard cuts from Kalweit and Sharma (Nuclex group).  It is deliberately broad to be inclusive and refinable later.


aohlson:
+1

PR #2661, opened 2017-09-19
add distance to the closest bad cell in floating point

dsekihat:
add distance to the closest bad cell in floating point


kharlov:
+1

PR #2679, opened 2017-09-21
Fixing operator ambiguity

mfasDa:
Due to the implementation of the casting operators and
arithmetic operators with int/float types in TComplex,
there is an ambiguity on whether to use
- operator- from TComplex with TComplex and int
- operator- from int with TComplex implicitly casted to int
In order to fix the ambiguity the complex value needs to
be explicitly defined, or the integer can be converted to
complex using the value as real part of the complex number.
The proposed solution does the subtraction on complex level.

Needed for ROOT6 builds.


rbertens:
+1 @jaelpark this commit is needed so that your code can be compiled under ROOT6 (for which it currently breaks compilation). It should not change anything in functionality (a 'normal' 1 is replaced by its complex notation TComplex(1,0), but if you think it's problematic, please respond. 
Thanks,
Redmer

PR #2689, opened 2017-09-21
Add static method to check whether the AOD primary vertex is from TPC

mpuccio:


fprino:
+1 and many thanks for taking care of this implementation!

PR #2694, opened 2017-09-22
Make AOD vertex selection optional + fix integer overflow

mpuccio:
The selection of good AOD vertex is on by default, but it can be switched off by the user.


fprino:
+1

PR #2707, opened 2017-09-25
fix run number range for LHC16l in Lc->V0 TMVA task

jezwilkinson:
Run number range for LHC16l in AliAnalysisTaskSELc2V0bachelorTMVA::GetEstimatorHistogram() was incorrect, meaning that in LHC16l the multiplicity distribution was not being corrected at runtime. This patch should fix it.


zconesa:
+1

PR #2715, opened 2017-09-26
Fixes needed for compilation with Xcode 9

mfasDa:
Similar ambiguity issue introduced by Xcode 9 as in AliRoot PR 417 (https://github.com/alisw/AliRoot/pull/417). 

@akubera: Due to the ambiguity the constructor generation in AliFemtoConfigObject doesn't work for TString, as an explicit cast is needed in order to resolve the ambiguity. The constructor is added manually.


pzhristov:
+1

PR #2717, opened 2017-09-27
Pile-up rejection aligned with AliMultSelection for Run2 pp

mpuccio:


fprino:
Hi Max,

it seems from the diff files that you have changed the order of the data members in the header file.
Shouldn't one increase the class revision of the class in such cases?

Cheers, Francesco

mpuccio:
Ciao Francesco,

right, I fixed that. Thanks for checking!

Max

fprino:
+1

PR #2718, opened 2017-09-27
Improved jet extraction underflow bins

rhaake:
Code change is to give more options for jet extraction below the min pT threshold. In this case, a jet is pu into low-pT underflow bins under certain optional conditions.


ebruna:
+1

PR #2726, opened 2017-09-28
Fixed removal of dau tracks of the single candidate from q2 and added mass selection

fgrosa:
Before this fix for the first candidate its daughter tracks were removed from the q2, for the second candidate its daughter and those of the first candidate were removed and so on. With this fix, for each candidate only its daughter tracks are removed from the q2.


fprino:
+1

PR #2728, opened 2017-09-28
PWGDQ: EfficiencyTask with more resolution plots

carstooon:
PWGDQ:
added plots for different detector resolutions as a function of pT


otonvd:
+1

PR #2734, opened 2017-09-28
Optimize rotational background opening angle distribution.

moliver813:
Optimize rotational background opening angle distribution.
Make rotational background loop safer.


efranzis:
+test

dlodato:
+1

PR #2736, opened 2017-09-28
use ESDrun->InitMagnetic field instead of copy/paste

shahor02:
Just to avoid patching already copy-pasted code for the fix in the B-field initialization: it is better to use the source method which is already fixed.


gconesab:
+1

PR #2738, opened 2017-09-29
Install the full folder macros in the CMake file

dcolell:


dcolell:
Dear @ktf and @dberzano,
it's the first time I end-up with an error during test.
The compilation of AliPhysics on my laptop worked fine after the modification.
I would say the error is not connected with the change I did.
As far as I can see the last change to the code mentioned in the Log dates nine months ago.
Do you suggest me to send a mail to the task force mailing list?
Thank you for any help.

Cheers,
  Domenico.
  (cc @gbencedi)

PR #2767, opened 2017-10-03
Introduce a new YAML configuration class

raymondEhlers:
Introduces a new configuration class, AliYAMLConfiguration, that provides an interface which allows users to read and write values in YAML files. It is refactored and generalized from the functionality in the EMCal Corrections Framework.

Given the generalization, it was moved to PWG/Tools, with the hope that it can benefit other users. Documentation on usage is available in the class header.

It has been tested and validated on ROOT 5 and 6 on macOS (thanks @jdmulligan for help testing!), but given the use of templates, it may cause errors with gcc, so it may take a few iterations.


aiola:
+test

raymondEhlers:
Great - code review is appreciated!

That bring said, I'd appreciate it if this could be merged in the next day or two - I'm trying to add a feature which depends on this class. Thanks!

aiola:
+1

PR #2776, opened 2017-10-03
Added two more origin bins in MC histogram

amatyja:


gconesab:
Hi Adam,

no need to close, this is a general problem I have seen the last days, with other messages. Just be patient it will eventually pass. Let's see in your other commit.

Gustavo

amatyja:
Hi @gconesab , 
I was afraid it is my fault and I forgot to update something. I have just redo the commit after git pull updates to the most recent. But there is the same problem. I haven't noticed that error earlier. I leave it open now.
Regards
Adam

PR #2778, opened 2017-10-03
Added two more origin bins in MC histogram

amatyja:


dberzano:
Note that this error's unrelated, it will clear out automatically.

PR #2788, opened 2017-10-05
Small bug fixes for AliYAMLConfiguration

raymondEhlers:
Two minor bug fixes:
- Handle corner case in retrieving a specialized property
- Fix argument order for writing a configuration

@aiola - The corner case fixed here is subtle, but without the fix, it makes a common embedding configuration difficult, so could you please merge this before the tag? Thanks!

### DIFF
--- a/PWG/Tools/AliYAMLConfiguration.cxx
+++ b/PWG/Tools/AliYAMLConfiguration.cxx
@@ -41,6 +41,7 @@ AliYAMLConfiguration::AliYAMLConfiguration(const std::string prefixString, const
 bool AliYAMLConfiguration::AddEmptyConfiguration(const std::string configurationName)
 {
   YAML::Node node;
+  AliInfoStream() << "Adding configuration \"" << configurationName << "\" as an empty YAML node.\n";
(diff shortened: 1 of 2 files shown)


PWG/Tools/AliYAMLConfiguration.cxx
```diff
@@ -110,7 +111,7 @@ bool AliYAMLConfiguration::AddConfiguration(const YAML::Node node, std::string c
  *
  * @return True if write was successful.
  */
-bool AliYAMLConfiguration::WriteConfiguration(const int index, const std::string filename) const
+bool AliYAMLConfiguration::WriteConfiguration(const std::string filename, const int index) const
```
mfasDa: const std::string & ? (until at some point we would finally be allowed to use std::string_view) Would suppress a copy in case a std::string is passed as argument. Also in other cases.

mfasDa:
+test

aiola:
+1, approving now since I am not sure that I will be able to do it later. However it would be good to implement the trivial change proposed by Markus (just make a new PR)

PR #2789, opened 2017-10-05
Minor Jet Framework fixes

raymondEhlers:
Fix a ClusterAt() -> Cluster() in AliEmcalJet which I missed in a recent change. Note that it only would have caused problems for EMCal embedding, so it didn't adversely affect anyone's results.

Also enable ROOT 6 with the jet response maker

### DIFF
--- a/PWGJE/EMCALJetTasks/macros/AddTaskJetResponseMaker.C
+++ b/PWGJE/EMCALJetTasks/macros/AddTaskJetResponseMaker.C
@@ -101,8 +101,11 @@ AliJetResponseMaker* AddTaskJetResponseMaker(
   jetCont2->SetFlavourCut(jetTagging);
   jetCont2->SetMaxTrackPt(1000); // disable default 100 GeV/c track cut for particle level jets
 
-    
+#ifdef __CLING__
+  jetTask->SetMatching(static_cast<AliJetResponseMaker::MatchingType>(matching), maxDistance1, maxDistance2);
+#else
   jetTask->SetMatching(matching, maxDistance1, maxDistance2);
+#endif
   jetTask->SetVzRange(-10,10);
   jetTask->SetIsPythia(kTRUE);
   jetTask->SetPtHardBin(ptHardBin);
--- a/PWG/JETFW/AliEmcalJet.cxx
+++ b/PWG/JETFW/AliEmcalJet.cxx
@@ -498,7 +498,7 @@ AliVCluster* AliEmcalJet::GetLeadingCluster(TClonesArray* clusters) const
 {
   AliVCluster* maxCluster = 0;
   for (Int_t i = 0; i < GetNumberOfClusters(); i++) {
-    AliVCluster* cluster = ClusterAt(i, clusters);
+    AliVCluster* cluster = Cluster(i);
     if (!cluster) {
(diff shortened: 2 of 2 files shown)


PWGJE/EMCALJetTasks/macros/AddTaskJetResponseMaker.C
```diff
@@ -101,8 +101,11 @@ AliJetResponseMaker* AddTaskJetResponseMaker(
   jetCont2->SetFlavourCut(jetTagging);
   jetCont2->SetMaxTrackPt(1000); // disable default 100 GeV/c track cut for particle level jets
 
-    
+#ifdef __CLING__
+  jetTask->SetMatching(static_cast<AliJetResponseMaker::MatchingType>(matching), maxDistance1, maxDistance2);
+#else
   jetTask->SetMatching(matching, maxDistance1, maxDistance2);
+#endif
```
mfasDa: Why is this necessary? That cling requires the correct type is clear, but the static_cast should in principle do no harm under cint.

mfasDa:
+test

PWGJE/EMCALJetTasks/macros/AddTaskJetResponseMaker.C
raymondEhlers: I actually don't know the reason for doing this - I was inclined to just use the static_cast, but I noticed that you used this approach in a number of places, and I assumed that you had a reason for it. [For example, your commit here](https://github.com/alisw/AliPhysics/commit/363df1459f567e784a188bcc25616177af4cb9cf).  Do you recall what the CINT problem was?

In any case, I'm happy to remove it if it's not needed - just let me know what you prefer.

adriansev: just as a side comment : you can declare this function in a namespace along the .h/.cpp of AliJetResponseMaker code and convert your .C macro to a return MY_NS::AddTaskJetResponseMaker (my_signature ). This way will not matter if it is root 5 or 6 as it is compiled code (and for a usage point of view is equivalent as the AddTask macros are not modified locally, they are used anyway from the $ALICE software locations)

raymondEhlers: Thanks for the suggestion - that is actually what I usually do for my own tasks, but I didn't in this case as I don't really own this task.

@aiola Is it okay with you if I just convert this to a compiled AddTask?

And while we're on the topic - would anyone object to moving the sample task to a compiled AddTask?

raymondEhlers:
Hi All,

Since we're getting into some discussion about the AddTask and the other fix in this pull request to fix cluster access in the jet framework really should make it into the tag, I've separated that fix into another PR: #2801 . Can one you please approve that for the tag, and then we can address the AddTask here.

Best,
Raymond

PWGJE/EMCALJetTasks/macros/AddTaskJetResponseMaker.C
mfasDa: I agree with @adriansev proposal. It has been done now for a good fraction of new classes and I think if we come across we might fix the old ones as well.

raymondEhlers: Okay, I will this close this PR, move some tasks to compiled AddTasks, and then make a new PR.

PR #2801, opened 2017-10-05
JETFW: Fix cluster constituent access

raymondEhlers:
Fix a ClusterAt() -> Cluster() in AliEmcalJet which I missed in a recent
change. Note that it only would have caused problems for EMCal
embedding, so it didn't adversely affect anyone's results.


mfasDa:
+1

PR #2807, opened 2017-10-05
Avoid unnecessary copying and fix warnings

raymondEhlers:
Avoid unnecessary copying of strings by using references (originally
mentioned in #2788 - thanks @mfasDa ). Also fix warning related to signed vs unsigned
comparison


mfasDa:
+1

PR #2808, opened 2017-10-05
Compiled Sample Task and Response Maker AddTasks

raymondEhlers:
Move the EMCal-Jet Sample task and Jet Response Maker to compiled AddTasks, as discussed in #2789 .

@aiola - I'm imagine that you're quite busy right now, so no need to review it - I'm just tagging you to make sure that you're aware and okay with it!


aiola:
+test

raymondEhlers:
I'm assuming that the alibuild error is just a bug and that the PR is fine - let me know if that's not the case

aiola:
+1

PR #2809, opened 2017-10-06
EMCal CorrFW: Add additional config on LEGO train

raymondEhlers:
New function to retrieve an existing correction task so that it can be
configured further in a different wagon on the LEGO train.

To achieve this, a dummy task is created when the configure task is
called because AliAnalysisAlien requires that all wagons add a task.
Then, when Initialize(true) is called on the correction task, the dummy
task is removed.

cc: @mfasDa , @jdmulligan , @aiola

### DIFF
--- a/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionTask.cxx
+++ b/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionTask.cxx
@@ -4,7 +4,6 @@
 //
 
 #include "AliEmcalCorrectionTask.h"
-#include "AliYAMLConfiguration.h"
 #include "AliEmcalCorrectionComponent.h"
 
 #include <vector>
@@ -26,8 +25,6 @@
 #include "AliESDEvent.h"
 #include "AliAnalysisManager.h"
 #include "AliAODEvent.h"
-
-#include "AliAnalysisTaskEmcalEmbeddingHelper.h"
 
 /// \cond CLASSIMP
 ClassImp(AliEmcalCorrectionTask);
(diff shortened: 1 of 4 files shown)


aiola:
+test
Could you elaborate a bit on the goal of this update? Just for my understanding! Thanks.

raymondEhlers:
Hi @aiola ,

I thought that you were on the thread discussing this, but it seems that we missed you. I forwarded you the messages, but I'll also summarize here.

Markus expressed concern about testing and running multiple trains with different YAML configuration files at the same time (which I agree can be quite cumbersome). We discussed a few solutions and came up with this:

- There is a central EMCal Correction Task which configures general settings like physics selection, n cent bins, etc. This wagon should be in the user dependency chain, so something like: Physics Selection <- EMCal Correction Task <- Jet finders <- etc...
- Each individual user will have a "configuration" wagon which retrieves the EMCal Correction Task, sets their configuration file, and then initializes the correction task. This wagon will depend on the global correction task wagon, but nothing should depend on it.

The user documentation is available [here](http://alidoc.cern.ch/AliPhysics/master/_r_e_a_d_m_eemc_corrections.html#emcCorrectionsLEGOTrainWagon).

This way, each user can be responsible for their own config, but we can a mostly centralized wagon. The only downside I see is that it needed some hack to add and remove a dummy wagon (the configure wagon creates it, and Initialize(true) in the correction task removes it) to satisfy AliAnalysisTaskCfg, but it's not such a huge deal.

I've tested locally, but now that it's committed, I'm going to test on the trains.

And of course, if you have any suggestions, they would be greatly appreciated!

Best,
Raymond

PR #2810, opened 2017-10-06
tree and QA histograms are in TList, task name added

nirbhaygit:
Tree and QA historgrams are kept in common output container (TList)
Task name is can be set outside


sashingo:
+1

PR #2816, opened 2017-10-06
Fix AddTask

hzanoli:
The AddTask is case sensitive in the LEGO train system, so it is needed to match the exact name on the AddTask.


aiola:
+1

PR #2820, opened 2017-10-06
Primary vertex position for selected events with D mesons

jhamoncern:
Add histograms monitoring the primary vertex position for selected events with D mesons.
- When running on MC: histograms are filled if the event contains a true D meson (from MC truth).
- When running on Data: histograms are filled if the the D candidate pass through all the selection cuts (topo+PID) and if its reconstructed invariant mass is within +/- 40 MeV/c2 around the PDG value. Currently, it works only for D+ and D0.


fprino:
+1

PR #2828, opened 2017-10-07
resonance studies

marslandALICE:


marslandALICE:
+test

rbertens:
+test
Hi @marslandALICE ,
In the changes that you propose to your AddTask macro, you load libraries and set include paths:
`
 gSystem->Load("libANALYSIS");
 gSystem->Load("libANALYSISalice");
 gSystem->AddIncludePath("-I$ALICE_ROOT/include"); 
 gSystem->AddIncludePath("-I$ALICE_PHYSICS/include");`
`
Usually instructions like these are reserved for a steering macro, so that the addTask only contains the actual instantiation of the analysis task class and configuration of it; gSystem calls can later on interfere with e.g. the LEGO train system. Maybe you should consider removing these statements? 

Thanks,
Redmer

marslandALICE:
Hi Redmer,
The reason for the error in compiling is that the "AliAnalysisTaskFilteredTree.h" is in "$ALICE_PHYSICS/include" can someone add it in the path in cmake ? 
As for the Addtask, I will modify it next week. Right now i am in a workshop.
Cheers,
Mesut

aohlson:
Hi Mesut, 
I see thatAliAnalysisTaskFilteredTree is in the PWGPP directory.  When possible, we try to avoid cross-PWG compilations (i.e. adding a PWGPP or PWGLF directory to the CMakeLists in PWGCF).  Is the functionality that you need available anywhere else, or is this linking necessary?  If so, you can add it to PWGCF/EBYE/CMakeLists.txt.  
Thanks, 
Alice

marslandALICE:
Hi Alice,
Thanks for the answer. From my side i definitely need this class. It contains many useful features, which might be interesting for many people therefore we can, maybe,  put it somewhere else. I will discuss it with Marian. Since the internet connection is terrible here in china, i can make the pull request next week once i am back. 
Cheers,
Mesut

PR #2839, opened 2017-10-09
Nano AOD update

mazimm:
introduce new offline trigger and fired Trigger Classes variables to the NanoAODs. 
remove some compiler warnings from the code.


rbertens:
+1

PR #2842, opened 2017-10-10
Fill Et-M02 histo only for neutral clusters

ReneChristianSchach:


e-masson:
+test

e-masson:
Hi @ReneChristianSchach, there seems to be a problem in the tests, but I am not sure this is linked to your commit since this is not the first time I see this kind of error since yesterday...
Just to be sure, could you please double check your implementations?
Best,
Erwann

ReneChristianSchach:
Hi @e-masson , I cannot access the error log so I have no idea what went wrong, but I have tested the changes locally before committing them and there were no problems. The changes are also rather small, I do not see how they could be problematic. 

Since all checks seem to be successful anyway, I think this error message can be ignored.
Best,
Rene

e-masson:
Hi again @ReneChristianSchach, I think so. I cannot access the log either, but I think something is going wrong at an higher order (several mails were sent by people who had problems with AliRoot since yesterday).
Best,
Erwann

ReneChristianSchach:
Hi again @e-masson , can you approve the merging anyway?

Best,
Rene

e-masson:
Yes of course @ReneChristianSchach, sorry for missing!
Best,
Erwann

PR #2844, opened 2017-10-10
Updating improver task adding pulls sigma correction

afestant:
Add pt dependent correction of the sigma of the pulls and possibility to mimic in the MC the shift of the d0 mean observed in data (same sign)


arossi81:
+1

PR #2857, opened 2017-10-11
PWG/Cocktail: Added first cocktail parameterizations for pp5TeV

nschmidtALICE:
File contains interpolations to 5 TeV using 2.76 and 7 TeV as input. Included particles are pi+-, K+-, p/bar{p}, lambda/bar{lambda}, phi and K0*.
In addition, the pi0 and eta measured with EMCal are also included.


FriederikeBock:
+1

PR #2858, opened 2017-10-11
PWGGA/GammaConv  Updated CutString

AMechler:
Changed CutString for DCal analysis to new DCal eta range setting
Changed PCM-DCal CutString to new DCal eta range setting
Added NonLin variation for PCM-DCal


nschmidtALICE:
+1

PR #2851, opened 2017-10-10
Updated LMee cocktail

otonvd:
Updated LMee cocktail code:
 - DCA simulation
 - Virtual photon simulation
 - Resolution for runII included
 - Updated macro to run the simulation (runStandaloneLMeeCocktail.C)

PR #2862, opened 2017-10-11
Fix macOS compilation

raymondEhlers:
I still have a problem with "PWGDQ/dielectron/LMEE/AliAnalysisTaskLMeeCocktailMC.cxx", which I noted in the #2851

### DIFF
--- a/PWG/DevNanoAOD/AliAnalysisNanoAODCutsJet.cxx
+++ b/PWG/DevNanoAOD/AliAnalysisNanoAODCutsJet.cxx
@@ -1,4 +1,6 @@
 
+#include <TClonesArray.h>
+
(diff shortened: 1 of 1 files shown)


mazimm:
I committed another fix suggested by @mfasDa. You can see it here. It will be merged soon. 

https://github.com/alisw/AliPhysics/pull/2863

mazimm:
+test the other fix failed the last validation, so we can use this one. I cannot test this locally because I use only ROOT 5.

mazimm:
+1

raymondEhlers:
Thanks for the quick fixes! These combined with deleting my AliPhysics build directory and starting from scratch seemed to resolve the problems that I was experiencing.

PR #2865, opened 2017-10-12
HFEminiTree beauty task added and order changed

nirbhaygit:
HFEminiTree beauty task added and the order changed compared to last commit by Andrea


sashingo:
+test

PR #2868, opened 2017-10-12
PWG/Cocktail: fixed bad pi0 param for PCM pp 8TeV in cocktail

nschmidtALICE:
Unfortunately the low momentum part of the pi0 PCM parameterization for pp 8 TeV was significantly overestimated. This update fixes that problem.


FriederikeBock:
+1

PR #2870, opened 2017-10-12
fix memory leak

mazimm:


rhaake:
+1
Markus and I fixed a memory leak in the NanoAOD arraymaker. Tested to work with ROOT5.

PR #2873, opened 2017-10-13
Improvements to YAML config class

raymondEhlers:
Includes an improved interface and refactoring additional utility functions from the EMCal Correction Framework to the YAML configuration class


aiola:
+test, it looks fine to me, maybe @mfasDa wants to take a look as well

raymondEhlers:
That error is rather difficult to interpret, but I think I just fixed it with my latest commit (of course, it doesn't show up on clang, so it's a bit difficult to debug)

mfasDa:
+test

If the error is fixed I would think it should be approved.

aiola:
+1

raymondEhlers:
Thanks @aiola and @mfasDa !

PR #2893, opened 2017-10-13
Move YAML config class to PWG::Tools namespace

raymondEhlers:
Added to the namespace for safety


mfasDa:
+1

raymondEhlers:
@mfasDa - Any suggestions on how to resolve this? I tried to google around, but this seems to be a CINT problem, and I can't seem to fix it

raymondEhlers:
Okay, I finally found the solution (resolved by moving operator<< back into the global namespace. see: https://stackoverflow.com/a/38801633) - it should be fixed now.

aiola:
+1

PR #2907, opened 2017-10-15
change (eta vs. phi) histogram binning

dsekihat:
change eta vs. phi histogram binning. 3.2-> TMath::Pi(), 6.3->TMath::TwoPi()


kharlov:
+1

PR #2930, opened 2017-10-16
EMCal Embed: Implement yaml config in order to (optionally) set an em…

jdmulligan:
…bedded runlist

@raymondEhlers


raymondEhlers:
Looks good! Thanks for implementing!

PR #2936, opened 2017-10-17
Modified version to default tune (Monash 2013)

gbencedi:
Modified version to default tune (Monash 2013). 
Additional settings of BeamRemnants.


ycorrales:
+1

PR #2942, opened 2017-10-17
task for first look into He4 candidates in pp collisions

mhbib:


mpuccio:
Hello @mhbib!

The task that you are adding will not be compiled unless you modify the CMakeLists.txt and the LinkDef accordingly. Let me also suggest to have a look to the tasks that are already available in AliPhysics for the 4He analysis in order to avoid code duplication.

Cheers,
Max

mhbib:
Hello Max,
thank you for the comment! I added the task in the MakeLists and LinkDef. I will talk to Benjamin soon, but I hope it is okay to leave it like this for now - just to have a first look into the data.
Cheers,
Michael

jgronefe:
+1

PR #2950, opened 2017-10-17
Match AddTask with compiled version

raymondEhlers:
Update of my personal task.

The compiled AddTask was modified without modifying the corresponding pass
through call for the LEGO train.


raymondEhlers:
+1 I'm not waiting for review as this is just a trivial bug fix

PR #2952, opened 2017-10-17
update additional weighting to pi and K

dsekihat:
update additional weight to pion and Kaon in order to reproduce realistic K/pi ratio


kharlov:
+1

PR #2953, opened 2017-10-17
PWGDQ: Added option for V0And trigger and pileup rejection to Addtask

jjungIKF:
PWGDQ: Added option for pileup rejection and V0AND trigger to personal AddTask


pdillens:
+1

PR #2961, opened 2017-10-18
update K/pi ratio in P8 and P6 separately

dsekihat:
update K/pi ratio in Pythia8 and Pythia6 separately to reproduce real K/pi ratio


kharlov:
+1

PR #2966, opened 2017-10-18
Fix a bug where the normalized d0 was not stored as an asbolute value…

aiola:
… + throw an expection if the event pointer is not set (should not happen)


aiola:
+1

aiola:
+1

aiola:
+test

aiola:
+1

PR #2973, opened 2017-10-18
EMCAL: Fix EMCal container titles

raymondEhlers:
The titles were not getting the current minimum pt setting correctly.
Separating the definition and assignment seems to resolve the issue.
This is particularly important as the titles are relied upon to find
jet collections.

Also updates the branches print out

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliMCParticleContainer.cxx
+++ b/PWG/EMCAL/EMCALbase/AliMCParticleContainer.cxx
@@ -238,6 +238,7 @@ const AliMCParticleIterableMomentumContainer AliMCParticleContainer::accepted_mo
  */
 const char* AliMCParticleContainer::GetTitle() const
 {
-  static TString trackString = TString::Format("%s_pT%04d", GetArrayName().Data(), static_cast<int>(GetMinPt()*1000.0));
+  static TString trackString;
+  trackString = TString::Format("%s_pT%04d", GetArrayName().Data(), static_cast<int>(GetMinPt()*1000.0));
   return trackString.Data();
 }
--- a/PWG/EMCAL/EMCALbase/AliParticleContainer.cxx
+++ b/PWG/EMCAL/EMCALbase/AliParticleContainer.cxx
@@ -433,7 +433,8 @@ Int_t AliParticleContainer::GetNAcceptedParticles() const
  */
 const char* AliParticleContainer::GetTitle() const
 {
-  static TString trackString = TString::Format("%s_pT%04d", GetArrayName().Data(), static_cast<int>(GetMinPt()*1000.0));
+  static TString trackString;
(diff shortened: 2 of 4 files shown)


raymondEhlers:
I'll wait some time for review, but it must get into the next tag, as it breaks anything that looks for a jet collection which includes a minimum particle pt cut

aiola:
+1, tricky use of static variables... maybe we should drop the static variable and use a class member

raymondEhlers:
Hi @aiola,

Thanks! Yeah, this was rather tricky. That being said, I think the benefit of this is that we don't need to remember to update the title on update to the min pt - we can just format it when requested. But perhaps we can reconsider this approach if we need to make further changes

mfasDa:

I would also vote for moving away from static variables, and by the way I was wondering why to always recreate the title on the fly. On the other side in case of storing the title in the container an automatic update mechanism is needed in case some setting (like min. pt cut) changes.

Cheers 

Markus

aiola:
I was thinking more at an intermediate approach, where the title is still calculated on the fly, but it is stored in a class member rather than in a static variable. alternatively the method could return an std::string (or TString) by value so there is not need of a static variable nor of a class member

raymondEhlers:
I assumed that it returned a const char * because TObject::GetTitle() returns a const char * . But if we aren't attached to that, then return a std::string seems fine to me

aiola:
Ah you’re right. Then I would probably go for replacing the static variable with a class field

adriansev:
a few comments/questions : 
1. When the container is instantiated there is already an array name and the default pt cut values, isn' t it?
so, if the title is a member, maybe it would make sense to build the title at instantiation time (with the default values) and then in each setter that change the contents of title (like array name or pt cut) to add an UpdateTitle() like method ? This way the title would be always consistent with the actual settings of the container
2. i see that only min pt values is considered ... maybe would make sense to add also a max pt (which i think has a default value of 500) and have something in the form of _pT1-500 in the name (for the cases that only a slice of pt is taken into consideration - and this would be available not only to the base class but also to inheritors)
what do you think?
Thank you!

mfasDa:
Hi @adriansev ,

for 1. I think it is a nice recommendation and certainly something we should consider!
for 2. I am not sure, a) the names get terribly long and b) not sure this cut is varied. Probably at some point one might consider using hash value of all settings in order to identify objects (containers).

Cheers

Markus

adriansev:
for 2. i thought about the cases where the analysis is in defined jet pt bins... in my case i (re)construct the name of histogram branch but maybe it would more flexible to just use the title of container that is analyzed and use it for the name of histogram group ... (and because of that a hash would not be helpful because if it is written a root file a human could not read at first glance the type of contents in that branch) ... am i the only one with such a use case?
well, either way is fine, there were only some random suggestions... Thank you!

PR #2975, opened 2017-10-19
Implement handling for track status bits (i.e. hybrid tracks) in the …

mfasDa:
…AliEmcalAODFilterBitCuts

Handling implemented as done for the track cuts. 
Functionality is covered by a unit test (based on
an analysis task)

Once functionality is well tested the filter bit cuts will replace the manual track selection in AliEmcalTrackSelectionAOD, and once all cuts can be expressed in terms of AliVCuts the track selections for ESDs and AODs can be combined to 1 object.

In addition:
- Implementation of ITS pure stand alone track selection based on AliESDtrackCuts in AliEmcalTrackSelectionAOD
- Adding 2016 and 2017 datasets to the default hybrid track selection

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliEmcalAODFilterBitCuts.cxx
+++ b/PWG/EMCAL/EMCALbase/AliEmcalAODFilterBitCuts.cxx
@@ -16,23 +16,61 @@
 #include "AliAODTrack.h"
 
 /// \cond CLASSIMP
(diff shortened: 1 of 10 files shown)


mfasDa:
@aiola Test covers at the moment logical or of the hybrid tracks global contrained and tpc constrained. Do you think there is a need for a test implementation of global / tpc constrained separately?

PR #2985, opened 2017-10-19
PWGDQ: Update of efficiency task

carstooon:
There was a bug in the pairing step. Kinematic cuts where not applied properly.


pdillens:
+1

PR #2988, opened 2017-10-19
Feature jetconstituent

mfasDa:
New classes for constituent (charged and neutral) handling. Direct access to particles without indices in arrays.

### DIFF
--- a/PWG/JETFW/AliEmcalJet.cxx
+++ b/PWG/JETFW/AliEmcalJet.cxx
@@ -19,6 +19,8 @@
 
 #include "AliParticleContainer.h"
 #include "AliClusterContainer.h"
+#include "AliEmcalClusterJetConstituent.h"
+#include "AliEmcalParticleJetConstituent.h"
 
 /// \cond CLASSIMP
 ClassImp(AliEmcalJet);
@@ -62,7 +64,9 @@ AliEmcalJet::AliEmcalJet() :
   fHasGhost(kFALSE),
   fGhosts(),
   fJetShapeProperties(0),
-  fJetAcceptanceType(0)
+  fJetAcceptanceType(0),
+  fParticleConstituents("PWG::JETFW::AliEmcalParticleJetConstituent", 500),
+  fClusterConstituents("PWG::JETFW::AliEmcalClusterJetConstituent", 500)
 {
   fClosestJets[0] = 0;
   fClosestJets[1] = 0;
@@ -112,7 +116,9 @@ AliEmcalJet::AliEmcalJet(Double_t px, Double_t py, Double_t pz) :
   fHasGhost(kFALSE),
   fGhosts(),
   fJetShapeProperties(0),
-  fJetAcceptanceType(0)
+  fJetAcceptanceType(0),
+  fParticleConstituents("PWG::JETFW::AliEmcalParticleJetConstituent", 500),
+  fClusterConstituents("PWG::JETFW::AliEmcalClusterJetConstituent", 500)
 {
   if (fPt != 0) {
     fPhi = TVector2::Phi_0_2pi(TMath::ATan2(py, px));
@@ -167,7 +173,9 @@ AliEmcalJet::AliEmcalJet(Double_t pt, Double_t eta, Double_t phi, Double_t m) :
   fHasGhost(kFALSE),
   fGhosts(),
   fJetShapeProperties(0),
-  fJetAcceptanceType(0)
+  fJetAcceptanceType(0),
+  fParticleConstituents("PWG::JETFW::AliEmcalParticleJetConstituent", 500),
+  fClusterConstituents("PWG::JETFW::AliEmcalClusterJetConstituent", 500)
 {
   fPhi = TVector2::Phi_0_2pi(fPhi);
 
@@ -216,7 +224,10 @@ AliEmcalJet::AliEmcalJet(const AliEmcalJet& jet) :
   fHasGhost(jet.fHasGhost),
   fGhosts(jet.fGhosts),
   fJetShapeProperties(0),
-  fJetAcceptanceType(jet.fJetAcceptanceType)
+  fJetAcceptanceType(jet.fJetAcceptanceType),
+  fParticleConstituents(jet.fParticleConstituents),
+  fClusterConstituents(jet.fClusterConstituents)
+
 {
   // Copy constructor.
   fClosestJets[0]     = jet.fClosestJets[0];
@@ -285,6 +296,8 @@ AliEmcalJet& AliEmcalJet::operator=(const AliEmcalJet& jet)
       fJetShapeProperties = new AliEmcalJetShapeProperties(*(jet.fJetShapeProperties));
     }
     fJetAcceptanceType  = jet.fJetAcceptanceType;
+    fParticleConstituents = jet.fParticleConstituents;
+    fClusterConstituents = jet.fClusterConstituents;
   }
 
   return *this;
@@ -664,6 +677,8 @@ void AliEmcalJet::Clear(Option_t */*option*/)
   fPtSub = 0;
   fGhosts.clear();
   fHasGhost = kFALSE;
+  fClusterConstituents.Clear();
+  fParticleConstituents.Clear();
 }
 
 /**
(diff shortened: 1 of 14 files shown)


mfasDa:
@aiola @raymondEhlers @lcunquei 

This is the pull request where to further discuss the constituent handling. Please review!

mfasDa:
+test

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
```diff
@@ -493,6 +495,7 @@ void AliEmcalJetTask::FillJetConstituents(AliEmcalJet *jet, std::vector<fastjet:
         AliError(Form("Could not find track %d",tid));
         continue;
       }
+      jet->AddParticleConstituent(t);
```
raymondEhlers: The embedding status should be set here as part of the function call. Should probably be something like:

~~~c++
jet->AddParticleConstituent(t, partCont->GetIsEmbedding());
~~~

Of course, that will require the modification of some constructors

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
```diff
@@ -553,6 +556,8 @@ void AliEmcalJetTask::FillJetConstituents(AliEmcalJet *jet, std::vector<fastjet:
       Double_t cPhi = nP.Phi_0_2pi();
       Double_t cPt  = nP.Pt();
       Double_t cP   = nP.P();
+      Double_t pvec[3] = {nP.Px(), nP.Py(), nP.Pz()};
+      jet->AddClusterConstituent(c, (AliVCluster::VCluUserDefEnergy_t)clusCont->GetDefaultClusterEnergy(), pvec);
```
raymondEhlers: Same idea about the embedding here

PWG/JETFW/AliEmcalJet.cxx
```diff
@@ -751,6 +764,65 @@ Int_t AliEmcalJet::ContainsCluster(AliVCluster* cluster, TClonesArray* clusters)
   return ContainsCluster(clusters->IndexOf(cluster));
 }
 
+const PWG::JETFW::AliEmcalClusterJetConstituent *AliEmcalJet::ClusterConstituentAt(unsigned int icl) const {
+  return (static_cast<int>(icl) < fClusterConstituents.size()) ? &fClusterConstituents[icl] : nullptr;
```
raymondEhlers: std::vector::size() is size_type, which is unsigned, right? In that case, I think you could remove the static_cast<int>?

PWG/JETFW/AliEmcalJet.cxx
```diff
@@ -751,6 +764,65 @@ Int_t AliEmcalJet::ContainsCluster(AliVCluster* cluster, TClonesArray* clusters)
...
+}
+
+const PWG::JETFW::AliEmcalParticleJetConstituent *AliEmcalJet::ParticleConstituentAt(unsigned int ipart) const {
+  return (static_cast<int>(ipart) < fParticleConstituents.size()) ? &fParticleConstituents[ipart] : nullptr;
```
raymondEhlers: Same question here

aiola:
My only comment would be to make this feature optional in AliEmcalJetTask. Since it implies allocation/deallocation of memory each event, I fear that it may add an overhead especially for PbPb. However if you have benchmarked it and verified that it does not slow down the execution significantly, then I would be ok with having it always on.

mfasDa:
@aiola Indeed there is a dynamic memory allocation, however the objects are really tiny (currently just one long + for the clusters Double + int) compared to tracks / clusters themselves. Furthermore by design we can have at maximum once constituent object per track / cluster. And they are handled via the std::vector, so once the jet gets destructed the constituent objects get destructed as well. So I don't expect a memory leak created by the new development.

mfasDa:
We found another use case where such a development is urgently needed: Consider the toy model where smearing to the constituent tracks is applied when doing the jet finding. The smearing is applied in GetMomentum, so dynamically once the function is called. The smearing is a random smearing. The itself should never change the objects (tracks) it handles. So during clustering the smeared momentum vectors are used, which contribute to pt, energy and mass. As long as one looks at global jet properties this works. The problem comes in for analysis tasks looking at jet substructure variables which are based on the constituent momentum vectors (i.e. girth, PtD, soft drop). When accessing the momentum vectors the *old* way GetMomentum is now called again, this time from the analysis task, which does again a random smearing, this time however with a new random value. Results will be meaningless in this way. This can only be overcome if a) the container would change the object, which is unsafe and leads to a problem what happens if the GetMomentum function is called twice, or b) the jet stores some basic information like proposed here + the four vector of the momentum.

Btw.: FastJet itself does it in a similar way.

mfasDa:
+test

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
```diff
@@ -493,6 +497,7 @@ void AliEmcalJetTask::FillJetConstituents(AliEmcalJet *jet, std::vector<fastjet:
         AliError(Form("Could not find track %d",tid));
         continue;
       }
+      if(fFillConstituents) jet->AddParticleConstituent(t);
```
raymondEhlers: My comments regarding setting embedding still apply here

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
```diff
@@ -553,6 +558,8 @@ void AliEmcalJetTask::FillJetConstituents(AliEmcalJet *jet, std::vector<fastjet:
       Double_t cPhi = nP.Phi_0_2pi();
       Double_t cPt  = nP.Pt();
       Double_t cP   = nP.P();
+      Double_t pvec[3] = {nP.Px(), nP.Py(), nP.Pz()};
+      if(fFillConstituents) jet->AddClusterConstituent(c, (AliVCluster::VCluUserDefEnergy_t)clusCont->GetDefaultClusterEnergy(), pvec);
```
raymondEhlers: Same here

aiola:
we can proceed, I leave it to @raymondEhlers just in case he has any additional remarks

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
mfasDa: Thanks for notifying! I just overlooked it. Will be fixed in a couple of minutes.

PWG/JETFW/AliEmcalJet.cxx
raymondEhlers: Can you also address these comments before committing?

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
raymondEhlers: Thanks! I'm also wondering about my other comments above regarding size_type

mfasDa: I am not sure. To be checked. If it works without compiler error I will add your fix.

mfasDa:
Fixed:
- Add flagging of constituents from embedded events
- Removing static_cast from comparison with std::vector::size()
@raymondEhlers please review!

mfasDa:
+test

raymondEhlers:
@mfasDa It looks good to me! Thanks!

mfasDa:

Thanks a lot for your constructive comments! It has significantly improved the quality of the committed changes! If there is nothing more to be done here I will merge the commit.

One more thing I would like to discuss is adding a common test. Recently I have added test classes for EMCAL containers, see for example PWG/EMCAL/EMCALtasks/TestAliEmcalTrackSelection.h . We might want to start similar tests for the jet framework as well. Questions are:
- What to be tested?
- How to implement a test for the embedding? @raymondEhlers, here your assistance might be needed.
- Do we want to put tests into a dedicated library?

Cheers

Markus

mfasDa:
+1

raymondEhlers:
@mfasDa , I'm totally in favor of testing although I'm not entirely sure where to start

PR #2993, opened 2017-10-19
modify my code 

ysekiguc:


rbertens:
+1 @ysekiguc in the future, it would be good if you could make your PR descriptions a bit more descriptive than 'modify my code', so that our 'future generations' will still understand changes made :) thanks, redmer

PR #2994, opened 2017-10-19
Fix root 6 compilation

raymondEhlers:
namespace std needs to be specified for vector


zconesa:
+test

zconesa:
+1

raymondEhlers:
@zconesa Thanks! I think it doesn't parse the +1 if given while the test is running, so it may need another +1

zconesa:
+1

PR #3003, opened 2017-10-20
Apply ZDC Gain Equalization and NUA correction for EP method

rihanphys:
Apply ZDC Gain Equalization and NUA correction for Event Plane method


rbertens:
+1

PR #3005, opened 2017-10-20
Event mixing in centrality

hzanoli:
Added event mixing in centrality bins and possibility to use TPC only analysis.


sashingo:
+1

PR #3013, opened 2017-10-21
extract full jet performance histograms into a new task

jdmulligan:
Extract general full jet studies from two of my personal tasks (DijetImbalance and EmcalVsPhos) into a separate user task.


raymondEhlers:
+1

PR #3019, opened 2017-10-22
change M02 energy cut to use nonlincorr energy

jdmulligan:


raymondEhlers:
Hi @jdmulligan , I haven't thought about this a lot, but would it be desirable / possible to make this user configurable like is done a few lines below?

jdmulligan:
Hi @raymondEhlers, I suppose one could get the default energy, but I'm not sure that's really the right thing to do (e.g. I might want HadCorr default energy, but the M02 cut to be applied based on NonLinCorr energy). I think NonLinCorr is the most natural choice, since where the M02 cut will be applied depends on particle composition at a given cluster energy. But anyway this is probably a temporary way of doing things, and once we finalize whether the M02 cut is feasible and what to do about the hadronic correction, it may be revised. But suggestions are certainly welcome!

raymondEhlers:
Okay, I was wondering about the possible hadcorr/non-lin mismatch, which is why I wasn't sure if that was a good suggestion. Thanks for educating me! :smile:

PR #3022, opened 2017-10-23
OADB/EMCAL: Update of LHC15n bad channel maps

nschmidtALICE:
Update of LHC15n bad channel maps with Inputs from Adrian Mechler. The following run ranges are covered with this update:
244340-244484
244531-244628


gconesab:
+1

PR #3024, opened 2017-10-23
Adding new files for LambdaNN search

amastros:
The new folder LambdaNN has been created. It contains the files for the LambdaNN search to be added to the analysis trains. Please note that it runs over the ESD.


mpuccio:
+1

PR #3025, opened 2017-10-23
PWGGA/GammaConv: In  NeutralMesonPi+Pi-Pi0 addition of pi- likesign m…

fjonasALICE:
…ixing and variable renaming

- added pi0pi-pi- likesign event mixing in addition to pi0pi+pi+ mixing
- fixed confusing variable names of different event mixing group


dmuhlhei:
+1

PR #3029, opened 2017-10-23
add fEmbEventID for embedding analyis

dsekihat:
add a member variable fEmbEventID for embedding analysis to allow re-shuffle combination between UE and MC events.


kharlov:
+1

PR #3037, opened 2017-10-24
Random eta gap in events w/o cand + q2Cand vs. q2Event histo + right dau rem and eta gap

fgrosa:
3 modifications:

1) If eta gap between D-meson and q2 enabled, same Delta eta (random) applied in events w/o candidates and for the global q2, in order to have a consistent multiplicity to compute q2

2) added TH2F for q2 candidate vs. q2 global event 

3) Daughter tracks NOT removed twice, if both eta gap and daughter removal enabled


fprino:
+1

PR #3043, opened 2017-10-24
PWG/Cocktail: new 8 TeV cocktail input with interpolated spectra

nschmidtALICE:
Now using interpolations with 0.9, 2.76, 5, 7 and 13 TeV spectra as inputs for all spectra besides pi0 and eta and wherever measurements are available.


FriederikeBock:
+1

PR #3052, opened 2017-10-24
Fix compilation for root6 (plus some warnings)

dstocco:


dstocco:
Hi @AudreyFrancisco ,
your commit was actually breaking the compilation on root6. I made some small fixes. Please make a git pull before modifying your code locally.

@dberzano : I thought that alibuild was testing the compilation against root6 as well. Am I wrong? If so, can this test be included in the continuous integration? Now that I'm on High-Sierra, I cannot get back to root5 any more.

Thanks in advance!
Ciao,
Diego

PR #3054, opened 2017-10-24
automatically loop through output lists, rather than requiring user t…

jdmulligan:
…o pass a list name

Update to post-processing script for pT-hard jet QA.


raymondEhlers:
+1

PR #3061, opened 2017-10-25
PWGDQ: Added initialization of new variables to constructor

carstooon:
Needed to ensure backwards compatibility but need to be set by setter to get physical values.


pdillens:
+1

PR #3089, opened 2017-10-26
PWG/EMCAL and OADB/EMCAL: custom bad channel map feature added to EMCtender

nschmidtALICE:
With this small addition, it is now possible to load a custom bad channel map for testing purposes. The previous feature to set a different path for all calibrations remains unchanged with this update.

### DIFF
--- a/PWG/EMCAL/macros/AddTaskEMCALTender.C
+++ b/PWG/EMCAL/macros/AddTaskEMCALTender.C
@@ -30,7 +30,8 @@ AliAnalysisTaskSE *AddTaskEMCALTender(
   Int_t  removeNMCGenerators  = 0,        // set number of accepted MC generators input (only for enableFracEMCRecalc=1)
   Bool_t enableMCGenRemovTrack= 1,        // apply the MC generators rejection also for track matching  
   TString removeMCGen1        = "",       // name of generator input to be accepted
-  TString removeMCGen2        = ""        // name of generator input to be accepted
+  TString removeMCGen2        = "",       // name of generator input to be accepted
+  TString customBCmap         = ""        // location of custom bad channel map (full path including file)
 ) 
 {
   // Get the pointer to the existing analysis manager via the static access method.
@@ -89,6 +90,8 @@ AliAnalysisTaskSE *AddTaskEMCALTender(
 
   if (pass) 
     EMCALSupply->SetPass(pass);
+  if (customBCmap!="") 
+    EMCALSupply->SetCustomBC(customBCmap);
 
   if (evhand->InheritsFrom("AliESDInputHandler")) {
     #ifdef __CLING__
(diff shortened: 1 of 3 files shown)


jdmulligan:
HI Nicolas,

The EMCal Tender is no longer maintained and has been superceded by the EMCal Correction Task (http://alidoc.cern.ch/AliPhysics/master/_r_e_a_d_m_eemc_corrections.html). Is there a particular reason you need/want to commit this to the Tender? Otherwise I would suggest to add it to the new correction framework instead. Namely, the function you modify is located at AliEmcalCorrectionComponent::InitBadChannels(). I am happy to help with the implementation (or do it myself).

Cheers,
James

loizides:
I'd like to keep the tender in sync with changes... so lets go ahead.

nschmidtALICE:
Hello James,

we have not yet made the switch to the Correction Task in our group as all functionalities are still available with the "old tender". As this small change does not affect anyone and only provides a small improvement to the tender, it would be great if it could be added.

Cheers,
Nicolas

gconesab:
I would still suggest to move to the new framework, in this way we will never abandon the tender.

loizides:
I do not see what is wrong with the tender...
it has been proven stabel for many many years, and there is a large group of people using it, and that does not need all the xtra features of corr framework

gconesab:
Hi Constantin, I will not maintain N frameworks. You can approve the request if you think it is the way to go.

mfasDa:
@nschmidtALICE , @loizides 

First of all I agree with Raymond and Gustavo that it is a bad idea to keep and old and unmaintained framework alive. There is no coordinated effort put into this, so whoever is using it is on their own. In addition there were extensive test done by Raymond and James showing the correction task output is the same as for the tender.

Second, could you please elaborate a bit why this bypass is needed? Having bypasses can become dangerous, in particular if they are misused (i.e. copying one macro from another).

Cheers

Markus

loizides:
The corr framework is much more than the tender... plus we can not afford now to change a functional system that is used by ~5 people out of which we have obtained 3 recent papers and 2 to come just because we do not want to add one easy change to the tender
I will approve the change

loizides:
+1

loizides:
Ok so it needs someone to approve with rights in all directories, gustavo can you please do it, in the end this stops nico from testing the maps he is creating!

gconesab:
As a matter of principle, I do not want to approve it but I will. Future use and developments of this task will have to be be discussed again within the emcal community.

gconesab:
+1

raymondEhlers:
From our perspective, this principle is problematic for a few reasons:

- When the Corrections Framework was deployed, it was agreed in the EMCal group that it would be the maintained framework, and we would cease to maintain the tender.
- No one has been maintaining the tender for the last year, and there have been recently found bugs from the old corrections that are fixed in the Correction Framework and may not be fixed in the tender.
- As noted by Markus, we put in a significant amount of work to validate the Correction Framework against the tender, and we know they agree. We also made those tools available, so others can use them to validate their own analyses.
   - If the tender and Correction Framework diverge now, then we don't benefit from shared code, and we're more susceptible to bugs and non-uniformity across the EMCal.
- By default, the Correction Framework does as little or as much as you want it to. If you only want the aspects that correspond to the tender, no problem - just enable those components and you can ignore everything else.

Raymond + James

PR #3120, opened 2017-10-30
Tree slipt level changed from 0 to 99

nirbhaygit:
HFE beauty tree slipt level changed from 0 to 99, autosave for each 100MB
A minor bug fixed to grid run macro.


sashingo:
+1

PR #3132, opened 2017-10-31
Fixes for ROOT6

mfasDa:
- Add missing forward declaration, remove default parameters
  from implementation
- Use TMacro instead of gROOT->LoadMacro


mfasDa:
+1

PR #3134, opened 2017-10-31
Fix filling number of trials for the case of AODs

mfasDa:
Method to fill number of trials and pt-hard
was missing for AODs. The detection whether
the task is running on AODs is based on the
result of the MCEvent() function and does
not access the MC handler directly as in the
current implementation for ESDs.


mfasDa:
+1

PR #3149, opened 2017-11-01
Fix handling of filtermask

mfasDa:
The case of both hybrid tracks (2 bits set) was handled
improperly: By using TestFilterMask the check is done
for equalty (which means AOD tracks must have both
bits set in order to be accepted), which is impossible
for hybrid tracks (can only be of one type). Changed to
TestFilterBits, which selects tracks in case any of the
bits is set.

In addition adding new datasets from run2 to the configuration.


mfasDa:
+1

PR #3163, opened 2017-11-03
Falcon mode for CF of D+/Dstar/Ds/Lc

yowatana:
Dear all,

The falcon mode worked successfully for D0 and Lc->pK0s, the similar changes are implemented for other mesons/baryons. 

Best regards,
Yosuke


fprino:
+1

PR #3167, opened 2017-11-03
Add quality flags to TOF calibration tree task

jezwilkinson:
Adds new flags to AliTOFAnalysisTaskCalibTree for the quality of a hit (used in case of fSaveCoordinates, not in default invocation of task in calib train).

Also updates setting of time resolution in AddTask (updates default value to 60ps, and replaces redundant "GetTOFT0maker()->SetTimeResolution()" with "task->SetTimeResolution()" (as the value set in the TOFT0maker is anyway overwritten by fTimeResolution in the task::InitEvent() function).


preghenella:
+1

PR #3188, opened 2017-11-06
Add one over pairefficiency squared, create sumw2 also for profiles

dweiser90:
Added squared pair efficiency (OneOverPairEffSq) to the VarManager. Create Sum of squares of weights also for profiles in the HistogramManager.


iarsene:
+1

PR #3197, opened 2017-11-07
Adding LHC16e1 to the list of known datasets

mfasDa:


mfasDa:
Clearly this handling is not ideal: Currently whenever we have a new dataset, no matter whether it is MC or data, it has to be added by hand in order to properly separate the types of hybrid tracks. This is error prone and cumbersome. New datasets typically follow the new definition, for which the standard should be to take the new filter bits unless the dataset is specifically marked. The more correct solution however would have been to have something like a version number store in the event which has a unique connection specific settings so the decoding could happen automatically and not manually.

mfasDa:
+1

PR #3229, opened 2017-11-08
Implemeting jet tagger based on KD-trees

mfasDa:
Re-implement AliAnalysisTaskEmcalJetTagger with
replacing the brute-force algorithm used for
geometric matching with a KD-tree based approach
(based on TKDTree).Also Remimplementation of the
true tag matching: Only 1 loop over base jets,
comparing match index in the arrays. Match function
is made atomic - only matching is handled, extra
parts like init and reset are moved to Run.

Some more fixes:
- Add new class to build
- Move from hand-made debug to AliDebug
- Factory function for adding the task and corresponding
  macro added


mfasDa:
+test

mfasDa:
+test

mfasDa:
+test

mfasDa:
+test

mfasDa:
+1

PR #3238, opened 2017-11-09
Add cluster container name to output container

rqwa:
The change is necessary to run more then one instance of AliAnalysisTaskEMCALPhotonIsolation in one train over different cluster selections.

### DIFF
--- a/PWGGA/EMCALTasks/macros/AddTaskEMCALPhotonIsolation.C
+++ b/PWGGA/EMCALTasks/macros/AddTaskEMCALPhotonIsolation.C
@@ -94,10 +94,10 @@ AliAnalysisTaskEMCALPhotonIsolation* AddTaskEMCALPhotonIsolation(
 
   TString myContName("");
   if(bIsMC){
-    myContName = Form("Analysis_Neutrals_MC");
+    myContName = Form("Analysis_Neutrals_MC_%s",clusName.Data());
   }
   else{
-    myContName = Form("Analysis_Neutrals");
+    myContName = Form("Analysis_Neutrals_%s",clusName.Data());
   }
   
   // For the 2012 EGA/L1 Analysis, only events with EGA/L1 recalc patches are considered


e-masson:
Hi @rqwa,
Out of curiosity, what is the link between this modification to the output name and "different cluster selections"? Is it useful only when you manually set the cluster container name (i.e. different from `"usedefault"`)?
In this case, could you please add a condition based on `clusName != "usedefault"` in order not to affect analyses which use only `"usedefault"`?
Best,
Erwann

e-masson:
Hi @rqwa,
I understand. I preferred to ask because such a modification may impact all the post-processing macros used by analysers, even if it is not really invasive.
@dlodato, what do you think?
Best,
Erwann

e-masson:
+test

e-masson:
+1

dlodato:
Hi @rqwa and @e-masson,
how about changing a bit the commit to :  
if (clusName=="usdefault" || clusName=="caloCluster") 
 myContName = Form("Analysis_Neutrals");
else
 myContName = Form("Analysis_Neutrals_%s",clusName.Data());

and same for MC?
This way it will be perfectly backward compatible while adding the new functionality?
Am I right?

If this is NOT possible, I guess we have to live with the change and the commit can go forward.

Cheers, 
Davide

e-masson:
Hi @dlodato,
Sorry I approved the commit so it can be included in the today's tag (in case Marco needs it quickly). In any case the change still can be reverted.
I think your idea should work, but as you suggest we will have to live with more and more changes, and when it will come to cleaning the AddTask parameters our wagons and macros will be affected for sure.
Best,
Erwann

rqwa:
Hi @dlodato , @e-masson ,
if any of your post-processing macros is affected I will prepare an additional commit for tomorrow, just let me know.
@e-masson thanks for the approve, I hope it will go into today's tag.
Regards
Marco

e-masson:
+1

PR #3251, opened 2017-11-10
Ntrklts vs. q2 vs. cent histos + minor changes

fgrosa:
1) Added Ntrklts vs. q2 vs. cent histos (when flag enabled)  
2) modified binning of M(q2) vs. cent histos 
3) default centrality binning of 1% when kEvShape enabled


fprino:
+1

PR #3263, opened 2017-11-12
support different trigger input in different period

dsekihat:
update trigger analysis to support different period, especially for L0


kharlov:
+1

PR #3278, opened 2017-11-14
implement 1-to-1 correspondence :  1 task - 1 (+2 sub) detector for Qn

dsekihat:
1 analysis task has 1 set of detectors to measure Qn vector


kharlov:
+1

PR #3288, opened 2017-11-14
zg and rg distributions for each hard splitting in the history

klapidus:
A couple of new histograms in the user task


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskSoftDrop.cxx
```diff
@@ -403,11 +407,19 @@ void AliAnalysisTaskSoftDrop::SoftDropDeepDeclustering(fastjet::PseudoJet jet, c
     Float_t pt1 = jet1.pt();
     Float_t pt2 = jet2.pt();
 
+    Float_t dr = sqrt( jet1.plain_distance(jet2) );
```
raymondEhlers: I think sqrt will needs an explicit namespace (?)

klapidus: I think it's ok as is (was already used in another method). Nevertheless, changed. Thanks.

raymondEhlers: Apologies - I really thought this was going to be an issue on root 6 - I guess I should have tested it! Thanks for bearing with me!

raymondEhlers:
+1

PR #3297, opened 2017-11-15
OADB/EMCAL: added LHC17p bad channel maps

nschmidtALICE:
This pull request contains the bad channel maps for LHC17p. The maps were created from runs 282050, 282051, 282078, 282098, 282099, 282118, 282119 and cover the run range 282008-999999 for now.


mfasDa:
+1

mfasDa:
This is a preliminary bad channel map provided by Eliane based on the first runs from LHC17p. It is left with open upper range in order to be used in the trigger QA to provide fast feedback on the upcoming runs reconstructed. Once the data taking is over the run range will be refined.

PR #3307, opened 2017-11-16
Bug fix for trigger mask and further debug info

raymondEhlers:
Updates to my personal task.

The trigger mask bug was a trivial reverse of the desired conditions


raymondEhlers:
+1

PR #3317, opened 2017-11-16
New task to recursively store jet splittings

haandrew:
New task to run over data jets to store iteratively the jet splitting information

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskRecursiveSoftDrop.cxx
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskRecursiveSoftDrop.cxx
@@ -70,7 +70,7 @@ AliAnalysisTaskRecursiveSoftDrop::AliAnalysisTaskRecursiveSoftDrop() :
   fTreeRecursive_True(0)
 
 {
-  for(Int_t i=0;i<nBranches;i++){
+  for(Int_t i=0;i<4;i++){
     fShapesVar_Det[i]=0;
     fShapesVar_True[i]=0;
   }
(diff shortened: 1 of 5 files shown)


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskRecursiveSoftDrop.cxx
```diff
@@ -0,0 +1,297 @@
...
+  TH1::AddDirectory(kFALSE);
+  TH1::AddDirectory(oldStatus);
+  //create a tree used for the MC data and making a 4D response matrix
+  const char* nameoutput = GetOutputSlot(2)->GetContainer()->GetName();
+  fTreeRecursive_Det = new TTree(nameoutput, nameoutput);
+  TString *fShapesVarNames_Det = new TString [nBranches];
+  const char* nameoutput2 = GetOutputSlot(3)->GetContainer()->GetName();
+  fTreeRecursive_True = new TTree(nameoutput2, nameoutput2);
+  TString *fShapesVarNames_True = new TString [nBranches];
+
```
mfasDa: Here you allocate memory without freeing it. I recommend replacing by
std::vector<TString> fShapeVarNames_Det(nBranches), nShapeVarNames_True(nBranches);

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskRecursiveSoftDrop.h
```diff
@@ -0,0 +1,81 @@
...
+class AliJetContainer;
+class AliEmcalJetFinder;
+class AliFJWrapper;
+
+#include "AliAnalysisTaskEmcalJet.h"
+#include "AliFJWrapper.h"
+#include "AliClusterContainer.h"
+#include "FJ_includes.h"
+
+const Int_t nBranches = 4;
```
raymondEhlers: This should be inside of your class - otherwise I think you are adding some new constant to the global namespace. I would suggest creating an enum for it in your class

mfasDa: And since we are already at that step please follow the naming convention for constants in order to not mix between local variables, global variables, class members and constants.

PR #3332, opened 2017-11-17
Do not process part level jets without matching det level jet

mfasDa:
Case of missing match at detector level should only happen at low pt,
and is anyhow handled by a cut at detector level pt. Storing all low pt
jets is much too much disk space consumptive.


mfasDa:
+1

PR #3336, opened 2017-11-18
using both fine log-binning and coarse nat-binning for m^2

bschaefe:
Also additional correlation function comparison histograms.


rbertens:
+1

PR #3337, opened 2017-11-18
Wrong json string parsing

victor-gonzalez:
There are two fixes included for two different issues.
- The boolean values are not correctly parsed: 'true' or 'false' are correctly taken as boolean but the parsing point is not progressed after taking them and an exception is thrown
- It is imposed a useless additional comma at the end of the map components if there are spaces before the right bracket: as an example, 
`"{type: \'AliFemtoAnalysisPionPion\' }"` throws an exception
`"{type: \'AliFemtoAnalysisPionPion\', }"` gets accepted
`"{type: \'AliFemtoAnalysisPionPion\'}"` gets accepted


akubera:
+1  Thanks; I noticed the boolean error, but procrastinated merging into master.  I hadn't noticed the missing whitespace error. I hope to get some tests working with this code to catch these errors quickly.

PR #3338, opened 2017-11-19
Change output to tree

hzanoli:
I have changed the output to trees instead of histograms. The total number of electrons is relatively small (for example ~270k electrons for 16k), so the tree output is still quite small. 
I need to change this to investigate if there is some pT (or any of the other cuts applied) in the matching residuals.

### DIFF
--- a/PWGPP/EMCAL/AliAnalysisTaskEMCALAlig.cxx
+++ b/PWGPP/EMCAL/AliAnalysisTaskEMCALAlig.cxx
@@ -152,10 +152,7 @@ void AliAnalysisTaskEMCALAlig::DoTrackLoop()
             
             TVector3 trackposOnEMCAL;
             trackposOnEMCAL.SetPtEtaPhi(440,track->GetTrackEtaOnEMCal(),track->GetTrackPhiOnEMCal());
-            
-            Float_t phidiff = TVector2::Phi_mpi_pi(trackposOnEMCAL.Phi()-emcphi);
-            Float_t etadiff = trackposOnEMCAL.Eta() - emceta;
-            
+                        
             Float_t xdiff = trackposOnEMCAL.X() - clustpos.X();
             Float_t ydiff = trackposOnEMCAL.Y() - clustpos.Y();
             Float_t zdiff = trackposOnEMCAL.Z() - clustpos.Z();
--- a/PWGPP/EMCAL/AliAnalysisTaskEMCALAlig.cxx
+++ b/PWGPP/EMCAL/AliAnalysisTaskEMCALAlig.cxx
@@ -39,34 +39,8 @@ AliAnalysisTaskEmcal(),
 fEMCALRecoUtils(NULL),
 fEMCALGeo(NULL),
 fPIDResponse(NULL),
-fNPartPt(NULL),
-fNPartPhi(NULL),
-fNPartEta(NULL),
-fPPartPt(NULL),
-fPPartPhi(NULL),
-fPPartEta(NULL),
-fTPCnSgima(NULL),
-fEOverP(NULL),
-fElectronPhiRes(NULL),
-fElectronEtaRes(NULL),
-fElectronXRes(NULL),
-fElectronYRes(NULL),
-fElectronPositronZRes(NULL),
-fPositronPhiRes(NULL),
-fPositronEtaRes(NULL),
-fPositronXRes(NULL),
-fPositronYRes(NULL),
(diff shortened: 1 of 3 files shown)


gconesab:
+test
a comment, I just spotted a couple of coding violations, data members must start with f (ElectronInformation, ElectronTree) and variables must be lower case (TPCNSgima).

gconesab:
+1

raymondEhlers:
Hi @gconesab,

If I may make a suggestion, I'd suggest that if we have some coding concerns then we should delay approving the PR until they've been rectified (I also had some comments that I wanted to make). That way, we can ensure that we always have high quality, consistent code that is easier to maintain. I think it's helped quite a bit after we started doing this for code in the PWG/EMCAL folder and in the JE group. What do you think?

gconesab:
Fine with me, next time.

hzanoli:
Hi @raymondEhlers,

Please let me know in case there are more comments additionally to the ones made by @gconesab. I can implement them in a next pull request.

PWGPP/EMCAL/AliAnalysisTaskEMCALAlig.h
```diff
@@ -43,54 +157,17 @@ class AliAnalysisTaskEMCALAlig : public AliAnalysisTaskEmcal {
...
-    TH2F        **fElectronXResRU; //! Electron matching residual in x as function of EMCAL SuperModule using propagation to cluster (done by EMCALRecoUtils)
-    TH2F        **fElectronYResRU; //! Electron matching residual in y as function of EMCAL SuperModule using propagation to cluster (done by EMCALRecoUtils)
-    TH2F        **fElectronPositronZResRU; //! Electron and positron matching residual in z as function of EMCAL SuperModule using propagation to cluster (done by EMCALRecoUtils)
-    
-    TH2F        **fPositronPhiResRU; //! Positron matching residual in phi as function of EMCAL SuperModule using propagation to cluster (done by EMCALRecoUtils)
-    TH2F        **fPositronEtaResRU; //! Positron matching residual in eta as function of EMCAL SuperModule using propagation to cluster (done by EMCALRecoUtils)
-    TH2F        **fPositronXResRU; //! Positron matching residual in x as function of EMCAL SuperModule using propagation to cluster (done by EMCALRecoUtils)
-    TH2F        **fPositronYResRU; //! Positron  matching residual in y as function of EMCAL SuperModule using propagation to cluster (done by EMCALRecoUtils)
+    TTree* ElectronTree;
```
raymondEhlers: You need to specify whether your variables will be streamed (///< or //!<!)

PWGPP/EMCAL/AliAnalysisTaskEMCALAlig.cxx
```diff
@@ -39,34 +39,8 @@ AliAnalysisTaskEmcal(),
...
-fElectronEtaResRU(NULL),
-fElectronXResRU(NULL),
-fElectronYResRU(NULL),
-fElectronPositronZResRU(NULL),
-fPositronPhiResRU(NULL),
-fPositronEtaResRU(NULL),
-fPositronXResRU(NULL),
-fPositronYResRU(NULL)
+ElectronInformation(ElectronForAlignment()),
+ElectronTree(NULL)
```
raymondEhlers: Gustavo's comment here is quite important for maintainability

PWGPP/EMCAL/AliAnalysisTaskEMCALAlig.cxx
```diff
@@ -122,126 +69,9 @@ void AliAnalysisTaskEMCALAlig::UserCreateOutputObjects()
...
-        fOutput->Add(fPositronYResRU[i]);
-        
-        
-    }
-    
-
-    
+    ElectronTree = new TTree("electron_information","electron_information");
+    ElectronTree->Branch("electrons", &ElectronInformation);
+    fOutput->Add(ElectronTree);
```
raymondEhlers: I'm not an expert here, but I was under the impression that it was preferred to have a separate output slot for trees. Perhaps someone more knowledgeable can comment?

aiola: Yes, definitely better to use a separate slot, rather than adding it to a TList. This way you can use a TChain to combine and analyse the result from multiple trains (e.g. if you want to analyse results from different datasets=trains combined together). It may also have some improvements in the performance, but not 100% sure that's the case.

PR #3339, opened 2017-11-19
Proper rendering in object browser

victor-gonzalez:
Proposal for render the configuration into the object browser.
The current look with the proposal incorporated is as attached.
[c1.pdf](https://github.com/alisw/AliPhysics/files/1485494/c1.pdf)


akubera:
+1

Looks good. It does funny stuff when trying to draw in a canvas/pad that already has some contents; I'll look into that.

PR #3342, opened 2017-11-19
implement timing cut in trigger analysis for LHC17p

dsekihat:
implement timing cut for PHOS trigger analysis in LHC17p


kharlov:
+1

PR #3364, opened 2017-11-21
V1 hf mu

AudreyFrancisco:
Additional development for single muon v1 task


zconesa:
+test

zconesa:
+1
@AudreyFrancisco you should consider to increase the classDef when you change the class data members, but I guess that since this is anyhow a first development that will go in the same tag you might not need it, so I approve it  ;)

PR #3366, opened 2017-11-21
add QA plots to ZDC EventPlane task

jmargutt:
add various QA plots in a dedicated output container ("ZDCQA")

tested to be backward compatible


jmargutt:
@AudreyFrancisco c'est parti

PR #3381, opened 2017-11-22
 change to unix line endings

rbertens:


rbertens:
hi @dberzano , @ktf ,
I tried to change the line terminators of three files that were merged yesterday from CRLF to unix style (to avoid issue with switching between branches), but it seems that this causes an issue in merging the PR. Could you take a look? 
Thanks,
Redmer

dberzano:
Hi @rbertens, can you rebase this PR as soon as #3395 is merged?

PR #3383, opened 2017-11-22
update embedding task : introduce nonlin before embedding

dsekihat:
introduce nonlinearity to adjust energy before embedding


kharlov:
+1

PR #3389, opened 2017-11-22
modified AddTask for single mu v1 to run without PS

AudreyFrancisco:
Modification of the AddTask for single mu v1  to run without the physics selection task


zconesa:
+1

PR #3390, opened 2017-11-22
Update variable names and trees

hzanoli:
I have updated the variable names to match the style as requested in #3338.
I have also changed the trees as a separate container, but they are always saved with the name declared in the task rather than the container name (is this the expected behaviour?). I have changed the names by adding a suffix. If there is a better way to do that, please let me know. But I need that the trees have different names in case I run more than one wagon at the same time.

Thanks!

### DIFF
--- a/PWGPP/EMCAL/AliAnalysisTaskEMCALAlig.cxx
+++ b/PWGPP/EMCAL/AliAnalysisTaskEMCALAlig.cxx
@@ -35,7 +35,8 @@ fEMCALRecoUtils(NULL),
 fEMCALGeo(NULL),
 fPIDResponse(NULL),
 fElectronInformation(ElectronForAlignment()),
-fElectronTree(NULL)
+fElectronTree(NULL),
+fTreeSuffix("")
 {
     
 }
@@ -46,7 +47,8 @@ fEMCALRecoUtils(NULL),
 fEMCALGeo(NULL),
 fPIDResponse(NULL),
 fElectronInformation(ElectronForAlignment()),
-fElectronTree(NULL)
+fElectronTree(NULL),
+fTreeSuffix("")
 {
     SetMakeGeneralHistograms(kTRUE);
     DefineOutput(2, TTree::Class());
@@ -56,7 +58,6 @@ fElectronTree(NULL)
 AliAnalysisTaskEMCALAlig::~AliAnalysisTaskEMCALAlig()
 {
     if (fEMCALRecoUtils) delete fEMCALRecoUtils;
-    if (fElectronTree) delete fElectronTree;
 }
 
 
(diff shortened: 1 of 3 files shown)


raymondEhlers:
At first glance, I think your suffix seems like a fine way to go - this is basically what is done for subwagons anyway. However, I think you made need to apply that suffix to the output container as well (?)

raymondEhlers:
+test

hzanoli:
@raymondEhlers I also added the suffix to the container names, otherwise the task crashes with multiple wagons. 
But when I create one TList as output, it is saved by default with the output container name (even if the name of the list is fOutput in the task, the name saved in the AnalysisResults.root file is what I define as name in the output container). The trees copy the name that I declare inside the task("electron_information"+suffix) instead of what I define as output container, that's what I found suspicious. The current solution works, I am not sure if there is any other preferable way of doing this. But maybe in case someone comments, I can also change this later.

gconesab:
+test, 
few more comments on minor coding violations
variables with "_" are not kosher: n_sigma_electron_TPC
variables must start with lower case: Isshared, EoverP

gconesab:
and SuffixForTree

raymondEhlers:
Ah, okay, I see what your concern is now. I haven't used trees enough with the analysis framework to know the answer - perhaps @aiola or @mfasDa can comment. My best guess would a GetName() vs GetTitle() difference?

Since I looked more closely at your AddTask, note that you can use subwagons on the LEGO train if you make "suffix" the last argument to your AddTask

raymondEhlers:
How did this get merged already? I'm rather confused, as I don't see a "+1" yet

@dberzano Is this expected behavior? Does "+test" twice = "+1"?

dberzano:
Uh, must be a bug, we'll have to address it :-/ sorry for the inconvenience.

dberzano:
For the record: @gconesab commented `+test` when there was nothing to test. Tests were OK and the build system requested to comment `+1` only. So, a temporary workaround for this misbehaviour is simply not to type `+test` when the PR has been tested already :-)

raymondEhlers:
Okay, no worries. @hzanoli When you get a chance, please address @gconesab comments and create a new PR

mfasDa:
@hzanoli : One more thing: Before opening a pull request please squash commits like this into one single commit.

mfasDa:
The reason why the trees ignore the name of the output container is because they are written directly to the file (with operator new) and circumvent the slot mechanism which is used otherwise. Therefore the tree has no clue about the slot name. A workaround is suggested in #3458: In this case one can access the output slot from the task, obtain the container from the slot, and set the name of the tree to the name of the slot.

PR #3395, opened 2017-11-22
Allow for fixing CRLF->LF line endings

dberzano:


dberzano:
This is related to #3381. Git automatic conversion of line endings is problematic in our environment: when Git creates the working area, with current settings it changes all CRLF to LF, leading to `git status` showing modified files even when they weren't.

This is the reason why #3381 does not work.

The correct solution is:

* Tell Git not to change anything,
* Add a PR check preventing CRLF from being committed at all.

PR #3396, opened 2017-11-22
T-mask treatment modified - better debugging

rolavick:
Before, script ended, when T-mask class was not in a run. Now, when T-mask class is not in a run, the script will only skip the lines, which treat T-mask classes.


ekryshen:
+1

PR #3398, opened 2017-11-23
Forcing float comparison for pt-hard bin determination

mfasDa:


mfasDa:
In new (PYTHIA8-based) datasets some (rare) events are assigned to the neighboring bin (downwards). The most likely explanation for this is that these are events at the bin edges which are sorted towards neighboring bins because of precision mismatch due to implicit casts (narrowing). Comparison is forced to be done with floating point precision.

PR #3399, opened 2017-11-23
Task for calculating the acceptance and contamination for charged par…

zkhabano:
…ticles


rbertens:
Hi @zkhabano , the day before yesterday a class with the same name was committed to AliPhysics by @pchrista (see PR #3373 ), i think on behalf of you. Could it be that there is a mix-up here? 
Thanks,
redmer

zkhabano:
Hi Redmer,

Panos did it for me because I had some problems before, but from now on I’m gonna modify these files, so there’s no mix-up.

I had to modify the code which was already committed and after the pull request was sent but not approved yet. I tried to use the git commit —amend command to apply the changes , but I am not sure if this is a right way to proceed in this case. Is that ok?

Thank you,
Zhanna

PR #3416, opened 2017-11-24
Update YAML config docs

raymondEhlers:
Update to YAML configuration docs to provide an explicit example before announcing it more broadly


aiola:
+1

PR #3427, opened 2017-11-24
Prepare substructure task for neutral jets

mfasDa:
- Use constants in AliJetContainer for jet type determination
- Adapt jet containers in the AddTask function to handle neutral jets + adapt acceptance


mfasDa:
+1

PR #3433, opened 2017-11-25
Fix for colliding containers names

victor-gonzalez:
Container names collide with other tasks container names which could cause troubles in LEGO train scenarios.

Also allow for more stringent z vertex cuts at the data collection level.


rbertens:
+1

PR #3449, opened 2017-11-27
Fix handling of neutral jets

mfasDa:
- Adapt jet container to generate proper name also for particle level neutral jets
- Add verbosity about jet branch names to AliEmcalJetTask
- Add AliDebugs to AliAnalysisTaskEmcalJetTagger
- Handle cases of failure of the Softdrop algorithm due to jets with single constituents only

### DIFF
--- a/PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
+++ b/PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
@@ -383,12 +383,12 @@ void AliEmcalJetTask::ExecOnce()
   if(auto partcont = GetParticleContainer(0)) {
     std::cout << "Found particle container with name " << partcont->GetName() << std::endl;
   } else {
-    std::cout << "Not found particle container" << std::endl;
+    std::cout << "Not particle container found for task" << std::endl;
   }
   if(auto clustcont = GetClusterContainer(0)){
     std::cout << "Found cluster container with name " << clustcont->GetName() << std::endl;
   } else {
-    std::cout << "Not found cluster container" << std::endl;
+    std::cout << "Not cluster container found for task" << std::endl;
   }
 
   // add jets to event if not yet there
--- a/PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
+++ b/PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
@@ -378,6 +378,18 @@ void AliEmcalJetTask::ExecOnce()
   }
 
   fJetsName = AliJetContainer::GenerateJetName(fJetType, fJetAlgo, fRecombScheme, fRadius, GetParticleContainer(0), GetClusterContainer(0), fJetsTag);
+  std::cout << GetName() << ": Name of the jet container: " << fJetsName << std::endl;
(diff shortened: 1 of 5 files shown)


PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
```diff
@@ -378,6 +378,18 @@ void AliEmcalJetTask::ExecOnce()
   }
 
   fJetsName = AliJetContainer::GenerateJetName(fJetType, fJetAlgo, fRecombScheme, fRadius, GetParticleContainer(0), GetClusterContainer(0), fJetsTag);
+  std::cout << GetName() << ": Name of the jet container: " << fJetsName << std::endl;
```
raymondEhlers: Would this be better as AliInfoStream()? (Same for below) Or is this because you want it to show up on the train?

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
```diff
@@ -378,6 +378,18 @@ void AliEmcalJetTask::ExecOnce()
...
+  std::cout << "Use this name in order to connect jet containers in your task to connect to the collection of jets found by this jet finder" << std::endl;
+  if(auto partcont = GetParticleContainer(0)) {
+    std::cout << "Found particle container with name " << partcont->GetName() << std::endl;
+  } else {
+    std::cout << "Not found particle container" << std::endl;
```
raymondEhlers: Trivial, but I would say "The particle container was not found" (same idea for below)

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
mfasDa: Indeed AliInfoStream would be the appropriate solution. As the lego train system will everything with a level lower than kError (hard coded in AliAnalysisManager!) this message will not appear. However I want it to appear, as it makes debugging easier in the lego train test and comes with little cost (only displayed once in ExecOnce) the only way to have this would be the usage of std stream (which violates the approach of having a dedicated logging handler for a project) or AliErrorStream (which is a completely inappropriate log level and therefor unlogical) - so I think using std stream is the less inappropriate of the two.

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
mfasDa: OK, but then I would rather have something like "No particle container found for this task". "The particle container was not found" indicates it was expected, but this is not necessarily the case.

raymondEhlers: Sure, "No particle container found for this task" is also good by me. I just found "Not found particle container" to sound a bit strange

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
raymondEhlers: Yeah, sounds good - I figured that was the case, but wanted to confirm

mfasDa:
+test

mfasDa:
+1

PR #3450, opened 2017-11-27
fix in fall-back to ITSSA vtx in absence of TRKvtx

shahor02:
Additionally, in case of cpass1 the updated object will take the range of run
for which the CalibObject is extrcted in case the object being updated is default one


shahor02:
+1

PR #3451, opened 2017-11-27
urgent bug fix, (Psi_n was stuck at n=2, now n is set via AddTask)

rihanphys:
urgent bug fix, (Psi_n was stuck at n=2, made n variable). Also restore Vz dependent NUA.


rbertens:
+1

PR #3453, opened 2017-11-27
Dijet mass histograms and fix unnecessary warnings in compilation

TWSman:
- Modify dijet mass calculation
-Fix warnings related to unused variables etc.


aohlson:
+1

PR #3458, opened 2017-11-27
Relate tree name to output slot

mfasDa:
+1


mfasDa:
Same issue as #3390: As trees are written directly to the file opened by a certain slot, the slot mechanism doesn't work (tree not sent to slot but created within the file by the operator new). One can however access the slot directly from the task, from this obtain the data container which has the output name.

mfasDa:
+1

PR #3471, opened 2017-11-28
Add possibility to switch on/off branches

mfasDa:
Branches sorted according to cathegories (rho, mass, soft drop, n-subjettiness) and part level jets


mfasDa:
+1

PR #3478, opened 2017-11-29
10 cent bins to shrink output

rvertesi:
AliHFJetsContainer: number of mult bins decreased from 100 to 10 to shrink the output size.
AddTaskEmCalJetBtagSV: allow to set verbosity level in function call


ebruna:
+1

PR #3520, opened 2017-12-01
EMCal CorrFW: Helper for accessing component

raymondEhlers:
Helper function which enables users to more easily access a
particular component, allowing them to access data such as the bad
channel map from an outside task.

@jdmulligan - I'm adding this helper function requested by @nschmidtALICE to access the EMCal RecoUtils (and therefore the bad channel map) from an outside task. Let me know if you have any thoughts!


jdmulligan:
Yea given that you have a comment that it shouldn't be used in general, it should be fine.

You probably know better than me, but is it sensible also to return the component as a const type?

raymondEhlers:
@jdmulligan That's a good question! As I see it, I think this task corresponds with CorrectionComponents(), which also returns non-const components. So I could see an argument for making it non-const.

On the other hand, const can certainly be safer, but then it would prevent using this function for making any modifications (which we in principle allow, even if it's discouraged). So probably the best would both const and non-const methods. But in any case, this feature is rare enough that I don't think it's worth worry about too much (and I don't really feel like fighting with const_cast right now).

So to summarize, it's a good question, I think you could make arguments both ways, but I also think it's fine right now, so I probably won't change it unless we really feel that we need to change it. Is that okay with you?

jdmulligan:
Yea those were more or less my thoughts -- but it's fine to me to leave it as is.

PR #3522, opened 2017-12-01
Corrected cut for sub-underflow extraction

rhaake:
Corrected a bug affecting the lowest jet pT in extraction task.


arossi81:
+1

PR #3529, opened 2017-12-03
Make optional the symmetrization for results size reduction

victor-gonzalez:
Today the results size is reduced by only incorporating half of the symmetrized data. 
For some particular tests the results should not be symmetrized so, the possibility of symmetrize data is made optional via a configuration parameter.


aohlson:
+test

aohlson:
+1

PR #3531, opened 2017-12-04
minor fix + option to use centrality in CF

yowatana:
Dear all,

1) fix missing initialisation
2) option to use centrality in CF task

Best regards,
Yosuke


zconesa:
+1

PR #3537, opened 2017-12-04
SP and ESE method in AliAnalysisMuMuFlow

AudreyFrancisco:
-New flow methods are available in MuMuFlow : Scalar product and event shape engineering
-New binning in vs out of plane
-Corresponding AddTask


aphecetche:
+1

PR #3582, opened 2017-12-08
Updates in the physics selection OADB for 2017 runs

ekryshen:
- tuned on-vs-of V0 and SPD correlation cuts
- special OADB for pp@5TeV
- relaxed cuts for Xe-Xe
- CINT7ZAC class added for kINT7 in Xe-Xe


ekryshen:
+1

PR #3589, opened 2017-12-09
Further separate PhotonHadron MC particle type, and plot its leading …

jdmulligan:
…energy contributor

(Updates to my user task).


aiola:
+1

PR #3599, opened 2017-12-12
change deltaPhi, fix event plane bin

moliver813:
Change deltaPhi from degrees to radians.
Correct event plane bin determination.


efranzis:
+test

efranzis:
+1

PR #3603, opened 2017-12-12
OADB/EMCal: Added additional bad channels for LHC15o

nschmidtALICE:
Several new bad channels were added based on a presentation from
Mike Sas. These channels are mostly in SM5 and SM8. In total about
30 new bad channels were added.


gconesab:
+1

PR #3605, opened 2017-12-12
Adding various hybrid definitions for AODs

mfasDa:
Hybrid definition targets on AODs mainly used in order to distinguish between with/without ITS-refit tracks in 2010. Hybrid definition is applied during filtering and cannot be expanded later (i.e. distinction will have no effect when using the 2011 hybrid definition)


mfasDa:
+1

PR #3612, opened 2017-12-13
[WIP] User information handling in the track selection

mfasDa:
The current ALICE cut framework only provides information whether an object (i.e. a track) is selected or not. Cut classes cannot pass the circumstance among which it was selected. For i.e. hybrid tracks this approach is too simple as hybrid tracks are categorized in different classes, where different classes need to be treated differently. So in addition to the selection status one needs to provide a user information encoding the classification of the object.

Using integer values can work as long as only one cut is applied at a time. Once the cuts are combined with other cuts (i.e. the geometrical cut) an integer value would no longer be well-defined any more as there will be an ambiguity about how each class will define its return value.

In any kind of implementation the interface needs to be changed, either with a different return value or with different function arguments. Compatibility to cut objects based on the existing ALICE cut framework can be reached with wrappers. The implementation here advocates the usage of an object as return value containing
- the selection status
- a reference to the object to be selected
- an (optional) user-defined object containing further information about the return value

This approach has multiple benefits: 
- By overloading operator bool the return value can be treated as simple boolean
- Combining results of multiple cuts will be trivial - a user-defined return object contains the return object of the separate pointers in a TObjArray.
- Clear interface: Arguments are input parameters, return value is the output

As the object contains only one bool and two pointers copy operations are cheap, and in addition c++11 move-semantics is supported as well. The handling of the user-information is done via reference counting: As the user information is depending on the cut class implementation the memory management must be dynamic in this case. This is best achieved with a shared pointer having all copies of the same selection result pointer sharing the same object. To overcome the issue that c++11 is not allowed in header files in order to allow compatibility with antique ROOT versions the shared pointer is re-implemented in the library (quite some unnecessary code overhead).

Various selections used in the AliEmcalTrackSelectionAOD/ESD are expressed as cut object inheriting from the new AliEmcalCutBase class - allowing for the combination of the selection functions. They are still separate, with however identical implementations, and will be combined to one in AliEmcalTrackSelection with removal of AliEmcalTrackSelectionESD/AOD at the same time at a later stage.

As usual for new object the classes are put into the namespace PWG::EMCAL in order to protect them from being re-defined in other translation units.

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliEmcalAODHybridTrackCuts.cxx
+++ b/PWG/EMCAL/EMCALbase/AliEmcalAODHybridTrackCuts.cxx
@@ -26,6 +26,7 @@
  ************************************************************************************/
 #include "AliAODTrack.h"
 #include "AliEmcalAODHybridTrackCuts.h"
+#include "AliEmcalTrackSelResultHybrid.h"
 
 /// \cond CLASSIMP
 ClassImp(PWG::EMCAL::AliEmcalAODHybridTrackCuts)
@@ -34,24 +35,39 @@ ClassImp(PWG::EMCAL::AliEmcalAODHybridTrackCuts)
 using namespace PWG::EMCAL;
 
 AliEmcalAODHybridTrackCuts::AliEmcalAODHybridTrackCuts():
-  AliVCuts(),
+  AliEmcalCutBase(),
   fSelectNonITSrefitTracks(kTRUE)
 {
-
+  fHybridFilterBits[0] = -1;
+  fHybridFilterBits[1] = -1;
 }
 
 AliEmcalAODHybridTrackCuts::AliEmcalAODHybridTrackCuts(const char *name):
-  AliVCuts(name,""),
+  AliEmcalCutBase(name,""),
   fSelectNonITSrefitTracks(kTRUE)
 {
   
 }
 
-bool AliEmcalAODHybridTrackCuts::IsSelected(TObject *o){
+AliEmcalTrackSelResultPtr AliEmcalAODHybridTrackCuts::IsSelected(TObject *o){
   AliAODTrack *aodtrack = dynamic_cast<AliAODTrack *>(o);
-  if(!aodtrack) return false;
+  if(!aodtrack) return AliEmcalTrackSelResultPtr(nullptr, kFALSE);
   bool selectionresult = aodtrack->IsHybridGlobalConstrainedGlobal();
   // Reject non-ITSrefit tracks if requested
   if((fSelectNonITSrefitTracks == false) && (!(aodtrack->GetStatus() & AliVTrack::kITSrefit))) selectionresult = false;
-  return selectionresult;
+  AliEmcalTrackSelResultPtr result(aodtrack, result);
+  // Create user object defining the hybrid track type (only in case the object is selected as hybrid track)
+  if(selectionresult){
+    AliEmcalTrackSelResultHybrid::HybridType_t tracktype = AliEmcalTrackSelResultHybrid::kHybridGlobal;
+    if(fHybridFilterBits[0] > -1 && fHybridFilterBits[1 > -1]) {
+      if(aodtrack->TestFilterBit(BIT(fHybridFilterBits[0]))) tracktype = AliEmcalTrackSelResultHybrid::kHybridGlobal;
+      else if(aodtrack->TestFilterBit(BIT(fHybridFilterBits[1]))){
+        if(aodtrack->GetStatus() & AliVTrack::kITSrefit) tracktype = AliEmcalTrackSelResultHybrid::kHybridConstrained;
+        else tracktype = AliEmcalTrackSelResultHybrid::kHybridConstrainedNoITSrefit;
+    }
+  }
+  result.SetUserInfo(new AliEmcalTrackSelResultHybrid(tracktype));
+
+  }
+  return result;
 }
--- a/PWG/EMCAL/EMCALbase/AliEmcalAODHybridTrackCuts.cxx
+++ b/PWG/EMCAL/EMCALbase/AliEmcalAODHybridTrackCuts.cxx
@@ -27,6 +27,8 @@
 #include "AliAODTrack.h"
 #include "AliEmcalAODHybridTrackCuts.h"
 #include "AliEmcalTrackSelResultHybrid.h"
+#include "AliLog.h"
+#include <iostream>
 
 /// \cond CLASSIMP
 ClassImp(PWG::EMCAL::AliEmcalAODHybridTrackCuts)
@@ -55,11 +57,11 @@ AliEmcalTrackSelResultPtr AliEmcalAODHybridTrackCuts::IsSelected(TObject *o){
   bool selectionresult = aodtrack->IsHybridGlobalConstrainedGlobal();
   // Reject non-ITSrefit tracks if requested
   if((fSelectNonITSrefitTracks == false) && (!(aodtrack->GetStatus() & AliVTrack::kITSrefit))) selectionresult = false;
-  AliEmcalTrackSelResultPtr result(aodtrack, result);
+  AliEmcalTrackSelResultPtr result(aodtrack, selectionresult);
   // Create user object defining the hybrid track type (only in case the object is selected as hybrid track)
   if(selectionresult){
     AliEmcalTrackSelResultHybrid::HybridType_t tracktype = AliEmcalTrackSelResultHybrid::kHybridGlobal;
-    if(fHybridFilterBits[0] > -1 && fHybridFilterBits[1 > -1]) {
+    if(fHybridFilterBits[0] > -1 && fHybridFilterBits[1 ] > -1) {
       if(aodtrack->TestFilterBit(BIT(fHybridFilterBits[0]))) tracktype = AliEmcalTrackSelResultHybrid::kHybridGlobal;
       else if(aodtrack->TestFilterBit(BIT(fHybridFilterBits[1]))){
         if(aodtrack->GetStatus() & AliVTrack::kITSrefit) tracktype = AliEmcalTrackSelResultHybrid::kHybridConstrained;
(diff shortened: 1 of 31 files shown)


mfasDa:
@saiola, @lcunquei , @raymondEhlers , @loizides 

Please have a look at the development. This should solve the issues with the hybrid tracks in particular with the distinction of the different hybrid track types which require different treatment i.e. in momentum determination which is in the current framework adapted completely to the AOD case and not transparently portable to ESD tracks.

mfasDa:
Work in progress - please do not approve!

aiola:
@raymondEhlers, does this development address the issue that you reported privately to me yesterday?

mfasDa:
@aiola I guess this goes to me, right? Yes, I mentioned this to you yesterday by mail. Since now the auto-approval is removed for EMCAL we can as well use a pull request to discuss this further and have the nice github features in addition.

aiola:
@mfasDa, no I was actually referring to @raymondEhlers!
Raymond found another issue with the selection of the hybrid tracks in case of embedding. I was just wondering whether this development would at least go in the direction of a solution also for his issue.

mfasDa:
Ah, thanks for clarification! I wasn't aware of this. Yes, I can imagine that the current implementation breaks down in case of embedding. It would be good if @raymondEhlers could expand a bit on his findings to see how we can find the best possible solution (maybe this cures it already to some extend).

PWG/EMCAL/EMCALbase/AliEmcalAODHybridTrackCuts.cxx
```diff
@@ -34,24 +35,39 @@ ClassImp(PWG::EMCAL::AliEmcalAODHybridTrackCuts)
...
+  if(!aodtrack) return AliEmcalTrackSelResultPtr(nullptr, kFALSE);
   bool selectionresult = aodtrack->IsHybridGlobalConstrainedGlobal();
   // Reject non-ITSrefit tracks if requested
   if((fSelectNonITSrefitTracks == false) && (!(aodtrack->GetStatus() & AliVTrack::kITSrefit))) selectionresult = false;
-  return selectionresult;
+  AliEmcalTrackSelResultPtr result(aodtrack, result);
+  // Create user object defining the hybrid track type (only in case the object is selected as hybrid track)
+  if(selectionresult){
+    AliEmcalTrackSelResultHybrid::HybridType_t tracktype = AliEmcalTrackSelResultHybrid::kHybridGlobal;
+    if(fHybridFilterBits[0] > -1 && fHybridFilterBits[1 > -1]) {
```
raymondEhlers: I think this is a bug: `fHybridFilterBits[1 > -1]` -> `fHybridFilterBits[1] > -1`

PWG/EMCAL/EMCALbase/AliEmcalAODHybridTrackCuts.cxx
```diff
@@ -34,24 +35,39 @@ ClassImp(PWG::EMCAL::AliEmcalAODHybridTrackCuts)
...
+      if(aodtrack->TestFilterBit(BIT(fHybridFilterBits[0]))) tracktype = AliEmcalTrackSelResultHybrid::kHybridGlobal;
+      else if(aodtrack->TestFilterBit(BIT(fHybridFilterBits[1]))){
```
raymondEhlers: This unfortunately won't fix the issue that I've encountered. I'll expand on the issue in another post momentarily.

PWG/EMCAL/EMCALbase/AliEmcalAODHybridTrackCuts.cxx
mfasDa: Thanks a lot - will fix it. Interesting that the compiler didn't complain.

mfasDa:
@raymondEhlers Yes, interface changes have only been performed up to AliEmcalTrackSelection. Users working only with AliTrackContainer will see no change in the interface.

raymondEhlers:
Hi @aiola , @mfasDa , all,

I observed an issue with hybrid track selection in data to data embedding (LHC11a embedded into LHC11h). After some investigation, I've found that is caused by a difference between the `constrainedfilterbit` (bit[1]) for 11a (bits 8, 4) and 11h (bits 8, 9). This issue only shows up if `AliTrackContainer::SetDefTrackCutsPeriod(...)` is set.

Before I go into more detail, a few points to be clear about:
- This is very much an edge case. Almost always, one is embedding MC into a period it was anchored to, which means the filter bits should be the same.
- This issue can be worked around by setting `SetTrackCutsPeriod("lhc11a")` for the 11a containers.

In detail, I found that if I set the global track cuts period to LHC11h then it was giving me a non-uniform phi track distribution for 11a track containers (amongst other issues). I verified that `IsHybridGlobalConstrainedGlobal()` works properly for the 11a AOD. The easiest way to think about the issue is to imagine what happens when you call `AliEmcalTrackSelectionAOD::IsTrackAccepted(...)` for a track that has bit 4 set, but fHybridFilterBits are set to 8 and 9 (the values for 11h, which it will get from the global track cuts period). The result is that the `cutcounter = 2` but `fTrackBitmap` is empty, which means that a track with bit 4 (a hybrid track in 11a) will be rejected despite being a hybrid track! (These details change slightly with @mfasDa's changes, but the issue is still there).

I don't see any clear way to fix this. I think the easiest thing is document the workaround. Does anyone have other suggestions?

PWG/EMCAL/EMCALbase/AliEmcalAODHybridTrackCuts.cxx
raymondEhlers: See this [post below](https://github.com/alisw/AliPhysics/pull/3612#issuecomment-351454454)

aiola:
Hello @raymondEhlers ,
thanks for bringing up this point and explaining it in detail.
I also don't see a simple solution. The only "solution" that I can see is to eliminate the default period completely. In this way by default hybrid tracks will be selected only using the IsHybridGlobalConstrainedGlobal method (but unable to distinguish between the different track types). This would fix the issue for embedding, but it won't work for ESD, unless one explicitly sets the period in each AliTrackContainer.
In the end we probably have to live with it, unless @mfasDa or others can propose other solutions.
In any case, whichever solution we go for, I agree with @raymondEhlers that this issue should be clearly documented.
Best,
Salvatore

mfasDa:

Thanks for the explanation - I almost expected this once I saw the comment about the filter bits!

Indeed the default periods string has to go out, but that will happen at a later development step. It creates more problems, i.e. also in the usage of derived datasets. In addition static variables represent global information on class level which should be avoided.

I can see some way to solve this, I am thinking of a source-sink model similar to the o2 layout where the track container work as kind of processors. Each dataset (LHC11a and LHC11h) in your example would represent a single data source and consequently have their own track container. This is necessary as you cannot apply the same selection process on selection where you have different selection mechanisms. In the jet finder (kind of representing the sink) they are combined to a single particle collection or grid. Drawback is that it requires again rather fundamental changes, this time in the jet finder, and a new abstraction layer (combined particle collection from multiple sources) is needed.

If that foreseen to be implemented it needs to be done in an upcoming implementation step.

Cheers

Markus

raymondEhlers:
Hi @aiola, @mfasDa,

Thanks for your comments!

I think removing the default period would probably be nice eventually, but given @aiola point about it breaking ESDs, it is probably not appropriate right now. Perhaps an intermediate solution could be to defer setting the track container cut period until `AliTrackContainer::SetArray()` and then not use the default period if the container is embedded. However, this will require some testing and development for a fairly negligible benefit. Plus, it may cause further confusion.

I think that @mfasDa's suggested development could be a good path in the future, but would require a lot of work to get there.

Given that it can be worked around pretty easily by setting the track cuts period for the containers and that data-to-data embedding doesn't seem to be super common, I think the best solution for now will be just to document it. I'll create another PR.

Best,
Raymond

mfasDa:
The framework is tested to a large extend and fundamental classes like AliEmcalTrackSelResultPtr are covered by unit tests assuring their performance. If nothing speaks against it I would like to merge the changes into master.

raymondEhlers:
Hi @mfasDa - just to confirm, are these tests run automatically if the user runs `ctest` without specifying the particular test? (to be certain that we don't regress in the future)

mfasDa:
Hi @raymondEhlers ,

If the user just types ctest without any argument all tests which are implemented (not only those by us!) will be executed. This is how ctest works. However this is not what runs in aliBuild. The command aliBuild uses is ctest -R load_library which runs only the load library tests. So even if our test fail they will not disturb the build (but if the test fail one usually wants to know).

For a more automatic way one could think that only the tests related to the library are executed. One does not necessary need to execute unit tests for commits in a library that has absolutely nothing to do with the test. In general I would be in favor of running unit tests more regularly in order to guarantee code stability and integrity. However there is unfortunately nothing we can do about this.

Cheers

Markus

aiola:
Hi @mfasDa,
it is fine to proceed for me as well.
Best,
Salvatore

mfasDa:
+1 

Thanks a lot everyone for the feedback! I will merge it into master.

Cheers 

Markus

mfasDa:
+test

mfasDa:
+test

mfasDa:
+test

Commits to make change ROOT5-compatible ;( will be squashed at a later stage into one commit once tests pass

mfasDa:
+test

The main issue seems to have been that the streaming operator was not forced to be in the namespace PWG::EMCAL and therefore was interpreted by gcc as to be in the global namespace leading to an undefined symbol. Despite of this ROOT5 is also unable to handle std::vector even as argument to the constructor - replaced by TObjArray. Now in local tests based on ROOT5 it compiles and then ctests pass - will be merged after successful build test.

mfasDa:
+1

lcunquei:
Hello,
I have ran a simple test with just jet finders and rho task and it works as usual for the tag of the 16/12,also for the 20/12,but it crashes for the tag of the 21/12 and following tags. The logs are not very revealing but I wonder if it can be connected to this update, I see no other commits in between.
The tests are 2821-2825 of Jets_EMC_PbPb, could you have a look please?

mfasDa:
Hello @lcunquei ,

the crash is fixed in #3683, however there was still a memory leak (connected to a missing SetOwner for the output TObjArray in the functions GetAcceptedTracks). This has been fixed (together with a move from TObjArray to TClonesArray for the combined track selection results reducing dynamic memory allocation) in #3695. I tested locally and now the memory profile stays approximately constant after the initialization. Fix will be in the tag from this evening on.

Cheers

Markus

lcunquei:
Hi Markus, 
I think there is still some problem with the implementation. Jets_EMC_PbPb, train 2848 with today's tag gives a strange output while anything prior to 20/12 works.  The problem appears when I select my trigger track (a track in a certain pT bin), I get the corresponding index (triggerHadronLabel), and then later in the code I retrieve the track from the container via:
 triggerHadron = static_cast<AliAODTrack*>(TrackArray->At(triggerHadronLabel));
where TrackArray = PartCont->GetArray();  
and PartCont = GetTrackContainer(0);
where track container at position zero is the default track container. 
It looks as if triggerHadronLabel points to a different track than the one that was selected. 

This is simple but something must have changed, could you please have a look? 
thanks
Leticia

mfasDa:
Hi @lcunquei ,

It might be that the different track containers have different track selections (maybe your's has none). In the current version non-selected tracks are rejected during NextEvent, so if two track containers have different track cuts there are gaps in one container for rejected tracks. However as the track selection returns AliEmcalTrackSelResultPtr also for rejected tracks it should be feasible to store them as well in the track container and there should be no indexing problem any more. I will have a look.

Cheers

Markus

PR #3621, opened 2017-12-13
Added new calibration OADBs for LHC16d and LHC16e

rderradi:
New multiplicity calibration OADBs for LHC16d and LHC16e based on kINT7 trigger selection.


ddobrigk:
+1

PR #3624, opened 2017-12-14
Changes to existing code for contamination and acceptance studies in …

zkhabano:
…data and MC

PR #3634, opened 2017-12-15
Changes to existing code for contamination and acceptance studies in …

zkhabano:
…data and MC


rbertens:
Hi @zkhabano , this PR is as answer to my comment on #3624 , but it's not exactly what I meant. What you do now, is make a new pointer to the object that's pointed to by aodTrack:

AliAODTrack* copy = dynamic_cast<AliAODTrack *>(aodTrack);

but this means that you still perform ::PropagateToDCA() on the same AliAODTrack instance (as copy and aodTrack now point to the same memory). What I meant was that you make a stack copy of the track, and perform the operation on that copy, i.e. replace the above line by

AliAODTrack copy(*aodTrack);

this creates a new track called copy, identical to the object that aodTrack points to. Then,  if you call copy.PropageToDCA(), the instance of AliAODTrack that aodTrack points to is not changed. Sorry for being so nit-picky :)

rbertens:
+1 thanks @zkhabano :)

PR #3635, opened 2017-12-15
fix memory safety of fHitFlags in AliTOFAnalysisTaskCalibTree

jezwilkinson:
patch to ensure re-initialising fHitFlags[] to 0 for each event (previously was being kept from event to event).

[NB I don't know why git decided to flag the entire file as "changed", but the relevant edit is line 267]


preghenella:
+1

PR #3658, opened 2017-12-18
PWGCF: "global" updates to char* parameters & ConfigObject & PionPion updates

akubera:
* Made `char *` parameters `const char *` in parameters for use in future C++11 environments.
* ConfigObject has new options for parsing and iteration.


akubera:
This has been split into smaller requests, so now only #3713 needs further review

PR #3673, opened 2017-12-21
Adding package loading for Therminator2 MC generation script

maciekniewielki:
Fixing an issue with Therminator2 packages not loading when executing the gen_therm2.sh script.


rbertens:
+1

PR #3674, opened 2017-12-21
PWGDQ: event selection and pile-up rejection for pp run2 analysis

ivorobye:
- code snippets are taken from https://twiki.cern.ch/twiki/bin/view/ALICE/AliDPGtoolsEventProp
- corresponding update of user's addtask and config scripts


tdahms:
+1

PR #3682, opened 2017-12-25
implement choices of pileup cuts and TOF cut in trigger analysis

dsekihat:
implement choices of pileup removal criteria and TOF cut in trigger analsyis


kharlov:
+1

PR #3683, opened 2017-12-25
Fix passing track selection result

mfasDa:
Pointer to local variable in loop over the track selections was passed to the combined track selection (introduced when fixing the compilation under ROOT5 when moving from std::vector to TObjArray).  Copy introduced and TObjArray made owner of the track selection result pointers. Eventually going to TClonesArray for more efficient memory handling.


mfasDa:
+test

mfasDa:
+1

PR #3694, opened 2017-12-29
Change of settings in EPOS-LHC running macro (MCGen_pp train)

gbencedi:
Changed environment variables for testing EPOS-LHC in the on-the-fly MC train.


mpuccio:
+1

PR #3695, opened 2017-12-29
TClonesArray / Fix memory leak

mfasDa:
- Move to TClonesArray for the combined track results
  in order to reduce dynamic memory allocation
- Fixing memory leak (SetOwner missing for TObjArray
  in both functions GetAcceptedTracks)
- Make storage suicide object: Once the ref count is 0
  the storage deletes itself automatically, not handled
  anymore by the pointers referencing it.


mfasDa:
+test

Fixes memory leak introduced with changes in #3612 .

mfasDa:
+1

PR #3698, opened 2017-12-29
Add missing self-destruction

mfasDa:
Should have been part of the previous commit when changing storage handler to self-destruction, was lost somehow. Essential part of the reference counting.


mfasDa:
+1

PR #3702, opened 2018-01-02
GetAcceptedTracks returning selection result for all tracks

mfasDa:
As discussed in #3612 GetAcceptedTracks should rather return
the selection status for all tracks in order to avoid gaps
in the track container.

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliEmcalTrackSelection.cxx
+++ b/PWG/EMCAL/EMCALbase/AliEmcalTrackSelection.cxx
@@ -143,11 +143,7 @@ TObjArray* AliEmcalTrackSelection::GetAcceptedTracks(const TClonesArray* const t
   }
 
   for(auto mytrack : *tracks) {
-    AliVTrack *track = static_cast<AliVTrack *>(mytrack);
-    PWG::EMCAL::AliEmcalTrackSelResultPtr selectionResult = IsTrackAccepted(track);
-    if(selectionResult) {
-      fListOfTracks->AddLast(new PWG::EMCAL::AliEmcalTrackSelResultPtr(selectionResult));
-    }
+    fListOfTracks->AddLast(new PWG::EMCAL::AliEmcalTrackSelResultPtr(IsTrackAccepted(static_cast<AliVTrack *>(mytrack))));
   }
   return fListOfTracks;
 }
@@ -163,10 +159,7 @@ TObjArray* AliEmcalTrackSelection::GetAcceptedTracks(const AliVEvent* const even
   }
 
   for(int itrk = 0; itrk < event->GetNumberOfTracks(); itrk++){
-    AliVTrack *trk = static_cast<AliVTrack*>(event->GetTrack(itrk));
-    PWG::EMCAL::AliEmcalTrackSelResultPtr selectionStatus = IsTrackAccepted(trk);
-    if(selectionStatus)
-      fListOfTracks->AddLast(new PWG::EMCAL::AliEmcalTrackSelResultPtr(selectionStatus));
+    fListOfTracks->AddLast(new PWG::EMCAL::AliEmcalTrackSelResultPtr(IsTrackAccepted(static_cast<AliVTrack*>(event->GetTrack(itrk)))));
   }
   return fListOfTracks;
 }
--- a/PWG/EMCAL/EMCALbase/AliEmcalTrackSelection.h
+++ b/PWG/EMCAL/EMCALbase/AliEmcalTrackSelection.h
@@ -198,19 +198,31 @@ public:
 
   /**
    * @brief Select tracks from a TClonesArray of input tracks
+	 * 
+	 * Running track selection for all tracks in the input array 
+	 * (by calling the IsTrackAccepted function) and storing the track 
+	 * selection result in the output container.
+	 * Attention: The output container contains the result for all
+	 * tracks, not only those selected. It is the user responsibility
+	 * to filter the selected tracks.
    *
    * @param[in] tracks TClonesArray of tracks (must not be null)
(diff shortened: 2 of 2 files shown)


mfasDa:
+test

@lcunquei this should fix the problem with the non-matching indices mentioned in #3612.  Note that the function GetAcceptedTracks now not only returns the selected tracks but all tracks, together with a selection status, which is a bit confusing. It might be worthwhile changing the name of the function to something like GetTrackSelectionStatus. In the former implementation the function only returned the selected tracks, but a second bitmap - with membership of the AliEmcalTrackSelection - had the selection status for all tracks.

Please confirm whether this is what you want to have.

Cheers

Markus

mfasDa:
+1

lcunquei:
Hi  @mfasDa mfasDa 
thanks!
 for my understanding: I was using function GetAcceptParticle() to loop over particles in the container to select a set of tracks and get their indexes. 
 So then, if I GetParticle(i) from the Particle Container array, I don't understand how I could get a different track, since the container is the same. 
When is GetAcceptedTracks called? 
cheers
Leticia

mfasDa:
Hi @lcunquei,

GetAcceptedTracks is called in NextEvent, which is handled by AliAnalysisTaskEmcal in RetrieveEventObjects (so before any kind of user code).

A problem which came in with the new implementation was that in contrast to the previous version non-selected tracks were filtered out. The original track container still had all tracks in the container and just flagged those not selected. When filtering out those we will get gaps in the track container, and with closing the gaps the indices change. This might have an effect if you use track containers in different tasks (i.e. one in the jet finder and one in a user task) with different track selections, here indices are not necessarily portable. Otherwise with standard forward iteration indices should always be consistent, as the track container is filled only once (in NextEvent).

I need more information here:
- You are looking at track level, right?
- Do you use embedding (adding @raymondEhlers)? Here one would have to check if the indexing has consequences on the embedding, which I however currently don't see.
- When and Where do you call GetParticle and GetAcceptParticle?

The changes only affect the track container and not the particle or jet container.

Cheers

Markus

raymondEhlers:
@mfasDa, I'm traveling today, so I'll try to take a closer look tomorrow. It wasn't clear to me how the track selection changed until now - I'm a bit concerned about it for embedding, but I'll have to think about it more closely

mfasDa:
Hi @raymondEhlers, thanks for the answer! Just to clarify: The old handling where the track container keeps all tracks is restored with this commit. From what I understand with this should be fine for the index map, but of course to be checked with you.

Cheers

Markus

raymondEhlers:
Thanks for the clarification! I believe I understand the changes better now and I _think_ this should be fine for embedding, but I'm going to run a test against a reference to confirm

raymondEhlers:
I finally found the time to check the changes within the embedding framework - integration tests seem to indicate that everything is working. Quoted below for reference is the test output (the cpu time mismatch will vary from run to run and can be ignored).

```
Tasks using the new correction framework:
    AliEmcalCorrectionCellCombineCollections_combined
    AliEmcalCorrectionClusterizer_combined
    AliEmcalCorrectionClusterExotics_combined
    AliEmcalCorrectionClusterNonLinearity_combined
    AliEmcalCorrectionClusterTrackMatcher_combined
    AliEmcalCorrectionClusterHadronicCorrection_combined
    AliAnalysisTaskEmcalEmbeddingHelper_histos
    AliAnalysisTaskEmcalJetSample_tracks_caloClusters_truthLevel_histos
    AliAnalysisTaskEmcalJetSample_tracks_caloClustersCombined_emcalCellsCombined_hybridJets_histos
    AliJetResponseMaker_hybridJets_AKTFullR020_tracks_pT3000_caloClustersCombined_E3000_pt_scheme_truthLevel_AKTFullR020_tracks_pT3000_caloClusters_E3000_pt_scheme_Bias0_BiasType1_EMCALfid_histos
    AliAnalysisTaskJetH_tracks_caloClustersCombined_new
Tasks using the old correction framework:
    AliJetResponseMaker_hybridJets_AKTFullR020_tracks_pT3000_caloClustersCombined_E3000_pt_scheme_truthLevel_AKTFullR020_tracks_pT3000_caloClusters_E3000_pt_scheme_Bias0_BiasType1_EMCALfid_histos_reference
    AliAnalysisTaskEmcalJetSample_tracks_caloClusters_truthLevel_histos_reference
    AliAnalysisTaskJetH_tracks_caloClustersCombined_new_reference
    AliEmcalCorrectionClusterizer_combined_reference
    AliEmcalCorrectionClusterNonLinearity_combined_reference
    AliEmcalCorrectionClusterTrackMatcher_combined_reference
    AliAnalysisTaskEmcalJetSample_tracks_caloClustersCombined_emcalCellsCombined_hybridJets_histos_reference
    AliAnalysisTaskEmcalEmbeddingHelper_histos_reference
    AliEmcalCorrectionCellCombineCollections_combined_reference
    AliEmcalCorrectionClusterExotics_combined_reference
    AliEmcalCorrectionClusterHadronicCorrection_combined_reference

Histograms hCPUTime in component AliEmcalCorrectionClusterizer_combined mismatch!
Old entries: 510.0, New entries: 510.0
subtracted max = 11.0
subtracted min = -8.0
Print the histogram with the "-p" option to investigate more closely.
Histograms hRealTime in component AliEmcalCorrectionClusterizer_combined mismatch!
Old entries: 510.0, New entries: 510.0
subtracted max = 5.0
subtracted min = -5.0
Print the histogram with the "-p" option to investigate more closely.

Number of histograms in new correction components: 1068
Number of matched histograms: 1068
Number of unmatched histograms due to not finding old component: 0
Number of unmatched histograms: 0
Number of correct histograms: 1066
Number of incorrect histograms: 2
Number of empty histograms: 346
```

PR #3713, opened 2018-01-04
PWGCF: Enforce const-correct character string parameters

akubera:
`ISO C++11 does not allow conversion from string literal to 'char *'`
^ This rule would cause errors when trying to construct an AliFemto
object using a standard string literal when using root6.

This commit changes `char *` parameters to `const char *` (and in the
process of saving, also removed trailing whitespace in affected files).

AliFemtoCorrFctnDPhiStarDEtaStar - Broke up long line


dberzano:
+test

dberzano:
+1

PR #3714, opened 2018-01-04
implement dummy run number to enable analysis of single particle simulation

dsekihat:
One member variable Int_t fRDN which is dummy run number to enable us to analyze single particle simulation have been introduced. As PHOS OCDB does not change in Run2, this modification allows us to analyze single particle simulation (for example LHC17i7[a,b,c][1,2]) for all period by reading proper OADB which one wants to use in PHOSTender.


peressounko:
+1 I agree

PR #3722, opened 2018-01-04
Fix for GetAcceptedTracks + fixes of Add-macros for ROOT6

mfasDa:
Fixing issue that GetAcceptedTracks returns also non-accepted tracks, introduced in #3702 (extra check needed as the track selection now returns the selection status of all tracks and not only the accepted tracks).

In addition two macros were made ROOT6-aware (mismatch in types + missing default arguments).

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliTrackContainer.cxx
+++ b/PWG/EMCAL/EMCALbase/AliTrackContainer.cxx
@@ -541,6 +541,7 @@ Bool_t AliTrackContainer::AcceptTrack(const AliVTrack *vp, UInt_t &rejectionReas
  */
 Bool_t AliTrackContainer::AcceptTrack(Int_t i, UInt_t &rejectionReason) const
 {
+  if(fTrackTypes[i] == kRejected) return false; // track was rejected by the track selection
   Bool_t r = ApplyTrackCuts(GetTrack(i), rejectionReason);
   if (!r) return kFALSE;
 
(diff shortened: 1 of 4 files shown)


mfasDa:
+test

mfasDa:
+1

PWG/EMCAL/EMCALbase/AliTrackContainer.cxx
```diff
@@ -628,6 +629,37 @@ PWG::EMCAL::AliEmcalTrackSelResultHybrid::HybridType_t AliTrackContainer::GetHyb
   return hybridDefinition;
 }
 
+Bool_t AliTrackContainer::CheckArrayConsistency() const {
```
raymondEhlers: I don't see where this function is called - did you just add it as a testing utility?

mfasDa: Exactly, this is a test function. It was used to investigate the problems Leticia mentioned in a different thread.

I would like to keep it in the compiled code as part of a future unit test, and in order to have access to both arrays to has to be a member function as direct access to at least one of the arrays is prohibited (I would actually prefer prohibiting direct access to both arrays - that's the task of the container). Array consistency is important as it will create problems when accessing tracks with indices.

raymondEhlers: Okay, thanks for the clarification!

PR #3723, opened 2018-01-04
Increase digits in the output string SetRangeUser

dcolell:
In my analysis code I'm using many AliCFcontainer objects.
When changing the range for an axis, the method SetRangeUser prints on the screen a sentence like:
W-AliCFGridSparse::SetRangeUser: fCFContCascadeCuts_minnTPCcls70_vtxlim10.0-0.0_minptdghtrk0.0_etacutdghtrk0.8_revertex_nTPCcls70_SelStep0 new range: 0.0 < Dca(cascade daughters) (cm) < 2.0
As you can see only the first decimal digit is displayed. 
I need more digits to be displayed, to check the value of the cut applied.


dberzano:
+1

PR #3724, opened 2018-01-04
Updated multiplicity calibration OADB for LHC15i using pass2

rderradi:
New calibration OADB reprocessed using pass2 for LHC15i


ddobrigk:
+1

PR #3741, opened 2018-01-08
Fixes TriggerHist calculation of event plane bin

moliver813:
Fixes TriggerHist calculation of event plane bin, now it is the same as the event plane bin determination for CorrVsManyThings.


e-masson:
+test

e-masson:
+1

PR #3756, opened 2018-01-09
switched PID deuteron cut from pp -13TeV-2015/2016/2017   to   p-Pb-2013

bschaefe:
Also added histograms and changed cut thresholds.


aohlson:
+1

PR #3771, opened 2018-01-10
PWGDQ: Update of new efficiency task

carstooon:
Update of new dielectron efficiency task:
- including ULS and LS histograms


pdillens:
+1

PR #3778, opened 2018-01-11
ZDCTreeTask Update

udmitrie:
Some variables were deleted from tree to reduce its size. Information about run number was included.


mbroz84:
+1

PR #3781, opened 2018-01-11
Option to skip cuts for MC event, default: apply cuts on MC event

rihanphys:
Option to skip cuts for MC event, default: apply cuts on MC event


jmargutt:
+1

PR #3783, opened 2018-01-11
PhotonQA: add possibility to reduce the event statistics in the tree

m-c-danisch:


dmuhlhei:
Could you please fix the following:

- ffillTree and kTree should be Double_t
- please increment the classdef

thanks :)

dmuhlhei:
+1

PR #3788, opened 2018-01-12
Added Jet core task (h+jet) with Emcal framework

jaimenorman:
Jet core analysis task and AddTask macro added for semi-inclusive h+jet analysis in run 2 (main basic functionality the same as AliAnalysisTaskJetCore, Emcal functionality copied from AliAnalysisTaskEmcalJetSample)

### DIFF
--- a/PWGJE/PWGJE/AliAnalysisTaskJetCoreEmcal.cxx
+++ b/PWGJE/PWGJE/AliAnalysisTaskJetCoreEmcal.cxx
@@ -1,0 +1,944 @@
+/**************************************************************************
+ * Copyright(c) 1998-2016, ALICE Experiment at CERN, All rights reserved. *
+ *                                                                        *
+ * Author: The ALICE Off-line Project.                                    *
+ * Contributors are mentioned in the code where appropriate.              *
+ *                                                                        *
+ * Permission to use, copy, modify and distribute this software and its   *
+ * documentation strictly for non-commercial purposes is hereby granted   *
+ * without fee, provided that the above copyright notice appears in all   *
+ * copies and that both the copyright notice and this permission notice   *
+ * appear in the supporting documentation. The authors make no claims     *
+ * about the suitability of this software for any purpose. It is          *
+ * provided "as is" without express or implied warranty.                  *
+ **************************************************************************/
+
+#include <TClonesArray.h>
+#include <TList.h>
+#include "TChain.h"
+#include "TTree.h"
+#include "TMath.h"
+#include "TH1F.h"
+#include "TH2F.h"
+#include "TH3F.h"
+#include "THnSparse.h"
+#include "TCanvas.h"
+#include "TRandom3.h"
+#include "AliLog.h"
+
+#include <AliAnalysisManager.h>
+#include <AliVEventHandler.h>
+#include <AliVEvent.h>
+#include <AliVCluster.h>
+#include <AliVParticle.h>
+#include <AliLog.h>
+
+#include "AliTLorentzVector.h"
+#include "AliEmcalJet.h"
+#include "AliRhoParameter.h"
+#include "AliJetContainer.h"
+#include "AliParticleContainer.h"
+#include "AliClusterContainer.h"
+
+#include "AliAnalysisTaskJetCoreEmcal.h"
+
+/// \cond CLASSIMP
+ClassImp(AliAnalysisTaskJetCoreEmcal);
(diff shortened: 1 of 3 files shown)


PWGJE/PWGJE/AliAnalysisTaskJetCoreEmcal.h
```diff
@@ -0,0 +1,122 @@
...
+class TH2F;
+class TH3F;
+class THnSparse;
+class TRandom3;
+
+#include "AliEventCuts.h"
+#include "AliAnalysisTaskEmcalJet.h"
+#include "THistManager.h"
+
+  enum JetShapeType {
```
raymondEhlers: This enum should be inside of your class

PWGJE/macros/AddTaskJetCoreEmcal.C
```diff
@@ -0,0 +1,76 @@
...
+  TString contname(name);
+	contname += Form("_%s",TString(listName).Data());
+	contname += Form("_%02d",Int_t(R*10+0.001));
+  contname += "_histos";
+  AliAnalysisDataContainer *coutput1 = mgr->CreateContainer(contname.Data(),
+      TList::Class(),AliAnalysisManager::kOutputContainer,
+      Form("%s", AliAnalysisManager::GetCommonFileName()));
+  mgr->ConnectInput  (task, 0,  cinput1 );
+  mgr->ConnectOutput (task, 1, coutput1 );
+	cout<<"emcal add task settings - return;"<<endl;
```
raymondEhlers: Needs to be `std::cout` for ROOT 6 compatibility. Generally good to use AliDebugStream(int), AliInfoStream(), etc.

PWGJE/PWGJE/AliAnalysisTaskJetCoreEmcal.cxx
```diff
@@ -0,0 +1,945 @@
...
+}
+
+/**
+ * Performing run-independent initialization.
+ * Here the histograms should be instantiated.
+ */
+void AliAnalysisTaskJetCoreEmcal::UserCreateOutputObjects()
+{
+  AliAnalysisTaskEmcalJet::UserCreateOutputObjects();
+	cout<<"in task - create output objects"<<endl;
```
raymondEhlers: Needs to be `std::cout` for ROOT 6 compatibility. Generally good to use AliDebugStream(int), AliInfoStream(), etc.

raymondEhlers:
+test

PWGJE/macros/AddTaskJetCoreEmcal.C
mfasDa: This is in a macro, AliDebug/Info/WarningStream won't work here (required to be called within non-static member functions).  As long it is in the user macro missing std:: - though not recommendable - can be tolerated. However in the compiled code (see above) this must be fixed because it will break compilation

jaimenorman: thanks for suggestions - this was leftover from testing and I don't need this output. I have removed in latest push

PWGJE/PWGJE/AliAnalysisTaskJetCoreEmcal.h
jaimenorman: Thanks, fixed

PWGJE/PWGJE/AliAnalysisTaskJetCoreEmcal.cxx
jaimenorman: Thanks, I didn't need this output so removed this line

PWGJE/macros/AddTaskJetCoreEmcal.C
raymondEhlers: @mfasDa - Yeah, you're right, I wasn't being super precise. One could use `AliDebugGeneralStream(...)` for such situations, but in any case, it's irrelevant here

raymondEhlers:
I didn't check it extremely closely, but from what I've looked at, it looks good to me now. I'll leave it open a bit longer so others can comment, but I'll approve it in a few hours if nothing else is mentioned

raymondEhlers:
+1

raymondEhlers:
Revising my previous comment - I wasn't keeping close track of the time and thought that we were already passed tagging time. Since we are not and this commit shouldn't cause any problems, I've approve it so that it can be included in the tag

PR #3793, opened 2018-01-12
introduce minimum energy for eta meson

dsekihat:
introduce minimum energy cut for eta meson analysis to reduce combinatorial background.


kharlov:
+1

PR #3795, opened 2018-01-13
EMCal CorrFW: Track to container map for had corr

raymondEhlers:
Creates a mapping for tracks to their corresponding container in the
hadronic correction when running with AODs. AOD clusters only
provide pointers to tracks, so this map is necessary to select the
proper container, which allows the proper track
selection to be applied to each track. This is important if the
containers have different track selection conditions (as might happen
in, for example, embedding pp data into PbPb).

I've verified using the correction framework testing tools and comparison script
that this change doesn't modify the behavior of the correction when not
using embedding.


raymondEhlers:
+1

raymondEhlers:
+1 Fixed warnings

PR #3805, opened 2018-01-15
Adding INEL>1 centrality estimator

afestant:
Adding new centrality estimator, INEL>1: classes of events defined on the basis of the number of charged tracks with pT>0.4 GeV/c and |eta|<0.8, selected through filter bit 32 and 64.


ddobrigk:
+1

PR #3816, opened 2018-01-16
Added Jet core task (h+jet) with Emcal framework - move to EmcalJetTa…

jaimenorman:
…sks folder + correct links

Task was originally comitted to PWGJE/PWGJE, I have moved to EmcalJetTasks since it is written using Emcal jet framework. I also added the relevant lines to CMakeLists.txt and PWGJEEMCALJetTasksLinkDef so the task will build. Finally a couple of changes are made to the task to remove compilation warnings.


mfasDa:
+1

PR #3821, opened 2018-01-16
Add Xe-Xe MC productions support

victor-gonzalez:
Also, simplify the configuration process in train scenarios in the associated macros


rbertens:
+1

PR #3832, opened 2018-01-17
Enabled mixed event analysis for data and MC in parallel

matarzil:
Enabled mixed event analysis for data and MC in parallel; combined multiplicity only.


c-andrei:
+test

c-andrei:
+1

PR #3836, opened 2018-01-17
Add macro for new efficiency task (Carsten's task)

rbailhac:
Add a macro to run the efficiency task of Carsten on MC ESDs for pp 7 TeV as cross-check


otonvd:
+1

PR #3845, opened 2018-01-18
Rolling back the includes inclusion in AddTask macro

victor-gonzalez:
It seems that the standard include paths are not ready at
AddTask macro execution time in train scenarios. They are in local
train test scenario!


aohlson:
+1

PR #3846, opened 2018-01-18
Fixes for EMCAL MC trigger selection

mfasDa:
- In trigger selection task fix jet thresholds for 2016 pp
   LHC16h-k
- In substructure tree task run MC trigger selection only
   when an EMCAL trigger is in the selection string
-  In substructure tree task move from TString to std::string
   / std::stringstream (works a bit better with lldb/gdb)


mfasDa:
+test

mfasDa:
+1

PR #3854, opened 2018-01-19
PWGDQ/reducedTree: added writing of additional histograms from dstTre…

laltenka:
…e.root into analysis output to runAnalysisTrain.C


iarsene:
+1

PR #3855, opened 2018-01-19
Prevent container name to be reset

mfasDa:
List objects must overwrite Clear, otherwise TNamed::Clear is called (by the analysis manager or input handler) which resets the name of the list object.

In addition name of the trigger decision container in user task configurable.


mfasDa:
+1

PR #3864, opened 2018-01-19
Further check for the single electron efficiency

rbailhac:
Give the possibility to calculate the single electron efficiency for electrons which does not form with their partner an invariant mass below fLimitmee. Per default this check if off so that nothing is changing for the users, if it does not see fCheckmee == kTRUE in his config file.


rbailhac:
Hi all,
does somebody know why the pull request seems to be still pending with "pending approval" ?
Thanks for the help,
Raphaelle

otonvd:
+1

PR #3867, opened 2018-01-20
Therminator2 parameters configuration

maciekniewielki:
This pull request adds the possibility to configure Therminator2 using custom parameters. It uses a python script to scan environment variables and set the key=value pairs in configuration files according to them.

The script works by scanning the configuration files line by line. When it finds a parameter_name=value pair, it checks if the corresponding THERM2_PARAMS_<parameter_name> env variable is set. If yes, the default value in the line is substituted for env variable value.

The scanning is first done on the general [events.ini](https://github.com/alisw/therminator/blob/alice/v2.0.3/events.ini) file. Then, according to the selected model, a proper .ini file in [fomodel](https://github.com/alisw/therminator/tree/alice/v2.0.3/fomodel) directory is changed.

An example of using the environment variables to set the Krakow model with non-default temperature of 140MeV:
`gSystem->Setenv("THERM2_PARAMS_FreezeOutModel", "KrakowSFO"); //Set in events.ini`
`gSystem->Setenv("THERM2_PARAMS_Temperature", "140.0"); //Set in fomodel/krakow.ini`

There is also a way to select a custom xml path (via the THERM2_PARAMS_XML_PATH) to the freeze-out hypersurface in case one of the hydro models is selected. It can either be a local path to the built-in xml files, relative to the fomodel directory (such as lhyquid2dbi/LHCPbPb5500c0005Ti500ti100Tf145.xml) or a GRID path to the custom xml, beginning with the "alien:" prefix (such as alien:/alice/cern.ch/user/m/mbuczyns/Therminator2_custom_hypersurface/custom_lhc_5.5TeV_c0005.xml)

### DIFF
--- a/PWG/MCLEGO/CMakeLists.txt
+++ b/PWG/MCLEGO/CMakeLists.txt
@@ -69,5 +69,7 @@ install(DIRECTORY AMPT
 # install THERMINATOR2 configurations
 install(DIRECTORY Therminator2
   DESTINATION PWG/MCLEGO
-  FILES_MATCHING PATTERN "*.ini"
+  FILES_MATCHING
+  PATTERN "*.ini"
+  PATTERN "*.py"
   )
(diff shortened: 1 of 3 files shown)


rbertens:
+1

PR #3872, opened 2018-01-21
add Emin also in embedding task and single sim

dsekihat:
add minimum energy cut also in embedding task and single simulation


kharlov:
+1

PR #3887, opened 2018-01-22
PWGDQ/reducedTree: fix for use of old trees

laltenka:
- fix in analysis task for use of trees w/o second track array
- slight modification of writing of histograms from tree to analysis output file


iarsene:
+1

PR #3894, opened 2018-01-22
PWGDQ: crash fix + generator name handling

carstooon:
Should prevent crashes because of uninitialized FillMap.
Added feature to only look  at particles from given generators.


wiechula:
+1

wiechula:
+1

PR #3900, opened 2018-01-22
PWGDQ/reducedTree: setup functionality in AliReducedEventCut to make …

laltenka:
…exclusion cuts based on event tag bits


iarsene:
Hi Lucas,
I checked the code. I see that the exclusion flag is applied globally to the whole event filter. Since in the event filter we store various decisions which are not necessarily related, we should be able to apply the exclusions independently for each filter bit. So, instead of one single boolean as you implemented now, we should either add an ULong64_t bit map or a Bool_t[64] to hold all possible exclusion flags.
Cheers, Ionut

laltenka:
Sorry to everyone I spammed, I accidentally included all commits between my fork and master, which is why I closed the pull request. Please ignore.

PR #3904, opened 2018-01-22
add fallback behavior to get nEvents, depending on which histograms a…

jdmulligan:
…re included in task.

(minor update to Jet QA post-processing script to scale pT-hard bins, to allow compatibility with AliEventCuts)


raymondEhlers:
+1

PR #3931, opened 2018-01-24
usage of data driven corrections and after corrections QA plots (Davide)

pchrista:


rbertens:
+1 @pchrista in the functions GetNu*Correction, dynamic cast, string manipulation, and string comparison are performed for - as far as I can see - each track. If really called for each track this would be very inefficient, maybe better static_cast and use a ThashList or so?

PR #3949, opened 2018-01-25
Fix dictionary problem with GCCv7

dberzano:


dberzano:
Along with alisw/AliRoot#551 this aims to make AliRoot/AliPhysics compatible with GCCv7 (used, for instance, on Ubuntu 17.10 and 18.04).

PR #3956, opened 2018-01-25
Fix multiple-deletes for TList in ROOT v6-12-04

mfasDa:
The promise that a TList will not delete objects
where it is not owner of seems to be invalid in
ROOT v6-12-04 as it tried to delete obeject in
a list where SetOwner was explicitly set to false.
When calling Clear it has to be protected with
the option "nodelete".


mfasDa:
+1

PR #3958, opened 2018-01-25
 Added possibility to study Lc->pK0S vs multiplicity

decaro:
Added possibility to study Lc->pK0S vs multiplicity.
The code modified is already used for D menso study vs multiplicity.


zconesa:
Ciao Annalisa,

I think the modifications you did would impact the other meson badly. 
In particular on the line 1067 of the new code, you are applying the selection bit only for Lc case, while it was and should be done for all mesons. I spotted this and didn't check the rest, could you please have a closer look and check these things?
I also saw you changed something on the header parameters, that does not need to be modified and could be maintained.
Could you please check them? Thanks !

Cheers, Zaida

decaro:
Ciao Zaida.

Sorry, maybe I am wrong.
But on the line 1067 of the new code I am applying the selection bit for all D meson and not for Lc case. And this is what I would like.
Thanks for your feedback and please, let me know if I can cause problems to the D meson analyses.
It is not my intention.
   Cheers
   Annalisa

zconesa:
+test

zconesa:
Ciao Annalisa,
You're right, my wrong interpretation, apologies! 
Will write you an email for another (not crucial) detail.
Thanks and cheers, 
Zaida

zconesa:
+1

PR #3960, opened 2018-01-25
introduce flexible trigger analyses

dsekihat:
choice of trigger analyses is introduced to measure rejection factor.


kharlov:
+1

PR #3961, opened 2018-01-26
Task for acceptance and contamination calculation in the data

zkhabano:


rbertens:
+1 @zkhabano for future commits ,it would be good if you could give a bit more descriptive message than 'Task for acceptance and contamination calculation in the data'. This can later make it easier to see which changes you introduced at what point in time. 
My best,
Redmer

PR #3965, opened 2018-01-26
Included sphericity in the event mixing analysis.

matarzil:
Included sphericity selections in the event mixing analysis as a function of particle multiplicity.


c-andrei:
+1

PR #3971, opened 2018-01-26
Optimising retrieval of correction objects (Davide)

pchrista:


rbertens:
+1

pchrista:
thanks @rbertens for the previous comment
@dcaffarr tried to optimize a bit the procedure
Maybe things can still be optimized further; if so more updates will come

PR #3989, opened 2018-01-28
EP vs. centr. vs. q2 histos filled with q2 percentile, if flag enabled

fgrosa:


arossi81:
+1

arossi81:
Ciao Fabrizio, I approved you changes. Just a comment (very low priority): I see that in the task you get the histograms by "TList::FindObject(char* histname)" method. It might be more efficient to add each histogram as data member, and just use the pointer when filling it. 
Ciao
Andrea

fgrosa:
Ciao Andrea,
thanks for the suggestion! Indeed I agree with you, however it was like that when I started to work on it and I never changed it (I continued in the same way, for consistency). I guess that it was done this way, since the number of some histograms (e.g. for EP resolution) depends on the number of centrality bins that one sets. I will think how to do that in a more efficient way. 
Ciao,
Fabrizio

PR #3992, opened 2018-01-29
Fix handling of charged / neutral constituents for MC truth

mfasDa:
For MC-true particles both charged and neutral particles are in
the same container. In order to look at charged / neutral
constituents rejecting the track container is not sufficient,
but the disctiction has do be on particle-by-particle.


mfasDa:
+1

PR #3998, opened 2018-01-29
Pre-select jets based on number of constituents

mfasDa:
For jet substructure jets must have at least 1 constituent
for reclustering, consequently reclustering fails if one
only looks at charged / neutral constituents and has none
of that type's constituents. A preselection makes sure the
jet has at least one allowed constituent.


mfasDa:
+1

PR #4003, opened 2018-01-29
Added possibility to make PID for impPar checks

decaro:
Added possibility to check impact parameter, in terms of mean and sigma, separately for identified particles (on the base of TPC and TOF informations)


fprino:
+1

PR #4004, opened 2018-01-29
Add splines for LHC17p pass1

wiechula:


wiechula:
@chiarazampolli , could you please approve.
I couldn't yet test it fully, since my AliRoot/Physics compilation is very slow. But I don't expect problems. In case I see problem I will not announce the new splines, but fix them for tomorrow.
Cheers,
Jens

chiarazampolli:
Done!

PR #4009, opened 2018-01-30
EMCal Embed: Handle pt hard = 0

raymondEhlers:
Such events likely correspond to grid or header issues. In such cases,
they should be rejected. This is performed via the external event
selection.


raymondEhlers:
+1

PR #4010, opened 2018-01-30
[WIP] EMCal Embed: Internal event selection

raymondEhlers:
__Work in Progress! Please do not commit!__

Previously, good embedded events were lost when the internal event
was outside of, for example, the desired centrality range. Now,
the user should check for
`AliAnalysisTaskEmcalEmbeddingHelper::GetIstance()->IsGoodEvent()`
to ensure that the internal event passes event selection and that
the embedded event was embedded successfully. If the internal event
did not pass event selection, the embedded event is saved for the
next internal event.

The code should be ready for code review, but it is still undergoing some
testing. @aiola - Would you mind reviewing this code? I think this simple solution is
fine, but I would appreciate another set of eyes which is familiar with the embedding
framework!

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
+++ b/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
@@ -34,6 +34,7 @@
 #include <TProfile.h>
 #include <TH1F.h>
 #include <TRandom3.h>
+#include <TList.h>
 
 #include <AliLog.h>
 #include <AliAnalysisManager.h>
@@ -77,6 +78,12 @@ AliAnalysisTaskEmcalEmbeddingHelper::AliAnalysisTaskEmcalEmbeddingHelper() :
   fRandomEventNumberAccess(kFALSE),
   fRandomFileAccess(kTRUE),
   fCreateHisto(true),
+  fUseInternalEventSelection(false),
+  fUseManualInternalEventCuts(false),
+  fInternalEventCuts(),
+  fEmbeddedEventUsed(true),
+  fCentMin(-999),
+  fCentMax(-999),
   fAutoConfigurePtHardBins(false),
   fAutoConfigureBasePath(""),
   fAutoConfigureTrainTypePath(""),
@@ -138,6 +145,12 @@ AliAnalysisTaskEmcalEmbeddingHelper::AliAnalysisTaskEmcalEmbeddingHelper(const c
   fRandomEventNumberAccess(kFALSE),
   fRandomFileAccess(kTRUE),
   fCreateHisto(true),
+  fUseInternalEventSelection(false),
+  fUseManualInternalEventCuts(false),
+  fInternalEventCuts(),
+  fEmbeddedEventUsed(true),
+  fCentMin(-999),
+  fCentMax(-999),
   fAutoConfigurePtHardBins(false),
   fAutoConfigureBasePath("alien:///alice/cern.ch/user/a/alitrain/"),
   fAutoConfigureTrainTypePath("PWGJE/Jets_EMC_PbPb/"),
@@ -309,6 +322,9 @@ bool AliAnalysisTaskEmcalEmbeddingHelper::GetFilenames()
       auto result = gGrid->Query(filePattern.Data(), fInputFilename.Data());
 
       if (result) {
+        // Retrieve the good run list (if specified)
+        fYAMLConfig.GetProperty("runlist", fEmbeddedRunlist, false);
+
         // Loop over the result to store it in the fileList file
         std::ofstream outFile(fFileListFilename);
         for (int i = 0; i < result->GetEntries(); i++)
@@ -404,6 +420,11 @@ bool AliAnalysisTaskEmcalEmbeddingHelper::IsGoodEmbeddedRun(TString path)
 
 }
 
+/**
+ * Initialize the YAML configuration with a potentially specified YAML configuration file.
+ *
+ * @return true if no yaml file or it exists and was successfully accessed.
+ */
 bool AliAnalysisTaskEmcalEmbeddingHelper::InitializeYamlConfig()
 {
   if (fConfigurationPath == "") {
@@ -412,16 +433,11 @@ bool AliAnalysisTaskEmcalEmbeddingHelper::InitializeYamlConfig()
   else {
     AliInfoStream() << "Embedding YAML configuration was provided: \"" << fConfigurationPath << "\".\n";
 
-    PWG::Tools::AliYAMLConfiguration config;
-    
-    int addedConfig = config.AddConfiguration(fConfigurationPath, "yamlConfig");
+    int addedConfig = fYAMLConfig.AddConfiguration(fConfigurationPath, "yamlConfig");
     if (addedConfig < 0) {
       AliError("YAML Configuration not found!");
       return false;
     }
-    
-    config.GetProperty("runlist", fEmbeddedRunlist, false);
-    
   }
   return true;
 }
@@ -908,6 +924,49 @@ void AliAnalysisTaskEmcalEmbeddingHelper::UserCreateOutputObjects()
 {
   SetupEmbedding();
 
+  // Setup AliEventCuts
+  // NOTE: Need to define the base name here so that the property path is not ambiguous (due to otherwise only being `const char *`)
+  std::string baseName = "internalEventSelection";
+  bool res = fYAMLConfig.GetProperty({baseName, "enabled"}, fUseInternalEventSelection, false);
+  if (fUseInternalEventSelection)
+  {
+    AliDebugStream(1) << "Configuring AliEventCuts for internal event selection.\n";
+    // Handle manual cuts
+    res = fYAMLConfig.GetProperty({baseName, "useManualCuts"}, fUseManualInternalEventCuts, false);
+    if (res && fUseManualInternalEventCuts) {
+      fInternalEventCuts.SetManualMode();
+      // Implement these cuts by retrieving the event cuts object and setting them manually.
+    }
+
+    // Centrality
+    // TODO: Move to YAML configuration function in post event selection update
+    std::vector <double> centralityRange;
+    bool res = fYAMLConfig.GetProperty({baseName, "centralityRange"}, centralityRange, false);
+    if (res) {
+      if (centralityRange.size() != 2) {
+        AliErrorStream() << "Passed centrality range with " << centralityRange.size() << " entries, but 2 values are required. Ignoring values.\n";
+      }
+      else {
+        AliDebugStream(1) << "Setting internal event centrality range to [" << centralityRange.at(0) << ", " << centralityRange.at(1) << "]\n";
+        fCentMin = centralityRange.at(0);
+        fCentMax = centralityRange.at(1);
+      }
+    }
+
+    // Trigger selection
+    bool useEventCutsAutomaticTriggerSelection = false;
+    res = fYAMLConfig.GetProperty({baseName, "useEventCutsAutomaticTriggerSelection"}, useEventCutsAutomaticTriggerSelection, false);
+    if (res && useEventCutsAutomaticTriggerSelection) {
+      // Use the autmoatic selection. Nothing to be done.
+      AliDebugStream(1) << "Using the automatic trigger selection from AliEventCuts.\n";
+    }
+    else {
+      // Use the cuts selected by SelectCollisionCandidates()
+      AliDebugStream(1) << "Using the trigger selection specified with SelectCollisionCandidates()\n.";
+      fInternalEventCuts.OverrideAutomaticTriggerSelection(fOfflineTriggerMask);
+    }
+  }
+
   if (!fCreateHisto) {
     return;
   }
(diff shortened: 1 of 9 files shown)


aiola:
Hello Raymond,
I haven't followed all the details of the development of the embedding framework, so I apologize in advance in case my suggestions are not feasible.
Having said that, my question is: why do you need to apply the selection of the internal event in the embedding helper class? The way I would have done it is to have the user task set a flag in the embedding helper to confirm that the current embedded event was used. Unless I am missing something, this solution would appear to me simpler and more robust. What do you think?
Best,
Salvatore

raymondEhlers:
Hi @aiola,

Thanks for your thoughts! I think there are pros and cons to each approach.

For your suggestion, the pros seems to be that any (potentially complicated) event selection doesn't need to go in the embedding helper. The cons seem to be that if users aren't careful, they'll just have the same event embedded over and over again (for example, if they forget to set the flag). This should be obvious, so hopefully it's okay, but I would think that ideally, we could be a bit more automatic. In terms of code implementation, I'll have to think over how to do it, because when I tried to code this up, requiring affirmative confirmation seemed more complicated than setting a bad event flag (which wouldn't be possible here).

For my proposed implementation, the pros seem to be that embedding can work without any user intervention. The cons are that you have to match event selections in the embedding helper, which may be a pain. (Any more?)

I'll try to code up your suggestion and see how it looks. Perhaps we can discuss them both once they are available.

Best,
Raymond

aiola:
Hello Raymond,

I see your point. Would it be possible to automatically set the "event-used" flag once any method that grants access to the external event properties is called?
This would include GetExternalEvent, GetEventHeader etc.

Salvatore

raymondEhlers:
That's a very clever idea! I'll think about the implementation and report back

raymondEhlers:
Hi @aiola , @jdmulligan ,

I've put together an implementation based on the above discussion. You can see the idea [here](https://github.com/alisw/AliPhysics/compare/master...raymondEhlers:embeddingEventSelection). It requires a bit more testing, but I think it is a bit cleaner.

I'll need to clean up the documentation as well, but here is the rough idea:

- The embedding helper will be configured to look for a signal from another task to move to the next embedded event. This is set via `AliAnalysisTaskEmcalEmbeddingHelper::SetRequireSignalToAdvanceEmbeddedEvent()`. If the user doesn't enable this mode, everything proceeds as previously.
- The user task is expected to let the embedding helper know that it is configured to send the signal by calling the function `AliAnalysisTaskEmcalEmbeddingHelper::RegisterTaskForEventUsedSignaling()` (name subject to change, as it's not very good). This is to ensure that the embedding helper doesn't repeat the same event indefinitely (basically a sanity check). It also passes the name of the task, because the embedding helper can alternatively be configured with a list of tasks that should signal. This is more or less hidden in the embedding QA task (although a user could do it by hand if desired).
- If the user task is configured to signal, then it will do so via `AliAnalysisTaskEmcalEmbeddingHelper::EmbeddedEventWasUsed()` when the embedded event is used. Again, this is hidden in the embedding QA task.

Due to some API changes in the embedding QA task, I had to modify @jdmulligan tasks, so please take look (it is extremely minimal, so I don't suspect it will be a problem).

I recognize that there is a bit more complexity than originally envisioned, but I think the sanity checks are important, as is the flexibility in configure the tasks. But I'm certainly open to thoughts here. (Generally, any feedback would be appreciated!)

What do you think?

Best,
Raymond

aiola:
Hello @raymondEhlers,
indeed this implementation became quite a bit complex!
I apologize for asking more questions, I just want to make sure that I understand the logic.
1) One thing that we should avoid is the possibility of user tasks interfering with each other. If a task can decide whether or not to use the embedded event based on the properties of the embedded event itself, this could bias the sample also for the other tasks. Instead if the decision about whether to use or not the embedded event is based solely on the properties of the internal event, then the embedded event sample will stay unbiased. Do you agree? Can you confirm that after reading the properties of the embedded event, there is no way that a user task can "change its mind" and tell the embedding helper to keep the event in the stack for the next internal event?
2) What is the advantage of this implementation compared to my proposal of simply signalling the "embedded-event used" flag as soon as any method of the embedding helper that reads off the properties of the event is called? I am not asking this because I necessarily want to push for my proposal, just to understand the need for this more complex implementation compared to other alternatives.
Thanks,
Salvatore

raymondEhlers:
Hi @aiola,

No need to apologize! I'm looking for ideas to improve things, so your questions are great!

1. Yes, I agree that this should only be used to select on the internal event. I imagine it as being called as one of the first things that I user would call in, for example, `Run()`. Yes, once a task calls `EmbeddedEventWasUsed()`, then the stack advances. There is no possibility to go back.
2. Here was my thinking regarding the very simple approach. For concreteness, let's use wanting to select 0-10% as an example.
    - Some functions are called too commonly to be helpful as a proxy. For example, GetExternalEvent() will be called by the correction task, but since the correction task doesn't have centrality selection, the embedded properties will be accessed, but the internal event could be rejected by the user task based on centrality. So we need something more selective.
    - My next thought was to use the EMCal embedding QA task as a proxy, because if that was called, then it almost certainly in `Run()` of a user task. But then I was concerned that there could be a task in the chain (say, the jet response maker), that may have broader event selection than the final user task. And I think this could be easy for the user to miss. So I wanted a way to turn on and off this "signaling" for a particular task. If this concern could be set aside, it could simplify the code quite a bit, so perhaps it is worth considering getting rid of this (but then it could take much more care in configuration).
     - Since the embedding QA helper isn't in AliAnalysisTaskEmcal (I guess it could...), I was not entirely thrilled about adding a bunch of accessor functions for each task (although I did add them for the existing tasks), so I wanted to allow it to either be configured in the task (via a setting in the QA helper), or via configuration in the embedding helper (via the task name(s) in yaml)
     - Given the complexity of the above, I also wanted a sanity check to ensure that if the user enabled this mode that they were definitely had enable advancing events. So hence the `RegisterTaskForEventUsedSignaling()`. And I then took advantage of this to handle the previous step too.

I'm not actually strongly attached to this approach - I just don't see the any other way at the moment. Do you have any ideas on how to simplify it? Would you just remove the additional task configuration options?

Best,
Raymond

aiola:
Hi Raymond,
thanks for elaborating on the details. It is much clearer now.
My approach would be to ignore all these possibilities. If an embedded event is used by _any_ task (including the correction task), it should not be re-used again. In fact, also the correction task has some QA plots and you may not want to have the same embedded event repeated over and over. On the other hand, if no task is going to use events outside the 0-10 centrality, then one can configure the correction task to only analyze those events in order to make the entire train more efficient (less time spent on looping over embedded events, as well as less time spent on applying corrections to events - internal and embedded - that are not going to be used).
These details can be explained in a section of the embedding documentation that could be called "Computing Optimizations" or something like this.
I do think that in these cases it is better to go for a simpler approach rather than devising a complex logic for a (likely) little benefit. Much easier to maintain and use also for the future!
This is my opinion based on past experiences, but of course you are the main developer and main user so I would leave the final decision up to you.
Best,
Salvatore

raymondEhlers:
Hi @aiola,

Thanks for your thoughts! I think you make some very good points about simplicity and maintainability.

A few things that I'm still trying to figure out:

- NextEvent() is called before IsEventSelected() in AliAnalysisTaskEmcal, which means that we call the embedded event before we apply event selection. I think rearranging this could be difficult.

- There is currently no event selection in terms of centrality, etc in  the correction task. Perhaps it would be best to apply it with something like a disabled by default AliEventCuts. Then we should add AliEventCuts to AliAnalysisTaskEmcal too to make matching easy. Modifying the base class is never fun, but I don't see any way around it for any chance of matching event selections.

James and I will continue to brainstorm solutions, but please let us know if you have any thoughts on the above.

(Apologies for typos - I'm on my phone)

Best,
Raymond

raymondEhlers:
Hi @aiola,

@jdmulligan and I discussed this for quite a while today, but I think we've found something that we're starting to be happy with. Here is what we propose:

- Circle back around to similar to my original implementation and apply event selection in the embedding helper via AliEventCuts (disabled by default)
    - Since we apply event cuts here, no need to play games with increment and decrementing the current event. If the internal event doesn't pass the selection, then it we just don't embed anything, and we set a property that the embedded event isn't used (specifically, `EmbeddedEventUsed()` will be false).
- Add a check in `IsEventSelected()` in `AliAnalysisTaskEmcal` which (if enabled), will reject events for which `EmbeddedEventUsed()` is false. This check will also need to be added to `AliEmcalCorrectionTask`.
    - The user can also handle this on their own if they are so inclined.

One of the main reasons that we propose this approach is in that we are rather concerned about matching event selection in every task. Since AliEventCuts is not implemented in `AliAnalysisTaskEmcal`, matching that to other user tasks could be difficult. In our case, everything follows the lead of the selection in the embedding helper.

I still see the appeal of the simplicity of your approach, but I see no way around the issue of NextEvent() being called before IsEventSelected() in `AliAnalysisTaskEmcal`, which means that we call `ExternalEvent()` before we apply event selection. But I'm happy to hear of alternatives.

I think I'll probably start implementing this, but comments are still definitely appreciated!

Best,
Raymond

aiola:
Hello @raymondEhlers and @jdmulligan,
this last solution that you propose does look much better!

If we wanted to use my approach a possible solution could be to call NextEvent only for the non-embedding containers in RetrieveEventObjects. Then the NextEvent method could be called for the embedding containers after IsEventSelected.
You would still need to implement the usage of AliEventCuts in the correction task, but you would not need it in the embedding helper. What do you think?

Anyway, both solutions have pros and cons so I am not arguing strongly one way or the other.

Best,
Salvatore

raymondEhlers:
Hi @aiola,

I think I would lean towards my most recent proposal for a few reasons:

- I really like that I won't need to mess around with increment and decrementing fCurrentEntry. It is of course possible to get this right, but I worry that I'll miss some edge cases. By applying the internal event selection before we even embed, we don't have to do any sort of additional book keeping
- Splitting out NextEvent() could resolve the IsEventSelected() ordering problem, but conceptually, it's not my favorite. That being said, either solution will require at least some embedding specific code, so while I slightly prefer my proposal here, I think this is probably a wash.
- The downside to having AliEventCuts in the correction task is that we still need to match the exact cuts in AliEventCuts vs those options in AliAnalysisTaskEmcal. In the case where we look for a flag set in the embedding helper, we are guaranteed to have the same event selection (as long as the event selection in the later tasks is not too restrictive compared to the embedding helper). (It would probably be good to implement AliEventCuts in AliAnalysisTaskEmcal at some point, but I think that could end up being a rather large task, depending on the implementation, but that is a different discussion).

So I think I'm going to implement my most recent proposal - I'll update this PR when I'm done.

Thanks for all of your thoughts and feedback!

Best,
Raymond

raymondEhlers:
+test

raymondEhlers:
Hi @aiola, Hi @jdmulligan,

I just pushed the discussed implementation. It definitely needs to be tested further, but a brief code review would be appreciated if you have time!

Thanks,
Raymond

jdmulligan:
Hi @raymondEhlers,

Looks good to me. Thanks for implementing!

aiola:
Hi @raymondEhlers 
looks good to me as well!

PWG/EMCAL/READMEemcEmbedding.txt
```diff
@@ -382,10 +382,33 @@ embeddingHelper->SetAutoConfigureIdentifier("myEmbedding");
...
+issue, the embedding helper allows for more complicated internal event selection via AliEventCuts (disabled by default).
+When enabled, The cuts can be configured through the YAML configuration for standard options such as centrality and
+z vertex, while more complicated manual configurations can be achieved by retrieving and configuring the members
+of the event cuts object. When an internal event is selected, `EmbeddedEventUsed()` will be true, allowing other tasks
+to only process events when this is the case. In doing so, the event selection that is applied in the embedding helper
+is effectively applied to all other tasks. Thus, be certain that any other event selection that you apply to other tasks
+is less restrictive than that in the embedding helper to ensure that no good embedded events are lost.
+
+To configure this mode, internal event selection must be enabled in the embedding helper via
+`SetUseManualInternalEventCuts(true)`, and then selection on the outcome from the embedding helper must be enabled
```
jdmulligan: I think you mean SetUseInternalEventSelection(true) here, yes?

raymondEhlers: Yeah, you're right. Tab completion failure on my part. I'll update it

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
```diff
@@ -908,6 +920,65 @@ void AliAnalysisTaskEmcalEmbeddingHelper::UserCreateOutputObjects()
...
+
+    if (fCreateHisto) {
+      // This list will be owned by fOutput, so it won't be leaked.
+      TList * eventCutsOutput = new TList();
+      eventCutsOutput->SetOwner(kTRUE);
+      eventCutsOutput->SetName("EventCuts");
+
+      // Add the event cuts to the output
+      fInternalEventCuts.AddQAplotsToList(eventCutsOutput);
+      fOutput->Add(eventCutsOutput);
```
jdmulligan: fOutput is not initialized until a few lines later -- need to move it before this section

raymondEhlers: Fixing now - I will push in a moment

raymondEhlers:
@jdmulligan Should be fixed now. I may have rebase a few days ago without thinking about it, so you may need to reset hard to pull my fixes (sorry!). I'll squash the commits when we're done

raymondEhlers:
+test

raymondEhlers:
+test . Thanks to @jdmulligan , we have successfully tested this and are ready to commit. I will approve around 15:00 if there are no other comments

raymondEhlers:
+1

PR #4012, opened 2018-01-30
EMCal CorrFW: Document LEGO train edge case

raymondEhlers:
LEGO train subwagons execute in alphabetical order, which must be
considered if specialized correction task wagons depend on each other.

@jdmulligan - What do you think? Does this make the issue clear?


jdmulligan:
Hi @raymondEhlers, 

Thanks for adding this. I think it is good, however I think that it may lack context -- would average users understand where this sub wagon trick comes from in the first place? I didn't see an explanation of it elsewhere in the documentation, did I overlook it? If not, a couple sentences along the lines of "A practical way to use specialization on the LEGO trains is to implement a correction task wagon with sub wagons for each specialization. By naming the sub wagon suffix the same as the specialization name, the function AliEmcalCorrectionTask::DetermineComponentsToExecute will extract the appropriate components for each correction chain from a single YAML file. [Perhaps reference a train number with an example]. However, ... [your text]."

raymondEhlers:
Hi @jdmulligan,

Thanks for your suggestions! I'm a bit reluctant to specify the function `DetermineComponentsToExecute()` because I think it's a bit too specific for most people. I didn't really say much about subwagons because I didn't want to dictate too much about how people set up trains, but now I think it would be good to at least mention it. So I would propose (modifying yours):

> A practical way to use specialization on the LEGO trains is to implement a correction task wagon with sub wagons for each specialization. By naming the sub wagon suffix the same as the specialization name, a specialized correction task will be created for each suffix. Each specialized correction task will extract the appropriate components for each correction chain from a single YAML file. For an example, see train 2860 on the Jets_EMC_PbPb train. However, ... [your text].

What do you think?

jdmulligan:
Yep, perfect

raymondEhlers:
+1 I updated the docs as discussed

PR #3985, opened 2018-01-27
Added code to perform D-meson ESE analysis using q2 percentiles

fgrosa:
1) Added setting to perform the ESE analysis with q2 percentiles in AliAnalysisTaskSEHFvn
2) Added macro to compute q2 percentiles using TSpline fit

PR #4028, opened 2018-01-30
Fixed bug in ESE w/o daughter tracks removed from q2

fgrosa:
Fixed bug to kEvShape method when daughters not removed from q2 and q2 percentile not used (bug introduced in #3985)


fprino:
+1

PR #4029, opened 2018-01-30
Removed not useful cut and Code beautify

BongHwi:
Remove a cut when used for saving the tree output.
Beautify the code using the formatter. (indent)


mpuccio:
+1

PR #4030, opened 2018-01-30
AliEventCuts pointer accessible from outside

iravasen:
Modification to permit the changing of the default settings of AliEventCuts event selection from the RunTask without modifying the task.


bdoenigu:
+1

PR #4033, opened 2018-01-31
Fix extraction of downscale correction parameter

mfasDa:
- Incorrect token was used to separate trigger classes in
trigger string  - fixed
- Lookup of downscale weight from database moved to
after event selection


mfasDa:
+1

PR #4037, opened 2018-01-31
[WIP] Fix needed to overcome double-delete issue with ROOT v6-12-04

mfasDa:
It turned out that the double-delete issue with ROOT v6-12-04
appears for objects, in particular inheriting from TCollection,
which are bound to a TFile but not delete in the task destructor.
Once deleting them in the destructor the crash disappears.

This affects the destruction at the end of the ROOT session and
works as long as there is only one task writing to the same output
file. As soon as there is a second instance of a task (same task
or different) writing to the same output file the segfault reappears.

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcal.cxx
+++ b/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcal.cxx
@@ -294,6 +294,9 @@ AliAnalysisTaskEmcal::AliAnalysisTaskEmcal(const char *name, Bool_t histo) :
 
 AliAnalysisTaskEmcal::~AliAnalysisTaskEmcal()
 {
+  if(fOutput) delete fOutput;
+  if(fAliAnalysisUtils) delete fAliAnalysisUtils;
+  if(fPythiaInfo) delete fPythiaInfo;
 }
 
 void AliAnalysisTaskEmcal::SetClusPtCut(Double_t cut, Int_t c)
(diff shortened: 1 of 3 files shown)


mfasDa:
**Work in progress - don't merge!**

Fixing the double delete issue observed with ROOT v6-12-04: Objects - even bound to a file - must be deleted manually, otherwise a segfault appears when destructing the session due to TList's new RecursiveRemove trying to delete objects which were already deleted.

mfasDa:
+test

adriansev:
i did not look in detail over the use cases but i was wondering : wouldn't be possible to convert all used pointers to shared_ptr? (both root5 and root6 are compiled with cxx11 by alibuild) .. all test and delete code could be eliminated entirely ...

aiola:
Hello @adriansev,
I believe that this is not possible. Since we need to keep root5 compatibility the header files (and therefore the class member declarations) cannot contain C++11 code. CINT must be able to process the class declaration and it cannot understand C++11.
Best,
Salvatore

mfasDa:
Hello @adriansev ,

@aiola is right - we cannot use shared pointers (unfortunately) in header files as for ROOT5-compliance header files must be free from c++11 code unless its protected off by the __CINT__ guard, which is however not applicable to data members.

The RecursiveRemove introduces quite some troubles on our side which have to be fixed one-by-one.

Cheers

Markus

adriansev:
yeah, i forgot completely about cint compatibility problems :(

mfasDa:
+1

For all people using the EMCAL framework in case of problems please report.

Cheers

Markus

PR #4040, opened 2018-01-31
removed duplicate/redundant class from PWGCF/FLOW/macros

rihanphys:


rbertens:
+1

rbertens:
@rihanphys hey Rihan, I clicked '+1' a bit too soon I think. You remove a class, but you do not remove the corresponding compilation instructions in CmakeLists.txt and PWGflowTasksLinkDef.h , or was this code that was never compiled?

rihanphys:
@rbertens   I mistakenly copied the class to macros and there the class did not get compiled. However, I have the correct copies of the class in Task folder where they are compiled. Therefore, I do not  have to edit the CMakeList and PWGflowTasksLinkDef

PR #4042, opened 2018-01-31
Fix for FillStatics issue of the Physics Seleciton with ROOT v6-12-04

mfasDa:
The physics selection crashes under ROOT6 v6-12-04 in FillStatistics
when filling fHistStat. The problem seems to be due to a temporary
histogram used when extending the histogram axis, which seems to get
connected to a directory. Releasing the directory status around the
function seems to cure the problem.

### DIFF
--- a/OADB/AliPhysicsSelection.cxx
+++ b/OADB/AliPhysicsSelection.cxx
@@ -500,7 +500,6 @@ void AliPhysicsSelection::FillStatistics(){
   fHistStat->SetCanExtend(TH1::kAllAxes);
 #endif
   fHistList.Add(fHistStat);
-  TH1::AddDirectory(oldStatus);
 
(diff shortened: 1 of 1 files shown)


mfasDa:
This is connected to https://github.com/alisw/alidist/issues/994: It seems that for extending an/multiple ax(i/e)s temporary histogram(s) are created which are get bound to directories if gDirectory is set. With the new TList::RecursiveRemove they get deleted twice. Moving the closing directory guard to the end of the function, so around the full filling process, seems to cure the problem on ALICE side, though I think the more appropriate solution would be to fix this in ROOT as this will generally appear in case a directory is open when the axis range gets extended.

mfasDa:
problem disappeared with the latest release of ROOT (v6-14-00), introduced in alisw/alidist#1212. Closing the PR.

PR #4047, opened 2018-01-31
Additional check if particle has mother

rhaake:
This code fixes a bug that potentially leads to seg. fault if the the particle has no mother.


ebruna:
+1

PR #4064, opened 2018-02-01
connect jet container in AddTask

jaimenorman:
Connect jet container in AddTask, to allow to run on LEGO trains


raymondEhlers:
+1

PR #4066, opened 2018-02-01
EMCal Embed: Fix combining cell collections

raymondEhlers:
There were some indexing bugs which prevented the cell collections from
being combined successfully. Those were fixed, with new functionality
added to verify they were successfully combined.

Note that the length of the combined array should always been the same
length as the combination of the lengths of the cell objects, but that
doesn't mean that the arrays are the same. This can only be verified by
checking each cell by hand.

cc: @jdmulligan


raymondEhlers:
+1

raymondEhlers:
+1

raymondEhlers:
+1 (Didn't successfully push the fix the first time)

PR #4068, opened 2018-02-01
Background calculation check always against charged jets in jT analysis

TWSman:
Fixes a problem where background was incorrectly calculated for full jets when background region is outside EMCAL acceptance


aohlson:
+test

aohlson:
+test

aohlson:
+test

aohlson:
+1

PR #4069, opened 2018-02-01
EMCal CorrFW: Update hist titles for clarity

raymondEhlers:
@jdmulligan - Any objections? (I think I have all of the right units, but confirmation would be good!)


jdmulligan:
@raymondEhlers I believe the timing histograms are in (s), not (ns). The uncalibrated peak is plotted at ~0.6*10^-6, and should correspond to 600ns.

raymondEhlers:
You're absolutely right - I was apparently on autopilot. I'll update it. Thanks!

raymondEhlers:
+1

PR #4070, opened 2018-02-01
enforce the correct reset on the NanoAODs for every event loop

mazimm:
add dummy arrays to the NanoAODArrayMaker so that the analysis arrays will be reseted for every event

### DIFF
--- a/PWG/DevNanoAOD/AliAnalysisNanoAODCuts.cxx
+++ b/PWG/DevNanoAOD/AliAnalysisNanoAODCuts.cxx
@@ -120,13 +120,17 @@ void AliNanoAODSimpleSetter::SetNanoAODHeader(const AliAODEvent * event   , AliN
 
   TString firedTriggerClasses = header->GetFiredTriggerClasses();
 
-  static const char * validatorString[] = {"Centr","CentrTRK","CentrCL0","CentrCL1", "MagField", "OfflineTrigger", "RunNumber", "FiredTriggerClasses", 0};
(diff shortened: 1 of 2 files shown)


raymondEhlers:
+test

PWG/DevNanoAOD/AliAnalysisNanoAODCuts.cxx
```diff
@@ -120,13 +120,17 @@ void AliNanoAODSimpleSetter::SetNanoAODHeader(const AliAODEvent * event   , AliN
 
   TString firedTriggerClasses = header->GetFiredTriggerClasses();
 
-  static const char * validatorString[] = {"Centr","CentrTRK","CentrCL0","CentrCL1", "MagField", "OfflineTrigger", "RunNumber", "FiredTriggerClasses", 0};
+  UShort_t bunchCrossNumber = header->GetBunchCrossNumber();
+  UInt_t orbitNumber = header->GetOrbitNumber();
+  UInt_t periodNumber = header->GetPeriodNumber();
+
+  static const char * validatorString[] = {"Centr","CentrTRK","CentrCL0","CentrCL1", "MagField", "OfflineTrigger", "RunNumber", "FiredTriggerClasses", "BunchCrossNumber", "OrbitNumber", "PeriodNumber", 0};
```
mfasDa: Why static, and why const char *? Better use array of const TStrings (in particular const std::array< const TString, ...>), non-static.

mfasDa:
+test

raymondEhlers:
It seems they didn't fix the test + test = +1 bug...

PR #4071, opened 2018-02-01
Add option to use trigger selection for data

mfasDa:
With the trigger seleciton in addion to the
online trigger result a patch above threshold
is require. This is used to cleanup noisy
triggers and to overcome the configuration issue in LHC16l.


mfasDa:
+1

PR #4078, opened 2018-02-02
Fix output for histogram and include event containers

a-mathis:
- Fixes a bug that crashes the run merging
- Include event containers


dmuhlhei:
+1

PR #4082, opened 2018-02-02
Pile-up rejection based on correlation between kTPCout tracks and VZERO multiplicity implemented

mfaggin:
I re-open a new pull request with a very small change in the .cxx file
Maybe is not the right procedure...


sashingo:
+1

PR #4088, opened 2018-02-02
EMCal Embed: Add configure function for LEGO train

raymondEhlers:
This functionality is similar to that developed for the EMCal Correction Task.

Also clarified and fixed some documentation

(cc @jdmulligan)


raymondEhlers:
+1

PR #4089, opened 2018-02-02
EMCal Embed: Add macro for config LEGO train func

raymondEhlers:
Forgot this macro needed for use on the LEGO train in the initial PR


raymondEhlers:
+1

PR #4090, opened 2018-02-03
PWGCF: AliFemtoCorrFctn - Implement missing Clone methods

akubera:
The `Clone()` method, used to copy an object without knowing its type, was left to the default base-class implementation of just returning NULL for most correlation function objects, despite their copy constructors being implemented. This changeset makes `Clone()` pure-virtual and `const` in the base class `AliFemtoCorrFctn`, and provides trivial implementations for classes missing them.

Also:
- Fixed out-of-order initialization warnings
- Merged `if` statements checking for pair-cut existence and if pair passes cut
- Fixed warning in AliFemtoCorrFctnDEtaDPhiSimpleWithCorrections copy constructor
- Auto-removed trailing whitespace


dberzano:
+1

PR #4091, opened 2018-02-03
update standard ESD parameters after the smearing

ffionda:
Update all parameters ("standard" and "constrained") of ESD tracks after the smearing. Impact parameters are also updated.


arossi81:
+1

PR #4093, opened 2018-02-03
Fix bug in cluster container that removes M02=0 clusters

jdmulligan:
Please do not approve yet.

This commit fixes a bug in PR #2680: https://github.com/alisw/AliPhysics/commit/3374c97df3cd20a7d3623c4efaaa877e4c6a3ec7. The constant DBL_MIN gives the minimum positive float, not -infinity. This had the effect of rejecting clusters with M02=0 (which can amount to up to ~10% of clusters, depending on thresholds etc.). 

This affects EMCal analyses that do not apply an M02 cut (and which use cluster containers). In particular, full jet analyses by default do not apply an M02 cut. Analyzers should be aware that between Sep 21 and now, an M02 cut at M02=0 was applied by the cluster container, and that results should be re-run with the fix.

Note: A separate discussion should determine whether a small-M02 cut _should_ be applied in jet analyses, but at present the intended default behavior is to not apply any cut, and this commit restores that behavior.

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliClusterContainer.cxx
+++ b/PWG/EMCAL/EMCALbase/AliClusterContainer.cxx
@@ -46,7 +46,7 @@ AliClusterContainer::AliClusterContainer():
   fIncludePHOSonly(kFALSE),
   fPhosMinNcells(0),
   fPhosMinM02(0),
-  fEmcalMinM02(DBL_MIN),
+  fEmcalMinM02(-1.),
   fEmcalMaxM02(DBL_MAX),
   fEmcalMaxM02CutEnergy(DBL_MAX)
 {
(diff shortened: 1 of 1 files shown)


jdmulligan:
+test

gconesab:
Hi James, if I understand the comment, this should not be a problem since exotic clusters have m02 < 0.1 so this is a recommended cut. For "real" clusters this would affect to clusters of 1 cell which anyway are recommended to be cut.
Other thing would be that you cut on m20=0, this would be bad.

jdmulligan:
Hi Gustavo,
Indeed, if one applies the cut M02<0.1, this is not a problem. However that cut is not applied in the "default" framework (maybe it should be, but for jet analyses this was never done as far as I know). The 1-cell clusters however should be removed already by the Fcross exotic cut, which is applied in the default framework, and occurs before the presently discussed cut. I'm not exactly sure what these extra M02=0 clusters are, but I do see them. Perhaps >1 cell but not included in the M02 calculation if ln(Ecell/Eclus)<w0?
 
There is no cut on M20, only M02.

jdmulligan:
Hi Gustavo,

I checked and indeed these are 1-cell clusters, but with leading Ecell < 4 GeV, so the Fcross exotic cut isn't applied. So the conclusion stands that if an M02 cut is applied in analysis, this bug has no effect, but if no M02 cut is applied, it can have a significant effect.

jdmulligan:
+1

PR #4096, opened 2018-02-05
Several Changes

efranzis:
1. Changed to new file format for OADB file creation. 
2. Updated run-by-run Analysis such that cells with <20% of bad runs will be unmasked 
automatically since they will be recovered in ~98% of the cases
3. Change of some gif output names to make it easier to upload them to the Twiki with unambiguous file names.


efranzis:
+1

PR #4097, opened 2018-02-05
OADB/EMCAL: Addition of BC maps for 16o and LHC16p

nschmidtALICE:
For LHC16o, four different run ranges were covered with this update.
The period LHC16p was split in two parts for the BC maps.


gconesab:
+1

PR #4098, opened 2018-02-05
add example macro to correct MC tracks 

ffionda:
Example macro to correct MC tracks based on the data / MC comparison for the impact parameter.


iarsene:
+1

PR #4099, opened 2018-02-05
update trigger analysis (logical OR only on active TRU acceptance)

dsekihat:
only consider cluster pairs only on active TRU acceptance in both data and M.C.


kharlov:
+1

PR #4123, opened 2018-02-06
Fix doxygen typos and issues

raymondEhlers:
The first `//` is eaten by doxygen in the EMCal corrections readme for
no apparent reason (as the exact same type of code is shown directly
below without the `//` being lost), so they are doubled up to ensure
that the comment is displayed properly.

Also fixes a few typos.

@jdmulligan - This fixes the issue that I commented on (although it's a rather ridiculous fix). Let me know if you have another idea - otherwise I will commit this


jdmulligan:
Yea things like that are basically the only way I know how to do battle with doxygen!

raymondEhlers:
+1

PR #4125, opened 2018-02-07
First version of aod skim task for emcal

loizides:
First working version of aod skim task able to run data and MC

### DIFF
--- a/PWG/EMCAL/EMCALtasks/AliAodSkimTask.cxx
+++ b/PWG/EMCAL/EMCALtasks/AliAodSkimTask.cxx
@@ -1,0 +1,175 @@
+#include <Riostream.h>
+#include <TChain.h>
+#include <TGraph.h>
+#include <TList.h>
+#include <TMath.h>
+#include <TProfile.h>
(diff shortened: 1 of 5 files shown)


loizides:
+1

loizides:
+test

gconesab:
What is the purpose of this task that is not present  or that could be added to the existing tasks doing AOD/ESD filtering?

mfasDa:
Some more comments:
- Please when opening pull requests squash related commits into one in order to not unnecessarily blow up the history
- Please avoid if(1) within code as this might on some newer platforms raise a warning about a statement with no effect and since EMCAL libraries are build with Werror consequently end in compiler errors. For the purpose obvious from the code just putting code into curled brackets is sufficient for limiting scoping.
Minor:
- LICENSE disclaimer missing (BSD-license)
- ClassDef/ClassImp should be surrounded by /// \cond CLASSIMP and /// \endcond since EMCAL libraries are enabled in doxygen and missing guards result in error messages (which however not stop the doxygen build)

PR #4126, opened 2018-02-07
Separate PHOSTender configuration from PHOS_LHC16_pp analysis setup

kqf:
PHOS tender should be added to the separate wagon in lego trains. A new macro is needed to setup custom map of bad channels (will be useful for LHC17 QA analysis) and ZS threshold.


kharlov:
+1

PR #4129, opened 2018-02-07
Hyperon Femtoscopy Analysis in pPb - AliFemtoDream 

bhohlweger:
Femtoscopy analysis design to select charged tracks (Protons, Pions, Kaons), two particle v0 Decays (e.g. Lambdas) and Cascades (e.g. Xis), which get buffered and are then used to calculate the Same Event and Mixed Event statistics for the calculation of the correlation function. This analysis will be used to analyse Run2 p-p and pPb Data and is therefore build in a modular way. 

The particles are selected by the corresponding cut object, which each contains a class for data output and Monte Carlo output. Additionally classes to configure the behavior of the event mixing, to handle the output of the particle pairing, and at last to store the particles. 

Additionally included in this commit is the steering macro for the analysis task of Oliver Arnold (PLamAnalysisPP) since it was requested from the Analysis Review Committee.

### DIFF
--- a/PWGCF/FEMTOSCOPY/CMakeLists.txt
+++ b/PWGCF/FEMTOSCOPY/CMakeLists.txt
@@ -14,5 +14,6 @@ add_subdirectory(K0Analysis)
 add_subdirectory(UNICOR)
 add_subdirectory(V0LamAnalysis)
 add_subdirectory(PLamAnalysisPP)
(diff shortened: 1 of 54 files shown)


dberzano:
+test

lgraczykCern:
+1

dberzano:
+1

PR #4132, opened 2018-02-07
Adding DeltaRg to the jet substructure tree

mfasDa:
DeltaRg := difference in R between the
subjet vectors of the last splitting


mfasDa:
+test

mfasDa:
+1

PR #4138, opened 2018-02-07
Add V0C,V0A options for centrality in FlowEventCuts object

rihanphys:
Add V0C, V0A options for centrality estimators in FlowEventCuts object

### DIFF
--- a/PWG/FLOW/Tasks/AliAnalysisTaskCMEV0.cxx
+++ b/PWG/FLOW/Tasks/AliAnalysisTaskCMEV0.cxx
@@ -1946,8 +1946,9 @@ void AliAnalysisTaskCMEV0::UserExec(Option_t *)
 
    //------------ eta depedence -----------------
    Double_t fEtaCent = -100.;
+   Int_t nBinsX = fEtaBinFinderForQA->GetNbinsX();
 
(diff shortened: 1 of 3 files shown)


rihanphys:
@rbertens Just to let you know, I added enum options in the AliFlowEventCuts class to select V0C or V0A as centrality estimators (to be used to pPb system).

jmargutt:
+1

rbertens:
Hi @rihanphys @jmargutt ,

Please realize that this commit changes the value of kSPD1clusters and kZDC
Old case: kVZERO = 2, kV0=kVZERO = 2, kSPD1clusters = 3, kZDC = 4
New case: kVZERO =2, kV0=kVZERO= 2, kV0M=kVZERO =2, kV0C =3, kV0A =4, kSPD1clusters =4, kZDC =5

I grepped through the package and cannot find a situation where this would lead to a problem (i.e. conversion from the enum to an int), but in principle this is not really a safe commit,it's better to add the new enumerators at the end of the list to ensure backwards compatibility. 

My best,
redmer

rihanphys:
@rbertens  thanks for pointing this out. I would place the new enums at the end of the list and commit asap.

PR #4147, opened 2018-02-07
bug fix first mc event. proper handling of pysex file. Fix trig sel

loizides:
Should be the first version with which we filter


loizides:
+1

PR #4148, opened 2018-02-08
Change the max number of mix events from 1 to -1 (no limit).

moliver813:
Necessary for mixed events to work due to semi-recent change in AliEventPoolManager


efranzis:
+1

PR #4156, opened 2018-02-08
New entry for LHC17n pass1 with 56 ps resolution (P. Antonioli)

njacazio:
- Updated TOFPIDParams.root file with LHC17n parameters


fprino:
+1

PR #4159, opened 2018-02-08
removed ofstream objects so they will not be written

bschaefe:


akubera:
+test

akubera:
This actually failed to build with root6 defaults on my machine. I'll look into this.

akubera:
Ok, it will compile if you change `ofstream` back to `std::ofstream` OR in your cxx file move your `using namespace std;` somewhere before `#include "AliAnalysisTaskCorPIDTOFQA.h"`. Alternatively add `using std::ofstream;` before the include.

akubera:
+1

looks good

PR #4171, opened 2018-02-09
EMCal CorrFW: Fix typo

raymondEhlers:
Typo introduced by 65da1fc95aaf9672ec1e27282875dd881b5a2144. The timing
label was fixed in the cell time calibration component, but missed in
the cluster non-linearity


raymondEhlers:
+1

PR #4180, opened 2018-02-12
Allow yaml-cpp to convert TStrings

raymondEhlers:
Otherwise, TString cannot be encoded or decoded for use with
AliYAMLConfiguration


aiola:
+1

PR #4183, opened 2018-02-12
Add optional additional trigger selection via trigger decision container

mfasDa:
Optional trigger selection implemented for
EMCAL triggers, requiring a recalc trigger
patch above nominal threshold


mfasDa:
+test

PR #4184, opened 2018-02-12
Add file with producion name <> pass name correspondence

wiechula:
This is required to automatically extract the anchor pass name for a give MC production


chiarazampolli:
+1

PR #4199, opened 2018-02-12
EMCal CorrFW: Update docs to avoid link to YAML NS

raymondEhlers:
The YAML namespace is automatically linked at every mention of YAML,
which is desired, so it is disabled with "YAML" -> "%YAML" (the %
prevents linking)


raymondEhlers:
+1

PR #4200, opened 2018-02-12
EMCal CorrFW: Move create histo init to base class

raymondEhlers:
This field is present for every component, so it should be handled in
the base class initialization instead of in each component

cc: @jdmulligan


raymondEhlers:
+1

PR #4201, opened 2018-02-12
Adding kNone option for ESD track cuts configuration

afestant:
kNone option for SPD requirements in the ESD track cuts configuration added


fprino:
+1

PR #4210, opened 2018-02-13
Fix hybrid track selection on ESDs

mfasDa:
Hybrid tracks were selected but the payload
defining the hybrid track type was not attached
to the track selection result ptr. In this case
the track container is not able to distinguish
the hybrid track type and will reject all tracks.
The hybrid type is now attached in case the track
was selected as hybrid track.

The track cuts generator was up to now not using the
new hybrid cuts wrapper but instead defined hybrid
cuts as set of ESD cuts. Also in this case the payload
defining the hybrid track type was missing and the
distinction in the track container was imposible. The
different hybrid track types are however already
implemented, and selecting one of the types gives the
expected result. The track cuts generator was modified
to ues the hybrid cuts wrapper with the proper hybrid
definition for the period provided so users don't
need to select the hybrid track type by hand.

In addition:
- Indentation fix in AliEmcalAODHybridTrackCuts
- Some AliDebugStreams for debugging purpose
- Move track cuts generator to the namespace
  PWG::EMCAL (will be done step-by-step also with
  other classes in the EMCAL core framework)


mfasDa:
+test

mfasDa:
+1

PR #4178, opened 2018-02-11
draft of crosstalk emulation correction component

jdmulligan:
Please do not approve yet.

@gconesab, @raymondEhlers: Here is an initial draft of the crosstalk emulation correction component. I didn't test it at all besides that it compiles and doesn't crash while running. That being said, the component is integrated into the correction framework and will run if you enable it in your user yaml file (see available configuration options and defaults in AliEmcalCorrectionConfiguration.yaml). The main function is Run().

 A few notes:

- I removed all of the code about digits, and attempted to directly write new cells. I added two new functions AddInducedEnergiesToExistingCells() and AddInducedEnergiesToNewCells() to do this. Gustavo, I am not super familiar with how your code works, so please take a look here to see if it has the right idea. You should also check carefully the details, since I included only a fairly minimal amount of code from AliAnalysisTaskEMCALClusterize. But the draft should at least give a good starting point... 

- I placed the component to execute after the cell calibrations, but before clusterization.

PR #4215, opened 2018-02-13
EMCal CorrFW: Add config for crosstalk task

raymondEhlers:
As promised in #4178, this PR allows array initialization for the new crosstalk correction. This configuration was a bit more complicated than usual, but a little templates magic makes it work and should be fairly intuitive. I tried to document it and configure it as you have in your .C, but both should be checked more closely. Please let me know if it is unclear.

cc: @gconesab , @jdmulligan


raymondEhlers:
+1

PR #4228, opened 2018-02-14
Final corrections to Cross Talk emulation component

gconesab:
- Fix major bug, reset of temporal arrays with induced energy per channel at the end of the event
- Minor fixes like double semicolon and printf instead of AliInfo when print parameters is requested
- Add author mail
-Set default value for energy conservation as true


gconesab:
+1

PR #4229, opened 2018-02-14
Fixing the problem with the Resolution Function

hahassan7:
I made a mistake in reconstructing the resolution function used for the jet probability, and now it is fixed.


fcolamar:
+1

PR #4231, opened 2018-02-15
Update yaml-cpp to 0.6.1

raymondEhlers:
There was a quick update to fix an installation bug (which I don't
believe affected AliPhysics, but it is a minor update that is worth
having in case it makes a difference). Imported from commit
5d5bb52ec2825eb4c0c932a9c2d50dc3699f932f


mfasDa:
+test

mfasDa:
+1

PR #4251, opened 2018-02-15
fixed typo in addtask

haandrew:


raymondEhlers:
+1 (I'd recommend making your AddTask a static method in your class so you can benefit from it being compiled)

haandrew:
Hi Raymond,

Would you be able to inform me as to how I can add the AddTask as a static method? Do I include the function in the Public declaration of my class?

raymondEhlers:
Hi @haandrew ,

Yeah, you want to declare it as a public static method in your class. For an example, see AliEmcalJetTask (specifically, the declaration is [here](https://github.com/alisw/AliPhysics/blob/master/PWGJE/EMCALJetTasks/AliEmcalJetTask.h#L131)). Note that you'll still need a .C AddTask macro for the LEGO train (unfortunately, it can't call static methods directly), but it can be extremely minimal (and you still get the benefits of the compiled AddTask). For an example, see the EMCal Correction Task [AddTask](https://github.com/alisw/AliPhysics/blob/master/PWG/EMCAL/macros/AddTaskEmcalCorrectionTask.C).

Let me know if you have any questions.

Best,
Raymond

PR #4255, opened 2018-02-16
EMCal Embed: Move file access check to Initialize

raymondEhlers:
When file access was checked in SetupInputFiles(), it would cause
unnecessary load to the grid due to AccessPathName calling a command in
xalienfs (`gapi_stat()`), which then performed an `ls -l` on every
single directory. If we instead move it to Initialize(), such checks
will only have to be performed once.

cc: @jdmulligan

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
+++ b/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
@@ -56,6 +56,54 @@
 ClassImp(AliAnalysisTaskEmcalEmbeddingHelper);
 /// \endcond
 
+/**
+ * Helper function to connect to AliEn (so the code doesn't need to be duplicated).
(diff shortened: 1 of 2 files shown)


raymondEhlers:
+test

raymondEhlers:
@jdmulligan , I think I'm going to approve this so I can go back to sleep - we can always make another PR if you spot a problem

jdmulligan:
@raymondEhlers , I just started testing so I can deal with it (haven't seen your approval yet). Thanks for staying up late to implement this!

raymondEhlers:
Sorry, I'm a bit confused - you'll approve it? Or you're fine with me approving?

jdmulligan:
I will approve it before the tag if not otherwise done!

raymondEhlers:
Okay, thanks. I'll leave it to you so you have as much time to test as possible

raymondEhlers:
+1 (I think it has to be the first statement)

PR #4256, opened 2018-02-16
Revert moving to namespace EMCAL because of a problem with constants …

mfasDa:
…used in macros appearing in some cases under ROOT5

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliEmcalESDTrackCutsGenerator.h
+++ b/PWG/EMCAL/EMCALbase/AliEmcalESDTrackCutsGenerator.h
@@ -39,12 +39,12 @@
 
 class AliESDtrackCuts;
 class TString;
+class AliEmcalTrackSelection;
 
 namespace PWG {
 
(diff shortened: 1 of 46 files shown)


mfasDa:
+test

mfasDa:
@sashingo @gconesab  Could you both please do a "+test"?

gconesab:
+test

mfasDa:
Was reported by some users accessing track type constants in macros that constant name including the namespace was misinterpreted, leading to an error that the constant was not found. The error happens purely under ROOT5, and happens only under certain conditions as the case reported was reproducable but other macros do run normally without errors and use the proper constant value. 

As long as the reason is not understood the change will be reverted.

mfasDa:
@sashingo @arossi81 @ginesmartinez @fcolamar @zconesa @rbailhac @ebruna @fprino 

Coud someone of the HF conveners please approve the HF part of the commit!

zconesa:
+1

PR #4257, opened 2018-02-16
Make a TPC QA macro compatible with ROOT5/6

dberzano:
ROOT 5 does not like `const char *` even if it is the right thing to do


dberzano:
+1 This is being discussed in [ALIROOT-7685](https://alice.its.cern.ch/jira/browse/ALIROOT-7685) and it works around a CINT bug.

PR #4273, opened 2018-02-16
EMCal Embed: Fix yaml config reinitialization

raymondEhlers:
Without reinitialization, the yaml config does not stream properly


raymondEhlers:
+1 (@jdmulligan )

PR #4274, opened 2018-02-17
EMCal Embed: Add yaml config and docs

raymondEhlers:
Allows the embedding helper to be almost entirely configured from YAML.
Also adds documentation on usage of YAML with the embedding helper

@jdmulligan


raymondEhlers:
+1

PR #4297, opened 2018-02-19
Use always pt-hard bin from filename

mfasDa:
Using always pt-hard bin from filename in histograms relevant for
merging. Reason: Insufficient precision in variable used for the
pt-hard value (float) leading to migration into the lower bin for
pt-hard values close to the bin limits, for which weights in the
reweighting get screwed up.


mfasDa:
+test

mfasDa:
+test

Removed (now ununsed) pt-hard bin detection from pt-hard value in header.

mfasDa:
+1

PR #4212, opened 2018-02-13
Move track selection to namespace PWG::EMCAL

mfasDa:
Move track selection and dowscale factors to namespace
PWG::EMCAL. No functional change, just move to namespace
and adaptions in classes utilizing the track selection

PR #4299, opened 2018-02-19
Added addtask as static member method of class RecursiveSoftDrop

haandrew:


raymondEhlers:
Hi @haandrew ,

This looks good, but one suggestion to get the full benefit: in your .C macro, just call the static function that you just added to your class. That way, you have a minimal amount of code in CINT.

Best,
Raymond

haandrew:
Hi @raymondEhlers 

I have updated the .C file now to just call the function from my task.

Thanks for the help,

Harry

raymondEhlers:
+1 Looks good!

haandrew:
Hi @raymondEhlers 

Small problem. When I try and call the add task now I get the following error:

Error: Function kEmb) is not defined in $AddTaskRecursiveSoftDrop("","JetMCOnlyPartLevel_AKTChargedR040_MCSelectedParticles_pT0000_E_scheme","JetMCOnly_AKTChargedR040_PicoTracks_pT0150_E_scheme","Jet_AKTChargedR040_PicoTracks_pT0150_E_scheme","Jet_AKTChargedR040_PicoTracks_pT0150_E_schemeConstSub",0.4,"Rho","","MCSelectedParticles","MCSelectedParticles","PicoTracks","TracksConstSub","TPC","V0M",kPhysSel,"","","",AliAnalysisTaskRecursiveSoftDrop::kConstSub,AliAnalysisTaskRecursiveSoftDrop  /home/alitrain/train-steer/generate.C:358:

Which I have seen before when there is a mistake calling the the addtask, i.e a typo in the class name. However if I simply replace the line: return AliAnalysisTaskRecursiveSoftDrop::AddTaskRecursiveSoftDrop(njetsData,njetsTrue,njetsDet,njetsHybridUs,njetsHybridS,R,nrhoBase,ntracksData,ntracksTrue,ntracksDet,ntracksHybridUs,ntracksHybridS,type,CentEst,pSel,trigClass,kEmcalTriggers,tag,jetShapeSub,fjetType);


with my actual addtask code, as it was before, it runs okay. Have I made a mistake that you can see in my cleaned up AddTask?

Thanks in advance,

Harry

raymondEhlers:
Hi @haandrew ,

This is almost always due to an argument order or type issue that is often hidden by CINT. You should check these very closely, and I'll try to take a look a bit later today.

Best,
Raymond

mfasDa:
Hi @haandrew Hi @raymondEhlers,

I think the issue is similar to what was reported in #4212 and later on the EMCAL mailing list, for which I had to revert the commit as it points to a problem in CINT itself with accessing constants from nested namespaces including classes. In my case the issue points to CINT itself as I can provide a minimal example of the macro accessing the constant which does not crash, but the add-macro mentioned, which compiles fine (so obvious coding mistake) and uses the same constant in the macro, does.

Maybe for the moment try to replace the symbolic constant (kEmb) with a number (note: normally not recommended) to work around the problem.

Cheers

Markus

raymondEhlers:
Hi @mfasDa ,

Thanks for the suggestion! Looking back more closely at my AddTask, I use `UInt_t` in the case of enums. I didn't remember the reason (and I was surprised because I would usually avoid such an approach), but this must be it! So @haandrew , if you change to `UInt_t`, I think everything should work. Since I can only use ROOT 6, it's a bit difficult for me to check.

Sorry for the disruption!

Best,
Raymond

PR #4307, opened 2018-02-19
fix tracking efficiency and res to plot correctly for PbPb MC

jdmulligan:
(minor fix in QA plotting macro)

@efranzis


raymondEhlers:
+1

PR #4315, opened 2018-02-20
Update AddTaskDStarSpectra.C 

sjaelani:
Update AddTaskDStarSpectra.C for d0Imp study -> d0Imp = kTRUE


zconesa:
Hi,
I am wondering if the flag for the impact parameter could be implemented as an option of the addtask that you can set as external parameter, and have it off by default. Having it on by default in all analyses could slow down a lot the studies on pPb and PbPb data uselessly. I seem to recall we had problems with it in the past in PbPb, but I might be wrong. What do you think about?
Cheers,
Zaida

sjaelani:
Hi Zaida,

yes, you're right. We have the memory problems in the past in PbPb. I think to be safe, I will set as an external parameter as you suggested.

Best,
Syaefudin

dberzano:
Hi, this PR looks pretty inactive. Can we close it please?

PR #4323, opened 2018-02-20
Fixed AddTask by adding missing static method

haandrew:
Hi @raymondEhlers and @mfasDa 

Thanks for your suggestions about how to fix the addTask. I had a look into it and compared with your add task Raymond and I think the problem was more simple than it seemed. I hadn't defined my AddTask as static in my class and I think this caused knock on effects. I have tested this version locally and it seems to work.

Cheers,

Harry


raymondEhlers:
+1 Ah, yes, that would do it! I'm glad that it was an easy fix

PR #4333, opened 2018-02-21
Including and separating EMCAL triggers in track QA task

mfasDa:


mfasDa:
+1

Tested for various triggers under run2-conditions

PR #4342, opened 2018-02-21
Automatic pthard scaling: Do not scale further QA histograms

rhaake:
In the automatic scaling, do not scale the following histograms:
fHistTrials, fHistXsection, fHistEvents, fHistTrialsAfterSel, fHistXsectionAfterSel, fHistEventsAfterSel.
They might be needed for QA purposes.


mfasDa:
+test

Looks OK to me - indeed the weighting histograms should not be scaled, otherwise there would be no way to recalculate the scale factor offline. Leaving the PR open until tomorrow noon so others can check as well.

mfasDa:
+1

PR #4348, opened 2018-02-22
implement cuts used for NanoAOD filtering

sebaleh:
implement cuts used for NanoAOD filtering, need to be applied in AnalysisTaks when running on MC (MC is not NanoAOD)


aaroncapon:
+test

aaroncapon:
+1

PR #4377, opened 2018-02-23
loosen pT cuts and remove shared ITS cluster cut

sebaleh:


miweberSMI:
Hi @sebaleh,

in the description you write "remove shared ITS cluster cut", but I didn't see that in the diff file. Am I missing something?

Cheers,
Michael

miweberSMI:
After cross checking with @sebaleh, this PR is fine and I am merging it.

miweberSMI:
+1

PR #4381, opened 2018-02-23
Possibility to change the multi-vertexer settings

jhamoncern:
Possibility to configure the multi-vertexer (MV) in the RDHF cut objects. So far the MV settings were always the same, namely the one tuned for p-Pb 2013 (https://twiki.cern.ch/twiki/bin/view/ALICE/AliDPGtoolsPileup), and could not be changed.

**Important points:**
The default value of the Chi2/ndf data member 'fMaxVtxRedChi2', useful for the MV but not used in D2H framework (_as far as I can see!_), has been changed from 1e6 to 5 (the default MV value).
Could you confirm that the change of the fMaxVtxRedChi2 default value will not affect our analyses? Otherwise an additional data member, dedicated to the MV, could be created.

Cheers,
Julien

### DIFF
--- a/PWGHF/vertexingHF/AliRDHFCuts.cxx
+++ b/PWGHF/vertexingHF/AliRDHFCuts.cxx
@@ -65,9 +65,10 @@ AliRDHFCuts::AliRDHFCuts(const Char_t* name, const Char_t* title) :
 AliAnalysisCuts(name,title),
 fMinVtxType(3),
 fMinVtxContr(1),
-fMaxVtxRedChi2(5.),
+fMaxVtxRedChi2(1e6),
 fMaxVtxZ(10.),
 fMinSPDMultiplicity(0),
+fMaxVtxChi2PileupMV(5.),
 fMinWDzPileupMV(15.),
 fRejectPlpFromDiffBCMV(kFALSE),
 fTriggerMask(AliVEvent::kAnyINT),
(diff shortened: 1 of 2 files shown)


fprino:
Hi Julien,

so far we were using the multivertexer pileup with the default cuts of AliAnalysisUtils.
These are:
  fMinPlpContribMV(5),
  fMaxPlpChi2MV(5.),
  fMinWDistMV(15.),
  fCheckPlpFromDifferentBCMV(kFALSE),

The fMaxVtxRedChi2 data member was not intended as a parameter for the pileup tagging.
It was already there before the implementation of the multivertexer pileup.
I guess the original idea was to use it to cut on the chi2 of the secondary vertex of a D meson, but it is not used by any analysis from what I could see in the cut objects.

I think the idea of having the possibility to set the parameters of the multivertexer is good to have, but I would add a new data member of the chi2 of the pileup vertex, in case sooner or later we want to try the .,fMaxVtxRedChi2 for the selection of candidates.

Cheers, Francesco

jhamoncern:
Hi Francesco,

Thanks for the advice. I created a new data member dedicated to the multi-vertexer Chi2.
By default, the MV parameters are set to their default AliAnalysisUtils values. They can be modified by means of setters.

Julien

fprino:
+1

PR #4384, opened 2018-02-24
Allow minimum shared momentum fraction

raymondEhlers:
The requirement of a minimum amount of shared momentum fraction
is in addition to already requiring jets to match geometrically

NOTE: This first iteration will currently only work with
embedding real pp data.


raymondEhlers:
+1

PR #4432, opened 2018-02-28
Improvements to jet-hadron correlations and performance tasks

raymondEhlers:
Includes:
- Move Correlations task to PWGJE::EMCALJetTasks
- Preserve axis order in the Performance task (following the order that they are defined in YAML)
- Performance task bug fixes
- Move to custom retrieval of leading hadron momentum to ensure that the proper cluster momentum is selected
- Add documentation on cluster energy (it only retrieves the raw E()) to leading hadron momentum function in the Jet FW


raymondEhlers:
+1

PR #4451, opened 2018-02-28
Fix includes typo

raymondEhlers:
It still included the desired header eventually, but also added an
undesired dependence on fastjet


raymondEhlers:
+1

PR #4453, opened 2018-02-28
add outlier removal option

jdmulligan:


PWGJE/EMCALJetTasks/macros/JetQA/scalePtHardHistos.py
```diff
@@ -220,7 +235,136 @@ def ScaleAllHistograms(obj, scaleFactor,f):
...
+          
+  count = 0.
+  average = 0.
+  for i in range(index - numberOfCountsBelowIndex, index + numberOfCountsAboveIndex + 1):
+    # Avoid going over histogram limits
+    if i < 0 or i >= hist.GetNcells():
+      continue
+    # TEMP - Hack due to entries in pt hard value = 1 for large pt hard bins for no apparent reason
+    if i == 1:
+      continue
```
raymondEhlers: You probably should remove this temp hack. I've addressed it in the embedding helper, so it shouldn't be an issue anymore

PR #4454, opened 2018-02-28
in case of charged jets, allow charged background computations

jdmulligan:
(update to my user task)

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
@@ -285,7 +285,7 @@ void AliAnalysisTaskEmcalJetPerformance::AllocateJetHistograms()
     
     // (Centrality, pT, NEF)
     Int_t nbinsx = 20; Int_t minx = 0; Int_t maxx = 100;
-    Int_t nbinsy = nPtBins; Int_t miny = 0; Int_t maxy = fMaxPt;
+    Int_t nbinsy = fMaxPt; Int_t miny = 0; Int_t maxy = fMaxPt;
     Int_t nbinsz = 50; Int_t minz = 0; Int_t maxz = 1.;
     
     histname = TString::Format("%s/JetHistograms/hNEFVsPtEMCal", jets->GetArrayName().Data());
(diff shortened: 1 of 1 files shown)


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
```diff
@@ -1886,53 +1886,56 @@ void AliAnalysisTaskEmcalJetPerformance::ComputeBackground()
...
-      deltaR = GetDeltaR(&clus, etaEMCalRC, phiEMCalRC);
-      if (deltaR < jetR) {
-        clusPtSumEMCalRC += clusPt;
-      }
-      
-      // (6)
-      deltaR = GetDeltaR(&clus, etaDCalRC, phiDCalRC);
-      if (deltaR < jetR) {
-        clusPtSumDCalRC += clusPt;
+    if (jetCont->GetArrayName().Contains("Full")) {
```
raymondEhlers: It may not be super efficient to have TString::Contains() in your event loop, but I don't think it's a huge deal (if you want to change it, maybe you could check if the cluster container attach to the jet is null?)

jdmulligan: True, was just out of laziness. Your suggestion is anyway more robust, I'll make a fix. Thanks

raymondEhlers:
+1

raymondEhlers:
I had one other thought (since avoiding dependence on the cluster container in the jet container would be better if possible) - can't you just check if you have a cluster container attached to your task? Or would it be possible to have charged jets but still want to calculate the background?

jdmulligan:
What I intend to do is loop through all jet containers attached to the task, and for each, compute e.g. background fluctuations. The type of calculation for the background fluctuations most naturally corresponds to the jet container, rather than the containers in the task. For example, it allows to compute both full and charged background fluctuations in a single instance of the task (which wouldn't be possible if I base the computation on whether the task itself has a cluster container). What is the reason you consider having dependence on the cluster container in the jet container problematic?

raymondEhlers:
Okay, thanks for enlightening me! My suggestion is not at all helpful then! (Sorry for the diversion)

Ideally, we'd remove cluster and particle containers from the jet container, so minimizing any dependence now would make that easier (although I don't think it's super likely to happen soon). So I figured I would ask now if it was a trivial change, but since it's not, no need to worry about it!

PR #4455, opened 2018-02-28
Including charm template corrections

efgauger:
Plots added to check weighting D meson pT distribution in MC, charm species ratios in MC, etc.


sashingo:
+1

PR #4456, opened 2018-02-28
EMCal Embed: Update jet taggers

raymondEhlers:
The tasks use `AliEmcalJet::TrackAt(int index, TClonesArray * tracks)`,
with the particles accessed via the particle container attached to the
jet container. This will cause harmless warnings (the proper particle
is _always_ returned) as explained in the docs of `AliEmcalJet` and
of the embedding framework. However, the tasks should be updated to
silence these warnings under the new embedding framework.

This update causes no change in functionality.

Two compiler warnings for unused variables are also fixed.

@lcunquei , @mfasDa I'm switching over to these classes to better handle jet matching and cut on the shared momentum fraction. This modification is trivial, but allows the use of these classes without warnings under the new embedding framework. If possible, I'd like to get this into the tag for tomorrow - please let me know what you think!

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetTagger.cxx
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetTagger.cxx
@@ -222,8 +222,8 @@ void AliAnalysisTaskEmcalJetTagger::Init(){
    if(!cont1 || !cont2) AliError("Missing jet container");
    
    // when full azimuth, don't do anything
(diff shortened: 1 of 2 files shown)


raymondEhlers:
+test

raymondEhlers:
One follow up: I ran my integration tests and the output histograms are the same

mfasDa:
Fine from my side - please go ahead!

raymondEhlers:
Thanks @mfasDa !

@lcunquei - do you have any objections to me approving this for today's tag?

raymondEhlers:
+1 I'm approving as I haven't heard any objections

PR #4474, opened 2018-03-01
Substructure tree: Add trigger cluster information

mfasDa:
Adding index for a trigger cluster to the substructure tree
- 0: CENT
- 1: CENTNOTRD without CENT
- 2: CALO
- 3: CALOFAST without CALO
-> CENTNOTRD = CENT || (CENTNOTRD without CENT)
-> CALOFAST = CALO || (CALOFAST without CALO)


mfasDa:
+1

PR #4494, opened 2018-03-03
Add mode to use fixed mass mean, sigma 

moliver813:
Add mode to use fixed mass mean,sigma (per analysis pT bin)
Also, adding self to author line :)


efranzis:
+1

PR #4495, opened 2018-03-03
removal of the includes at the beginning, since these crash my train

bhohlweger:
add task macro didn't work with trains due to the includes on top - root includes should work ?!


rbertens:
+1

bhohlweger:
thanks redmer :)

PR #4499, opened 2018-03-04
Update leading hadron calculation method

raymondEhlers:
The functions in AliEmcalJet do not use the proper energy in the
momentum calculation of the leading cluster. Thus, the jet constituents
should be used. Due to a bug in retrieving the leading cluster and
particle, they are calculated in the utils class instead


raymondEhlers:
+1

mfasDa:
Bug referred to is fixed in #4522, local tests look OK.

PR #4502, opened 2018-03-04
Revert "Incorporated resonances/conversions rejection functionality"

victor-gonzalez:
Wrong binning configuration. A fix will follow.


aohlson:
+1

PR #4515, opened 2018-03-05
add a function to measure PHOS trigger efficiency in MB

dsekihat:
add a function to measure PHOS trigger efficiency in MB.


kharlov:
+1

PR #4516, opened 2018-03-05
Improve memory behaviour of AliAnalysisTaskPLFemto

a-mathis:
o Lightweight option to initialize only a subset of the histograms
o Employ smaller data types


rbertens:
+1

PR #4517, opened 2018-03-05
Bugfixes and Minimal booking 

bhohlweger:
A minimal booking implementation all over the task for memory optimization, where several bugs had to be fixed. Additionally a bug in the multiplicity dependent binning of the same event and mixed event distribution was fixed.


rbertens:
+1

bhohlweger:
:3

PR #4522, opened 2018-03-05
[Bugfix] Handling of return address for leading constituents

mfasDa:
Use const auto & in range-based for loop in order to
prevent copying which would result in a pointer to a
local variable.

Fixing #4499


mfasDa:
+1

PR #4529, opened 2018-03-05
Update leading hadron calculation and docs

raymondEhlers:
Update method committed in #4499 since the standard method was fixed
in #4522. Also update docs in the jet container to better describe
various options to retrieve the leading hadron momentum.

Thanks @mfasDa for the fix!


raymondEhlers:
+1

PR #4536, opened 2018-03-06
Possibility to use generated pT and physical primaries

afestant:
Adding possibility to use generated pt and select physical primaries when running on Monte Carlo


fprino:
+1

PR #4541, opened 2018-03-06
PWGGA/GammaConv added bin configurations for EMC and PCM-EMC PbPb for…

mikesas:
… lowering memory consumption


FriederikeBock:
+1

PR #4549, opened 2018-03-06
Using ClassDefOverride for ROOT6

mfasDa:


mfasDa:
+test

Not sure this works properly under root5, eventually the part using override needs to be converted back to virtual.

mfasDa:
Doesn't work with ROOT5 (as expected) - will remove cxx11 mode with final and override and only use it for the constructors in order to suppress the user implementation of the default constructor.

PR #4550, opened 2018-03-07
EMCal CorrFW: Fix additional dictionary warning

raymondEhlers:
Add dictionary entry for pair of AOD track and particle container.
Introduced in a0e3356 and partially fixed in 8de9429.


raymondEhlers:
+1

PR #4569, opened 2018-03-07
update PHOS trigger analysis

dsekihat:
update PHOS trigger analysis. make T-P distance wider in trigger efficiency.


kharlov:
+1

PR #4570, opened 2018-03-08
Included additional info in QA histogram title for reference

rderradi:
Additional info about the production name and whether the calibration OADB exists for the run being analyzed were included in the title of the QA fHistEventCounter histogram for future reference.


fprino:
+1

PR #4571, opened 2018-03-08
Add a new Pt min cut   < 0.05 MeV

pgonzale83:
Add case 6 which set Pt min  = 0.05 MeV for primary electron selection


dmuhlhei:
+1

PR #4582, opened 2018-03-08
[feature,doxygen] Add pp 2012 triggers + document task

mfasDa:
Add 2012 triggers and a predefined list of supported dataset.
Include 2017 and 2018 pp in the run2 pp definition.
Improve documentation of the task.


mfasDa:
+1

PR #4584, opened 2018-03-08
Global Track Array Changes and Minimal Booking

bhohlweger:
Wrong order of initialization in the CorrHists caused undefined behaviour. Improvement of the handeling of the too small global track array size, only a warning will be thrown and the task will still run. The track and v0 candidates are rejected then.


rbertens:
+1

PR #4597, opened 2018-03-08
extend cluster time plot to negative values

jdmulligan:
Old framework is also updated, only for the sake of making automatic comparison script still agree.

@raymondEhlers


jdmulligan:
+1

PR #4605, opened 2018-03-09
Improved PID bachelor selection in case of combined probability

decaro:
Improved PID bachelor selection in case of combined probability used (fPidSelectionFlag==9)
Combined probability cut variable according to the bachelor momentum


fprino:
+test

fprino:
+1

PR #4617, opened 2018-03-09
[Feature] Add cluster-patch tagging

mfasDa:
-  Get selected patches from the trigger selection
   task (must always run)
- Make correlation cluster energy - tag patch
   energy: From multiple patches matched to a
   cluster the leading patch is selected


mfasDa:
+1

PR #4618, opened 2018-03-09
Update EMCal, EMCal-Jet run macros for ROOT 6

raymondEhlers:
Tested that all run on ROOT 6, but additional testing on ROOT 5 would be
helpful.

@aiola , @mfasDa, @jdmulligan  - I received a request to update these for ROOT 6, so I gave it a quick shot. Seems to work fine on ROOT 6, but testing on ROOT 5 or general comments are welcome. I updated the trigger QA run macro because I ran across it, but I could remove the changes if desired

### DIFF
--- a/PWG/EMCAL/macros/runEMCalSampleTask.C
+++ b/PWG/EMCAL/macros/runEMCalSampleTask.C
@@ -1,3 +1,15 @@
+/// \file runEMCalSampleTask.C
+/// \brief Example macro to run a EMCal-Framework-based analysis
+///
+/// \ingroup EMCALFWTASKS
+/// This macros illustrates how to run an EMCal analysis using the EMCal analysis
+/// framework. It can run locally or on grid (test/full/terminate modes).
+/// The script runEMCalSampleTask.sh in the same folder allows to easily set
+/// the input values
+///
+/// \author Salvatore Aiola <salvatore.aiola@cern.ch>, Yale University
+/// \date Mar 09, 2016
+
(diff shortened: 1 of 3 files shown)


raymondEhlers:
+test (Won't be meaningful, but doing it now anyway so merging will go faster)

adriansev:
Wouldn't be easier to add Create{AOD,ESD}Chain utilities in a general lib? (maybe even in ALIROOT in AliAnalysisManager.h as static methods as they are general functions to create chains to be used with StartAnalysis method of AliAnalysisManager)?

raymondEhlers:
Sure, it would be better, but I was trying to make the most minimal changes possible. At the first glance, I think the two functions could possible be (mostly) combined together by just adding a few variables. Unfortunately, I don't have the time to push through and test such a change, so I propose this as the next best option. But I would certainly support such a change if someone else wanted to take the initiative

mfasDa:
Hi @raymondEhlers,

I see no obvious problems for ROOT5 in the changes, and I can definitely confirm that the ifdef __CLING__ guards will execute properly the ROOT6-specific code under cling and ifndef __CLING__ will properly execute the ROOT5-specific code under cint (other direction with __CINT__ is not the case) - this part I have tested.

While I agree that it is the right way to go away from macros in favor of compiled code also for helpers like chain creators I think it is not urgent. I think the change could be merged.

Cheers

Markus

raymondEhlers:
+1 Thanks to everyone for the comments!

PR #4619, opened 2018-03-10
Fixes needed to work with both ROOT6 and ROOT5

chiarazampolli:
After the changes for ROOT6, some things didn't work anymore in ROOT5. This should fix it.


kharlov:
+1

PR #4620, opened 2018-03-10
EMCal Embed: Example run macro and update docs

raymondEhlers:
The example run macro supports both full and charged jets.


raymondEhlers:
+1

PR #4622, opened 2018-03-10
fixing the conflict with multiselection, jt/iaa analysis code runhead…

dongjokim:
…er problems, event cuts for 15o, cleain up jfilter


aohlson:
+1

PR #4627, opened 2018-03-11
Allow compilation without AliEn

ktf:


ktf:
Notice I did not check if it still works with AliEn. My assumption is that WITHALIEN is actually defined during build as it is in root.

ktf:
+test

dberzano:
Hi @ktf, what is the use case? In the meanwhile I am double-checking whether the variable is available.

ktf:
Building AliPhysics with `--defaults o2`.

PWGPP/AliOfflineTrigger.cxx
```diff
@@ -639,15 +641,20 @@ void   AliOfflineTrigger::ExtractSelected(const char *rawList, const char * trig
...
     rawArray=new TObjArray(nFiles);
     while( coll->Next()){
       rawArray->AddLast(new TObjString(coll->GetTURL()));
     }
+#endif
   }else{
     rawArray = gSystem->GetFromPipe(TString::Format("cat %s", rawList)).Tokenize("\n");  
   }
+  if (rawArray == 0) {
+    ::Error("AliOfflineTrigger::ExtractSelected","No AliEn support compiled-in");
```
dberzano: I'd raise a fatal error here, and maybe move it into an `#ifdef ... #else ... #endif` block.

ktf:
I've changed it to Fatal and adapted the error message. I do not think an ifdef is needed, since the code works correctly also in the case the gSystem fails (right now it would crash without an error).

dberzano:
+1 @ktf thanks for your update, I see you have also changed the error message (you removed references to AliEn), so that now your change makes more sense to me :-)

PR #4635, opened 2018-03-11
Change access specifier to access AliCaloCellsQA field in derived classes

kqf:
Now it's impossible to update `AliCaloCellsQA` without copying the existing code. The `private` hides the variables so they can't be accessed in derived (child) classes.


kharlov:
+1

PR #4641, opened 2018-03-12
[Hotfix] Fix auto-configuration for periods

mfasDa:
Small bugs (improper call of config function for periods)
prevented configuration of the matching period after the
recent update for pp 2012 and 2017/18.

In addition further cross check for for availability of
trigger decision objects in cluster task


mfasDa:
+1

PR #4644, opened 2018-03-12
Fixing output name

chiarazampolli:
Hello,
There was a mistake in building the name of the output. This is from the attempt to fix the previous version of the macro.
@njacazio , @fbellini : can you check how you wanted to use the "partName"? I have a proposal commented out in the macro, which I am not committing as I might then break the post-processing. Please, change if needed, now the output will be called as before.

Chiara


fbellini:
Ciao Chiara,
   thanks for the fix. What you committed is just fine. The "partName" is already used as suffix in the listName whenever needed thus no need to add it to the folder name. Sorry for forgetting to remove it.
The change as is can be merged and the PR closed.
Thank you and cheers,
   Francesca

chiarazampolli:
Ciao Francesco,

Ok, then you have an extra string before I started fixing for ROOT6. 

Good to know. I will remove the commented part.

Cheers,

Chiara

PR #4651, opened 2018-03-12
Herwig: Automatic matchbox file generation from template with JINJA

mfasDa:
Matchbox files contain several keys which need to be handled on-
the-fly (number of events, random seed). The easiest solution
to configure matchbox file from static and dynamic settings is
treating Matchbox.in files as JINJA templates with the JINJA
conventions and use a small python script called in gen_herwig.C
to create the Matchbox.in file from the Matchbox.tmpl file.
For configuration settings the python script assumes the AliDPG
naming convention (global variables, starting with CONFIG_).

Attention: The script should be called on the workers. Requires
the python package JINJA to be installed on the grid sides. The
commit here represents only an example at the current stage and
might not work on every grid site.

### DIFF
--- a/PWG/MCLEGO/Herwig/LHC-MB.tmpl
+++ b/PWG/MCLEGO/Herwig/LHC-MB.tmpl
@@ -1,0 +1,52 @@
+# -*- ThePEG-repository -*-
+
+################################################################################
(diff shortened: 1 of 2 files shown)


rbertens:
This PR was discussed and will be followed up on the meeting on Wednesday. For future reference we should update the egroup for this specific folder

PR #4659, opened 2018-03-12
Add pT vs Mass vs Cent Histos to AliAnalysisTaskGammaHadron

moliver813:
Adds histogram for examining mass vs pt in different centrality bins.
Changes the final centrality bins.


e-masson:
+test

e-masson:
+1

PR #4662, opened 2018-03-13
OADB/EMCAL: IMPORTANT change to OADB instructions for EOS

nschmidtALICE:
Please use the different global variable to access the OADB files as with OADB_PATH several people ran into problems with the PhysicsSelection.

Please use:
ALICE_DATA=/path/to/my/local/oadb
instead of:
OADB_PATH=/path/to/my/local/oadb/OADB


mfasDa:
+1

Note that the real problem is the usage of OADB_PATH directly by some components and in general it would be better to fix those. This is not necessarily only the physics selection but other users might access OADB_PATH directly as well.

PR #4667, opened 2018-03-13
[EMCAL-135] Disable temperature recalib due to memory leak

mfasDa:
It has been observed that the tender and the correction framework
leak ~100 MB once a run is changed, leading to problems for jobs
processing multiple runs. 70% of the leaking is produced from
loading of the temperature recalibration. As the temperature
recalibration is not used in run2 initializing the recalibration
factors is disabled in order to prevent opening the OADB file untill
the memory leak in the OADB container is fixed.

### DIFF
--- a/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellEnergy.cxx
+++ b/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellEnergy.cxx
@@ -221,6 +221,21 @@ Int_t AliEmcalCorrectionCellEnergy::InitRunDepRecalib()
 {
   if (!fEventManager.InputEvent())
     return 0;
+
+  // MF Disable temperature calibration for run2 for the moment
+  // A memory leak in the AliOADBContainer was observed when the 
+  // container is deleted. As consequence when processing multiple
+  // runs the correction task leaks ~100 MB per run included in the 
+  // job, where 70% of the memory leak comes from the temperature
+  // calibration. Untill a fix is provided the temperature calibration
+  // has to be disabled for run2 runs. Disabling has to be done before
+  // opening the OADB container
+  // 
+  // For more information see https://alice.its.cern.ch/jira/browse/EMCAL-135
+  if(fEventManager.InputEvent()->GetRunNumber() > 197692){
+    AliError("Temperature recalibration disabled for run2");
+    return 0;
+  }
   
   AliInfo("Initialising recalibration factors");
   
--- a/TENDER/TenderSupplies/AliEMCALTenderSupply.cxx
+++ b/TENDER/TenderSupplies/AliEMCALTenderSupply.cxx
@@ -1180,6 +1180,21 @@ Int_t AliEMCALTenderSupply::InitRunDepRecalib()
   
   if (!event) 
     return 0;
+
+  // MF Disable temperature calibration for run2 for the moment
+  // A memory leak in the AliOADBContainer was observed when the 
+  // container is deleted. As consequence when processing multiple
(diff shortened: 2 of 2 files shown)


mfasDa:
Disabling temperature recalibration for run2 as discussed in https://alice.its.cern.ch/jira/browse/EMCAL-135

mfasDa:
+test

mfasDa:
@raymondEhlers Done. @gconesab Can you please approve ASAP. Some trains seem to be in trouble with memory consumption and need this urgently to be in AliPhysics in order to not being blocked.

raymondEhlers:
Thanks @mfasDa! Can you point me to an example train where this is a
problem? I was under the impression that was not ideal, but also that we
didn't have any real world examples where it actually caused problems.
Thanks!

nschmidtALICE:
@raymondEhlers, here you can find an example train from the GA group where this problem can be seen quite well.
http://alitrain.cern.ch/train-workdir/PWGGA/GA_pPb/721_20180228-1145/test/GammaCaloMerged_8TeV_CFSpecial_101_CFS500A100W5/syswatch.png

raymondEhlers:
Hi @nschmidtALICE  ,

Thanks! This is from a test of a metadata set, right? My understanding is that sort of test is basically a worst case scenario - I see that it includes 4 correction tasks, and the metadata includes files from each dataset, which means it will leak as many times as there are metadata sets. So while no doubt this is a problem that we want to fix eventually (and I have no objection to this PR), I think is a more relevant measure - specifically, how often does this occur during a train job as opposed to a test? I don't have any concrete information on that, but I was under the impression that it isn't terribly common.

Best,
Raymnod

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellEnergy.cxx
```diff
@@ -221,6 +221,19 @@ Int_t AliEmcalCorrectionCellEnergy::InitRunDepRecalib()
...
     return 0;
+
+  // MF Disable temperature calibration for run2 for the moment
+  // A memory leak in the AliOADBContainer was observed when the 
+  // container is deleted. As consequence when processing multiple
+  // runs the correction task leaks ~100 MB per run included in the 
+  // job, where 70% of the memory leak comes from the temperature
+  // calibration. Untill a fix is provided the temperature calibration
+  // has to be disabled for run2 runs. Disabling has to be done before
+  // opening the OADB container
```
raymondEhlers: Hi @mfasDa , I'm just looking at the code now - could you add the link here too?

mfasDa: Done

nschmidtALICE: Hello @raymondEhlers,
yes, this is a meta dataset with multiple correction tasks and therefore, of course, the worst case. 
Here is also an example of a meta dataset, but with the Tender: http://alitrain.cern.ch/train-workdir/PWGGA/GA_pPb/719_20180212-1953/test/__ALL__/syswatch.png
This shows the same behaviour.
Cheers, Nicolas

raymondEhlers: Thanks @mfasDa

loizides:
+1

loizides:
@gconesab , can you please approve also for the second directory so we have the (temp) fix in todays tag!

gconesab:
+1

PR #4670, opened 2018-03-13
updates two-particle factorization analysis

cbourjau:
General updates; mainly concerning event selection cuts and MC stuff


aohlson:
+1

PR #4676, opened 2018-03-13
PWGDQ: effi task v2: added generator separation functionality

carstooon:
- Adds the possibility to separate different generators from each other
- Bugfix for filling of support histograms


pdillens:
+1

PR #4682, opened 2018-03-14
options for different run period are added

nirbhaygit:
Options for different run period are added. AliEventCuts only to be used for Run2.


akubera:
+1

PR #4685, opened 2018-03-14
 Added downsampling + daughter-removal from q2, q2TPC vs. q2V0C histos and D0 from Dstar analysis

fgrosa:
Added:
1) possibility to enable downsampling of tracks from q2 and daughter-removal from q2 at the same time
2) q2TPC vs. q2V0C histo to get correlation between the two q2
3) possibility to analyse only D0 coming from Dstar


fprino:
+test

fprino:
+1

PR #4690, opened 2018-03-14
PID bufix and added Flags to set harmonics for 3-particle correlator

rihanphys:
PID bufix and added Flags to set harmonics for 3-particle correlator


jmargutt:
+1

PR #4707, opened 2018-03-15
Added a new class BadChannelAna_V2 for the bad channel finding. This …

jokonig:
…class uses a different analysis technique then the existing class BadChannelAna


raymondEhlers:
+test

efranzis:
@jokonig 
Is this really necessary to add a completely new class? I assume they are still the same for 80-90% of their code. Maybe you can just add another member function with a V2 name or add a switch for things like your BuildHitAndEnergyMean_scaled(Int_t crit, Double_t emin, Double_t emax). I will have to go through this in more detail - which will probably not happen before Monday.

PR #4713, opened 2018-03-15
New Task to check for the raw correlation signal of the MC particles …

bhohlweger:
…without reconstruction effects within full four pi and within acceptance.


rbertens:
+1

PR #4737, opened 2018-03-17
Wrong size of the Cent bins in the CollConfig and binning

bhohlweger:
The size was taken from the mult bins. Removal of an uneeded AddTask (is now in FEMTOSCOPY/macros)


rbertens:
+1

PR #4739, opened 2018-03-17
remove kCanRebin

shkhan09:


rbertens:
+1 @shkhan09 thanks for the quick action !

PR #4760, opened 2018-03-19
Some cosmetics and implementation of pt cuts for the v0 and the xi in the cascade cut class

bhohlweger:
Minor bugfixes and cosmetics to the plotting classes, changes to some of the binning, and implementation of a pt range cut to the cascade class on the xi and v0 candidate.


rbertens:
+1

PR #4767, opened 2018-03-19
change <vector> (ROOT6 doesn't support) to arrays

rihanphys:
change vector (ROOT6 doesn't support) to arrays


rbertens:
+1

mfasDa:
ROOT6 does very well support std::vector!

You were missing a std::

rihanphys:
hi @mfasDa thanks for the info. I do not have ROOT6 yet in my mac (Having some trouble to install). I'll soon check with std::vector.

PR #4770, opened 2018-03-20
fix THnSparseF definition

jaimenorman:
change member variable THnSparseF -> THnSparse to be compatible with classdef


raymondEhlers:
+1

PR #4805, opened 2018-03-22
Removed my module dependencies to others in same directory

hyeonjoongkim:
Since I'm the only user using QnVector module, so I've made a separate shared library.


aohlson:
+test

PR #4828, opened 2018-03-23
large scale changes for the QM approvals

bhohlweger:
New way of obtaining the uncorrelated sample by rotating the angle of pairs in phi. Additional QA for tracking the effective mixing depth in the mixed event and delta phi vs. delta eta plots for the particles.


rbertens:
+test

rbertens:
+1

PR #4852, opened 2018-03-25
pull request

eumaka:


akubera:
+test

aohlson:
Hi @eumaka 
The compilation fails because of the line #include "AliLightV0vertexer".  If absolutely necessary, we could add a link the PWGLF/Strangeness directory of AliPhysics, in order for the include file to be found.  However, I don't see that AliLightV0vertexer is used anywhere in your class.  If it's not necessary, could you remove the include statement?  I notice there seem to be many unnecessary includes, in fact.  Could you try reducing them?  
Thanks!

akubera:
+test

aohlson:
+test

aohlson:
+1

PR #4858, opened 2018-03-25
Reducing memory consumption & task for pp

hyeonjoongkim:
Removing unused histograms to reduce memory consumption
Task fix for pp


aohlson:
+1

PR #4871, opened 2018-03-27
Print and plot time to execute InitTree()

jdmulligan:
Please do not approve yet!

@raymondEhlers , when you have a moment please take a look. I still don't know where to write to so I just printed to stdout and histogram (as you pointed out, from the successful jobs it will at least be interesting to see the distribution). I think the commit is harmless, but best for you to take a look and then approve :)

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
+++ b/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
@@ -35,7 +35,6 @@
 #include <TH1F.h>
 #include <TRandom3.h>
 #include <TList.h>
-#include <TStopwatch.h>
 
 #include <AliLog.h>
 #include <AliAnalysisManager.h>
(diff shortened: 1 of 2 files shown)


raymondEhlers:
I haven't looked at this in any depth, but a few quick comments:

- Please make it possible to enable / disable. Otherwise, we can't keep it from blowing up the logs
- More trivial, but change the stopwatch to an object instead of a pointer. Unless I'm missing something, I don't see why it needs to be a pointer, and I'd say it's usually best to avoid pointers unless necessary.
- Increment the ClassDef()

jdmulligan:
Hi @raymondEhlers , Thanks for the comments, I made modifications to the commit. Let me know if it looks okay to you (get to it when you can, I know you're very busy).

raymondEhlers:
Thanks for the changes! Two more comments and then otherwise I think it looks good:

- I had originally envisioned printing an absolute time stamp (for example, the unix time). I don't think there is any additional information vs what you've done, so I think you're implementation is good (and also easier to plot!). Can you think of any additional benefits? If not, let's keep it as it is (I want to check since my tired mind might have missed something)
- Really trivial, but please make the option configurable in YAML and add fPrintTimingInfoToLog to the Print() function. The former is not so important (but would be nice for consistency), but the latter is useful for keeping track of configurations, since it prints during train configuration (and I use it often when inspecting trains later).

jdmulligan:
Hi @raymondEhlers , 

YAML is in. Regarding the time stamp, I can't think of any reason we'd need the absolute time, since we just want to measure this one step. If later we find we want it we can of course add it, but I can't think of a reason.

Thanks for taking a look so quickly! I'll approve it then before the tag tomorrow if I don't hear otherwise.

raymondEhlers:
Great, thanks! I'll leave the actual approval up to you, but it looks good to me

jdmulligan:
+1

PR #4885, opened 2018-03-28
Remove absolute values on pdg labels

kqf:
It also fixes typos in class names for PHOS neutral mesons analysis @ 13 TeV


kharlov:
+1

PR #4893, opened 2018-03-28
PWGGA/GammaConv small Bugfix

AMechler:
There was an Bug in DCal Analysis which made them crash
This should fix that


dmuhlhei:
+1

AMechler:
Problem was introduced because of log axis starting with 0

PR #4896, opened 2018-03-28
Minor modifications in D0 from Dstar analysis

fgrosa:
1) Modified invariant-mass range of Dstar candidates used to select D0 from Dstar (from 5 to 3 sigma)  
2) Modified control of not double counting D0 candidates (one D0 can build different Dstars)
3) Added histogram to monitor number of Dstar candidates built using the same D0 candidate


fprino:
+1

PR #4906, opened 2018-03-29
Test pile-up rejection - study pT dependence

a-mathis:
The last attempt did not yield any further insight apart from the fact that the correlated v0 background is most pronounced at low pT. Therefore now test different pT cuts to see the impact on the v0 purity and the Lambda-Lambda correlation function


akubera:
+1

PR #4917, opened 2018-03-29
Reducing memory consumption

hyeonjoongkim:
* Reducing memory consumption by limiting number of bins in THn
* Adding Minimum pTA cut for merging
* Adding flag for run 2 dataset
* EP detector V0A -> V0C


akubera:
+test

PR #4922, opened 2018-03-30
Cluster E: added histos and bug fix. 

efgauger:
Added cluster E histograms and took out line that double-counts in existing cluster E histos.


sashingo:
+1

PR #4934, opened 2018-03-30
Bug fix: pseudorapidity cut removed from event selection

gbencedi:
Pseudorapidity cut has been removed from event selection.


ycorrales:
+1

PR #4972, opened 2018-04-02
modify streaming syntax of several members

jdmulligan:
@mfasDa @raymondEhlers 

Hi Markus,

I propose to update the streaming syntax of a couple members -- do you see any problem? To be honest I don't understand quite what the difference between "//!" and "//!<!" is, but my impression is that "//!<!" is more encouraged. I am experiencing some strange train issues when running this task, which may be due to streaming problems -- so I wonder if this can somehow be the cause.

Best,
James

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliEmcalJetTaggerTaskFast.h
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliEmcalJetTaggerTaskFast.h
@@ -216,10 +216,10 @@ class AliEmcalJetTaggerTaskFast : public AliAnalysisTaskEmcalJet {
   TH2            **fh2PtJet1VsDeltaR;            //!<! \f$ p_{t}\f$ jet 1 vs dR
   TH2            **fh2PtJet2VsFraction;          //!<! \f$ p_{t}\f$ jet 1 vs shared fraction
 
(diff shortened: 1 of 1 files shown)


raymondEhlers:
Hi @jdmulligan ,

I don't have any problem with this PR, but I also don't think this can have an impact. For an explanation, see [here](https://dberzano.github.io/alice/doxygen/#data_members).

Best,
Raymond

mfasDa:
+test

Hi @jdmulligan,

as @raymondEhlers already mentioned the change in streamer is due to ROOT6 compatibility, so I wonder whether this will solve your problem, nevertheless it could be merged.

Cheers

Markus

jdmulligan:
Hi @mfasDa , @raymondEhlers ,

I see, okay, so it shouldn't make a difference. Thanks for your explanations. We may as well merge it nevertheless. 

Markus, have you seen any streaming problems or odd behavior using the fast tagger? I am not 100% sure my problems are due to the tagger, but I am running a fairly simple train with only correction task, embedding helper, jet finders, and jet taggers. All of which I've used many times, with the exception of the tagger which I've used only a couple times, so I wonder if there may be some subtle problem there.

Best,
James

jdmulligan:
Hi @raymondEhlers , @mfasDa ,

Can one of you please approve? (So we don't leave this open)
Thanks!
James

raymondEhlers:
+1

PR #4973, opened 2018-04-02
update PHOS L1 trigger analysis

dsekihat:
this contains two commits (update PHOS L1 trigger analysis and pi0 weighting in HIJING)


kharlov:
+1

PR #4975, opened 2018-04-02
Comment out cout statement and line that rejects non-MC.

efgauger:
Comment out cout statement and line that rejects non-MC.


sashingo:
+1

PR #4998, opened 2018-04-04
Make Clones of external TList and set Flags, add QA after NUA

rihanphys:
Make Clones of external TList and set Flags, add QA after NUA


rbertens:
+1

PR #5021, opened 2018-04-06
[EMCAL-135] OADB/EMCAL: fix of AliOADBContainers and new maps for LHC16hijlqt and LHC17il

nschmidtALICE:
The OADB files for EMCal are now updated on EOS with a fix necessary to remove a memory leak from AliOADBContainer. Details about this can be found on JIRA EMCAL-135 and the PR 4792.

New BC maps for LHC17i,l were added and additional bad channels were added to the BC maps of LHC16hijlqt.


mfasDa:
+1

PR #5036, opened 2018-04-06
EMCal CorrFW: Documentation update

raymondEhlers:
Note that intermediate LEGO train tests can fail is using a configure
wagon. Such tests failing is not necessarily a problem - the results
that matter are the full train test.

@jdmulligan - Just so you're aware and can point people to it if necessary


raymondEhlers:
+1

PR #5043, opened 2018-04-08
Fix for std=c++14 not supported error

victor-gonzalez:
AliPhysics compilation broken


akubera:
+1

akubera:
Sorry I was getting a little experimental with that code yesterday. I'm quite surprised that it allowed the second version to pass while the first one (with the lambda implemented) threw the compiler error. Again, sorry for any inconvenience.

victor-gonzalez:
No problem!

PR #5045, opened 2018-04-08
K*+- set as: not a physical primary particle

pragatisahoo:
Before K*+- set as a physical primary particle(which was incorrect).


akalweit:
Are you sure?

pragatisahoo:
The histogram is not filled up when K*+- set as isPhysPrim. So I changed it as done for phi and K*0 particles.

PR #5059, opened 2018-04-09
[EMCAL-143] [ALIROOT-7754] Move some macros and code from AliRoot to AliPhysics

gconesab:

PR #5064, opened 2018-04-09
EMCAL - Fixed unsigned integer comparison error

akubera:
These `imc` indexes as `Int_t` would cause compiler error when using g++-6.3


raymondEhlers:
+1 Thanks!

akubera:
Oh no; this is the same problem the original commit encountered (#5059). AliPhysics currently needs AliRoot@master (or rather, ref v5-09-27-rc1), but the latest tag is v5-09-26 right now. @dberzano had merged it manually. I'll guess this will have to wait.

PR #5069, opened 2018-04-10
Fix EMCAL dependency and error in a QA script

dberzano:


dberzano:
Force-merging. This fix requires a more recent AliRoot version than the one we use for tests, that would fail.

PR #5078, opened 2018-04-11
Switched to High Multimplicity selector.

matarzil:


ycorrales:
+1

ycorrales:
I have approved it, btw I suggest to include a flag variable to select MB or HM trigger accordingly without change every time.

PR #5100, opened 2018-04-12
Add 2D histograms and responses to Jet Jt Analysis

TWSman:
-Adds 2D jT histograms (jT, pT)
-Adds 2D response matrices (3D histograms with histogram index as the fourth dimension)
-Add AliJTH3D to AliJHistManager
-Code cleanup


rbertens:
+1

PR #5120, opened 2018-04-12
new PID parametrization for low magnetic field configuration

iravasen:
PID parametrization for low magnetic field configuration. New parameters for PHOBOS parametrization. Pions and electrons use the old parametrization for nominal field.


bdoenigu:
+1

PR #5126, opened 2018-04-13
PWGGA/GammaConv  Bugfix for Calo CellIDvsCellE Histo axis

AMechler:
I think this should work as intended.


FriederikeBock:
Hi Adrian,

no you just have to remove the -1 after the i. Could you do that quickly?

Regards,
Fredi

PR #5127, opened 2018-04-13
PWGGA/GammaConv/ added configs for smallest cent intervals

mikesas:


FriederikeBock:
the last config would not be necessary. That you have already with a01.

FriederikeBock:
+1

PR #5130, opened 2018-04-13
"HFE tree creation updated to include field, event plane, etc"

mvolkl:
The heavy-flavour electron reduced event trees are updated to include more information and the event plane correction.


sashingo:
+1

PR #5143, opened 2018-04-13
PWGDQ Added default mc jpsi signal definitions 

pdillens:


pdillens:
Some default mc jpsi signals definition were added to the AliDielectronSignalMC class. They can be selected by an enum including a name scheme.

PR #5144, opened 2018-04-13
[EMCAL-51] Remap TRU index to STU convention

mfasDa:
Tests indicate that the abs fastor indexing works with
the STU convention while the DCS uses the TRU indexing.
In order to mask FastORs correctly based on the trigger
DCS config from OCDB the TRU indexing needs to be
remapped, otherwise masked FastORs appear in the wrong
position.


mfasDa:
+1

PR #5153, opened 2018-04-14
correct beta*gamma cut for pions and electrons

iravasen:
correct beta*gamma cut for pions and electrons. Correction of the bug.


bdoenigu:
+1

PR #5157, opened 2018-04-16
implement 15o pileup cuts, check +/- eta, and require mc matched jet acc

jdmulligan:
(updates to my user task)


raymondEhlers:
+1

PR #5166, opened 2018-04-16
[EMCAL-174] Compatibility for INT8 triggers

mfasDa:
- Add INT8/EMC8/DMC8 triggers to the list of supported
   triggers
- Add options ot relax vertex requirement when SPD is
   not in readout
- Adapt recalc patch task to look for INT8 triggers
   (handle them as INT7 triggers)

Task does not handle EMCAL triggers without
interaction trigger.


mfasDa:
+1

PR #5173, opened 2018-04-16
add LED event rejection, start Pt resolution, some typos, clean up z calculation

fibnumb:


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetSpectra8TeVTriggerQA.h
```diff
@@ -66,12 +82,20 @@ class AliAnalysisTaskEmcalJetSpectra8TeVTriggerQA : public AliAnalysisTaskEmcalJ
...
     Bool_t                      fUseRecalcPatches;          ///< Switch between offline (FEE) and recalc (L1) patches
     Bool_t                      SelectSingleShowerPatch(const AliEMCALTriggerPatchInfo *patch) const;
     Bool_t                      SelectJetPatch(const AliEMCALTriggerPatchInfo  *patch) const;
     THistManager                fHistManager                                      ;///< Histogram manager
+    Double_t                    GetZ(const Double_t trkPx, const Double_t trkPy, const Double_t trkPz, const Double_t jetPx, const Double_t jetPy, const Double_t jetPz) const;
+    Double_t                    GetSmearedTrackPt(AliVTrack *track)                ; ///< smeared Pt resolution with gaussian
 
 private:
+    
+    AliEMCALGeometry              *fGeom                     ;//!          EMCal goemetry utility
```
raymondEhlers: This shadows the base class. And you should update the streaming comment (either `///<` or `//!<!`

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetSpectra8TeVTriggerQA.h
```diff
@@ -66,12 +82,20 @@ class AliAnalysisTaskEmcalJetSpectra8TeVTriggerQA : public AliAnalysisTaskEmcalJ
...
+    AliEMCALRecoUtils             *fRecoUtil                 ;//!          Reco utility
+    TF1                           *fClusterEResolution       ;//!          Parameterization of cluster energy resolution from 2010 test beam results a = 4.35 b = 9.07 c = 1.63
+    Double_t                      fVaryTrkPtRes              ;//!          Variation of tracking momentum resolution
```
raymondEhlers: Should also update class def

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetSpectra8TeVTriggerQA.cxx
```diff
@@ -729,6 +748,13 @@ Bool_t AliAnalysisTaskEmcalJetSpectra8TeVTriggerQA::Run()
         }
     }
     
+    fClusterEResolution = new TF1("fClusterEResolution","sqrt([0]^2+[1]^2*x+([2]*x)^2)*0.01");
+    fClusterEResolution->SetParameters(4.35,9.07,1.63);
+    
+    //fGeom =  AliEMCALGeometry::GetInstance("EMCAL_COMPLETEV1");
+    fGeom =  AliEMCALGeometry::GetInstance("EMCAL_COMPLETE12SMV1");
```
raymondEhlers: This should be loaded automatically based on the period. You don't need to perform it yourself (unless you are doing something special with the geometry)

raymondEhlers:
+test

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetSpectra8TeVTriggerQA.h
raymondEhlers: Hi @rbertens , I'm not 100% certain here, but here is my thinking: Since they are transient, the new members won't change what is streamed. However, this does change the memory layout of the class (trivially), so I could imagine it having an impact when the object is created after streaming. I haven't dug into the details of the streamers enough to know the answer, so I tend to error on the side of updating the ClassDef for transients, even though it may be unnecessary. But perhaps I'm misguided here. What do you think?

In any case, for new code, I think the updated streamer markers (`///<` and `//!<!`) should be used for new code - otherwise, we'll never update them

rbertens: Hi @raymondEhlers,
As far as my knowledge goes, when read or written , persistent members are resolved by name (or hash), and not order, so changing the list of non persistent member doesn't require a classdef update. The streamer markers are another matter of course :)

raymondEhlers: Hi @rbertens ,
Okay, I see. That's probably a more clever strategy anyway - I just didn't know that they implemented it that way :-) Thanks for the explanation! I withdraw this comment

fibnumb:
Hi.  I believe I fixed the issues and is ready to test again

raymondEhlers:
+test

raymondEhlers:
(I'm fine with approving as is, but you might want to consider that chnage for maintainability)

raymondEhlers:
+test

fibnumb:
should be ready to test again

raymondEhlers:
+test

raymondEhlers:
+1

PR #5185, opened 2018-04-17
Jet bgrd. toy, extractor refinements

rhaake:
Changes to allow running the jet extractor on toy data. Used in jet background study.


rhaake:
+1

ebruna:
+1

PR #5186, opened 2018-04-17
Fix for proper tref handling

loizides:
Avoid copying objects that are being referenced via TRef. Instead either clone or move them from one to the other array. The latter was easily achieved use AbsorbObjects member of TClonesArray.
This is work related to EMCAL-142


loizides:
+1

PR #5191, opened 2018-04-17
[EMCAL-172] Exotic case for runs without INT trigger

mfasDa:
Optionally support events without INT trigger as exotic
case (no INTx or EMCx in trigger string). Such cases are
only of relevance for PHYSICS_2 runs or quiet beam runs.
By default disabled, due to memory constraints.


mfasDa:
+1

PR #5193, opened 2018-04-17
added zdc, dimuons, and versioning info

loizides:
Added ZDC, dimuon branches
Versioning of aliphysics and settings for task
version 1.0


loizides:
+1

PR #5202, opened 2018-04-18
fix problem with merging on grid

loizides:
On the grid one has to either load the data in the terminate function from the output list 
or avoid to access it. The latter is now implemented in this fix.


loizides:
+1

PR #5221, opened 2018-04-19
[EMCAL-174] Support for no-INT events on patch side

mfasDa:
- Do not perform event selection in the trigger maker
  as it is a service task
- Fix DCAL L0 trigger string in trigger base task
- Adopt recalc patch task to handle no INT triggers
- A few AliDebugs


mfasDa:
+1

PR #5225, opened 2018-04-19
Update AliAnalysisTaskSimSpectraLF.cxx

gbencedi:
Adding protection against negative values for TMath::Log() in rapidity selection.


bdoenigu:
+1

PR #5227, opened 2018-04-20
PWGGA/GammaConv Bugfix

AMechler:
Bugfix: Number of Bins for ECell vs CellID Histo for all Calos


FriederikeBock:
Hi Adrian,

this wouldn't fix the issue. I fixed it in another PR.
Thanks for looking into this.

Regards,
Fredi

PR #5234, opened 2018-04-20
PWGPP/EMCAL: update of OADB macros including OADBContainer fix

nschmidtALICE:
This PR contains the current macros I use for the creation of the OADB files. The most important addition is the ownership test at the end of each macro.
The BC map macro was rewritten from ground up and is now much more user friendly and powerful.


gconesab:
+1

PR #5235, opened 2018-04-20
add TRD PID Parameters for XeXe @ 5.4 TeV (LHC17n)  to TRDPIDResponse.root

kagarner:
- Only reference parameters available due to limited statistics. Threshold parameters set to -999.
- Reference parameters generated for combined charges.


chiarazampolli:
+1

PR #5243, opened 2018-04-22
Added alternative logic for D0 from Dstar analysis

fgrosa:
Added possibility to select D0 from Dstart, starting from D0 candidates and combining them with soft pions. Soft pions can be selected using standard selections for soft pions (3 points in ITS, pt > 0.1 GeV/c, ITS refit, SPD kAny) or requiring the TPC (filterbit 4).


fprino:
Hi Fabrizio,

in the method IsSoftPionSelected you can use AliESDtrackCuts::IsSelected, which calls AcceptVTrack where the AOD -> ESD track conversion is done, instead of doing the conversion locally.
In particular, this allows to avoid to call RelateToVertex, which modifies the track parameters.

Ciao, Francesco

fprino:
+1 thanks!

PR #5245, opened 2018-04-23
Moved soft pion AliESDtrackCuts in LocalInit() and fixed error in selection of Dstar inv mass

fgrosa:
1) Moved soft pion AliESDtrackCuts from class contructor to LocalInit() 
2) Fixed error in selection of Dstar inv mass for D0 from Dstar analysis (Dstar invariant mass interval to tag D0 from Dstar)


fprino:
+1

PR #5249, opened 2018-04-23
Improve alignment information

hzanoli:
In this commit I have changed:

- The propagation of the tracks is done by AliTrackerBase instead of AliEMCALRecoUtils. With this change I hope to achieve better precision on the edges of the EMCAL.
- Now the tree saves only XYZ positions of the track using the method above and the XYZ positions of the cluster. This saves space and may allow a more complete study. The phi and eta can of course be calculated straightforward.
- The default matcher is removed (saves a bit of space), since it is clearly not good enough for this study.
- The trees now have informations about the Z of the primary vertex
- A flag (is_in_fid_region) that looks if the cluster is or not in the edge was added.
- Check the distance to the bad channels and saves it to the tree (@gconesab can check if the implementation is ok for the last two)
- Changed the method to get the Energy from GetNonLinCorrEnergy() to E() (suggested by electrons experts) 

Please let me know in case you think I should add more variables or if I missed something from the discussion last week.

### DIFF
--- a/PWGPP/EMCAL/AliAnalysisTaskEMCALAlig.cxx
+++ b/PWGPP/EMCAL/AliAnalysisTaskEMCALAlig.cxx
@@ -24,6 +24,7 @@
 #include "AliEMCALGeometry.h"
 #include "AliExternalTrackParam.h"
 #include "TVector3.h"
+#include "AliTrackerBase.h"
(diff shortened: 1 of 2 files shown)


gconesab:
+1

PR #5269, opened 2018-04-25
Event weight information in the output

aiola:
This PR implements the event weight in the output for 3 tasks:

- The base class AliAnalysisTaskEmcalLight
- The inclusive jet analysis AliAnalysisTaskEmcalJetTree
- The D-meson tagged jet analysis AliAnalysisTaskDmesonJets


aiola:
+1

PR #5279, opened 2018-04-25
Avoid warnings by adding (not-implemented) private definitions of cop…

aiola:
…y constructors and assignment operators


aiola:
+1

PR #5282, opened 2018-04-26
[EMCAL-180] Apply EMCAL trigger bit selection to all cases

mfasDa:
Instead of separate &= operations use check for the triggerbit
for all INT (INT76, INT8, NoINT) cases


mfasDa:
+1

PR #5293, opened 2018-04-27
Add support for EMCAL gamma trigger

mfasDa:
Add trigger selection of the EMCAL gamma triggers to the
jet substructure tree and the jet spectrum task.


mfasDa:
+1

PR #5320, opened 2018-04-30
Centrality output now includes V0A, V0C and V0M

ewillsher:
Include all V0 centrality estimators in the tree outputs for comparisons


ycorrales:
+1

PR #5332, opened 2018-05-01
Added Abs() to MC label, put inv. mass calc after eID cuts again

efgauger:
Added Abs() to MC label, put inv. mass calc in data after eID cuts again.


sashingo:
+1

PR #5344, opened 2018-05-02
[EMCAL-51] Fix mapping for small supermodules in run2

mfasDa:
TRUs in the small supermodules still follow the linear
mapping between channels and bits in reg mask, used
in run1 and not the mapping used for the TRUs  in the
other supermodules (EMCAL full or DCAL 2/3rd). The mask
handler is created accordingly based on the TRU index
based on the L0 convention.


mfasDa:
+test

mfasDa:
+1

PR #5349, opened 2018-05-03
Modified track rotation and mixed event for Ds using pi and phi

fgrosa:
Modified track rotation and mixed event for Ds considering it as 2 prong, using phi and pion instead of kaons and pion:

1) for track rotation the pion is rotated and the phi (kaon pair) not
2) for mixed event the pion is taken from an event and the phi (kaon pair) from another one


fprino:
+1

PR #5359, opened 2018-05-04
Generic Framework code to run on MC trains

Katka19:
Task to get cumulants results with Generic Framework from standalone PYTHIA generated events; to be run on MC trains.


loizides:
+1

PR #5361, opened 2018-05-04
PWGGA/GammaConv adding pT scan in small steps for nominal B and lowB,…

amaringarcia:
… renaming the lowB config numbers


amaringarcia:
I know but the 5 TeV cuts are in between and it will be more difficult to follow

amaringarcia:
and having different numbers for low and nominalB make it more difficult to remember

nschmidtALICE:
+1

PR #5364, opened 2018-05-04
Several minor fixes, simplifications, and improvements to TPC tracking QA. Add some special features, that are disabled by default, which allow to compare HLT/Offline tracks one by one

davidrohr:


davidrohr:
@mkrzewic @wiechula @shahor02: Hi, this has been open for quite a while. Could someone have a look? Not sure who is responsible. It is only some minor fixes to the TPC-only tracking QA, + some optional extensions for HLT comparison, which are disabled at compile-time by default.

pzhristov:
+1

PR #5379, opened 2018-05-07
[EMCAL-56, EMCAL-180] Exclude/Require trigger coincidence

mfasDa:
Require or exclude coincidence of a trigger with a certain
list of triggers. Needed in order to check overlap of the
jet trigger with the gamma trigger.


mfasDa:
+1

PR #5388, opened 2018-05-08
[EMCAL-180] Adding task for patch multiplicity correlation

mfasDa:
In order to investigate the weird turnon behaviour at the
EG1 threshold in EJ1 events observed in LHC16k the
correlation between the multiplcities of firing (above
threshold) EJ1 and EG1 patches, as well as the number of
isolated (no overlap with firing JE patch) gamma patches
is monitored, assuming the turnon behaviour comes from
additional background.


mfasDa:
+test

mfasDa:
+1

PR #5393, opened 2018-05-09
Adding multiple planes for finite evp resolution correction

hyeonjoongkim:
* Extended output planes to 6, previous was 2 (In, out)
* Macro for lego train & plugin was modified
* Separating trigger pT bins and Associated pT bins


aohlson:
+test

PR #5394, opened 2018-05-09
[EMCAL-180] Monitoring for leading jets and leading constituents

mfasDa:
Monitoring for leading jets
- THnSparse with Lorentz vector
- Leading cluster
- Leading track

Monitoring for leading constituents
- pt(jet)
- NEF(jet)
- pt(const)
- z(const)
- DR(jet,const)

In addition add support for charged and neutral jets


mfasDa:
+test

mfasDa:
+1

PR #5420, opened 2018-05-11
New tasks with Generic Framework for flow of charged and identified particles

Katka19:
Tasks to calculate flow and other correlations using Generic Framework.


loizides:
+1

rbertens:
hi @Katka19 ,

I think you miss a  library dependency in your CMake instructions, i.e. you have to add OADB to
set(LIBDEPS ANALYSISalice)

Also, could you please make sure that the code is properly indented? The use of white space is rather inconsistent , which makes it hard to read the code. 

Thanks,
Redmer

PR #5426, opened 2018-05-13
Using AliExternalTrackParam::BetheBlochAleph instead of AliMathBase::…

pzhristov:
AliMathBase::BetheBlochAleph is just a wrapper that creates useless dependency between STEERBase and STAT, replacing its call by the original method AliExternalTrackParam::BetheBlochAleph.


pzhristov:
+test

PR #5429, opened 2018-05-14
Modified V0Reader cut string to be the same in all AddTasks (calo, co…

jokonig:
…nv, convcalo). This should avoid problems when using wagons from different groups within the same train.


nschmidtALICE:
+1

PR #5442, opened 2018-05-16
Added kinematic cuts in ME background for Ds

fgrosa:


fprino:
Hi Fabrizio,

thanks!
Before giving +1 I have a question: wasn't it possible to use the methods  AliAODRecoDecayHF3Prong::CosPiKPhiRFrame and AliAODRecoDecayHF3Prong::CosPiDsLabFrame instead of duplicating their code in the task?

Ciao, Francesco

fgrosa:
Hi Francesco,

yes it is possible, the only reason why I re-implemented them is that in the method for the ME TLorentsVector is used instead of AliAODRecoDecay, and therefore I duplicated the methods not to build additional objects and directly use the TLorentzVectors. 

Anyway if you think is more practical to build AliAODRecoDecays I can modify it, no problem!

Thanks!
Ciao,
Fabrizio

fprino:
+1
OK, then let' go on with the duplicated code.

PR #5447, opened 2018-05-16
PWGGA/GammaConv: LHC10 PHOS NonLin adjustments

fjonasALICE:
- added a const. NonLin correction for ConvCalo
- added case for external PHOS NonLin + additional correction to account for ConvCalo


dmuhlhei:
+1

PR #5448, opened 2018-05-16
Remove c++11 keywords from header, fix add macro

mfasDa:
As observed in a different task ROOT5 has problems with
with c++11 keywords default and delete in function
specification even when circumventing for ROOTCINT.
Therefore c++11 part is removed (temporarily).

Furthermore CINT seems to have problems with the macro
interpretation in case the macro doesn't end with an empty
line. Therefore an empty line is added at the end of the
add macro.


mfasDa:
+1

PR #5449, opened 2018-05-17
Update AddTask for PbPb and Chic option

pgonzale83:
New histograms for Chic analysis and updated AddTask for PbPb of the Dalitz anaylsis


dmuhlhei:
+1

PR #5450, opened 2018-05-17
Fixes in spectrum and energy scale task

mfasDa:
- Only handle trigger clusters on data, fall back to ANY
  for MC
- Add cluster energy cuts for cluster container in
  energy scale task in order to fix the jet container
  names
- Add histos with accepted and matched particle level
  jets for efficiency x acceptance
- Energy scale task forced to use analysis utils
- Add some debug printouts and display name of the
  jet containers in generation process


mfasDa:
+1

PR #5454, opened 2018-05-17
Fixes in EMCal cross talk emulation

gconesab:


PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellEmulateCrosstalk.cxx
```diff
@@ -450,31 +464,77 @@ void AliEmcalCorrectionCellEmulateCrosstalk::AddInducedEnergiesToExistingCells()
   Int_t     mclabel = -1;
   Double_t  efrac   = 0.;
   
-  for (Int_t icell = 0; icell < fCaloCells->GetNumberOfCells(); icell++)
+  // Add the induced energy to the cells and copy them into a new temporal container
+  // used in AddInducedEnergiesToNewCells() to refill the default cells list fCaloCells
+  // Create the data member only once. Done here, not sure where to do this properly in the framework.
+  //
+  if ( !fAODCellsTmp )
```
raymondEhlers: Just for future reference (since I didn't notice this before), there is a correction component function `ExecOnce()` which will be executed only once during execution, before `Run()`.

PR #5455, opened 2018-05-17
[EMCAL-209] Trigger correlation in MC

mfasDa:
New task studying the correlation of EMCAL
triggers from offline patches (for MC).


mfasDa:
+1

PR #5457, opened 2018-05-17
Filtering task: load local OCDB at the right time

dberzano:
localOCDBaccessConfig.C should be executed after both an OCDB default storage
and run number are set


dberzano:
@mkrzewic any comment? Loading it in the runFilteringTask macro is not effective since at that point neither a default storage, nor a run number are set.

dberzano:
+1

PR #5458, opened 2018-05-17
force changing the cluster data member energy with the non linearity …

gconesab:
…correction, if requested, not done by default

### DIFF
--- a/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionClusterNonLinearity.cxx
+++ b/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionClusterNonLinearity.cxx
@@ -40,7 +40,7 @@ AliEmcalCorrectionClusterNonLinearity::AliEmcalCorrectionClusterNonLinearity() :
   fEnergyTimeHistBefore(0),
   fEnergyDistAfter(0),
   fEnergyTimeHistAfter(0),
-  fSetClusterE(0)
+  fSetForceClusterE(kFALSE)
 {
 }
 
@@ -63,7 +63,7 @@ Bool_t AliEmcalCorrectionClusterNonLinearity::Initialize()
   GetProperty("nonLinFunct", nonLinFunctStr);
   UInt_t nonLinFunct = fgkNonlinearityFunctionMap.at(nonLinFunctStr);
 
-  GetProperty("setClusterE", fSetClusterE);
+  GetProperty("setForceClusterE", fSetForceClusterE);
   
   // init reco utils
   if (!fRecoUtils)
@@ -131,14 +131,17 @@ Bool_t AliEmcalCorrectionClusterNonLinearity::Run()
         if (fRecoUtils->GetNonLinearityFunction() != AliEMCALRecoUtils::kNoCorrection) {
           Double_t energy = fRecoUtils->CorrectClusterEnergyLinearity(clus);
           clus->SetNonLinCorrEnergy(energy);
-          if ( fSetClusterE ) clus->SetE(energy);
+          if ( fSetForceClusterE ) clus->SetE(energy);
         }
       }
 
       // Fill histograms only if cluster is not exotic, as in ClusterMaker (the clusters are flagged, not removed)
       if (fCreateHisto && !clus->GetIsExotic()) {
-        fEnergyDistAfter->Fill(clus->GetNonLinCorrEnergy());
-        fEnergyTimeHistAfter->Fill(clus->GetNonLinCorrEnergy(), clus->GetTOF());
+        Float_t energy = clus->GetNonLinCorrEnergy();
+        if(fSetForceClusterE) energy = clus->E();
+        
+        fEnergyDistAfter->Fill(energy);
+        fEnergyTimeHistAfter->Fill(energy, clus->GetTOF());
       }
     }
   }
--- a/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionClusterNonLinearity.cxx
+++ b/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionClusterNonLinearity.cxx
@@ -39,8 +39,8 @@ AliEmcalCorrectionClusterNonLinearity::AliEmcalCorrectionClusterNonLinearity() :
   fEnergyDistBefore(0),
   fEnergyTimeHistBefore(0),
   fEnergyDistAfter(0),
-  fEnergyTimeHistAfter(0)
-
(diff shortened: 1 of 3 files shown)


gconesab:
please @raymondEhlers  and @jdmulligan  check if you agree

jdmulligan:
Hi @gconesab ,
What is the reason for this feature? In general I would be inclined to avoid adding special cases like this. I think we have a clear, documented scheme of how to store the cluster energy at various stages of correction, and this may add confusion and bulk (especially over time, when more and more special cases are added). But if it is a big headache for you otherwise, I think we can also be practical and live with it.

gconesab:
Hi @jdmulligan ,
I have expressed several times my discomfort with the current approach. I think the cluster->E() method should return the energy after all corrections applied and all my analysis code is based on this premise. If one wants to recover the original cluster energy, it is quite simple just summing up the energies of the cells attached to the cluster. Now I realised again when comparing my clusterization task and the correction tasks that the correction framework was not doing what I expected. I prefer to add this simple feature for those not relying on the EMCal analysis framework where I assume this is handled properly.
Let me know if you think there is another better way to do it.

jdmulligan:
Hi @gconesab ,

Okay, then I won't object, I have no (non-disruptive) better way. Ideally we would settle on a single convention and document it (I would be fine with either convention), but that's the usual problem all over Aliphysics. To be honest I think it would probably be less confusing if in the current approach we called it e.g. RawE() instead of E(), to avoid such confusion. But go ahead from my side, I think however Raymond has a few minor technical comments.

Best,
James

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionClusterNonLinearity.cxx
```diff
@@ -39,8 +39,8 @@ AliEmcalCorrectionClusterNonLinearity::AliEmcalCorrectionClusterNonLinearity() :
   fEnergyDistBefore(0),
   fEnergyTimeHistBefore(0),
   fEnergyDistAfter(0),
-  fEnergyTimeHistAfter(0)
-
+  fEnergyTimeHistAfter(0),
+  fSetClusterE(0)
```
raymondEhlers: Please init with `kFALSE` so that we don't have to rely on implicit conversions (not that this would cause a problem here, but best to be consistent)

PWG/EMCAL/config/AliEmcalCorrectionConfiguration.yaml
```diff
@@ -149,6 +149,7 @@ ClusterNonLinearity:                                # Cluster non-linearity corr
     enabled: false                                  # Whether to enable the task
     createHistos: false                             # Whether the task should create output histograms
     nonLinFunct: kBeamTestCorrectedv3               # Sets the non-linearity function. Possible options are defined in the header of the cluster non-linearity correction component
+    setClusterE: false                              # Modify the cluster energy data member
```
raymondEhlers: We suggest naming this `setForceClusterE`, and I would suggest making this clear that it is for backwards compatibility and not generally recommended. Perhaps the comment could be:

```
# Only for backwards compatibility, force cluster->E() to be set to the cluster non-linearity corrected energy. Off by default. For the standard methods, see: http://alidoc.cern.ch/AliPhysics/master/READMEcontainers.html#emcalContainerClusterEnergyCorrections
```

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionClusterNonLinearity.h
```diff
@@ -43,6 +43,8 @@ class AliEmcalCorrectionClusterNonLinearity : public AliEmcalCorrectionComponent
   TH1F                  *fEnergyDistAfter;           //!<!energy distribution after
   TH2F                  *fEnergyTimeHistAfter;       //!<!energy/time distribution after
   
+  Bool_t                 fSetClusterE;               ///< Stored it in cluster->E() and not cluster->GetNonLinCorrEnergy()
```
raymondEhlers: Note that this comment isn't quite right, because we still set the cluster non-linearity energy regardless of the setting.

Anyway, in conjunction with the rename in the YAML config, I would suggest that this should be `fForceSetClusterE` and the comment should be along the lines of

```
// Only for backwards compatibility, force cluster->E() to be set to the cluster non-linearity corrected energy. Off by default. For the standard methods, see: http://alidoc.cern.ch/AliPhysics/master/READMEcontainers.html#emcalContainerClusterEnergyCorrections
```

gconesab:
Thanks @raymondEhlers and @jdmulligan 
I have implemented your comments, plus I changed the before the correction histogram, in case of forcing it checks the E() not the other method. If not agreed, I could implement the equivalent histograms. Let me know.

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionClusterNonLinearity.h
```diff
@@ -43,6 +43,8 @@ class AliEmcalCorrectionClusterNonLinearity : public AliEmcalCorrectionComponent
   TH1F                  *fEnergyDistAfter;           //!<!energy distribution after
   TH2F                  *fEnergyTimeHistAfter;       //!<!energy/time distribution after
   
+  Bool_t                 fSetForceClusterE;          ///< // Only for backwards compatibility, force cluster->E() to be set to the cluster non-linearity corrected energy. Off by default. For the standard methods, see: http://alidoc.cern.ch/AliPhysics/master/READMEcontainers.html#emcalContainerClusterEnergyCorrections
```
raymondEhlers: This is super trivial, but this has a double comment marker (`///< //`) - could you please remove the `//`? I know this is trivial, but I'd like to keep things as clean as possible

jdmulligan:
Hi @gconesab ,
Thanks. No objection from me, but aren't these energies exactly the same value at that point?

gconesab:
Thanks for spotting this stupid mistake.

gconesab:
Hi @jdmulligan 
yes, they are, just in case, call me paranoid ...

raymondEhlers:
Thanks! One point here: you actually implemented it in the after hists, which seems fine to me. So there is a typo somewhere, but I'm not sure what you intended

gconesab:
Hi @raymondEhlers 
I want to check after the correction the method I am interested in modifying when I requested the forcing which is E(), I am not sure what typo you mean.

gconesab:
Ah yes, I meant after not before, just my hidden dislexia :)

gconesab:
+1

PR #5460, opened 2018-05-18
[EMCAL-209] Hotfixes in MC trigger correlation task

mfasDa:
- Fix off-diagonal elemets (copy-paste error in label index)
- Fix name of static add function
- Fix typos / names in Add macro


mfasDa:
+1

PR #5461, opened 2018-05-18
remove low E cell and out of time cell selection

gconesab:
before feeding the clusterizer. This selection is already done there. Let those cells still be in the list of cells. In line with what is done in the EMCal correction framework


gconesab:
+1

PR #5464, opened 2018-05-21
removed import statements from addtask macro

CalPolyPat:
Removed import statements from AddTaskPSHFE so as to run on the lego train without errors.


raymondEhlers:
+1

PR #5469, opened 2018-05-21
Fix pool size in correlation task

rhaake:
This commit is to fix a change in behavior that was introduced with commit 4403510 (https://github.com/alisw/AliRoot/commit/4403510cda26490ba19991968de7b3206c84743b#diff-34966647f689d13061df3e7a1dc2b3a3)

With the above mentioned commit, AliEventPoolManager has a built-in event limiting feature.
To restore the original (desired) behavior, the AliEventPoolManager needs to be initialized with a poolsize of -1.


akubera:
+1

PR #5480, opened 2018-05-22
Add MC Particle Information

moliver813:
Adds MC Particle information to Pi0 Candidate mode.
Fixes error where rotational background was also applied to mixed events.
Fixes several compilation warnings.


gconesab:
+1

PR #5490, opened 2018-05-23
[EMCAL-56] Effect of CEMC8EG1 on other EMCALtriggers

mfasDa:
Task checking effect of CEMC8EG1 on other EMCAL triggers
both at L0 and L1, purely looking at trigger correlation
matrix.


mfasDa:
+1

PR #5502, opened 2018-05-24
[EMCAL-56] Handle T0 triggers non-exclusively

mfasDa:
If-else condition made handling of T0 triggers exclusive.
Replaced by three separate if conditions. Fix also the
case that different INT7 and INT8 triggers fire together
by using the new EMCAL trigger string decoder and analyzing
each class separately, not only the enitre trigger string.
First trigger classes are inspected, afterwards they are
assigned to L0 conditions. Code responsible for the trigger
selection by patches has been removed as it is now part of
AliEmcalTriggerDecisionContainer.


mfasDa:
+1

PR #5519, opened 2018-05-25
comment out PID differential histograms (and remove unused histograms)

rihanphys:
comment out PID differential histograms (and remove unused histograms)


jmargutt:
+1

PR #5524, opened 2018-05-25
Update MC information methods

moliver813:
Replace MC SharedGrandMother with MC SharedAncestor approach.
Add Single MC Particle match category.


gconesab:
+1

PR #5534, opened 2018-05-27
move TestCPV from protected to public

dsekihat:
This modification allows to use TestCPV function from external codes.
This will simplify to compute distance between a cluster and a track in mixed event.


chiarazampolli:
Hello,

Why did you change the ClassDef? Is it needed in this case?

Chiara

kharlov:
+1

dsekihat:
Thank you for your comments.
I changed ClassDef 9->8.

peressounko:
+1

PR #5538, opened 2018-05-27
Update default xTalk parametrization 

gconesab:
and add example yaml for data and MC in PWGGA/CaloTrackCorrelations

### DIFF
--- a/PWG/EMCAL/config/AliEmcalCorrectionConfiguration.yaml
+++ b/PWG/EMCAL/config/AliEmcalCorrectionConfiguration.yaml
@@ -66,38 +66,59 @@ CellTimeCalib:                                      # Cell Time Calibration comp
 CellEmulateCrosstalk:                               # Component to emulate crosstalk
     enabled: false                                  # Whether to enable the task
     createHistos: false                             # Whether the task should create output histograms
-    printConfiguration: false                       # Print the configuration after initialization
+    printConfiguration: true                        # Print the configuration after initialization
     conservEnergy: true                             # Activate cluster energy conservation
-    randomizeTCardInducedEnergy: false              # Randomize the energy fraction induced by the TCard
+    randomizeTCardInducedEnergy: true               # Randomize the energy fraction induced by the TCard
     inducedTCardMinimumCellEnergy: 0.01             # Minimum cell energy induced by the TCard
     inducedTCardMaximum: 100                        # Maximum energy induced by the TCard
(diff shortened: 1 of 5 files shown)


gconesab:
+1

PWG/EMCAL/config/AliEmcalCorrectionConfiguration.yaml
```diff
@@ -66,38 +66,59 @@ CellTimeCalib:                                      # Cell Time Calibration comp
...
-    randomizeTCardInducedEnergy: false              # Randomize the energy fraction induced by the TCard
+    randomizeTCardInducedEnergy: true               # Randomize the energy fraction induced by the TCard
     inducedTCardMinimumCellEnergy: 0.01             # Minimum cell energy induced by the TCard
     inducedTCardMaximum: 100                        # Maximum energy induced by the TCard
-    # Note on using the following section:
-    # Each setting can be enabled or disabled. If disabled, the settings will _not_ be applied
-    # The values are formatted as a map, with the keys as either "all", correspoding to all SMs
-    # or the super module number. The format of the value is specified in the comment after the field.
-    #
-    # NOTE: If you specify "all", it will take precedence over any individual SM values, which will be ignored.
```
raymondEhlers: I'd prefer if we could keep this documentation somewhere. I think information such as `NOTE: If you specify "all", it will take precedence over any individual SM values, which will be ignored.` is fairly useful. But perhaps I'm missing something?

gconesab: Hi Raymond,
I think this comment was already there, but yes, this should be better documented. I will see to put it in the comments of the component class. 
Gustavo

raymondEhlers: Hi Gustavo,

To clarify, I mean that I think we should keep this comment here instead of removing it. We can also put it in the component, but I think it will be most useful here since this is where someone will see it when they go to actually configure it. But I'm also open to changes if you have other ideas about how to do it.

Best,
Raymond

gconesab: Ah, I did not realize I removed it, I will put it back
Thanks ... Gustavo

raymondEhlers: Thanks Gustavo!

PR #5540, opened 2018-05-27
implement track matching in mixed event

dsekihat:
implement track matching in mixed event to estimate photon purity


kharlov:
+1

PR #5541, opened 2018-05-28
implement pT-dependent artificial track efficiency feature in jet finding

jdmulligan:
Hi @raymondEhlers, All,

Please take a look at my implementation of pT-dependent artificial track efficiency in jet finding. This builds on the previous implementation of a constant artificial track efficiency, e.g. SetTrackEfficiency(0.98) to account for the ~2% reduced efficiency in PbPb compared to pp. The idea with this commit is that the user can now provide a pT-dependent efficiency via a file containing a TF1 by calling the function LoadTrackEfficiencyFunction() in the AddTask customization. The TF1 should be a parameterization of the "additional" track efficiency you want to apply, typically (PbPb track efficiency / pp track efficiency).

I tested that the feature has no effect if disabled (which is default), and that if a constant efficiency is supplied (either via SetTrackEfficiency() or LoadTrackEfficiencyFunction() ) the behavior is the same as before. I tested also a couple cases of toy TF1s that I supplied, and the behavior is as expected. Further testing is always appreciated.

Please share any comments, I propose to leave the PR open for a day or so and then approve.

Best,
James

### DIFF
--- a/PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
+++ b/PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
@@ -384,6 +384,11 @@ void AliEmcalJetTask::ExecOnce()
   
   // If a constant artificial track efficiency is supplied, create a TF1 that is constant in pT
   if (fTrackEfficiency < 1.) {
+    // If a TF1 was already loaded, throw an error
+    if (fApplyArtificialTrackingEfficiency) {
+      AliError(Form("%s: fTrackEfficiencyFunction was already loaded! Do not apply multiple artificial track efficiencies.", GetName()));
+    }
+    
     fTrackEfficiencyFunction = TF1("trackEfficiencyFunction", "[0]", 0., fMaxBinPt);
     fTrackEfficiencyFunction.SetParameter(0, fTrackEfficiency);
     fApplyArtificialTrackingEfficiency = kTRUE;
@@ -970,27 +975,27 @@ Bool_t AliEmcalJetTask::IsJetInPhos(Double_t eta, Double_t phi, Double_t r)
  * @param path Path to the file containing the TF1
  * @param name Name of the TF1
  */
-void AliEmcalJetTask::LoadTrackEfficiencyFunction(const char* path, const char* name)
+void AliEmcalJetTask::LoadTrackEfficiencyFunction(const std::string & path, const std::string & name)
 {
   TString fname(path);
   if (fname.BeginsWith("alien://")) {
     TGrid::Connect("alien://");
   }
   
-  TFile* file = TFile::Open(path);
+  TFile* file = TFile::Open(path.data());
   
   if (!file || file->IsZombie()) {
-    ::Error("AliEmcalJetTask", "Could not open artificial track efficiency function file");
+    AliErrorStream() << "Could not open artificial track efficiency function file\n";
     return;
   }
   
-  TF1* trackEff = dynamic_cast<TF1*>(file->Get(name));
+  TF1* trackEff = dynamic_cast<TF1*>(file->Get(name.data()));
   
   if (trackEff) {
-    ::Info("AliEmcalJetTask::LoadTrackEfficiencyFunction", "Artificial track efficiency function %s loaded from file %s.", name, path);
+    AliInfoStream() << Form("Artificial track efficiency function %s loaded from file %s.", name.data(), path.data()) << "\n";
   }
   else {
-    ::Error("AliEmcalJetTask::LoadTrackEfficiencyFunction", "Artificial track efficiency function %s not found in file %s.", name, path);
+    AliErrorStream() << Form("Artificial track efficiency function %s not found in file %s.", name.data(), path.data()) << "\n";
     return;
   }
   
--- a/PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
+++ b/PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
@@ -18,6 +18,8 @@
 #include <TClonesArray.h>
 #include <TMath.h>
 #include <TRandom3.h>
+#include <TGrid.h>
+#include <TFile.h>
 
 #include <AliVCluster.h>
 #include <AliVEvent.h>
@@ -71,6 +73,8 @@ AliEmcalJetTask::AliEmcalJetTask() :
   fTrackEfficiency(1.),
   fUtilities(0),
   fTrackEfficiencyOnlyForEmbedding(kFALSE),
+  fTrackEfficiencyFunction(),
+  fApplyArtificialTrackingEfficiency(kFALSE),
   fLocked(0),
   fFillConstituents(kTRUE),
   fJetsName(),
@@ -107,6 +111,8 @@ AliEmcalJetTask::AliEmcalJetTask(const char *name) :
   fTrackEfficiency(1.),
   fUtilities(0),
   fTrackEfficiencyOnlyForEmbedding(kFALSE),
+  fTrackEfficiencyFunction(),
+  fApplyArtificialTrackingEfficiency(kFALSE),
   fLocked(0),
   fFillConstituents(kTRUE),
   fJetsName(),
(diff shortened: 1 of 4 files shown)


PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
```diff
@@ -372,8 +381,19 @@ Bool_t AliEmcalJetTask::GetSortedArray(Int_t indexes[], std::vector<fastjet::Pse
  */
 void AliEmcalJetTask::ExecOnce()
 {
+  
+  // If a constant artificial track efficiency is supplied, create a TF1 that is constant in pT
   if (fTrackEfficiency < 1.) {
-    if (gRandom) delete gRandom;
+    fTrackEfficiencyFunction = TF1("trackEfficiencyFunction", "[0]", 0., fMaxBinPt);
```
raymondEhlers: It looks like it would be possible for a user to set a track efficiency function as well as a `fTrackEfficiency`. Since I think we want to avoid that behavior, can you check if fTrackEfficiencyFunction is already valid (I'm not quite sure the best way to accomplish this) and throw an error if it is?

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
```diff
@@ -944,6 +964,43 @@ Bool_t AliEmcalJetTask::IsJetInPhos(Double_t eta, Double_t phi, Double_t r)
   }
   return kFALSE;
 }
+  
+/**
+ * Load the artificial tracking efficiency TF1 from a file into the member fTrackEfficiencyFunction
+ * @param path Path to the file containing the TF1
+ * @param name Name of the TF1
+ */
+void AliEmcalJetTask::LoadTrackEfficiencyFunction(const char* path, const char* name)
```
raymondEhlers: I would suggest to use `const std::string &` here unless there is a compelling reason not to.

PWGJE/EMCALJetTasks/AliEmcalJetTask.h
```diff
@@ -180,7 +183,9 @@ class AliEmcalJetTask : public AliAnalysisTaskEmcal {
   Double_t               fGhostArea;              ///< ghost area
   Double_t               fTrackEfficiency;        ///< artificial tracking inefficiency (0...1)
   TObjArray             *fUtilities;              ///< jet utilities (gen subtractor, constituent subtractor etc.)
-  Bool_t                 fTrackEfficiencyOnlyForEmbedding; ///<tituent Apply aritificial tracking inefficiency only for embedded tracks
+  Bool_t                 fTrackEfficiencyOnlyForEmbedding; ///< Apply aritificial tracking inefficiency only for embedded tracks
+  TF1                    fTrackEfficiencyFunction;///< Function that describes the artificial tracking efficiency to be applied on top of the nominal tracking efficiency, as a function of track pT
+  Bool_t                 fApplyArtificialTrackingEfficiency; ///< Flag to apply artificial tracking efficiency
```
raymondEhlers: Can you add a bit more documentation somewhere (maybe into doxygen)? For example, the information that you include in the PR seems useful to have somewhere near the code. It doesn't need to be super extensive - I just think it would be good to add something.

I'd suggest we either start adding more extensive documentation into `AliEmcalJetTask` or perhaps just add (labeled) section [here](http://alidoc.cern.ch/AliPhysics/master/READMEjetfw.html)?

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
```diff
@@ -944,6 +964,43 @@ Bool_t AliEmcalJetTask::IsJetInPhos(Double_t eta, Double_t phi, Double_t r)
...
+{
+  TString fname(path);
+  if (fname.BeginsWith("alien://")) {
+    TGrid::Connect("alien://");
+  }
+  
+  TFile* file = TFile::Open(path);
+  
+  if (!file || file->IsZombie()) {
+    ::Error("AliEmcalJetTask", "Could not open artificial track efficiency function file");
```
raymondEhlers: I'd suggest using `AliErrorStream() << "Message\n";` and `AliInfoStream() << "Message\n";` for this function. You won't have to specify the class name, and I think it's better practice to use streams

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
```diff
@@ -944,6 +964,43 @@ Bool_t AliEmcalJetTask::IsJetInPhos(Double_t eta, Double_t phi, Double_t r)
...
+    return;
+  }
+  
+  TF1* trackEff = dynamic_cast<TF1*>(file->Get(name));
+  
+  if (trackEff) {
+    ::Info("AliEmcalJetTask::LoadTrackEfficiencyFunction", "Artificial track efficiency function %s loaded from file %s.", name, path);
```
raymondEhlers: Same for these print outs

raymondEhlers:
Also, the function definition is somewhat different than I was anticipating. Given that we're used a ratio in the past, I was expecting the function to just be the ratio as a function of pt. I guess either would work, but could you fill me in a bit about why you decided to define it that way?

jdmulligan:
Hi @raymondEhlers,

Thanks for the comments, I added the changes.
The function definition that I wrote was indeed a mistake, I edited it now (this of course is immaterial for the commit, it rejects tracks in exactly the same way as the previous feature). 

Best,
James

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
jdmulligan: Done

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
jdmulligan: Done. But really does it matter?

PWGJE/EMCALJetTasks/AliEmcalJetTask.h
jdmulligan: I added to the standalone documentation pages, since there are now two ways to apply this track efficiency it is not very conducive to explaining in the class itself.

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
jdmulligan: Done

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
jdmulligan: Done

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
raymondEhlers: Thanks. Both will work fine, but I would argue that if we're adding new code, we should use the standards when possible to make maintenance easier. `std::string` is the proper c++ string, so I think it is preferable to use it

PWGJE/EMCALJetTasks/AliEmcalJetTask.h
raymondEhlers: Sounds great - thanks!

jdmulligan:
Hi @raymondEhlers ,

Sounds good, thanks!

raymondEhlers:
+1 since there are no additional comments

mfasDa:
Hi @jdmulligan,

I suspect that with this commit you introduce a double delete in the jet finders at least under ROOT5 (not yet checked under ROOT6). Have a look at train tests 1223 to 1225 in Jets_EMC_pp where I just run a simple jet finder even without tracking efficiency simulation. It starts crashing  with the version from May 29th, which is the first tag with this commit, and don't see any other commit related to jet finder or one of its base classes or components it uses. I don't yet have a clue what is the origin of the double delete. Since the logs point to TList I suspect the double delete comes from the TF1 not being pointer which might have some TLists internal. In general I also would stay away from deleting global variables.

Cheers

Markus

jdmulligan:
Hi @mfasDa,
Thanks (see my comment on PR 5584). One question: What are you referring to when you say stay away from deleting global variables?
Best,
James

mfasDa:
Hi @jdmulligan,

you delete gRandom. This might be nasty since you never know who else might be using it. It was one of my two suspicions that might have been responsible for the double delete apart from the TF1 (since the ROOT session might have deleted it as well). For random number generators better make them class members and use it from there instead of relying on gRandom.

Cheers

Markus

raymondEhlers:
Hi @mfasDa ,

You make a good point about `gRandom` - we should fix it. However, I suspect this isn't the cause of the problem (nor really James' fault), since the use of `gRandom` here is quite old.

Best,
Raymond

mfasDa:
Hi @raymondEhlers ,

No, this was not the reason for the crash, the crash was due to the non-pointer TF1 member and is purely a ROOT5 issue - that's why it didn't show up in James' test. Unfortunately we still need for a while to test changes in the framework classes both under ROOT5 and under ROOT6 as there is not yet a ROOT6 build available on the LEGO trains, please don't forget. ROOT5 has quite some pitfalls :( - I had to go through this as well ... . On mac it is probably easiest to use a docker container - I use a cc7 based image which works quite smooth.

For what concerns the gRandom variable: Suppose you use one engine in the jetfinder and some jet analysis replaces this by a different random engine (I think we even have such cases in some jet analyses), you are screwed! Even though it is not directly related to the crash relying on global variables is super dangerous, and even though ROOT is exporting them we are by no means forced to use them ...

Cheers

Markus

raymondEhlers:
Hi @mfasDa ,

Thanks for looking into this - I'm not surprised to hear it's a ROOT 5 issue. As you note, ROOT 5 is a pain to test on macOS, but you're right that it is still important to test, so docker it is. Better yet would be if we could run a test in the CI, but I know this is out of reach right now.

I totally agree with you regarding `gRandom`! I was just pointing out that it was already an existing problem, and so was unlikely to be the direct cause of the issue in this commit. In any case, let's fix it in conjunction with this issue.

Best,
Raymond

PR #5543, opened 2018-05-28
modification to pid for pion and proton

nirbhaygit:
Proton pid reverted to previous version. Some extra pid strategies are added for pion.


rbertens:
+1

PR #5556, opened 2018-05-29
Fixing value of a parameter and some minor changes

mlesch:


rbertens:
+test

PWG/FLOW/Tasks/AliAnalysisTaskStudentsML.cxx
```diff
@@ -35,7 +35,7 @@
 #include <vector>
 #include "TMath.h"
 #include "TF1.h"
-#include "AliMultSelection.h"
```
rbertens: Hi @mlesch , is it on purpose that you remove the header here? I think you do create an instance of AliMultSelection in the task, right ?

mlesch: Hi @rbertens ! 

Yes it’s on purpose that I remove the header. I’ve already included it some lines above but didn’t notice it. So I removed the second one now. 

Cheers,
Marcel

PR #5561, opened 2018-05-30
Updated multiplicity selection calibration of LHC16g/h/i/j

rderradi:
Updated calibration OADBs for periods LHC16g/h/i/j (standard calibration using MB sample).


ddobrigk:
+1

PR #5565, opened 2018-05-30
[EMCAL-109,99] Upload of updated EMC time calibs LHC17m,k

nschmidtALICE:


gconesab:
+1

mfasDa:
Hi @nschmidtALICE ,

one tiny remark: Please always cite the full ticket ID (i.e. EMCAL-99 instead of 99) when referring to a ticket. The bot can only match PRs with tickets when the full ticket ID is available as it doesn't know that 99 refers to a ticket, and to which project the ticket belongs to.

Cheers

Markus

PR #5579, opened 2018-05-31
Fixes in track spectrum task

mfasDa:
- Fix filter bits for global tracks
- Release default dependency on trigger patch container
- Make matched cluster histograms optional


mfasDa:
+1

PR #5584, opened 2018-05-31
Temporarily disable pt-dependent tracking efficiency

mfasDa:
In local test it turned out that the double delete introduced in #5541 is due to the tracking efficiency function. I comment out this part for the moment in order to unblock jet trains, leaving it up to @jdmulligan to find a better solution.

### DIFF
--- a/PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
+++ b/PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
@@ -73,7 +73,7 @@ AliEmcalJetTask::AliEmcalJetTask() :
   fTrackEfficiency(1.),
   fUtilities(0),
   fTrackEfficiencyOnlyForEmbedding(kFALSE),
-  fTrackEfficiencyFunction(),
+  //fTrackEfficiencyFunction(),
   fApplyArtificialTrackingEfficiency(kFALSE),
   fLocked(0),
   fFillConstituents(kTRUE),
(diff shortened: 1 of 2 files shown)


mfasDa:
+test

jdmulligan:
Hi @mfasDa,

Apologies, the behavior was normal when testing on my system. Thanks for the temporary disable. I'll look into it (perhaps not for a few days since I'm traveling).

Best,
James

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
```diff
@@ -255,7 +255,7 @@ Int_t AliEmcalJetTask::FindJets()
       if (fApplyArtificialTrackingEfficiency) {
         if (fTrackEfficiencyOnlyForEmbedding == kFALSE || (fTrackEfficiencyOnlyForEmbedding == kTRUE && tracks->GetIsEmbedding())) {
           
-          Double_t trackEfficiency = fTrackEfficiencyFunction.Eval(it->first.Pt());
+          Double_t trackEfficiency = 1.;//fTrackEfficiencyFunction.Eval(it->first.Pt());
```
raymondEhlers: Can you please set this to `fTrackEfficiency` instead of 1? That way, we can continue to use the previous behavior until James can provide a fix.

mfasDa: Sure - no problem! I will rebase before.

Cheers

Markus

mfasDa:
+test

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
raymondEhlers: Thanks!

mfasDa:
+1

Thanks @raymondEhlers for your comment!

PR #5592, opened 2018-06-01
Change in y-axis size for visualization purpose

strogolo:
Changes in y-axis size of vertex coordinates histograms for visualization purpose


mmasera:
+1

PR #5609, opened 2018-06-04
Update the nonlinearity parametrization for all productions. 

kqf:
Add histogram names and the parameter values to the logs.


kharlov:
+1

PR #5615, opened 2018-06-05
[PWGPP-429] Adapt new hybrid track selection for TRD test in EMCAL fr…

mfasDa:
…amework

Hybrid track selection can be done only on ESDs, therefore
only ESD hybrid track selection is modified. Track cuts
are the same as in the 2011 definition, except
- Number of TPC crossed rows cut reduced to 30
- Cut on the number of TPC + TRD clusters set to at least
  120
Since the cut on the combined number of crossed rows +
TPC number of clusters is not yet available in the
AliESDtrackCuts the cut is implemented as a temporary
hack directly in AliEmcalESDHybridTrackCuts. The cut
is applied for both global and complementary hybrid
tracks. The cut is connected with the flag kDef2018TRD.


mfasDa:
+test

mfasDa:
+test

@aiola Please review!

mfasDa:
+1 

Code preforming well in local tests.

PR #5622, opened 2018-06-05
Fix ROOT 6 build problem

dberzano:


dberzano:
+1

@sebaleh your PR #5618 breaks build on ROOT 6, this commit fixes it. Incidentally, I have noticed that in your PR there's plenty of `std::cout`: you should never use them, use `Ali*` instead (like `AliInfo`). Please provide a fix.

PR #5623, opened 2018-06-05
ALiBSDDijetTask update

kimbyumkyu:
AliBSDDijetTask and its corresponding AddTaskDijetTask file updated,


aohlson:
+test

aohlson:
+1

mfasDa:
Dear @kimbyumkyu ,

there is a problem in the lambda function you use as comparators for jet lorentz vectors with gcc 4.8 which is still among the supported compilers according to the gcc-toolchain build recipe (newer compilers are not affected). The reason is that you use a non-constant reference in the lambda function as second argument, however gcc4.8 uses a const pivot element in its stl quicksort implementation for which the comparator is applied in both directions (so the pivot element can be first or second argument).  gcc4.9 and newer use non-constant pivot elements, therefore no compiler error was raised in the automatic build test. A fix would be to make both references constant in the comparator - as the lambda doesn't modify the arguments this is safe. 

Cheers

Markus

fprino:
Hi,

with the modifications of these commits, the compilation of AliPhysics on my system, was crashing in AliBSDiJetTask  with this error:
/home/prino/AliSoft/sw/SOURCES/AliPhysics/0/0/PWGCF/Correlations/JCORRAN/Pro/AliBSDiJetTask.cxx:510:93: note:   no known conversion for argument 2 from ‘const TLorentzVector’ to ‘TLorentzVector&’

I applied a simple patch, you can see it in this PR: 
https://github.com/alisw/AliPhysics/pull/5707

Best regards, Francesco

PR #5625, opened 2018-06-05
MCStatus with LCA and EMCal/DCal flag

moliver813:
Modifies MC Status information to use a better classification method involving the Lowest Common Ancestor of the two clusters.
Also adds flag to select only main EMCal or only DCal.


efranzis:
+test

efranzis:
+1

fcatalan92:
On Root6 line 2595 of AliAnalysisTaskGammaHadron.cxx doesn't compile. Probably is missing a std::
Can you please fix it ? cc @dberzano

moliver813:
@fcatalan92 that line is "iMinHeight = min(lAncPart_1.size(),lAncPart_2.size());"
Replacing "min" with "std::min" should probably fix it then?  I only just started trying to compile with root6, so I don't know for sure if that will fix the issue, I'm about to submit a pull request that should make this change (and get rid of two compile time warnings)

fcatalan92:
@moliver813  I replaced "min" with "std::min" and I can confirm that compile on root6, at least for me

PR #5632, opened 2018-06-06
add addictional cut on minimum induced energy to avoid strong non linearity

gconesab:
and compress the cell energy induce calculation for better understanding


gconesab:
+1

PR #5646, opened 2018-06-07
Update AddTaskDStarSpectra.C

cbedda:
small modification to the D* AddTask to use the same cut files in different centralities


fprino:
+1

PR #5647, opened 2018-06-07
Allow NULL pointer for MCEvent->GetTrack()/Particle()

shahor02:
In case a MC track for dummy label (3151493) was requested, the
AliMCEvent will return 0 pointer. Protection is added for this case


shahor02:
+1

PR #5652, opened 2018-06-08
PWG/Cocktail: added cocktail feature for wider y range and eta gamma cut

nschmidtALICE:
Added the possibility to generate the cocktail particles in user defined rapidity range (default is still |y|<1)
Added the possibility to cut differently on photons and hadrons. For this, an eta cut has been added which is applied when it is set via the last argument of the AddTask. 
For example, AddTask_GammaCocktailMC(kFALSE,"80_90"); will apply a cut of  |y|<0.8 on hadrons and |eta|<0.9 on photons. This feature is now implemented for GammaCocktail and HadronicCocktail.
If the same cut on photons and hadrons should be used, then for example "80_80" and "80" would result in the same output.


FriederikeBock:
+1

PR #5653, opened 2018-06-08
PWGGA/GammaConv introducing EOverP cut in TM

hpoppenb:
new: CaloMerged task gives JJ weights & mcEvent to MatchTracksToCluster
     introducing EOverP veto in track matching, so that jets (charged tracks with large E/p) are not matched away
     new histos for data & mc:
         - TH2 trackP vs clusterE for track matches
         - TH2 trackP vs clusterE for track matches after EOverP veto
     for mc only:
         - TH1 true pi0 cluster spectrum before/after track matching
         - TH2 trackP vs clusterE for track matches in case of cluster = true pi0 cluster
     TM cuts ("c",...,"h") for different values of allowed maxEOverP(999.9, 3., 2., 1.75, 1.5, 1.25); took dEta,dPhi cuts from default case "7"
     mEMC addtask: took configs "17x" for EOverP cut studies in 8TeV pp
modified: only one eta-phi map of electron clusters before/after TM, instead of three and now only for high pt, for less memory consumption
          also, dont fill extQA(1,3,5) histograms and dont do MatchTracksToClusters if clusterE < 10 GeV AND fIsPureCalo>0 (= merged analysis)
          omit booking of three histograms if fIsPureCalo != 0, because they are only used in conjunction with PCM

### DIFF
--- a/PWGGA/GammaConv/AliCaloPhotonCuts.h
+++ b/PWGGA/GammaConv/AliCaloPhotonCuts.h
@@ -296,7 +296,7 @@ class AliCaloPhotonCuts : public AliAnalysisCuts {
     Bool_t      ClusterQualityCuts(AliVCluster* cluster,AliVEvent *event, AliMCEvent *mcEvent, Int_t isMC, Double_t weight, Long_t clusterID);
 
     Bool_t      MatchConvPhotonToCluster(AliAODConversionPhoton* convPhoton, AliVCluster* cluster, AliVEvent* event, Double_t weight=1.);
-    void        MatchTracksToClusters(AliVEvent* event, Double_t weight=1., Bool_t isEMCalOnly = kTRUE);
+    void        MatchTracksToClusters(AliVEvent* event, Double_t weight=1., Bool_t isEMCalOnly = kTRUE, AliMCEvent *mcEvent = 0x0);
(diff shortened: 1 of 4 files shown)


FriederikeBock:
+1

PWGGA/GammaConv/AliCaloPhotonCuts.h
```diff
@@ -563,6 +562,14 @@ class AliCaloPhotonCuts : public AliAnalysisCuts {
...
     TH2F*     fHistClusterEvsTrackSumEPrimaryButNoElec; //
 
+    TH1F*     fHistClusETruePi0_BeforeTM;               // for checking the false positives: how much true pi0s are matched away?
+    TH1F*     fHistClusETruePi0_Matched;                //
+
+    // histograms for studying track matching veto with track momentum vs. cluster energy
+    TH2F*     fHistMatchedTrackPClusE;                // track P vs cluster E in case of matching with a cluster
+    TH2F*     fHistMatchedTrackPClusEAfterEOverPVeto; // track P vs cluster E for matched tracks surviving the E/P veto
+    TH2F*     fHistMatchedTrackPClusETruePi0Clus;     // track P vs cluster E in case of matching with a true pi0 cluster
+
```
dmuhlhei: ClassDef increment is missing in this header, can you please fix it

PR #5672, opened 2018-06-10
Reverse order of retrieving values from config

raymondEhlers:
If values are retrieved in the order that the configuration files are
added, then it is impossible to override a value after a file has been
added. Instead, we now look in reverse order of the way they were added,
so a value can be overridden just be adding a new config file.

It's been tested and validated with some unit tests and the EMCal Correction Framework.


raymondEhlers:
+1 (For the EMCal part, but I can't approve the part in PWG/Tools)

mfasDa:
+1

PR #5680, opened 2018-06-11
fix Lc MC ID

jaimenorman:
Function used to ID Lc->pKpi was missing a small fraction of Lc when running over MC. Changed to be compatible with correct implementation in CF task and TMVA task.


fprino:
+1

PR #5685, opened 2018-06-11
New multiplicity selection calibration OADB for LHC16f

rderradi:
Standard calibration OADB produced using MB sample


chiarazampolli:
+1

PR #5727, opened 2018-06-14
PWGGA/GammaConv: further fixes for pT binning for MC histograms in GammaConv task

nschmidtALICE:
QA binning and standard pT binning was mixed in some places


nschmidtALICE:
+test

PR #5733, opened 2018-06-14
PWGPP/TRDPID: AliTRDPIDTree: add further pile-up cut

kagarner:
Test pile-up cut taken from AliAnalysisTaskBFPsi.cxx on correlation of multiplicity of VZERO detector and the number of TPCout tracks.


tdietel:
+1

PR #5735, opened 2018-06-14
Fixed a compilation error related to Fastjet.

osaarima:
A compilation error occurs when building with root5 but without the Fastjet package. This fix solves the problem.


aohlson:
+test

aohlson:
+1

PR #5750, opened 2018-06-15
AliConvEventCuts: modify pileup cuts and spd plot

m-c-danisch:
add possibility for PbPb to apply V0M-TPC pileup cut without applying pp pileup cuts using SPD
change range of spd plot according to values in AliAnalysisTaskGammaConvV1


FriederikeBock:
+1

PR #5753, opened 2018-06-15
introduce new cuts to restrict or open the induced energy leaks

gconesab:
set as default the limit on the induced leak at 50 MeV, like the default clusterization cut.
If one uses higher clusterization cuts, it has to be changed.

The calculation of the induced energy has been simplified, all is done by a single method for each of the correlated cells instead of replicating the lines for each correlated cell.

Update also the yaml files for the new cuts. In the PWGGA/CaloTrackCorrelations/yaml directory 3 yaml files with the cases: same as before, with the limit at the clusterization cell E cut, and adding the low energy leaks. For the particular case of V1 clusters and 100 MeV cell energy cut.


gconesab:
+1

PR #5754, opened 2018-06-15
Fix ROOT5 issue with pT-depedent artificial tracking efficiency

jdmulligan:
See PR #5541 and #5584. This commit fixes the ROOT5 issue by switching the TF1 back to using pointers. It was tested in both ROOT5 and ROOT6, and verified to agree in all cases. As before the feature is off by default.

The use of gRandom was also discontinued here, as discussed in 5541.


mfasDa:
+test

mfasDa:
+1

PR #5755, opened 2018-06-15
kMuonCalo added. CALO triggers moved to kMuonCalo

ekryshen:
Also the structure of the browseAndFill macro rearranged


ekryshen:
+1

PR #5758, opened 2018-06-17
Added configuration to CF for D0, D+ and D*+ ESE analysis

fgrosa:
Added configuration to CF for D0, D+ and D*+ for the ESE analysis, with pt, y, centrality, multiplicity, local multiplicity and q2 axes. The local multiplicity is defined as the number of tracks passing FilterBit 8||9 with DeltaEta<0.1 and DeltaPhi<0.1 wrt the D meson. The q2 is computed using TPC tracks with 0.2<pt<5 GeV/c, |eta|<0.8 and FilterBit 8||9.


fprino:
+1

PR #5776, opened 2018-06-19
Dalitz: Add cuts for R scan with low B field

pgonzale83:
Add cuts for low B field analysis in pp collisions at 13 TeV with Dalitz


dmuhlhei:
+1

PR #5784, opened 2018-06-19
Added features to CF with kESE configuration (D0, D+ and D*+)

fgrosa:
1) only HIJING tracks are used to compute q2 (to avoid autocorrelations with D-meson daughters)
2) local multiplicity definition changed to Ntracks in R<0.4 wrt the D meson
3) slightly modified axis dimensions


fprino:
+1

PR #5788, opened 2018-06-19
Added constituent cut for jets and macro arguments.

osaarima:
In addition I streamlined the different jets and dijets in a single table.


aohlson:
+test

aohlson:
+1

PR #5803, opened 2018-06-21
MC Closure Test + Centrality Binned CF

bhohlweger:
Centrality binning for the SE and ME Distributions to obtain the CF in Cent bins, pT, Phi and Eta resolution, and minor bugfixes


lgraczykCern:
+1

PR #5808, opened 2018-06-22
Added bool to turn on/off filling of sparse with eID variables.

efgauger:


sashingo:
+1

efgauger:
Error caused by a typo: meant to have a comma instead of a period. Should be fixed with my latest commit.

sashingo:
+1

PR #5825, opened 2018-06-22
MC and Embedding functionality added, for inclusive and recoil response

jaimenorman:
MC and embedding functionality. Running with the embedding framework allows to match detector level embedded and true jets, and fill response (histograms and trees) for both recoil and inclusive jets. Aim is to 2d unfold the pT/DeltaPhi distribution

### DIFF
--- a/PWGJE/EMCALJetTasks/macros/AddTaskJetCoreEmcal.C
+++ b/PWGJE/EMCALJetTasks/macros/AddTaskJetCoreEmcal.C
@@ -1,5 +1,7 @@
 AliAnalysisTaskJetCoreEmcal* AddTaskJetCoreEmcal(
 								 const char *njetsBase,
+								 const char *njetsTrue,
+								 const char *njetsPart,
 						     const Double_t R,
 								 const char *type,
 								 Int_t jetShapeType = AliAnalysisTaskJetCoreEmcal::kData,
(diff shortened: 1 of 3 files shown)


PWGJE/EMCALJetTasks/macros/AddTaskJetCoreEmcal.C
```diff
@@ -31,32 +33,69 @@ AliAnalysisTaskJetCoreEmcal* AddTaskJetCoreEmcal(
...
+	AliParticleContainer *trackContPartLevel = 0x0;
+	AliParticleContainer *trackContTrueLevel = 0x0;
+
+	if(jetShapeType == AliAnalysisTaskJetCoreEmcal::kMCTrue) {
+		trackCont = task->AddMCParticleContainer("mcparticles");
+	}
+	else if (jetShapeType == AliAnalysisTaskJetCoreEmcal::kData) {
+		trackCont = task->AddTrackContainer("tracks");
+	}
+	else if (jetShapeType == AliAnalysisTaskJetCoreEmcal::kDetEmbPart) {
```
raymondEhlers: If you're intending to use this with the new embedding framework, you'll need to `SetIsEmbedding(true)` on at least some of these containers

jaimenorman: I was doing this in my tests by getting the relevant particle container from the task after definition, then calling SetIsEmbedding (I guess this could be done in the macro customisation as well) - would this be ok?

raymondEhlers: Yeah, that's totally fine - I only briefly glanced at your code and didn't see it around where you were doing your configuration, so I just wanted to make sure that it wasn't lost!

jaimenorman: ok all good, thanks for the thought!

raymondEhlers:
+test

raymondEhlers:
+1

PR #5857, opened 2018-06-25
Fixed a compatibility problem with root5 in macro.

osaarima:
Giving a vector as an argument had troubles with root5 so I changed the argument to a TString which is then parsed to a vector. Also added checks for the centrality bin vector so that it is properly set.


aohlson:
+1

PR #5858, opened 2018-06-25
Fixed binning, load AOD using AODEvent().

vkucera:
Fixed number of bins in fh2QAV0NClRows.
Added loading AOD using AODEvent() in case InputEvent() fails.


vkucera:
+1

PR #5861, opened 2018-06-25
PWGDQ: added histo with gen+smeared from rec tracks

carstooon:
Adds histogram to compare reconstructed pT spectra with generated+smeared spectra


rbailhac:
+1

PR #5875, opened 2018-06-26
QATasks for Pi/K/p identification in the rTPC

omvazque:
Task used for calibrating the TPC signal Vs. pseudorapidity. Also in the output, the needed histograms to do Pi/K/p identification in the relativistic rise of the TPC is stored.


bdoenigu:
+1

PR #5879, opened 2018-06-26
Add support for virtual trigger selection

mfasDa:
Tasks inheriting from AliAnalysisTaskEmcal can
reimplement the function IsTriggerSelected to have a
custom trigger selection applied during the event
selection. The default implementation of the function
is identical to the previous trigger selection part in the
IsEventSelected function in order to guarantee backward
compatibility. Custom trigger selection is in particular
necessary for the EMCAL triggers when calculating the
vertex efficiency.

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcal.cxx
+++ b/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcal.cxx
@@ -1095,7 +1095,7 @@ Bool_t AliAnalysisTaskEmcal::IsEventSelected()
     } else {
       const AliAODEvent *aev = dynamic_cast<const AliAODEvent*>(InputEvent());
       if (aev) {
-	 res = ((AliVAODHeader*)aev->GetHeader())->GetOfflineTrigger();
+	      res = ((AliVAODHeader*)aev->GetHeader())->GetOfflineTrigger();
       }
     }
     if ((res & fOffTrigger) == 0) {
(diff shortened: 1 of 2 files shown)


mfasDa:
@lcunquei, @raymondEhlers, @aiola Please review! This is a proposed change to the framework allowing the users to implement a custom trigger selection which is evaluated during the method IsEventSelected. The default implementation of IsEventSelected is the current trigger selection part from IsEventSelected in order to ensure backward compatibility. Custom implementation of the trigger selection within the function IsEventSelected is in particular of relevance for the EMCAL triggers which are not completely handled by the physics selection bits in order to calculate the vertex finding efficiency for the bin0 correction. The proposed changes are not expected to have influence on existing code.

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcal.cxx
```diff
@@ -1294,6 +1235,57 @@ Bool_t AliAnalysisTaskEmcal::IsEventSelected()
...
+  // to trigger selection). Users should re-implement
+  // this function in case they have certain needs, in
+  // particular for EMCAL triggers
+  if (!fTrigClass.IsNull()) {
+    TString fired = InputEvent()->GetFiredTriggerClasses();
+    if (!fired.Contains("-B-")) return kFALSE;
+
+    std::unique_ptr<TObjArray> arr(fTrigClass.Tokenize("|"));
+    if (!arr) return kFALSE;
+    Bool_t match = 0;
```
raymondEhlers: I know you're just moving around the code, but since we're looking at this, can we change the values that we are setting in this bool to `true` and `false` (as appropriate)? I think it's preferable to use the values associated with the type.

mfasDa: Sure, thanks for spotting. I anyhow wanted to fix this but then forgot about this. I will add it to the commit.

mfasDa:
+1

PR #5883, opened 2018-06-27
Implement new virtual trigger selection function

mfasDa:
Implement virtual trigger selection function introduced
in #5879 for AliAnalysisTaskEmcalJetEnergySpectrum
and AliAnalysisTaskEmcalJetSubstructureTree. This is
basically moving code from Run/TriggerSelection to the new function.


mfasDa:
+1

PR #5887, opened 2018-06-27
Minor update to task, also new cent & zvtx histo.

osaarima:
Made CalculateJetsDijets function so that it is easier to use as a stand
alone function. In addition, centrality and z-vertex histograms were
added.


aohlson:
+1

PR #5893, opened 2018-06-28
JetPerformance: Let min pT in jet histograms be configurable

jdmulligan:
Update to my user task.

The purpose is to allow the jet pT to go negative due to background fluctuation, since we may want to consider these jets in the unfolding.


raymondEhlers:
+1

PR #5904, opened 2018-06-28
[PWGPP-429] Prepare PWGJE tracking QA task for subwagons

mfasDa:
This is in particular of relevance for running the tracking
QA task on EMCAL triggered events where one wants
to study the 4 EMCAL triggers in parallel.


mfasDa:
+1

Small change which is needed to run the task as subwagons. This is in particular needed for the EMCAL triggers in order to run all 4 EMCAL triggers in parallel. No functional change in the code

PR #5906, opened 2018-06-28
Make PIDqa macro compatible with ROOT6

lbariogl:
Using ROOT6, the macro MakeTrendingPIDQA.C gave some errors related to the output tree:
`Error in <TBranch::TBranch::WriteBasketImpl>: basket's WriteBuffer failed`.
Creating the output file before the output tree solved the problem.


shahor02:
+1

PR #5908, opened 2018-06-28
Allow checking the exact trigger string

mfasDa:
If EXACT is found in the trigger string then the full trigger
string must be equal to the requested trigger string, and
not only a substring (find). This is in particular necessry
to reject events which triggers only CENTNOTRD and
not CENT.

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalLight.cxx
+++ b/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalLight.cxx
@@ -890,8 +890,16 @@ Bool_t AliAnalysisTaskEmcalLight::IsEventSelected()
   Bool_t acceptedTrgClassFound = kFALSE;
   if (fAcceptedTriggerClasses.size() > 0) {
     for (auto acc_trg : fAcceptedTriggerClasses) {
+      std::string teststring(acc_trg);
+      bool fullmatch(false);
+      auto posexact = acc_trg.find("EXACT");
+      if(posexact != std::string::npos) {
+        fullmatch = true;
+        teststring.erase(posexact, 5);
+      }
       for (auto fired_trg : fFiredTriggerClasses) {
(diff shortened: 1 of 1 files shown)


mfasDa:
Hi @aiola ,

I need to make a change in the AliAnalysisTaskEmcalLight in the trigger selection: For the TRD reconstruction test reconstruction we need to select events which triggered the cluster CENT as in CENTNOTRD the TRD was not read out. This creates troubles for the high-threshold triggers as CENT is a subset of CENTNOTRD. With the current implementation this ends in the following cases:
- If I use the full trigger string (i.e. "CEMC7EG1-B-NOPF-CENT") in the accepted trigger class this will select both clusters CENT and CENTNOTRD (due to std::string::find) since CENT is a substring of CENTNOTRD
- If I add "NOTRD" to the rejected trigger classes this will reject all EMCAL triggered events since all CENT triggering events also trigger CENTNOTRD.
In order to overcome this a method where the exclusive match for the full trigger class is possible is needed. In order to not add a struct where in addition to the trigger string a flag for exact matching is passed, which might add streaming issues under ROOT5, I prefer to add a key "EXACT" to the accepted / rejected trigger (for the comparison a temporary object where the key is erased is used). This key decides on whether I use operator== or std::string::find for string matching.

Are you fine with this change?

Cheers

Markus

aiola:
Hi Markus,
sounds like a good solution to me. Please go ahead.
Best,
Salvatore

mfasDa:
+test

Hi @aiola,

Thanks for the quick reply. In the mean time I recalled there would be also a second option, using TRegExp, which allows maybe a more general solution. On the other hand TRegExp needs a special treatment for wildcards. Let me first merge this, and one can think about TRegExp at a second stage.

Cheers

Markus

mfasDa:
+1

mfasDa:
Selection with the "EXACT" tag working in test to select events from a certain trigger cluster - PR merged.

PR #5909, opened 2018-06-28
PDGCode for Particle is always positive, Xi minus now is properly set…

bhohlweger:
… as 3312 and the Xi plus as -3312. Additionally MC Protection for not fully set particles to not access the whole information.


akubera:
+test

akubera:
+1

PR #5916, opened 2018-06-29
Bringing the fitting and plotting macros up to date

mkofarag:
The fitting and plotting macros are changed to represent the version used for the publication of the 2.76 TeV data and the preliminary status of the 5.02 TeV data.


aohlson:
+1

PR #5921, opened 2018-06-29
Changes/updates to QATask for the rTPC analysis.

omvazque:
I also modified a line on PWGLF/SPECTRA/CMakeLists.txt  to run on the LEGO train


bdoenigu:
+1

PR #5934, opened 2018-07-01
WIP Allow to run tests without having a preset env or with SIP enabled (Mac)

aphecetche:
DO NOT MERGE (yet) - using the PR check to x-check my mods on linux...

Should probably be discussed first, but the idea here is to be able to get : 

cd  sw/BUILD/AliPhysics-.../AliPhysics
ctest 

to succeed even if there's no env. setup whatsover. Works also with SIP enabled.

This uncovered a few underspecified lib dependencies (mostly on fastjet)


dberzano:
@aphecetche thanks for this PR, I am wondering if some boilerplate can be reduced by creating something like an `ali_add_test` macro/function?

aphecetche:
@dberzano yes, that should be possible. But can we already +test this one first ?

dberzano:
+test I didn't start tests because I saw WIP in the title 😄

ktf:
Is this still relevant? If not, can we close it? Thanks.

aphecetche:
The idea is still relevant but I don't have the stamina to apply it to Run2 stuff, so I'm closing indeed.
Might try again for O2 though.

PR #5950, opened 2018-07-03
Fix in TOF QA trending

njacazio:
- Fixed the MakeTrendingTOFQAv2 to avoid using histograms after closing the file on which they are based
- Formatted whole MakeTrendingTOFQAv2
- Updated TOFqa script to remove .so and .d files after the trending is done


fbellini:
+1

PR #5962, opened 2018-07-03
Update AliAnalysisTaskNetLambdaTrad.cxx changes to cuts and pt binning

eumaka:


aohlson:
+1

eumaka:
Does  any know why the test failed on some part? I don't see that the updated code is now on AliPhysics.

PR #5971, opened 2018-07-05
distributions for mom. and z resolution+fix for refl. bug

btrzecia:
Distributions for momentum and z resolutions added + fix for a bug in the reflection part.


aiola:
+1

PR #5974, opened 2018-07-05
Fix empty pointer PrimaryVertex

lvermunt:
Pointer was empty with cand->GetOwnPrimaryVtx(), so didn't pass if statement. Now getting primary vertex from AOD


arossi81:
+1

PR #6040, opened 2018-07-11
Modify the status of the output histograms.

cimordas:
* Put all the pointers to the output histograms from // to //!


rbertens:
+1

PR #6043, opened 2018-07-11
Test for new mass cut, Dalitz 5TeV2017

eddoming:
A new cut in the invariant mass for Dalitz in the different range of pT was added.


dmuhlhei:
+1

PR #6046, opened 2018-07-11
This is for debugging and might not conform to best c++ standards

fibnumb:
This is for debugging and might not conform to best c++ standards, my code is crashing on the lego train yet runs fine locally with Root5 and Root6.  The changes made in this PR will only be temporary


raymondEhlers:
It also works fine on a local test train?

fibnumb:
Yes. I think it is some kind of memory leak that I’m not hitting by using a local machine.

raymondEhlers:
+test

raymondEhlers:
Humm, I suspect that should be quite rare, but it's your task, so this commit is fine by me

raymondEhlers:
+1

PR #6060, opened 2018-07-12
Update syst Dstar pp@5TeV

lvermunt:
Update of the QM18 preliminary systematics for Dstar analysis in pp@5TeV, discussed in D2H 10/07


fcolamar:
+1

PR #6064, opened 2018-07-12
Added new multiplicity calibration OADBs for LHC18b/d/e

rderradi:
LHC18b: standard calibration based on MB sample
LHC18d/e: stitched OADBs from MB and VHM samples


ddobrigk:
+1

PR #6102, opened 2018-07-17
Added more macro arguments and a couple of histos.

osaarima:
Added macro arguments for testing different background subtraction
settings. Histograms for kt-jets and raw constituent cut jets also
added.


aohlson:
+1

PR #6108, opened 2018-07-18
PWGHF: Updated strategy for tracking eff systematic for Dstar

lvermunt:
As discussed in the D2H of 10/07/18, the implementation of the new strategies to estimate the tracking efficiency systematic for Dstar.

1) ITS-TPC matching: Deleted the contribution from the soft pion, as it is not prolonged from TPC to ITS.
2) Track quality: Added a flag to be able to estimate the effect of TPC track quality variations separately for the soft pion and the D0 daughters.


fprino:
+1

PR #6112, opened 2018-07-18
added option of rejection factor when filling inclusive jets from tree

jaimenorman:
plus fix Dphi error message


raymondEhlers:
+1

PR #6149, opened 2018-07-21
JetPerformance: allow to use jettagger for geometrical matching

jdmulligan:
(without mc fraction requirement)

Update to my user task.

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
@@ -94,8 +94,6 @@ AliAnalysisTaskEmcalJetPerformance::AliAnalysisTaskEmcalJetPerformance() :
   fMedianDCal(0.),
   fkEMCEJE(kFALSE),
   fEmbeddingQA(),
-  fMinSharedMomentumFraction(0.5),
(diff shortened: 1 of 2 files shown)


raymondEhlers:
+1 I'm sure that you already know this, but just as a reminder, I don't think we've sorted out the grid issues when embedding (although Jamie's trains have been successful so far, so maybe it's okay?)

jdmulligan:
Hi @raymondEhlers,

Yea, I'm going to run a since train to try it again -- this time the configuration will be a bit different, since I won't use MC fraction matching. The whole reason I make this change is that I realized the response maker actually doesn't apply the jet container cuts when doing the matching...which can create some problems, particularly when using embedding.

raymondEhlers:
Ah, yeah, that's right. I guess this means it did the matching without the leading track bias, right?

jdmulligan:
Yea -- I noticed some small changes in the efficiency after changing the minPt thresholds on the jet finders, which led me to dig back into this...the design is pretty confusing, the ResponseMaker AddTask takes all this container information like the leading track bias, but doesn't actually use it for the matching...

raymondEhlers:
That's a good catch, and not an ideal design for the response maker. I guess we should just recommend the tagger if we can get it working consistently with embedding. In any case, I guess I'll need to re-run my RM too, so I'll eagerly await the results of your tests!

PR #6166, opened 2018-07-23
Hadron contamination functions for systematics in 30-50% and 60-80% added, HFE Pb-Pb 5TeV (mfaggin, 23-Jul-2018)

mfaggin:


sashingo:
+1

PR #6183, opened 2018-07-24
QATask for 4TPC analysis

omvazque:
1.- Contains the updated calibration parameters obtained from using all the HIR runs.
2.- PID histograms for the low pT are included
3.- The MC part is also included


njacazio:
+1

PR #6192, opened 2018-07-24
Updated AliAnalysisTaskSELc2V0bachelor.cxx, in order to have a tree with smaller size in aoutput

emeninno:
…chelorNew.cxx

Modified analysis task, in order to reduce the size of the tree created.


fprino:
Hi Elisa,

do you want to add a new cxx file, AliAnalysisTaskSELc2V0bachelorNew.cxx?
Note that this task would not be compiled, since there is no header file and it is not included in the CMakeLists. Therefore, the new features that you added to have a reduced tree, would not be usable on the grid.
Cheers, Francesco

PR #6193, opened 2018-07-24
introducing usage of run dependent mean multiplicity

afestant:
Introducing per run estimate of the mean multiplicity used in low multiplicity pp collisions to extract the luminous region parameters from the fits to the vertex coordinates distributions (x and y).


shahor02:
+1

PR #6224, opened 2018-07-26
Add possibility to accept declared as warm/dead/hot channels 

gconesab:
For calibration analysis it is important to try to understand if some channels declared as bad can be recovered, in particular those declared as "warm". 

With this commit, I changed the method AliEMCALRecoUtils::GetEMCALChannelStatus() that so far returned the status and in most of the code where it is used, if it was non 0 (good channel) this channel was rejected. Now instead, it returns true or false depending on internal conditions to select or not a bad channel depending its status but also the status is recoverable as one of the parameters.

The EMCal correction framework component dealing with bad channels is updated to consider this possibility, in the yaml file one should add the property

      acceptWarm: true

in the part related to bad channels.

In other frames or codes, one should recover the pointer to AliEMCALRecoUtils and do
fRecoUtils->SetWarmChannelAsGood() 

Similar actions for hot or dead are implement, even if it is unlikely its use.

Do not approve immediately in case of comments. I will do it tomorrow.

### DIFF
--- a/PWG/CaloTrackCorrBase/AliCalorimeterUtils.cxx
+++ b/PWG/CaloTrackCorrBase/AliCalorimeterUtils.cxx
@@ -1892,10 +1892,10 @@ Bool_t AliCalorimeterUtils::IsMCParticleInCalorimeterAcceptance(Int_t calo, TPar
     Bool_t ok = GetEMCALGeometry()->GetAbsCellIdFromEtaPhi(particle->Eta(),particle->Phi(),absID);
     if(ok)
     {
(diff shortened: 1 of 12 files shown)


raymondEhlers:
Correction framework looks good to me!

FriederikeBock:
PCM part looks also fine, thanks Gustavo for adapting this!

FriederikeBock:
+1

gconesab:
Let's approve this
+1

gconesab:
+1

PR #6244, opened 2018-07-27
Added jet extractor task based on flat trees

rhaake:
Implementation of jet extractor task and wrapper class for flat trees for jet properties.


rhaake:
+1

PR #6253, opened 2018-07-28
JetPerformance: Add some jet matching histos, and reduce output lists…

jdmulligan:
… to allow running more subwagons

(update to my user task)

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
@@ -97,6 +97,7 @@ AliAnalysisTaskEmcalJetPerformance::AliAnalysisTaskEmcalJetPerformance() :
   fMCJetContainer(nullptr),
   fDetJetContainer(nullptr),
   fJetMatchingR(0.),
+  fPlotJetMatchCandThresh(1.),
   fUseAliEventCuts(kTRUE),
   fEventCuts(0),
   fEventCutList(0),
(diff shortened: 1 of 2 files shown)


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
```diff
@@ -2493,12 +2532,24 @@ AliEmcalJet* AliAnalysisTaskEmcalJetPerformance::GetMatchedJet(AliEmcalJet* detJ
...
     return 0;
   }
   if (matchedTruthJetCand->ClosestJet() != detJet) {
     return 0;
   }
+  
+  // Fill matching distance between combined jet and pp det-level jet, before imposing deltaR cut (centrality, pT-truth, R)
+  TString histname = "MatchedJetHistograms/hMatchingDistance";
+  if (fForceBeamType != kpp) {
+    fHistManager.FillTH3(histname, fCent, matchedTruthJetCand->Pt(), detJet->ClosestJetDistance());
```
raymondEhlers: Be really careful with this. This is the kind of thing that I could imagine would cause the TH3 fill error that you saw in embedding

raymondEhlers:
+1

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
jdmulligan: Sure but I thought we both agreed that the tagger problem can't possibly have anything to do with my task? I will keep an eye for the error showing up elsewhere, but I don't see any reason to suspect FillTH3 misbehaves, I have been using it extensively for a long time.

raymondEhlers: Yeah, you're right, my bad - it struck me at first glance, but you're right that it shouldn't matter

jdmulligan: Ok, cool -- thanks for the review :)

PR #6265, opened 2018-07-30
 Added alternative parametrisation of double gaussian function for signal 

fgrosa:
Added k2GausSigmaRatioPar option for signal function (2 gaussian function that has the ratio of the two gaussian sigmas as a parameter, instead of the 2nd sigma)


fprino:
+1

PR #6272, opened 2018-07-31
Added a new run info histogram.

osaarima:
New histogram saves information about the run so that the settings used
can be checked easily later. Also changed fh_events histogram to have
alphanumeric bin labels.


aohlson:
+1

PR #6277, opened 2018-07-31
Added HF-jet options for jet extractor

rhaake:
Added possibility to extract secondary vertices/impact parameters correctly. Eventually, this will replace the JetExtractorHF task.


rhaake:
+1

PR #6307, opened 2018-08-02
JetPerformance: Add option to plot a couple trigger response histos in MC

jdmulligan:
(update to my user task)

@mfasDa I would appreciate if you take a look

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
@@ -93,6 +93,7 @@ AliAnalysisTaskEmcalJetPerformance::AliAnalysisTaskEmcalJetPerformance() :
   fMedianEMCal(0.),
   fMedianDCal(0.),
   fkEMCEJE(kFALSE),
+  fDoTriggerResponse(kFALSE),
   fEmbeddingQA(),
   fMCJetContainer(nullptr),
   fDetJetContainer(nullptr),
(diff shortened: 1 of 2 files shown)


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
```diff
@@ -1301,6 +1322,70 @@ void AliAnalysisTaskEmcalJetPerformance::DoTriggerSimulation()
...
+    AliTLorentzVector part;
+    Double_t partEta;
+    Double_t partPhi;
+    Double_t partE;
+    for (auto mcpart:fGeneratorLevel->accepted_momentum()) {
+      part.Clear();
+      part = mcpart.first;
+      partEta = part.Eta();
+      partPhi = part.Phi_0_2pi();
+      partE = part.E();
```
mfasDa: From what I understand you sum up the energy of all particles including hadrons. Hadrons however don't deposit the full energy in the EMCAL. You either have to apply an effective hadronic correction similar to the correction applied in the hadronic correction component or you take the parameterized response from Mateusz. Only for electrons and photons you should assume full energy deposit in the EMCAL. Note this is only relevant at generator level, at cluster level the detector response has already been applied.

mfasDa:
Hi @jdmulligan ,

I have just one comment which is that at generator level the energy has to be corrected for everything else than photons and electrons, otherwise the patch energy will be significantly overestimated as normally it only contains the neutral energy part of the jet. Hadrons only deposit part of their energy in the detector. There are effective detector responses available for pions (muons), kaons and protons, you should apply them.

Cheers

Markus

jdmulligan:
Hi @mfasDa,

My understanding was that these effects would be taken care of on Mateusz's side, and from me he just wants the response to the total truth energy. I'll check.

Cheers,
James

mfasDa:
+1

PR #6315, opened 2018-08-03
Restruction trigger cluster indexing

mfasDa:
As the clusters CENT and CENTNOTRD have a certain
overlap the indexing must reflect the overlap rather
than the absolute trigger name.


mfasDa:
+1

PR #6319, opened 2018-08-04
include converted electrons at large R into photons

dsekihat:
electrons from conversion at large R are considered as photons.


kharlov:
+1

PR #6332, opened 2018-08-06
A small bugfix for the task macro.

osaarima:
The goodbit for stringstream is not true for the last instance of the
list so instead !fail has to be used.


aohlson:
+1

PR #6333, opened 2018-08-06
Update Dstar tracking syst pp5TeV 2017

lvermunt:
Updated values for Dstar tracking systematic after implementation/fixes strategy for soft pion


arossi81:
+1

PR #6338, opened 2018-08-06
Implemented changes in Zvtx and SPD-vertex-based cuts in IsEventSelected

fgrosa:
1) swapped order of Zvtx and SPD-vertex-based cuts
2) set default of flag fCutOnzVertexSPD=3
3) added flag to apply |Zvtx|>X cm cut on SPD vertex

### DIFF
--- a/PWGHF/vertexingHF/AliRDHFCuts.cxx
+++ b/PWGHF/vertexingHF/AliRDHFCuts.cxx
@@ -113,7 +113,7 @@ fApplySPDDeadPbPb2011(kFALSE),
 fApplySPDMisalignedPP2012(kFALSE),
 fMaxDiffTRKV0Centr(-1.),
 fRemoveTrackletOutliers(kFALSE),
-fCutOnzVertexSPD(0),
+fCutOnzVertexSPD(3),
 fKinkReject(kFALSE),
 fUseTrackSelectionWithFilterBits(kTRUE),
 fUseCentrFlatteningInMC(kFALSE),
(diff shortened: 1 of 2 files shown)


PWGHF/vertexingHF/AliRDHFCuts.cxx
```diff
@@ -632,52 +634,65 @@ Bool_t AliRDHFCuts::IsEventSelected(AliVEvent *event) {
...
+          }
+        }
       }
     }
   }
 
+  Double_t zvert = -999;
+  if(!fApplyZcutOnSPDvtx && vertex) zvert=vertex->GetZ();
+  else if(fApplyZcutOnSPDvtx) {
+    if(!vSPD){
```
fprino: Hi Fabrizio,

thanks for this implemtentation.
Just one doubt: at this point shouldn't we also check that the contributors to the SPD vertex are >0?
Cheers, Francesco

PWGHF/vertexingHF/AliRDHFCuts.cxx
```diff
@@ -632,52 +634,65 @@ Bool_t AliRDHFCuts::IsEventSelected(AliVEvent *event) {
...
+      accept=kFALSE;
+      fEvRejectionBits+=1<<kBadSPDVertex;
+    }
+    else zvert = vSPD->GetZ();
+  }
+  
+  if((!fApplyZcutOnSPDvtx && vertex) || (fApplyZcutOnSPDvtx && vSPD)) {
```
fprino: Do we need this if condition?
if the conditions (!fApplyZcutOnSPDvtx && vertex) || (fApplyZcutOnSPDvtx && vSPD) are not satisfied, then zvert is -999 and will not enter in the fMaxVtxZ region.

PWGHF/vertexingHF/AliRDHFCuts.cxx
fgrosa: Hi Francesco, 
right! I forgot it, I add it now with an additional commit. Thanks!

PWGHF/vertexingHF/AliRDHFCuts.cxx
fgrosa: Yes, you are right! I remove it in the new commit. Thanks again!

fprino:
+1
Thanks for implementing my suggesstions!

PR #6339, opened 2018-08-07
Fixing memory leaks in QA script

Steffanic:
I had issues with memory leaks on large datasets in the QA task. These cleanup methods should fix the issue.

@efranzis @jdmulligan


jdmulligan:
Hi @Steffanic,

Looks reasonable to me (I think these are changes to recent additions by @efranzis, so she should confirm).

I have just one request: Can you please combine these two commits into a single one? It doesn't really make sense for them to be separate. Moreover I don't think the title "fixing OADB" is accurate. 

Cheers,
James

PR #6359, opened 2018-08-08
Added ntuple for studies on primary-vertex selections

fgrosa:


PWGHF/vertexingHF/AliAnalysisTaskCheckEvSel.cxx
```diff
@@ -295,6 +310,24 @@ void AliAnalysisTaskCheckEvSel::UserExec(Option_t */*option*/){
   Double_t zvSPD=vertexSPD->GetZ();
   Double_t zvTRK=vertex->GetZ();
 
+  if(fEnableVertexNtuple) {
+    Float_t wrej4ntuple = (Float_t)wrej;
+    if(isEvSel) wrej4ntuple = -1;
+    Float_t vertextype=0;
+    TString vertextitle = vertex->GetTitle();
```
fprino: To be safe, it would be better to move the line
TString vertextitle = vertex->GetTitle();
inside the condition if(vertex).

PWGHF/vertexingHF/AliAnalysisTaskCheckEvSel.cxx
```diff
@@ -295,6 +310,24 @@ void AliAnalysisTaskCheckEvSel::UserExec(Option_t */*option*/){
...
+    if(vertex) {
+      //Vtx tracks
+      if(vertextitle.Contains("VertexerTracks")) vertextype=3;
+      //vtx SPD
+      else if(vertextitle.Contains("Z")) vertextype=1;
+      else if(vertextitle.Contains("3D")) vertextype=2;
+    }
+      
+    Float_t vec4ntuple[6] = {(Float_t)zvSPD,(Float_t)zvTRK,(Float_t)vertexSPD->GetNContributors(),(Float_t)vertex->GetNContributors(),wrej4ntuple,vertextype};
+    fNtupleZvtxDistVsWhyRej->Fill(vec4ntuple);
```
fprino: Here you probably need to add a condition
if(fNtupleZvtxDistVsWhyRej) or if(fEnableVertexNtuple) before filling the ntuple, since in the case in which fEnableVertexNtuple is false, fNtupleZvtxDistVsWhyRej is a null pointer.

PWGHF/vertexingHF/AliAnalysisTaskCheckEvSel.cxx
fgrosa: Hi Francesco, 
right thanks! I committed your suggestion!

PWGHF/vertexingHF/AliAnalysisTaskCheckEvSel.cxx
fgrosa: Hi Francesco,
indeed my intention was to create the ntuple only if the flag is kTRUE, so I changed the if in the UserCreateOutputObjects (it was a typo the condition in the if)

Thanks again!

PR #6361, opened 2018-08-08
JetPerformance: Allow to fill EMCal response when trigger patches are missing, and update jet matching

jdmulligan:
(update to my user task)

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
@@ -1323,13 +1323,19 @@ void AliAnalysisTaskEmcalJetPerformance::DoTriggerSimulation()
   fHistManager.FillTH2(histname.Data(), nBkgPatchesDCal, kDCal);
   
   // Fill max patch response and total EMCal energy response, if requested
-  if (fGeneratorLevel && fDoTriggerResponse && maxPatch) {
+  if (fGeneratorLevel && fDoTriggerResponse) {
     
-    Double_t phiMinMaxPatch = maxPatch->GetPhiMin();
-    Double_t phiMaxMaxPatch = maxPatch->GetPhiMax();
-    Double_t etaMinMaxPatch = maxPatch->GetEtaMin();
-    Double_t etaMaxMaxPatch = maxPatch->GetEtaMax();
+    Double_t phiMinMaxPatch;
+    Double_t phiMaxMaxPatch;
+    Double_t etaMinMaxPatch;
+    Double_t etaMaxMaxPatch;
     Double_t truthEnergyPatch = 0;
+    if (maxPatch) {
+      phiMinMaxPatch = maxPatch->GetPhiMin();
+      phiMaxMaxPatch = maxPatch->GetPhiMax();
+      etaMinMaxPatch = maxPatch->GetEtaMin();
+      etaMaxMaxPatch = maxPatch->GetEtaMax();
+    }
     
     Double_t phiMinEMCal = fGeom->GetArm1PhiMin() * TMath::DegToRad(); // 80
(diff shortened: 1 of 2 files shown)


jdmulligan:
Hi All, I forgot that this PR was still open. I pushed a new commit now to my user task, unrelated to the first one. Please go ahead and approve both.

Thanks,
James

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
```diff
@@ -1323,13 +1317,19 @@ void AliAnalysisTaskEmcalJetPerformance::DoTriggerSimulation()
...
+  if (fGeneratorLevel && fDoTriggerResponse) {
     
-    Double_t phiMinMaxPatch = maxPatch->GetPhiMin();
-    Double_t phiMaxMaxPatch = maxPatch->GetPhiMax();
-    Double_t etaMinMaxPatch = maxPatch->GetEtaMin();
-    Double_t etaMaxMaxPatch = maxPatch->GetEtaMax();
+    Double_t phiMinMaxPatch;
+    Double_t phiMaxMaxPatch;
+    Double_t etaMinMaxPatch;
+    Double_t etaMaxMaxPatch;
```
raymondEhlers: You should initialize these values

raymondEhlers:
+test

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
jdmulligan: They are initialized right below, if the option is enabled to use them. Is that bad practice? Otherwise if I initialize them immediately C++ gives warnings that I have unused variables.

raymondEhlers: I would argue that it's bad practice because if you don't have a max patch, you then have a bunch of uninitiated variables floating around. Right now it's manageable, but what if you add a set of statements between them in the future? It's easy to lose track of such things.

I'm on my phone, but it looks like you only use them once, so why not call the methods directly and get rid of the variables?

jdmulligan: Okay, I just called the methods directly now, with the below commit.

But still a bit confused, suppose I needed to call those variables 10 times, what would you suggest? Initially I defined them inside of an if statement, but the compiler complains about unused variables (I thought it was smarter than that?). 

Anyway PR should be ready to approve now.

raymondEhlers: Humm - well, since you have them uninitialized if you don't have a max patch, I'd suggest that it indicates that you don't really need them there. Instead, I'd suggest that you refactor it into a closer logical block. I think that you should always be able to do that (although I'm not quite clear on how/when you received the warnings)

raymondEhlers:
+1

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
jdmulligan: Yea, ok, I guess you are right -- I can always in principle initialize the variables in the first "if statement" where I use them. It makes the code less readable imo, but fair enough. Anyway thanks!

raymondEhlers: Yeah, I can definitely see you point on readability (I'm fine with it, but I think there are a lot of reasonable ways to look at it). (As an aside, when I was checking into this, I found [this](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-always), which I think this a helpful discussion on the topic). Thanks for bearing with me on this!

PR #6381, opened 2018-08-10
New Histogram plus new Config

elobejko:
Add new folder without invariant mass cut and add in Config fail hisogram


pkarczma:
+1

PR #6398, opened 2018-08-13
Allow running on AOD MC particles

rhaake:
AliAODMCParticles cannot simply be copied, they need to be wrapped to AliAODtracks.


rhaake:
+1

PR #6409, opened 2018-08-13
Extract parton type from AliEmcalPythiaInfo object

rhaake:
Extract jet type from AliEmcalPythiaInfo object for MC data


rhaake:
+1

PR #6452, opened 2018-08-16
Added new calibration OADB for LHC18c (standard calibration MB sample)

rderradi:
Calibration based on 53 runs from LHC18c_pass1_FAST production - kINT7 trigger selection.


fprino:
+1

PR #6471, opened 2018-08-17
Numerous fixes for Temp and Led readout and processing

loizides:
Most important changes
a) Led and LedMon tree now properly read
b) Online-offline index conversion in create tree


loizides:
+1

PR #6479, opened 2018-08-19
PWGJEQA: cleanup prints, fix python 2 compatibility, fix outlier pdf naming bug

jdmulligan:
@efranzis please take a look and give your blessing. these are all superficial changes, mainly making the terminal output way more readable. also I fixed the outlier pdf bug, the fix is a bit hacky, but at least it works now.


efranzis:
@jdmulligan though shall have my blessing

raymondEhlers:
+1

PR #6487, opened 2018-08-20
Fix folder problem with multiple wagons & syst. cut vars.

a-mathis:
o Adding systematic cut variations
  - Disable rotated sample
  - Clean up options of the femto config


rbertens:
+1

PR #6505, opened 2018-08-21
First commit to calculate hadron-deuteron correlations with MC trains 

loizides:


loizides:
This was locally tested and works. Since I did not touh anyone elses code I hope it can just be approved?

bdoenigu:
+1

loizides:
This was locally tested and works. Since I did not touh anyone elses code I hope it can just be approved?

PR #6516, opened 2018-08-22
JetPerformance: require mc matching to impose deltaR cut on truth jet

jdmulligan:
and allow geom matching to use accepted jets.

(update to my user task)


raymondEhlers:
+1

PR #6517, opened 2018-08-23
Fixing bug in determination of AP + change in the code to read merged files

rderradi:
Modification in the code to allow to run the macro on merged output files instead of a list of per-run files.


chiarazampolli:
+1

PR #6518, opened 2018-08-23
Added option to use clusterizer v3 in EMCal correction task

rhaake:
This commit allows usage of EMCal clusterizer v3 when using the EMCal correction task.


mfasDa:
+1

Just adds additional option to enable the clusterizer v3 (which is now tagged), no change in existing code, can be approved right away.

PR #6529, opened 2018-08-23
Expands mass range, small changes

moliver813:
Expands mass range, removes some cluster E bins
Fixes tiny error in trigger counting.
Updates trigger pt bins.


efranzis:
+test

PR #6530, opened 2018-08-24
AliAnalysisTaskNetLambdaTrad.cxx update DCA cuts 

eumaka:


rbertens:
+1 @eumaka the cuts that you use are now hardcoded in your task, so that you have to update the task every time you want to change the cuts. To make changing cuts easier, you could think of introducing data members for the values you cut on, and adding some setter functions that set the values of these members. Then you don't have to commit changes and open a PR every time that you change a cut :)

PR #6540, opened 2018-08-24
Updated bin label for TOF QA

njacazio:
- Updated bin labelling for T0 mask plot, minor modification (first commit)
- Formatted task with clang-format (second commit)


chiarazampolli:
+1

chiarazampolli:
Ciao Niccolo', 

Can you open a PR for this change, so that we know it will go in v5-09-37? Just in case we see any difference in the RelVal output (which I don't expect, since you just changed the labeling, right?)

Cheers,

Chiara

PR #6543, opened 2018-08-24
New trainconf to test results in Mass,TPC de/dx and psipair for e+e- cuts

eddoming:
Just more cases and different trains config for Dalitz


dmuhlhei:
+1

PR #6549, opened 2018-08-25
Fixing a problem with pointers

hahassan7:
I fixed a problem with pointers.


fprino:
Hi Hadi,

are you sure that this modification is OK?
If you set to NULL the pointers before deleting them, then the delete command won't delete the pointers liberating the memory.
What is the problem with the pointers that you are trying to fix?

Cheers, Francesco

hahassan7:
Hi Francesco,

I tried to do just delete, but the code failed to execute (Break segmentation violation). I tried it this way and it worked.
I just want to free the memory after using each pointer so the memory consumption is very low.

Best Regards
Hadi

fprino:
Hi Hadi,

yes, it is correct to delete the pointers that you have allocated to free some memory.
However, after checking the whole cxx file (my comment before was posted after checking only the diff)
I have seen that the 2 pointers prodVtxDaughterPos are not allocated by your task, but are taken from the AOD tracks.
So, you should not delete these pointers, which are managed by AliAODtrack.

Cheers, Francesco

fprino:
+1

PR #6558, opened 2018-08-27
Major task modification: support for pt and eta bin-wise analysis.

nirbhaygit:
Major task modification is done to support p and pT options. Differential pt and eta analysis can be done in one go.


aohlson:
+test

aohlson:
+1

PR #6563, opened 2018-08-27
Remove prints from PHOSTriggerQA task to reduce the size of logs

kqf:
This task produces too big log file (it prints trigger classes for all events) so it's impossible to run this task in trains.


kharlov:
+1

PR #6576, opened 2018-08-28
Fix for ROOT6

davidrohr:
cannot include macro with $ALICE_PHYSICS in #include directory, and le is not in $ROOT_INCLUDE_PATH.

As discussed with @dberzano , this is a better fix as compared to alisw/alidist#1246, as it does not pollute the global include space.


davidrohr:
+test

dmuhlhei:
Hi, may I ask why there is a change from AliInfo() to Info() for the AddTask_ConversionAODProduction.C?

davidrohr:
AliInfo needs ClassName(), which is not available in the macro

dmuhlhei:
So this is only an issue for a macro? Since we use AliInfo in various places

davidrohr:
Yes, in any class that inherits from TObject it is OK. But the macro is a plain C function, so ClassName is not defined. Since the macro was never compiled, and apparently that error message was never triggered, we didn't see the problem with ROOT5.

dmuhlhei:
+1

fprino:
+1

PR #6577, opened 2018-08-28
 fix of a bug in the JES syst. evaluation

btrzecia:
Fix for a bug in the evaluation of the sys. unc. for the JES, track rejection was in a wrong place.


aiola:
+1

PR #6580, opened 2018-08-28
[EMCAL-191] PWG/EMCAL & TENDER: Implementation of Run2 EMCal/DCal temperature calib in OADB

nschmidtALICE:
The interface for the new temperature calibrations for Run2 is now available!

As soon as the calibration files are available on EOS (EMCALTemperatureCalibSM.root and EMCALTemperatureCalibParam.root) one can activate the temperature calibration in one of two ways:
1. Using the Tender: Set the last argument of AddTaskEMCALTender.C to kTRUE
2. Using the correction framework: Set doTempCalibRun2: true in the CellEnergy category.

The standard settig in both, tender and CF, is that no temperature correction is applied for Run2. Therefore nothing changes for run2 analyzers at the moment (except if they themselves activate it).

In both calibration tools, the implementation first checks the runnumber to be greater than 197692 to be certain that Run2 data is being handled. Then for Run2, two OADB containers are being loaded. One with the temperature per supermodule and one with slope and offset of the temperature dependent correction. It was made sure that in the CF, the std::unique_ptr is used to avoid memory leaks. 

In case of questions, ask me for the implementation of the OADB changes or @FriederikeBock and @loizides for the temperature calibration procedure which is documented in [EMCAL-191].

### DIFF
--- a/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellEnergy.cxx
+++ b/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellEnergy.cxx
@@ -49,7 +49,10 @@ Bool_t AliEmcalCorrectionCellEnergy::Initialize()
   AliWarning("Init EMCAL cell recalibration");
   
   if(fFilepass.Contains("LHC14a1a")) fUseAutomaticRecalib = kTRUE;
-  
+
+  // check the YAML configuration if the Run2 calibration is requested (default is false)
+  GetProperty("enableRun2TempCalib",fUseRunDepTempCalibRun2);
+
   if (!fRecoUtils)
     fRecoUtils  = new AliEMCALRecoUtils;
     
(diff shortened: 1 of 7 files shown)


nschmidtALICE:
@raymondEhlers, i have addressed all your proposed changes. The variable in the YAML file is now called "doTempCalibRun2" and the GetProperty() call is done in Initialize().
The coding style has been adjusted to be consistent with the rest of the code.
The comments and AliInfos have been improved and the containers are checked and ownership is properly handled.

raymondEhlers:
@nschmidtALICE Thanks! Looks great to me!

raymondEhlers:
+1 (I reviewed the Correction Framework, but not the tender. Since other approvals are needed, my +1 alone won't lead to it being merged immediately, so I think it's fine to add it now)

loizides:
Hi Nico, lets also add the switch to the tender as discussed yesterday. Constantin

FriederikeBock:
+1

lcunquei:
+1

PR #6581, opened 2018-08-28
make it possible to run the generate inside the task 

loizides:
this allows to run with several settings of coalescence parameters


bdoenigu:
+1

PR #6604, opened 2018-08-29
Added another way to choose kt-jets for background

osaarima:
This way the DeltaPhi cut will be taken into account when choosing what
kt-jets will be used in the rho estimation. In addition  I did some
minor bug fixes.


aohlson:
+1

PR #6620, opened 2018-08-29
The train have to be divided to no satured the memory of train...

eddoming:
We divide the trainconfig 118, in two trains, that because I have enough statistics from 17pq and test for the 15n, and maybe the train memory can be saturated.


dmuhlhei:
+1

PR #6635, opened 2018-08-30
Strip header paths for rootcint

dberzano:


dberzano:
+1 this is related to the ROOT 6 migration

dberzano:
@davidrohr @chiarazampolli we should make it work with ROOT 5 as well now 😄

davidrohr:
@dberzano : are you working on the ROOT5? I have just set up a root5/6 testbed for AliTPCCommon and have nothing to do tonight :|

davidrohr:
@dberzano @chiarazampolli : That was a pretty hard one: The class AliT0AnalysisTaskQA exists both in AliRoot and in AliPhysics, so it depends on the order of the include paths which one is taken. The AliPhysics version is just an extended version of the AliRoot one. Before, it worked by chance, and after our changes this somehow failed, so ROOT is now loading the AliRoot header, which misses some members.

To me it seems the AliRoot version is not used, and it was probably forgotten when we split of AliPhysics, so I would suggest to simply remove it in AliRoot. Otherwise, we'll have to rename one of the tasks.

dberzano:
This problem should be fixed in the corresponding AliRoot PR as you can see above. This PR will be merged automatically when the corresponding AliRoot PR passes the tests, so we just need to wait.

dberzano:
@davidrohr so this PR will not be merged automatically even if the corresponding AliRoot PR passes the tests. The reason is that we test AliPhysics PRs against tagged AliRoot and not master. In other words, we need an AliRoot tag before we can merge this one (but we can run the validation with this commit cherry-picked in any case, even if it's not merged).

dberzano:
+1

PR #6646, opened 2018-08-31
[EMCAL-191] TENDER+PWG/EMCAL: Correction of temperature calibration factor Run2

nschmidtALICE:
Small fix:
The inversion of the correction factor was missing for the Run2 correction.


FriederikeBock:
+1

lcunquei:
+1

PR #6652, opened 2018-08-31
Added new calibration OADB for LHC18h (standard calibration based on …

fprino:
…MB sample)


fprino:
Also VHM sample is used in the calibration

fprino:
+1

PR #6653, opened 2018-09-01
Added template sparses for varied D meson weight (for systematics).

efgauger:
Also added ability to specify cluster matching distance in the addtask call.


sashingo:
+1

PR #6655, opened 2018-09-01
Add PID/HM Trigger/Non-Systematic option to AddTask macro

BongHwi:
Add PID option for each daughter particles
Add HM Trigger option for pp HM event
Add Non-Systematic option for memory issue


mkruegerGitHub:
+1

PR #6689, opened 2018-09-04
Refactor the code; Add cpv, dispersion cuts to E/p ratio; Add distancе…

kqf:
… to the bad channel to the cluster cuts


kharlov:
+1

PR #6703, opened 2018-09-05
Tracking of magnetic field variable, more coarse binning to fix memory usage 

bhohlweger:


rbertens:
+1 @bhohlweger we should send a mail to see why your merge rights are not yet active ....

dberzano:
The permissions problem should have been fixed by now, apologies for the confusion.

PR #6717, opened 2018-09-05
Small bug fix on AliJIaaAna.cxx

hyeonjoongkim:
Line 423 should be changed to continue;
Adding pttbin for fPhiS Histogram


aohlson:
+1

PR #6718, opened 2018-09-06
Adding Set functions for tracking and PID cuts for systematics

gyronaut:
In-line set functions and new variables added to header to allow for systematic checks on Kaon PID and tracking cuts.


aohlson:
+1

PR #6743, opened 2018-09-07
Remove spacing after function name

BongHwi:
Just removed one spacing between function and "{"


BongHwi:
Is it some problem in my request?
It looks like it's pending again...

bdoenigu:
+1

PR #6750, opened 2018-09-09
extend uppper limit of Ncluster

dsekihat:
extend upper limit of Ncluster and add one more basic histogram in embedding


kharlov:
+1

PR #6754, opened 2018-09-10
[EMCAL-185,EMCAL-292,EMCAL-300,EMCAL-304] Upload of additional EMC bad channels for LHC16i,g LHC17g LHC18b

nschmidtALICE:


gconesab:
+1

PR #6764, opened 2018-09-10
PWGGA/GammaConv: removed absolute value over alpha in HeavyMesonTask

redeboer:
Removed 5 occurrences of `TMath::Abs` in `Fill` calls in `AliAnalysisTaskHeavyNeutralMesonToGG.cxx`. This is to solve a bug where negative \alpha values projected in the positive plane.


nschmidtALICE:
+1

PR #6771, opened 2018-09-11
PWGGA/GammaConv: added pPb GammaHeavyMeson trainconfigs without 0-20% centrality

redeboer:
Created 'thinner' copies of p-Pb GammaHeavymeson trainconfigs (all three methods) that contain only 0-100% centrality


nschmidtALICE:
+1

PR #6776, opened 2018-09-11
Updated DrawTrendingPIDQA.C for TOF plots

njacazio:
- Changes affect only PID for TOF only plots
- Added missing return if function is completed successfully
- Improved labelling and range in TOF plots
- For TOF plot showing if the data point is higher or lower than the plotting range
- Added new function for single detector customization


fprino:
+1

PR #6781, opened 2018-09-11
PWGGA/GammaConv: changed isHeavyIon and cutnumberEvent for pPb

redeboer:
Set `isHeavyIon` and `cutnumberEvent` so that it can handle p-Pb


nschmidtALICE:
+1

PR #6808, opened 2018-09-13
PWG/EMCAL + TENDER: Fix for temperature calibration factor EMCal in CF and Tender

nschmidtALICE:
The existing calibration factor is now being loaded and the temperature calibration factor is being multiplied to it.


gconesab:
+1

PR #6830, opened 2018-09-14
Change the output container name with input

BongHwi:
changed the output container name with HM trigger mode


BongHwi:
It looks like it's pending again.

bdoenigu:
+1

PR #6842, opened 2018-09-14
modifying correlation THnSparse axes in hadron-phi analysis

gyronaut:
changing handling of zvertex (move from being a sparse axis to being separate histograms), as well as reducing number of bins per axis due to memory issues when merging.


aohlson:
+1

PR #6844, opened 2018-09-15
AliFemto: removing checking onfly status for Xis and adding pile-up

lgraczykCern:
removing cut on on-the-fly status for cascades and adding out-of-bunch pile-up rejection track based cuts


lgraczykCern:
+1

PR #6853, opened 2018-09-17
PWGGA/GammaConv: added SPD requirement for omega analysis

fjonasALICE:
added requirement of hit in any of the SPD layers for charged pions to check influence of out-of-bunch pileup


dmuhlhei:
+1

PR #6860, opened 2018-09-17
Berkeley group ntuplizer

yslai:
Ntuplizer for inclusion within the PWG-JE


raymondEhlers:
+test

loizides:
where do we stand with this?

mfasDa:
At least the register keywords need to be removed - otherwise this will end in compilation issues under macOS.

For the direct access to CGAL which is currently breaking the compilation I think this needs to be protected in a similar way to what is done with fastjet (and RooUnfold soon) only building your ntupleizer classes if CGAL is found. Furthermore since you access CGAL directly you need to set the inlude path to the CGAL header and the library path in the CMakeLists.txt of the PWGJE/EmcalJetTasks library. CGAL is a build requirement for fastjet so it is available on the grid. In case you need a newer CGAL version than what is provided by alidist you need to propose a pull request there as well.

ktf:
Is this still relevant? If not, can we close it?

ktf:
Closing this since there was no answer to my inquiry on the 4th of May on wether this is actually still  being worked on. Feel free to reopen if still needed.

PR #6869, opened 2018-09-19
Adding Set function for track filter bits, new histos for Kaon PID

gyronaut:
Set functions for filter bits of Kaon tracks, associated hadrons, and trigger hadrons. Kaon PID and distribution histos added for systematic & efficiency checks.


rbertens:
+1

PR #6881, opened 2018-09-20
PWGGA

jluehder:
Changes for AOD production macro
Use of an open sigma cut (8th number in Cutstring to d)
Use of Chi2=50 (17th number in Cutstring to 1)


FriederikeBock:
+1

PR #6883, opened 2018-09-20
Allow to filter physical primaries in MC

rhaake:
Optional filtering of physical primaries in the input array (when running on MC)


rhaake:
+1

PR #6884, opened 2018-09-20
adapt trigger selection 

sebaleh:
adapt trigger selection in efficiency task to the one used in the ML tree maker


miweberSMI:
+1

PR #6897, opened 2018-09-21
Simulation of the EMCAL energy scale shift in the trigger electronics

mfasDa:
A constant scale shift (in GeV) is added to each cell
when filling the offline simple ADC data grid. The scale
shift is applied before conversion to FastOR ADC values.


mfasDa:
+1

PR #5489, opened 2018-05-23
Reactivate TPC HV dip detection for all data

wiechula:
This commit is a follow up of https://github.com/alisw/AliRoot/pull/678.

It also unifies the treatment of HV dips between the different event types (ESD, AOD, MC).

PR #6898, opened 2018-09-21
Fix AliTriggerAnalysis: TPC HV dip detection

loizides:
Revert partially change of 
https://github.com/alisw/AliRoot/pull/678
https://github.com/alisw/AliPhysics/pull/5489
to avoid calling IsDetectorOn for MC

Discussion by email attached below:
> Hi all,
>
> sorry for the confusion.
> You are right. If you use ESD MC indeed the events can be rejected with the addition of realistic time stamps, which was not there in the past.
>
> Then we should add back the check for MC. Sorry, this I didn't think about when modifying the code. I was fooled since I thought about AliMCEvent, which is of course not for the reconstructed MC data ...
>
> For the data analysis, of course, this is what we want, since those events will have bad PID. This in turn can lead to different results. The rejection in MC, however, should not have an influence on the analysis, except statistical, since we don't use the voltages in the simulation.
>
> Cheers,
> Jens
>
> On 21.09.2018 14:39, Constantin Loizides wrote:
>> Hi all,
>>
>> let me add that reading the code / commits it seems to me that unlike on AOD for ESD
>>
>> the change in the AliTriggerAnalysis::IsHVdipTPCEvent function would change between
>>
>> these two tags. (So I am betting a few beers that somehow on MC the function does not
>>
>> always return false)
>>
>> Constantin
>>
>>
>>
>> On 09/21/2018 02:37 PM, Daniel Mühlheim wrote:
>>> Dear all,
>>>
>>> The paper results were produced using ESDs because of much better grid efficiency compared to AODs. Hence, we are talking about ESDs and we do run PS.
>>>
>>> We are sure that it is not a grid glitch since our results are stable until 20180524 (we checked many tags from mid-2017 until 20180524) and it is persistent beyond 20180525 up to latest tags from today (the results are identical since 20180525, again checked with many tags in between).
>>>
>>> I'm trying to demonstrate the issue now locally. If I do not suceed in this I will run a train for further demonstration. I'll keep you updated.
>>>
>>> Cheers,
>>> Daniel
>>>
>>> On 21.09.2018 09:56, Evgeny Kryshen wrote:
>>>>
>>>> Dear all,
>>>>
>>>> I am also surprised to see this strange 5% effect in MC. Two questions:
>>>>
>>>> * Do you run PS on ESDs or AODs?
>>>>
>>>> * Are you sure that this 5% effect is not due to some glitch in the grid analysis? Could you take one ESD/AOD chunk and demonstrate the problem explicitly? If the effect is 5%-tish, it should be visible with one chunk.
>>>>
>>>> Cheers,
>>>>
>>>> Evgeny
>>>>
>>>>
>>>> On 09/20/2018 09:08 PM, Constantin Loizides wrote:
>>>>>
>>>>> To check it I would propose to add
>>>>>
>>>>> if (fMC) return kFALSE;
>>>>>
>>>>> in the Bool_t AliTriggerAnalysis::IsHVdipTPCEvent(const AliVEvent* event) {
>>>>>
>>>>> function...
>>>>>
>>>>> This should not hurt and could be tested with the next tag.
>>>>>
>>>>> Constantin
>>>>>
>>>>> On 09/20/2018 06:56 PM, Daniel Mühlheim wrote:
>>>>>> Dear Jens, all,
>>>>>>
>>>>>> Thanks for your explanations, I understand that in principle nothing should have changed. But the problem is that we see a difference and no other commit touched anything concerning trigger.
>>>>>>
>>>>>> Does anyone know if it is possible to create a tag like vAN-20180920-2 containing a revert of this TPC HV commit in addition? Then we could verify easily.
>>>>>>
>>>>>> Cheers,
>>>>>> Daniel
>>>>>>
>>>>>> On 20.09.2018 17:07, Jens Wiechula wrote:
>>>>>>> Hi all,
>>>>>>>
>>>>>>> the following two pull requests are of interest:
>>>>>>> https://github.com/alisw/AliRoot/pull/678
>>>>>>> https://github.com/alisw/AliPhysics/pull/5489
>>>>>>>
>>>>>>> The changes were done in a way that MC should not be affected. It is steered via
>>>>>>> AliVEvent.h
>>>>>>> virtual Bool_t IsDetectorOn(ULong_t /*detMask*/) const { return kTRUE; }
>>>>>>>
>>>>>>> which is only implemented for AliESDEvent and AliAODEvent
>>>>>>>
>>>>>>> git grep IsDetectorOn
>>>>>>> STEER/AOD/AliAODEvent.h:  virtual Bool_t IsDetectorOn(ULong_t detMask) const { return fHeader ? fHeader->IsDetectorOn(UInt_t(detMask)) : kTRUE; }
>>>>>>> STEER/AOD/AliAODHeader.h:  Bool_t IsDetectorOn(ULong_t detMask) const    { return (fDetectorStatus&detMask)>0; }
>>>>>>> STEER/AOD/AliNanoAODHeader.h:  virtual Bool_t IsDetectorOn(ULong_t /*detMask*/) const { return kTRUE; }
>>>>>>> STEER/ESD/AliESDEvent.h:  Bool_t IsDetectorOn(ULong_t detMask) const {return (fDetectorStatus&detMask)>0;}
>>>>>>> STEER/STEERBase/AliVAODHeader.h:  virtual Bool_t IsDetectorOn(ULong_t detMask) const = 0;
>>>>>>> STEER/STEERBase/AliVEvent.h:  virtual Bool_t IsDetectorOn(ULong_t /*detMask*/) const { return kTRUE; }
>>>>>>> TPC/TPCrec/AliTPCtracker.cxx:  //printf("Status %d \n", event->IsDetectorOn(AliDAQ::kTPC));
>>>>>>>
>>>>>>> So I don't see how this could influence MC since
>>>>>>>
>>>>>>> Bool_t AliTriggerAnalysis::IsHVdipTPCEvent(const AliVEvent* event) {
>>>>>>>   // The function IsDetectorOn is implemented in AliESDEvent and AliAODEvent
>>>>>>>   //     by default it returns kTRUE (so also for MC). Therefore no extra treatment is required
>>>>>>>   //
>>>>>>>   return !event->IsDetectorOn(AliDAQ::kTPC);
>>>>>>> }
>>>>>>>
>>>>>>> will always return false and not events will be rejected.
>>>>>>>
>>>>>>> There were also changes in other files between vAN-20180524..vAN-20180525
>>>>>>> git diff --stat vAN-20180524..vAN-20180525
>>>>>>>
>>>>>>>
>>>>>>> Please let me know in case of further questions.
>>>>>>>
>>>>>>> Cheers,
>>>>>>> Jens
>>>>>>>
>>>>>>>
>>>>>>> On 20.09.2018 14:17, Constantin Loizides wrote:
>>>>>>>> Hi everyone,
>>>>>>>>
>>>>>>>> thanks for this big step forward in understanding where the discrepancy comes from.
>>>>>>>>
>>>>>>>> Jens, in addition to the specifc question from Daniel, is there a potential dataset (or readout-cluster)
>>>>>>>>
>>>>>>>> dependence?
>>>>>>>>
>>>>>>>> Could run1 pPb datasets be affected?
>>>>>>>>
>>>>>>>> thanks
>>>>>>>>
>>>>>>>> Constantin
>>>>>>>>
>>>>>>>>
>>>>>>>> On 09/20/2018 10:11 AM, Daniel Mühlheim wrote:
>>>>>>>>> Dear Jens, Evgeny,
>>>>>>>>>
>>>>>>>>> we tried to validate our results on photon and meson production in pp collisions at 8 TeV (LHC12a-i and LHC15h1a-i + LHC15h2a-i) which entered
>>>>>>>>> arXiv earlier this year for the purpose of journal publication. However, we observed a disagreement on the level of approximately 5% compared
>>>>>>>>> to our old output.
>>>>>>>>>
>>>>>>>>> After some extensive studies, we found that this disagreement appears between AliPhysics tags 20180524 and 20180525 (both share the same
>>>>>>>>> AliRoot tag). At the same time we observe an increase of 5% of the total number of MC events being rejected due to some trigger condition. On
>>>>>>>>> the other hand, the data is completely unaffected.
>>>>>>>>> In the attachment, you can find the spectrum of our Pi0 measurement. "PUB" (neutral meson paper) denotes the spectrum entering publication in
>>>>>>>>> August 2017, "PUB dirG" (paper on direct photon production) entered arXiv in March this year. Some fluctuations are visible compared to both
>>>>>>>>> paper versions because we did not run full MC set to save resources during debugging.
>>>>>>>>>
>>>>>>>>> During these two tags, 24.05. and 25.05., a commit concerning TPC HV dip detection was made which we consider a possible cause for this
>>>>>>>>> behavior. In between these two tags, no other commit relevant to the trigger condition was made and there are no other commits we could
>>>>>>>>> identify that may cause this behavior.
>>>>>>>>>
>>>>>>>>> We are puzzled about this behavior, could you please have a look?
>>>>>>>>>
>>>>>>>>> Best regards,
>>>>>>>>> Daniel
>>>>>>>>
>>>>>>
>>>>>
>>>>
>>>
>>

### DIFF
--- a/OADB/AliTriggerAnalysis.cxx
+++ b/OADB/AliTriggerAnalysis.cxx
@@ -1156,6 +1156,8 @@ Bool_t AliTriggerAnalysis::IsHVdipTPCEvent(const AliVEvent* event) {
   // The function IsDetectorOn is implemented in AliESDEvent and AliAODEvent
   //     by default it returns kTRUE (so also for MC). Therefore no extra treatment is required
   //
+
+  if (fMC) return kFALSE;
   return !event->IsDetectorOn(AliDAQ::kTPC);
 }
 
--- a/PWGLF/NUCLEX/Nuclei/NucleiKine/AliAnalysisTaskNucleiKineCor.cxx
+++ b/PWGLF/NUCLEX/Nuclei/NucleiKine/AliAnalysisTaskNucleiKineCor.cxx
@@ -21,6 +21,9 @@
 
 ClassImp(AliAnalysisTaskNucleiKineCor);
 
+using std::cout;
+using std::endl;
+
 AliAnalysisTaskNucleiKineCor::AliAnalysisTaskNucleiKineCor(const char* name) :
   AliAnalysisTaskSE{name},
   fPdgCodes{211, -211, 321, -321, 2212, -2212, 2112, -2112, 1000010020, -1000010020,3122,-3122,3312,-3312,3334,-3334},
@@ -73,6 +76,9 @@ void AliAnalysisTaskNucleiKineCor::UserCreateOutputObjects()
   fHists[2] = new TH1D("hNdeuterons",";Number of deuterons at |Y|<1; Events",15,-0.5,14.5);
   fHists[2]->Sumw2();
   fOutputList->Add(fHists[2]);
+  fHists[3] = new TH1D("hEtaHadrons",";#eta; Events",100,-fAcc,fAcc);
+  fHists[3]->Sumw2();
+  fOutputList->Add(fHists[3]);
 
   fHists[10] = new TH2D("hHadHad","Hadron-hadron correlation;#Delta#varphi;Associated p_{T} (GeV/c)",
 			128,-TMath::Pi()/2,3*TMath::Pi()/2,100,0.,10);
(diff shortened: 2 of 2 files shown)


chiarazampolli:
Dear @wiechula , @ekryshen ,

I see that you had an exchange via email with Daniel and Constantin. Do I understand correctly that you already verified that this will fix the issue? If so, I will approve the change (or @ekryshen can do it).

Cheers,

Chiara

loizides:
Sorry this has to be taken out again... will ask Markus tomorrow how to do that

PR #6902, opened 2018-09-22
study of particle distribution

suppathak:


rbertens:
Hi @suppathak ,

In this PR and #6903 , you add independently a header and an implementation of a class. These  cannot be compiled for testing as they depend on one another (so you would need to make a PR in which you add these files simultaneously). 

Also, there are no changes in the CMake instruction, so for now, this class would be ignored when building - if it should be considered for building, you have to add the corresponding instructions in PWGCF/EBYE/CMakeLists.txt and class definition in PWGCF/EBYE/PWGCFebyeLinkDef.h . 

Hope this helps,

Redmer

suppathak:
Thank you for the informations. I have now updated it.

PR #6912, opened 2018-09-24
[EMCAL-190] New beam test data parametrization v4

gconesab:
constraining better energies larger than 100 GeV - P. Ganoti


gconesab:
+1

PR #6924, opened 2018-09-25
EMCal CorrFW: Add additional non-linearity enums

raymondEhlers:
Each enum value needs to be explicitly added to the string map to allow the value to be set from the YAML config. Adds `kBeamTestCorrectedv4` (from #6912) and some PCM parameterizations.


raymondEhlers:
+1

PR #6927, opened 2018-09-25
Fix AliTriggerAnalysis: TPC HV dip detection

dmuhlhei:
clean version of PR #6898


chiarazampolli:
I would like @wiechula to confirm that it is fine.

Chiara

dmuhlhei:
Yes, of course. Thanks a lot

wiechula:
Yes, this is fine for me. However, we should still understand why this could have an effect.
I don't understand it to be honest.

chiarazampolli:
Ok, I am giving a +1 then, while you follow it up.

Chiara

chiarazampolli:
+1

PR #6951, opened 2018-09-26
Add option for trigger track requirement

rhaake:
This commit adds the option to select events according to a trigger track requirement and saves the relevant relative information for each extracted jet.


rhaake:
+1

PR #6961, opened 2018-09-27
modify_weight_func

tomasuzuki:


dberzano:
For the record I am trying to figure out why the error log is not available, hold on.

PR #6963, opened 2018-09-27
Remove par. limits for Landau fit in TRD PID QA

martenole:
This makes the TRD QA trending parameters for the Landau fit more stable (as was observed for a few runs in LHC18l pass1).


tdietel:
+1

PR #6981, opened 2018-09-28
Allow trigger track label selection

rhaake:
New option: Trigger track in extractor will only be used if it has a label above a value (e.g. when from MC/embedding)


rhaake:
+test

rhaake:
+1

PR #7002, opened 2018-10-01
Fix bins of event mixing pool

BongHwi:
Event mixing pool has been changed.
Before: with zbin + tracklet multiplicity bin
After: with zbin + V0M multiplicity percentile bin(10 level)


bdoenigu:
+1

PR #7013, opened 2018-10-02
modified pile up cuts

naghmehmohammadi:


rbertens:
+1 @naghmehmohammadi not in time for the new tag im' afraid though ....

PR #7025, opened 2018-10-04
PWGGA: Added Option and Train Cfg to shift the TPC Sigma for clean 0 and clean 1

jluehder:
Added Option and Train Cfg to shift the TPC Sigma for clean 0 and clean 1, as in these Productions the dEdx distributions are shifted


nschmidtALICE:
+1

PR #7037, opened 2018-10-05
EMCal CorrFW: Default same min cellE and crosstalk induced energy, add docs

raymondEhlers:
This implements the recommendation that the cross-talk induced T-card minimum should be equal to the clusterizer min energy by default. I also added documentation to clarify how to use the shared parameters properly.  In particular, note that a user can't directly override a shared parameter - setting just the shared parameter is not sufficient! Instead, one must set both the value, as well as which values will share it, in their user config.

There is more information in the commit messages and documentation, including explicit examples.

If we think this is too complicated, we don't have to merge it. @gconesab , @jdmulligan, what do you think?

### DIFF
--- a/PWG/EMCAL/READMEemcCorrections.txt
+++ b/PWG/EMCAL/READMEemcCorrections.txt
@@ -280,7 +280,11 @@ There are a number of useful advanced options to make the Corrections Framework 
 
 #### Shared parameters
 
-Often, a user will want to change some parameters in unison. Say, if a pt cut is changed, it should be changed everywhere. In such a case, it is useful to able to define a variable so that one change will change things everything. This can be accomplished by defining a parameters in the "shared parameters" section of the %YAML file. The name of the parameter defined in the shared parameters section can be referenced in other areas of the file by prepending ``sharedParameters:`` to the parameter name. Consider the example below:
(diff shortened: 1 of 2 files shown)


raymondEhlers:
@jdmulligan and I chatted, and he thinks that the subtly described in the commits and documentation is more trouble than the benefit of specifying the value only once. So he would suggest to not merge this commit (I really wish this subtly was easily fixable, but it's a tremendous amount of design and coding work for a tiny benefit). I think both arguments have merit, so I'm fine either way.

So no rush, but when you get a chance, please let us know what you think.

(Regardless of how we decide, I'll merge the additional documentation unrelated to just changing to value)

gconesab:
Hi @raymondEhlers  and @jdmulligan,
I see the approach by Raymond is "simple" but also I agree with James that it could be quite error prone for most users, experienced or not, since this implies the modification on the way an important parameter used in the clusterization is called, just for a minor gain for something used only for the MC and not in all analysis for the moment. I would not merge this for the moment and just make sure the users set properly the parameter for the MC.

raymondEhlers:
Hi @gconesab , thanks for your comments - I agree. I'll close this for now and we can revisit it in the future if we decide it's needed

PR #7047, opened 2018-10-05
Adding task to fill trees with candidates info and tree handler class

afestant:
Adding task to fill trees with candidates info, the tree handler class and update of CMakeLists.txt and PWGHFvertexingHFLinkDef.h


fprino:
+1

PR #7056, opened 2018-10-07
Add fill function for the event mixing pool

BongHwi:
Add "FillTrackToEventPool" function to make mixing pool.
It was done in GoodTracksSelection() before, but it has filled 'current' event track list to the mixing pool. As a result, mixed background contains "real" signal, so that I changed order of making mixing pool after use it.


ycorrales:
+1

PR #7057, opened 2018-10-07
[EMCAL-403] Add histograms for the maximum cluster

mfasDa:
Add histograms for the maximum cluster
in  order to get estimates for centrality-
dependent rejection curves. ET-
dimension is removed, trigger cluster
dimension made optional.


mfasDa:
+1

PR #7060, opened 2018-10-07
CDBconnect : convert AddTask to static methods

adriansev:


shahor02:
Hi @adriansev ,
Why is this needed?
Cheers,
 Ruben

adriansev:
Hi! Well, this method will eliminate the LoadMacro usage and the possible headaches in root6. There are already a big number of tasks that moved to static method, i just wanted to eliminate the rest of LoadMacros from my analysis script. Of course the normal AddTask macro remains for compatibility with the current trains system.

shahor02:
+1
OK, thanks

PR #7062, opened 2018-10-07
AliMultSelectionTask : convert AddTask to static methods

adriansev:
In order to easily use AddTask in ROOT6, the AddTask function was moved into the source (AddTaskMultSelection.C macro remains for compatibility reasons)


chiarazampolli:
+1 should be the same as: https://github.com/alisw/AliPhysics/pull/7060

PR #7064, opened 2018-10-08
New analysis task  

lhusova:
Correlations with identified V0 as trigger particles ad unidentified charged hadrons as associated particles at high pt


akubera:
+1

akubera:
+1

PR #7076, opened 2018-10-08
Fix including headers from global include path (needed for ROOT6)

mfasDa:
Headers are installed in global include directory and
have to be taken from the global include dir. Incorrect
include paths show up as problems with symbols from the
dictionary in train tests using the ROOT6-based versions.


mfasDa:
+1

PR #7079, opened 2018-10-08
Add Set function to modify Cut parameters

BongHwi:
Add Set functions to the header file.
Add cut parameter variables to change them using AddTask macro.


bdoenigu:
+1

PR #7081, opened 2018-10-09
Modification of the gestion of the Q-vectors

cimordas:
Modify the computation of the Q-vectors using arrays and the functions using them accordingly. Add the computation of an alternative method to get the 2-harmonics symmetric cumulants.

### DIFF
--- a/PWG/FLOW/Tasks/AliAnalysisTaskTwoMultiCorrelations.cxx
+++ b/PWG/FLOW/Tasks/AliAnalysisTaskTwoMultiCorrelations.cxx
@@ -13,12 +13,14 @@
 * provided "as is" without express or implied warranty.                  * 
 **************************************************************************/
 
-/*******************************************************************************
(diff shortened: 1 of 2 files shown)


rbertens:
@dberzano checking the log
'collect2: fatal error: ld terminated with signal 9 [Killed]'
I'm a bit confused about the status of this PR - building failed, but it seemed that the build machine ran out of memory when linking ?

rbertens:
+1

dberzano:
@rbertens yes, er, we have seen the problem already randomly popping up when building ROOT 6 itself. As you have seen, the problem solved itself without any intervention. Note that PR checks only normally build (incrementally) your AliPhysics changes, but based on `alidist` master. In this case, an `alidist` PR altering the ROOT recipe was just merged. This is to say that it is not a systematic problem at least.

rbertens:
@dberzano ok thanks, all clear :) !

PR #7084, opened 2018-10-09
[EMCAL-403] Centrality dependence in trigger overlap

mfasDa:
Allow trigger correlation to be evaluated in classes of centrality.
Each class of centrality needs to be defined as separate wagon with
a range.


mfasDa:
+1

PR #7087, opened 2018-10-09
fix issue with mass calculation (SetUseKF), avoid event mixing on MC,…

sebaleh:
fix issue with mass calculation (SetUseKF), avoid event mixing on MC, adapt cuts to be consistent


miweberSMI:
+1

PR #7091, opened 2018-10-09
fixed the outlier removal .pdf creation

efranzis:
@jdmulligan, I fixed the broken .pdf creation. And I think outlier removal should be on by default


jdmulligan:
Okay, no objection from my side. 

Best,
James

raymondEhlers:
+1

PR #7107, opened 2018-10-10
D0 in finer pt bins for pp@5TeV

cterrevo:
Systematics for D0 in finer pt bins (pp@5TeV). Flag to standard bins set to kFALSE.


fprino:
+1

PR #7108, opened 2018-10-10
Incorporated support for MC p-Pb

victor-gonzalez:
Also
- multiplicty histograms binning according to analyzed system
- configuration activated p-Pb pile-up cut (for discarding it when MC)


rbertens:
+1

PR #7109, opened 2018-10-10
Cleaning up, make the syst cut Task to work, still need to work on th…

sartozza:
Cleaning up, make the syst cut Task to work, still need to work on the Minimalbooking Output


rbertens:
+1

PR #7114, opened 2018-10-10
few minor improvements to the rho task to analyze the 2016 data

efranzis:
Changed the way to check for ghost jets (was pT>0.1 now nTracks>0)
Changed the way the total covered area is determined (was sum of all jet areas (excluding two leading jets) now Area of the TPC)
Changed the way the total jet area that is used for the correction factor is used (was use all non-ghost jets w/o leading jets now use all jets that actually contribute to the rho calculation)


raymondEhlers:
+1

PR #7118, opened 2018-10-10
Add v0s, and other branches.

loizides:
Add switch on by default that the standard tracks are stored no extra stuff


loizides:
+1

PR #7119, opened 2018-10-10
Adding TString for Output MC

sartozza:
Added TString in the output container in AddTaskFemtoGranma.C in order to run in MC Lego trains.


akubera:
+1

PR #7131, opened 2018-10-11
Optimized purity estimation plots

panjinjin:
Optimized purity estimation plots


panjinjin:
Dear approvers, could you please +1 my request? so that the code can be integrated in today's AliPhysics tag. Thanks a lot.

rbertens:
+1 Hi @panjinjin , the PR came in slightly after 16h, so it was in any case too late for today's tag. But keep in mind, that we also have to actually review your code before giving a '+1', so if you want something to be included in the daily tag, please make the PR well before 16h

PR #7133, opened 2018-10-11
New trains for Dalitz analysis

eddoming:
New trains for Dalitz for the systematic error from Lucia for the 5TeV and 13TeV (Low and Normal Magnetic Field), plus studies on the electron cuts to establish a standard one and options for the chi decay.


FriederikeBock:
+1

PR #7136, opened 2018-10-12
Fix "THistManager" dependency error

BongHwi:
Change the cmake list to fix the dependency problem.
Especially, Add "PWGTools" to LIBDEPS.

Add QA histograms of Trigger Efficiency histogram in the code.


mkruegerGitHub:
+1

PR #7138, opened 2018-10-12
 PWG/EMCAL: Added possibilty to load custom Recalib and BadChannel OADB 

nschmidtALICE:
Due to a request by @loizides, @gconesab  and @FriederikeBock we now have the possibility to set custom OADB objects for the recalibration as well as bad channel maps. For this, simply set  one or both of the following values in the .yaml configuration:

```
CellEnergy:
    customRecalibFilePath: "" 
CellBadChannel: 
    customBadChannelFilePath: ""
```
In both cases, the full path including the .root filename must be provided. This allows also for custom names of the .root files.

### DIFF
--- a/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellEnergy.cxx
+++ b/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellEnergy.cxx
@@ -28,6 +28,7 @@ AliEmcalCorrectionCellEnergy::AliEmcalCorrectionCellEnergy() :
   ,fUseAutomaticRecalib(1)
   ,fUseAutomaticRunDepRecalib(1)
   ,fUseRunDepTempCalibRun2(0)
+  ,fCustomRecalibFilePath("")
 {
 }
 
@@ -52,6 +53,9 @@ Bool_t AliEmcalCorrectionCellEnergy::Initialize()
 
   // check the YAML configuration if the Run2 calibration is requested (default is false)
   GetProperty("enableRun2TempCalib",fUseRunDepTempCalibRun2);
+
+  // check the YAML configuration if a custom energy calibration is requested (default is empty string "")
+  GetProperty("customRecalibFilePath",fCustomRecalibFilePath);
 
   if (!fRecoUtils)
     fRecoUtils  = new AliEMCALRecoUtils;
@@ -149,6 +153,19 @@ Int_t AliEmcalCorrectionCellEnergy::InitRecalib()
     if (!recalibFile || recalibFile->IsZombie())
     {
       AliFatal(Form("EMCALRecalib.root not found in %s",fBasePath.Data()));
+      return 0;
+    }
+    
+    contRF = std::unique_ptr<AliOADBContainer>(static_cast<AliOADBContainer *>(recalibFile->Get("AliEMCALRecalib")));
+  }
+  else if (fCustomRecalibFilePath!="")
+  { //if custom recalib requested
+    AliInfo(Form("Loading custom Recalib OADB from given path %s",fCustomRecalibFilePath.Data()));
+    
+    recalibFile = std::unique_ptr<TFile>(TFile::Open(Form("%s",fCustomRecalibFilePath.Data()),"read"));
+    if (!recalibFile || recalibFile->IsZombie())
+    {
+      AliFatal(Form("Recalibration not file found. Provided path was: %s",fCustomRecalibFilePath.Data()));
       return 0;
     }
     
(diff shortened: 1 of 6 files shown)


FriederikeBock:
+1 
Thanks Nico for implementing these options!

PWG/EMCAL/config/AliEmcalCorrectionConfiguration.yaml
```diff
@@ -55,11 +55,13 @@ CellEnergy:                                         # Cell Energy correction com
     enabled: false                                  # Whether to enable the task
     createHistos: false                             # Whether the task should create output histograms
     enableRun2TempCalib: false                      # Whether the temperature calibration should be done for Run2
+    customRecalibFilePath: ""                       # Full path including .root file for custom recalibration object
     cellsNames:                                     # Names of the cells input objects which should be attached to the correction
         - defaultCells                              # This object is defined above in the cells section of the input objects
 CellBadChannel:                                     # Bad channel removal at the cell level component
     enabled: false                                  # Whether to enable the task
     createHistos: false                             # Whether the task should create output histograms
+    customBadChannelFilePath: ""                    # Full path to a custom bad channel OADB object
```
raymondEhlers: Since you note about that you need to include ".root" in the filename, should that also be done here?

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionComponent.cxx
```diff
@@ -337,6 +339,19 @@ Int_t AliEmcalCorrectionComponent::InitBadChannels()
     
     contBC = std::unique_ptr<AliOADBContainer>(static_cast<AliOADBContainer *>(fbad->Get("AliEMCALBadChannels")));
   }
+  else if (fCustomBadChannelFilePath!="")
```
raymondEhlers: Same question about `fBasePath` here

PWG/EMCAL/config/AliEmcalCorrectionConfiguration.yaml
raymondEhlers: Custom bad channels can also be set in the clusterizer (which is why the function is in the correction component base class), so it should also be possible to set that option there

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellEnergy.cxx
```diff
@@ -154,6 +158,19 @@ Int_t AliEmcalCorrectionCellEnergy::InitRecalib()
     
     contRF = std::unique_ptr<AliOADBContainer>(static_cast<AliOADBContainer *>(recalibFile->Get("AliEMCALRecalib")));
   }
+  else if (fCustomRecalibFilePath!="")
```
raymondEhlers: For my education, why not just make `fBasePath` settable from the YAML? As far as I can tell from looking at this quickly, I think it can achieve the same thing as this PR. We just never made fBasePath settable because we didn't know of any interest

nschmidtALICE: fBasePath would mean that all OADB objects would need to be within this path. this seems unreasonable if one only wants to set a single OADB object. But, of course, you can also make a setter for fBasePath if you need

PWG/EMCAL/config/AliEmcalCorrectionConfiguration.yaml
nschmidtALICE: As far as i understood, the CellBadChannel part will always be called before the clusterizer. Therefore, the custom bad channel maps should then also be automatically set for the clusterizer if the option is set in CellBadChannel. Please correct me if i'm wrong.

nschmidtALICE: I will edit the comment in the next PR.

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellEnergy.cxx
raymondEhlers: `fBasePath` is a member function - it's not static. So it's set per component. I think every component accesses only one OADB object per component (except for the energy calib, which accesses energy and temperature - these probably should be split anyway). So perhaps it is necessary for the energy calibration, but I don't think we need it for the bad channel

PWG/EMCAL/config/AliEmcalCorrectionConfiguration.yaml
raymondEhlers: While it's true that the bad channel component will tend to be called before the clusterizer, each component loads the request OADB container individually. So with the current implementation, one could load a custom bad channel map for the bad channel component, but then the standard bad channel map for the clusterizer would still be loaded

raymondEhlers:
As a general comment, please wait 24 hours before approve any changes to the correction framework. The framework is stable, so it's important to have time for code review.

PR #7150, opened 2018-10-14
init TGrid::Connect when external file provided

gconesab:
This is a simple fix but needs to be discussed to do this properly.
I will approve this to be able to work, but it can be rejected as soon as it is available at least in 1 tag.


gconesab:
+1

PR #7153, opened 2018-10-15
Add optional RooUnfold dependency

dberzano:
Modern CMake is used: target `RooUnfold::RooUnfold` is exported if found


dberzano:
@mfasDa @lcunquei fyi (see also [ALIROOT-8025](https://alice.its.cern.ch/jira/browse/ALIROOT-8025))

PR #7168, opened 2018-10-15
Improve AliYAMLConfiguration docs and printing

raymondEhlers:
Automatically reinitialize the YAML nodes when printing if they are not
initialized to ensure that configurations will be printed after
streaming to the grid.

As a future possible improvement, this approach could be added to
GetProperty() and WriteProperty(), which may avoid the need to call
Reinitialize(). However, such a change would require some careful
thinking and testing, so it won't be done now.

@jdmulligan


raymondEhlers:
@aiola Can you take a look? Thanks!

aiola:
+1

PR #7187, opened 2018-10-17
[EMCAL-403]  Trigger separation in case L0 is INT7

mfasDa:
Needed  for PbPb: As L0 trigger INT7 was used and not
0EMC/0DMC, so checkes for the E/DMC string failed and
the trigger selection was based purely on the trigger bits.


mfasDa:
+1

PR #7223, opened 2018-10-19
Study correlation between the Sigma0 and its daughters

a-mathis:
o Also look at the corresponding Lambda_Sigma - p and photon_Sigma - p correlation function


a-mathis:
+1

PR #7226, opened 2018-10-19
small update

ffionda:
Adding possibility to reject randomly signal events according to pT-dependent weights 
(same procedure as implemented in AliReducedAnalysisJpsi2ee)


iarsene:
+1

PR #7233, opened 2018-10-20
 fix minor bug: name of addtask function doesn't match defined functin

rbertens:


rbertens:
@qgp  see http://alitrain.cern.ch/train-workdir/PWGMM/MM_MCGen_validation/64_20181020-1814/test/__ALL__/stdout , this should hopefully fix it

PR #7234, opened 2018-10-20
PWGGA/GammaConv updates for mixing and opening angle study

mikesas:


dmuhlhei:
Could you please elaborate the change for the mixing a little bit for me?

mikesas:
> Could you please elaborate the change for the mixing a little bit for me?

Of course Daniel,
For PCM-EMC we always only did mixing between the current event V0's with the mixing pool of clusters. We did already store the pools of the V0, so we can also just mix the current event clusters with the V0 pool. This is now being tested =)

dmuhlhei:
Yes, but don't you override the current version?  Shouldn't you define a new case for testing this new mixing method and keep the old (standard) mixing untouched? Sorry if I'm wrong

mikesas:
> Yes, but don't you override the current version? Shouldn't you define a new case for testing this new mixing method and keep the old (standard) mixing untouched? Sorry if I'm wrong

Yes this is done, it is behind a switch now.

FriederikeBock:
+1

PR #7275, opened 2018-10-24
- Changed binning of rho related histograms in Jet performance Task f…

efranzis:
…or pp and pPb

- Changed !kpp to ==kAA check since we will study MB pPb collisions. If in the future pPb vs. centrality were to be analyzed this can be adapted.
- Added a check to remove RandomCone overlaps with kT-jets from the fluctuations calculation
- Added calculation of pure charged component background fluctuation
- Changed check of full jets (in background fluctuation calculation) based on the jet container name

Please @jdmulligan have a look at all the changes and check.

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
@@ -366,7 +366,7 @@ void AliAnalysisTaskEmcalJetPerformance::AllocateJetHistograms()
     
     // Jet rejection reason
     histname = TString::Format("%s/JetHistograms/hJetRejectionReason", jets->GetArrayName().Data());
-    if (fForceBeamType != kpp) {
+    if (fForceBeamType == kAA) {
       title = histname + ";Centrality (%);Rejection reason;#it{p}_{T,corr} (GeV/#it{c});counts";
       TH3* hist = fHistManager.CreateTH3(histname.Data(), title.Data(), 10, 0, 100, 32, 0, 32, nPtBins5, fMinPt, fMaxPt);
       SetRejectionReasonLabels(hist->GetYaxis());
@@ -379,8 +379,14 @@ void AliAnalysisTaskEmcalJetPerformance::AllocateJetHistograms()
     // Rho vs. Centrality
     if (!jets->GetRhoName().IsNull()) {
       histname = TString::Format("%s/JetHistograms/hRhoVsCent", jets->GetArrayName().Data());
-      title = histname + ";Centrality (%);#rho (GeV/#it{c});counts";
-      fHistManager.CreateTH2(histname.Data(), title.Data(), 50, 0, 100, 100, 0, 500);
+      if (fForceBeamType == kAA) {
+    	  	  title = histname + ";Centrality (%);#rho (GeV/#it{c});counts";
+    	  	  fHistManager.CreateTH2(histname.Data(), title.Data(), 50, 0, 100, 100, 0, 500);
+      }
+      else{
+          title = histname + ";Centrality (%);#rho (GeV/#it{c});counts";
+          fHistManager.CreateTH2(histname.Data(), title.Data(), 50, 0, 100, 250, 0, 50);
+      }
     }
     
     // (Centrality, pT, NEF)
@@ -389,7 +395,7 @@ void AliAnalysisTaskEmcalJetPerformance::AllocateJetHistograms()
     Int_t nbinsz = 50; Int_t minz = 0; Int_t maxz = 1.;
     
     histname = TString::Format("%s/JetHistograms/hNEFVsPtEMCal", jets->GetArrayName().Data());
-    if (fForceBeamType != kpp) {
+    if (fForceBeamType == kAA) {
       title = histname + ";Centrality (%);#it{p}_{T}^{corr} (GeV/#it{c});NEF";
       fHistManager.CreateTH3(histname.Data(), title.Data(), nbinsx, minx, maxx, nbinsy, miny, maxy, nbinsz, minz, maxz);
     }
@@ -400,7 +406,7 @@ void AliAnalysisTaskEmcalJetPerformance::AllocateJetHistograms()
     
     if (fPlotDCal) {
       histname = TString::Format("%s/JetHistograms/hNEFVsPtDCal", jets->GetArrayName().Data());
-      if (fForceBeamType != kpp) {
+      if (fForceBeamType == kAA) {
         title = histname + ";Centrality (%);#it{p}_{T}^{corr} (GeV/#it{c});NEF";
         fHistManager.CreateTH3(histname.Data(), title.Data(), nbinsx, minx, maxx, nbinsy, miny, maxy, nbinsz, minz, maxz);
       }
@@ -419,7 +425,7 @@ void AliAnalysisTaskEmcalJetPerformance::AllocateJetHistograms()
     
     // pT-leading vs. pT
     histname = TString::Format("%s/JetHistograms/hPtLeadingVsPt", jets->GetArrayName().Data());
-    if (fForceBeamType != kpp) {
+    if (fForceBeamType == kAA) {
       title = histname + ";Centrality (%);#it{p}_{T}^{corr} (GeV/#it{c});#it{p}_{T,particle}^{leading} (GeV/#it{c})";
       fHistManager.CreateTH3(histname.Data(), title.Data(), 10, 0, 100, nPtBins2, fMinPt, fMaxPt, fMaxPt, 0, fMaxPt);
     }
@@ -430,13 +436,13 @@ void AliAnalysisTaskEmcalJetPerformance::AllocateJetHistograms()
     
     // A vs. pT
     histname = TString::Format("%s/JetHistograms/hAreaVsPt", jets->GetArrayName().Data());
-    if (fForceBeamType != kpp) {
-      title = histname + ";Centrality (%);#it{p}_{T}^{corr} (GeV/#it{c});#it{A}_{jet}";
-      fHistManager.CreateTH3(histname.Data(), title.Data(), 10, 0, 100, nPtBins2, fMinPt, fMaxPt, 50, 0, 0.5);
+    if (fForceBeamType == kAA) {
+      title = histname + ";Centrality (%);#it{p}_{T}^{corr} (GeV/#it{c});#it{A}_{jet}/#piR^{2}";
+      fHistManager.CreateTH3(histname.Data(), title.Data(), 10, 0, 100, nPtBins2, fMinPt, fMaxPt, 75, 0, 3);
     }
     else {
-      title = histname + ";#it{p}_{T}^{corr} (GeV/#it{c});#it{A}_{jet}";
-      fHistManager.CreateTH2(histname.Data(), title.Data(), nPtBins2, fMinPt, fMaxPt, 50, 0, 0.5);
+      title = histname + ";#it{p}_{T}^{corr} (GeV/#it{c});#it{A}_{jet}/#piR^{2}";
+      fHistManager.CreateTH2(histname.Data(), title.Data(), nPtBins2, fMinPt, fMaxPt, 100, 0, 3);
     }
     
     // (Centrality, pT, z-leading (charged))
@@ -445,7 +451,7 @@ void AliAnalysisTaskEmcalJetPerformance::AllocateJetHistograms()
     nbinsz = 50; minz = 0; maxz = 1.;
     
     histname = TString::Format("%s/JetHistograms/hZLeadingVsPtEMCal", jets->GetArrayName().Data());
-    if (fForceBeamType != kpp) {
+    if (fForceBeamType == kAA) {
       title = histname + ";Centrality (%);#it{p}_{T}^{corr} (GeV/#it{c});#it{z}_{leading}";
       fHistManager.CreateTH3(histname.Data(), title.Data(), nbinsx, minx, maxx, nbinsy, miny, maxy, nbinsz, minz, maxz);
     }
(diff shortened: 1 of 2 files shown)


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
```diff
@@ -379,8 +379,14 @@ void AliAnalysisTaskEmcalJetPerformance::AllocateJetHistograms()
     // Rho vs. Centrality
     if (!jets->GetRhoName().IsNull()) {
       histname = TString::Format("%s/JetHistograms/hRhoVsCent", jets->GetArrayName().Data());
-      title = histname + ";Centrality (%);#rho (GeV/#it{c});counts";
-      fHistManager.CreateTH2(histname.Data(), title.Data(), 50, 0, 100, 100, 0, 500);
+      if (fForceBeamType == kAA) {
+    	  	  title = histname + ";Centrality (%);#rho (GeV/#it{c});counts";
+    	  	  fHistManager.CreateTH2(histname.Data(), title.Data(), 50, 0, 100, 100, 0, 500);
```
jdmulligan: Can you please fix the indentations to match the rest of the task?

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
```diff
@@ -893,11 +899,42 @@ void AliAnalysisTaskEmcalJetPerformance::AllocateBackgroundHistograms()
...
     title = histname + ";Centrality;Scale factor;counts";
     fHistManager.CreateTH2(histname.Data(), title.Data(), 50, 0, 100, 100, 0, 5);
-    
     histname = TString::Format("%s/BackgroundHistograms/hDeltaPtEMCal", jets->GetArrayName().Data());
-    title = histname + ";Centrality (%);#delta#it{p}_{T} (GeV/#it{c});counts";
-    fHistManager.CreateTH2(histname.Data(), title.Data(), 10, 0, 100, 400, -50, 150);
-    
+    if (fForceBeamType == kAA) {
+    		title = histname + ";Centrality (%);#delta#it{p}_{T} (GeV/#it{c});counts";
+    		fHistManager.CreateTH2(histname.Data(), title.Data(), 10, 0, 100, 400, -50, 150);
```
jdmulligan: Indentation again, please check all changes. And look through the PR on GitHub for these types of things before you make the PR!

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
```diff
@@ -2190,7 +2228,30 @@ void AliAnalysisTaskEmcalJetPerformance::ComputeBackground()
   AliJetContainer* jetCont = 0;
   TIter nextJetColl(&fJetCollArray);
   while ((jetCont = static_cast<AliJetContainer*>(nextJetColl()))) {
-    
+	  if(jetCont->GetNJets()<1)continue;//Don't enter if there are no jets in the container
```
jdmulligan: Can you please add a comment describing what this section of code is doing? (The whole block here, I mean)

jdmulligan: Please also in general follow the same coding conventions as in the rest of the task (spacing, brackets for if statements)

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
```diff
@@ -2304,21 +2376,27 @@ void AliAnalysisTaskEmcalJetPerformance::ComputeBackground()
...
     // Loop over clusters, if the jet container is for full jets. Sum the cluster ET:
     // (1) in the EMCal, (2) in the EMCal fiducial volume, (3) in the DCal, (4), in the DCal fiducial volume, (5) in the EMCal random cone, (6) in the DCal random cone
-    if (jetCont->GetClusterContainer()) {
-      AliClusterContainer* clusCont = GetClusterContainer(0);
-      
-      AliTLorentzVector clus;
+    TString jetContName = jetCont->GetName();
+    if (!jetContName.Contains("Charged")) {
+
+    	  AliClusterContainer* clusCont = GetClusterContainer(0);
```
jdmulligan: Spacing/indentation

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
```diff
@@ -2304,21 +2376,27 @@ void AliAnalysisTaskEmcalJetPerformance::ComputeBackground()
     
     // Loop over clusters, if the jet container is for full jets. Sum the cluster ET:
     // (1) in the EMCal, (2) in the EMCal fiducial volume, (3) in the DCal, (4), in the DCal fiducial volume, (5) in the EMCal random cone, (6) in the DCal random cone
-    if (jetCont->GetClusterContainer()) {
-      AliClusterContainer* clusCont = GetClusterContainer(0);
-      
-      AliTLorentzVector clus;
+    TString jetContName = jetCont->GetName();
+    if (!jetContName.Contains("Charged")) {
```
jdmulligan: This is not what we discussed! This is how I used to do the check, but changed to the above GetClusterContainer since it is more efficient. I think we should not change it, there is no reason it needs to be changed.

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
```diff
@@ -2344,19 +2422,23 @@ void AliAnalysisTaskEmcalJetPerformance::ComputeBackground()
         }
         
         // (5)
-        deltaR = GetDeltaR(&clus, etaEMCalRC, phiEMCalRC);
+        deltaR = GetDeltaR(&clusVec, etaEMCalRC, phiEMCalRC);
         if (deltaR < jetR) {
           clusPtSumEMCalRC += clusPt;
+          //Check if there is an overlap with a leading/subleading signal jet
+          //if so set the flag to 1 and not take this random cone into accound
+        	  if(IsSignalJetOverlap(0,clusID,jetCont,maxJetIds))
```
jdmulligan: spacing

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
```diff
@@ -2232,30 +2293,35 @@ void AliAnalysisTaskEmcalJetPerformance::ComputeBackground()
     Double_t clusPtSumDCalfid = 0;
     Double_t clusPtSumEMCalRC = 0;
     Double_t clusPtSumDCalRC = 0;
-    
+    Bool_t overlap=0;
     // Loop over tracks. Sum the track pT:
     // (1) in the entire TPC, (2) in the EMCal, (3) in the EMCal fiducial volume, (4) in the DCal, (5) in the DCal fiducial volume, (6) in the EMCal random cone, (7) in the DCal random cone
     // Note: Loops over all det-level track containers. For data there should be only one. For embedding, there should be signal+background tracks.
     AliParticleContainer * partCont = 0;
-    AliTLorentzVector track;
```
jdmulligan: Please be _very_ careful about changing these existing names, in general you should try to avoid that unless necessary. Check all the rest of the code to make sure it has no affect -- I don't have time to verify right now.

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
```diff
@@ -2395,12 +2477,26 @@ void AliAnalysisTaskEmcalJetPerformance::ComputeBackground()
     }
     
     // Compute delta pT, as a function of centrality
-    
-    // EMCal
+    // EMCal acceptance only charged component
```
jdmulligan: A general comment: I don't have time now to understand exactly what your new feature is doing and why. But can  you please make it a configurable option, defaulting to false? So that the code runs exactly as before otherwise? It is best if your studies are a bit separated from the "standard" background computations that are already implemented, both for code readability/comprehension and so that a bunch of histograms aren't unnecessarily duplicated (which I think would be the case in what you've implemented, if I run both a charged and full jet container).

efranzis: I have added now the flag fFillChargedFluctuations that enables the charged fluctuations to be filled. The full fluctuations stay unaffected.

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
efranzis: I carefully checked. It is used only in the subsequent lines below and in the function GetDeltaR(). It is changed consistently everywhere. I changed this because I needed to additionally get the real track object (not just the track vector) to access the ID in the container.

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
efranzis: I tried to fix it. I don't see this well in my editor

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
efranzis: This is not what we discussed since I double checked and there is **at the moment** no getter for the jetType. Markus is looking into it and I can update it once it is added to the container.

This what is changed is NOT what you did before. This check gets a fundamental property of the jet that must not be given to it in a secondary way. It is with it since its creation. The name is created by the type. If type is full jets the name contains "Full" if the type is charged jets the name contains "Charged". Just like it contains the radius in the name or the jet reconstruction algorithm. It is not more efficient to ask for a cluster container that one has to pass to the jet before. Likewise I could pass an element called clown to the jet and then ask if it contains an element clown, as a fancy way of encoding what kind of jet it is. I can likewise add a cluster container to a charged-jet container. That doesn't make it a full jet!
I am really trying to improve this here so please be a little more supportive!

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
efranzis: deleted line 2381. You are welcome

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
efranzis: you have also used such statements, like 1397 or 1505 btw. I have added now the bracket.

efranzis: the line that I have added has a comment attached to it. I didn't change anything else

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
efranzis: I tried to fix it. I don't see this well in my editor

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
efranzis: I tried to fix it. I don't see this well in my editor

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
jdmulligan: Thanks

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
jdmulligan: Yea okay thanks I just wanted to be sure that it was carefully checked

jdmulligan: (I realize you needed also the track object, my comment was directed towards the name modification)

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
jdmulligan: I think you misunderstand me. The check you implement was the check I initially implemented here a while back, but it was suggested and discussed to change to what it currently is since it is more efficient not to do the string comparison so often. Thinking about it again, I don't really care that much. I was surprised to see this change because we discussed a bunch yesterday and then you did something different.

In general it would be good if you open separate PR for new features compared to edits you want to make to the task. I try to be quick to review your new features so you can keep working (I am on vacation as you know), but for changes it's better to have them separate so we can slowly discuss. I know you want to make a lot of improvements, but I am very hesitant to change things since the analysis is nearly published and there is risk to doing these changes (we do not implement any tests to make sure behavior doesn't change -- if you really want to make a lot of changes the ideal thing would be to first implement tests). 

Anyway, I am fine now if you want to make the change after all, I don't have a preference.
But please in future make separate PR for changes to the current code. 
And please look at the diff more carefully in GitHub before the PR.

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetPerformance.cxx
jdmulligan: yes that's why I ask that you check it here to be sure. you may need to change your space/tab usage or settings.

raymondEhlers:
+1

PR #7280, opened 2018-10-25
Provide read access to jet definition at runtime

mfasDa:
Jet definition (algorithm, jet type, recombination scheme)
enum values are cached in the jet container, and getters
are provided. No setters are provided as the jet definition
is not supposed to change after the container is constructed.

In addition adding doxygen tag (///<) to persistent attributes
where missing.


mfasDa:
+1

PR #7283, opened 2018-10-25
PWGGA: add Option 5 for enableQAPhotonTask

jluehder:
enableQAPhotonTask option 2 and option 4 are both called, when option 5 is chosen.


FriederikeBock:
+1

PR #7286, opened 2018-10-25
adding flags to AddTask for setting PID histos, adding histos to Trac…

sartozza:
adding flags to AddTask for setting PID histos, adding histos to TrackCutsHist for PID


a-mathis:
+1

PR #7291, opened 2018-10-26
Sphericity cut as argument in the AddTask and optionally filling the …

pchrista:
…sphericity QA histogram if the flag is enabled


naghmehmohammadi:
+1

PR #7292, opened 2018-10-26
PWGGA Adding Unfolding and true meson jet histograms ROOT5

lizlamers:
Added MC true pi0/eta in jet histograms and MC true jet pt vs rec jet pt histograms in AliAnalysisTaskGammaCalo. Changed AliAnalysisTaskConvJet to be able to obtain MC true jet results from a MC true jet finder.


nschmidtALICE:
+test

nschmidtALICE:
+1

PR #7300, opened 2018-10-26
[ALIROOT-7981]: Fixes to D2H macros for ROOT6 compatibility

jezwilkinson:
fixes ROOT6 compilation errors for multiple PWGHFvertexingHF AddTask macros, and a segfault in the AliAnalysisTaskSEImproveITS task that appeared under ROOT6 due to the streaming of some TGraph objects [now uses TObject::Clone method instead of TGraph's copy constructor].


arossi81:
Hi @jezwilkinson ,
so we cannot use AliFatal and AliError anymore?
Cheeers,
Andrea

jezwilkinson:
Hi @arossi81,
Yes, those calls to AliError/Fatal stopped those macros from running under 6 (in fact I think that the calls to AliError/AliFatal in those macros were classed as syntax errors already under ROOT5; just that in interpreted mode it ran anyway and if it reached that line it crashed where it was "supposed to", while in root6 having them anywhere in the macro prevents compilation altogether)
--J

PR #7301, opened 2018-10-26
some cleanup for running with pp

jaimenorman:
Remove obsolete histograms and fix error messages when running on pp


raymondEhlers:
+1

PR #7333, opened 2018-10-30
attempt to fix centrality selection issue

sebaleh:
attempt to fix centrality selection issue (cannot be reproduced locally)


miweberSMI:
+1

PR #7336, opened 2018-10-30
Adds leading track pT cut

TWSman:
Add a cut to leading track reference jT. The pT of leading track is
required to be 20% of the jet pT.


rbertens:
+1

PR #7351, opened 2018-10-31
Filter for detector level MC simulation. Removes EPOS tracks from event

fkrizek:
This patch is supposed to allow to filter out EPOS part of event from MC simulations where we simulater PYTHIA+EPOS

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskJetExtractor.cxx
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskJetExtractor.cxx
@@ -1275,13 +1275,13 @@ AliAnalysisTaskJetExtractor* AliAnalysisTaskJetExtractor::AddTaskJetExtractor(co
   AliAnalysisManager *mgr = AliAnalysisManager::GetAnalysisManager();
   TString     trackArray         = "tracks";
   TString     jetArray           = "Jet_AKTChargedR040_tracks_pT0150_E_scheme";
-  TString     rhoObject          = "Rho";
+  TString     rhoObject          = "";
   Double_t    jetRadius          = 0.4;
   Double_t    minJetEta          = 0.5;
   Double_t    minJetPt           = 0.15;
   Double_t    minTrackPt         = 0.15;
   Double_t    minJetAreaPerc     = 0.557;
-  TString     suffix             = "_allJets";
+  TString     suffix             = "";
   if(taskNameSuffix != 0)
     suffix = taskNameSuffix;
 
@@ -1293,13 +1293,12 @@ AliAnalysisTaskJetExtractor* AliAnalysisTaskJetExtractor::AddTaskJetExtractor(co
 
   fYAMLConfig.GetProperty("TrackArray", trackArray);
   fYAMLConfig.GetProperty("JetArray", jetArray);
-  fYAMLConfig.GetProperty("RhoName", rhoObject);
+  fYAMLConfig.GetProperty("RhoName", rhoObject, kFALSE);
   fYAMLConfig.GetProperty("JetRadius", jetRadius);
   fYAMLConfig.GetProperty("MinJetEta", minJetEta);
   fYAMLConfig.GetProperty("MinJetPt", minJetPt);
   fYAMLConfig.GetProperty("MinTrackPt", minTrackPt);
   fYAMLConfig.GetProperty("MinJetAreaPerc", minJetAreaPerc);
-  fYAMLConfig.GetProperty("Suffix", suffix);
   fYAMLConfig.Print();
 
   // ###### Task name


PWGJE/EMCALJetTasks/macros/AddTaskEmcalJet.C
```diff
@@ -87,7 +87,7 @@ AliEmcalJetTask* AddTaskEmcalJet(
     AliMCParticleContainer* mcpartCont = new AliMCParticleContainer(trackName);
     partCont = mcpartCont;
   }
-  else if (trackName == "tracks" || trackName == "Tracks") {
+  else if (trackName == "tracks" || trackName == "Tracks" || trackName == "mytracks") {
```
raymondEhlers: In my opinion, this check should only be for standard track collections. Admittedly, it's a bit annoying to have to create a track container by hand, but I think it would also get unmanageable pretty quickly if we added lots custom track collections here.

However, if this is going to be done, the static function `AliEmcalJetTask::AddTaskEmcalJet(...)` must also be modified.

raymondEhlers:
+test

PWGJE/EMCALJetTasks/macros/AddTaskEmcalJet.C
fkrizek: Hi Raymond, in principle my container is supposed to remove EPOS tracks from detector level MC.  For the output "tracks" TClonesArray, I want that the jet finder handles this array as AliTrackContainer. Therefore i had to change also AddTaskEmcalJet.

raymondEhlers: Hi @fkrizek , what I'm trying to say is that I don't think you need to modify AddTaskEmcalJet. As far as I understand from the code and your comment, your task outputs a new TClonesArray named "mytracks" (by default). Is that right? To add that track collection as a track container to AliEmcalJetTask, I'm trying to suggest that you modify the way that you define your jet finder:

```
auto jetFinder = AliEmcalJetTask::AddTaskEmcalJet("mytracks", ...);
// Remove the existing particle container which you don't want.
jetFinder->RemoveParticleContainer(0);
// Add your track collection as a track container.
auto trackContainer = jetFinder->AddTrackContainer("mytracks");
 // Set any necessary options
trackContainer->SetParticlePtCut(0.15);
```

While this is a bit more work for you (and a bit redundant in terms of code), it also means that we don't need to modify a widely shared AddTask for just one task. This type of approach is what we use for the embedding framework. I think it is beneficial both in terms of simplify maintenance and keeping complexity down for shared tasks. What do you think?

fkrizek: Hi Raymond,   if this can be done like that, than indeed it is not necessary to change the AddTaskEmclaJet.   I guess that on the lego train one will then do for the jet finder task
__R_ADDTASK_->RemoveParticleContainer(0);
__R_ADDTASK_->AddTrackContainer("mytracks");

If this is so, then i do not insist on changing AddTaskEmcalJet.  How shall we proceed?
Can you push to the master just the Filter task + its steering macro?

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskSVtaskMCFilter.cxx
```diff
@@ -0,0 +1,191 @@
...
+ 
+
+   GetTrackContainer(0)->ResetCurrentID();
+   AliAODTrack* track = 0;
+   TString nameGen;
+   Int_t lab, mother;
+   AliAODMCParticle *mcpart = NULL;
+   Int_t n = fFilteredTracksArray->GetEntriesFast();
+
+   while ((track = static_cast<AliAODTrack*>(GetTrackContainer(0)->GetNextAcceptParticle()))) {
```
raymondEhlers: As noted [here](http://alidoc.cern.ch/AliPhysics/master/class_ali_particle_container.html#a6363a52ba754afdf0c4c357a281de6f1), this approach is deprecated because it can cause problems with multiple particle or track containers. Please update to using the iterator approach. See [here](http://alidoc.cern.ch/AliPhysics/master/READMEcontainers.html#emcalContainerIterateTechniques) for more information.

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskSVtaskMCFilter.h
```diff
@@ -0,0 +1,67 @@
...
+ private:
+
+    AliAnalysisTaskSVtaskMCFilter(const AliAnalysisTaskSVtaskMCFilter &source);
+    AliAnalysisTaskSVtaskMCFilter& operator=(const AliAnalysisTaskSVtaskMCFilter& source);
+    
+    ClassDef(AliAnalysisTaskSVtaskMCFilter, 1); 
+};
+
+#endif
+
```
raymondEhlers: You'll need to add this the `CMakeLists.txt` (as well as the `LinkDef.h`) to actually be able to run this task on the LEGO train!

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskSVtaskMCFilter.cxx
```diff
@@ -0,0 +1,191 @@
...
+* documentation strictly for non-commercial purposes is hereby granted   *
+* without fee, provided that the above copyright notice appears in all   *
+* copies and that both the copyright notice and this permission notice   *
+* appear in the supporting documentation. The authors make no claims     *
+* about the suitability of this software for any purpose. It is          *
+* provided "as is" without express or implied warranty.                  *
+**************************************************************************/
+//
+//
+// Task for selecting D mesons to be used as an input for D-Jet correlations
```
raymondEhlers: I think this is out of date

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskSVtaskMCFilter.cxx
```diff
@@ -0,0 +1,191 @@
...
+//
+//-----------------------------------------------------------------------
+// Authors:
```
raymondEhlers: Please add yourself

PWGJE/EMCALJetTasks/macros/AddTaskEmcalJet.C
raymondEhlers: Hi Filip,

Yeah, that's exactly how it can be done on the LEGO train.

To proceed, you'll need to modify your commit and then push to the master of your fork. Since modifying a commit will change the git history, you'll need to force push it. I added a few more comments below, so you might want to address them at the same time

PR #7371, opened 2018-11-01
PWGHF/jetsHF: Fixing a problem in the generator level jets

hahassan7:
Fixing a problem in the generator level jets.


fcolamar:
+1

PR #7376, opened 2018-11-01
Filter of detector level tracks (removes EPOS)

fkrizek:
Filter of detector level tracks that removes EPOS from the combined PYTHIA+EPOS simulations

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskSVtaskMCFilter.cxx
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskSVtaskMCFilter.cxx
@@ -1,0 +1,195 @@
+/**************************************************************************
+* Copyright(c) 1998-2009, ALICE Experiment at CERN, All rights reserved. *
+*                                                                        *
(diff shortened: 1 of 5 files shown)


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskSVtaskMCFilter.cxx
```diff
@@ -0,0 +1,195 @@
...
+    //fFilteredTracksArray->Delete();
+    fFilteredTracksArray->Clear();
+
+   // fix for temporary bug in ESDfilter
+   // the AODs with null vertex pointer didn't pass the PhysSel
+   if (!fAodEvent->GetPrimaryVertex() || TMath::Abs(fAodEvent->GetMagneticField()) < 0.001) return kFALSE;
+
+ 
+
+   GetTrackContainer(0)->ResetCurrentID();
```
raymondEhlers: Please remove this line - it was only needed for the old iteration technique and it can cause problems with other track containers

raymondEhlers:
+test

PR #7386, opened 2018-11-03
update TOF cut in LHC15o and purity estimation

dsekihat:
update TOF cut efficiency in LHC15o and purity estimation


kharlov:
+1

PR #7389, opened 2018-11-03
Fix for a wrong index assignment storing NUA and NUE corrections

victor-gonzalez:
The latest modification, for re-using weights when shortening the pT or
zvtx ranges, was wrongly implemented


rbertens:
+1

PR #7398, opened 2018-11-05
Fixes for D0 analysis

mfasDa:
- Consequently put objects which are used only locally within a function on the stack
- Fix indentation of the function CalculateBackground
- Increase invariant mass range for the D0 reconstruction mode


dmuhlhei:
+1

PR #7404, opened 2018-11-05
Use proper non-linearity function for MC

mfasDa:
The non-linearity function must be specified explicitly for simulation,
otherwise the default function kTestbeamCorrectedv3 is used, which is
not the correct one for simulations


mfasDa:
+1

@raymondEhlers , @jdmulligan This restores back the situation in the tender chain (see AddTaskEmcalPreparation.C): MC is using a different non-linearity correction than data. Without specifying one falls back to the default one, which is the one from the testbeam, which is not appropriate for simulations. Thanks @efranzis for spotting this! The config wagons in the MC train need to be fixed as well.

raymondEhlers:
Thanks @efranzis and @mfasDa

PR #7439, opened 2018-11-07
PWGHF/jetsHF: Increasing the number of bins for the IP 

hahassan7:
Increasing the number of bins for the IP to increase the number of working point to be studied.


fcolamar:
+1

PR #7495, opened 2018-11-09
Add JE train helpers for LEGO train configuration

raymondEhlers:
This allows for the extraction of AliEn production variables in the
LEGO train global variables. Control statements such as for loops are
not allowed in the global variables, so this function allows more
complicated extraction and assignment of variables.

Note that a class with static methods is used instead of a namespace
because the are easier to use with rootmaps.

@mfasDa , @jdmulligan


raymondEhlers:
+test

mfasDa:
Looks good to me. If there is more that doesn't work in the global variables it can be put there as well.

raymondEhlers:
+1 Thanks Markus!

PR #7525, opened 2018-11-12
PWGLF: Change setup for the HM trigger mode

BongHwi:
Remove "IsEventSelected" cut in HighMultiplicity Trigger event.
Adjust bin setup and get multiplicity function for HM mode.


bdoenigu:
+1

PR #7529, opened 2018-11-12
[EMCAL-411] Subtraction of online background estimate

mfasDa:
Online background estimated using the median mode
subtracted from the online ADC values. An area
of 1 subregion (4 TRUs per direction) is assumed
as area unit for the rho calculation. The background
estimate is only applied on request, and switched on
automatically in the PbPb2015 mode (covering all run2
PbPb).


mfasDa:
+1

PR #7533, opened 2018-11-12
Address assignment issues on the LEGO train

raymondEhlers:
It appears that we're not allowed to reassign global variables, so we
can't pass variables by reference (See [this test train stdout](http://alitrain.cern.ch/train-workdir/PWGJE/Jets_EMC_pPb/1683_20181112-1624/test/__BASELINE__/stdout)). To avoid this issue, we return a
std::vector from which we can assign the extracted values.

Local tests appear to work

@mfasDa , @jdmulligan

### DIFF
--- a/PWGJE/EMCALJetTasks/AliPWGJETrainHelpers.cxx
+++ b/PWGJE/EMCALJetTasks/AliPWGJETrainHelpers.cxx
@@ -13,9 +13,10 @@
  * provided "as is" without express or implied warranty.                  *
  **************************************************************************/
 
+#include <cstdlib>
 #include <iostream>
 #include <string>
-#include <cstdlib>
+#include <vector>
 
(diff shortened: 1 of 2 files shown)


mfasDa:
Interesting. I see no reason to forbid passing global variables as reference - sounds a bit like a bug.
Please go ahead with your solution - I don't see any obvious reason why it should not work.

raymondEhlers:
+1 I think it's because it's effectively a reassignment, which doesn't appear to be allowed. Below won't compile with clang:

```
// Fine
double test = 5;
// Compilation fails
test = 2;

int main() {
    return 0;
}
```

The way we implement global variables is a bit difference than this, so maybe it's fine. It also could be a ROOT bug. But if we can work around it with this new PR, we don't have to worry about it one way or another

PR #7546, opened 2018-11-13
PWGGA/GammaConv: added new MC flags for isolated photons & more

fjonasALICE:
- added MC flags for isolated photon analysis
- added enum for upcoming gamma-Jet MC production (LHC18b11c)
- fixed electron from frac photon loop in AliAODConversionPhoton.cxx


FriederikeBock:
+1

PR #7553, opened 2018-11-13
PWGHF: Update norm systs pp5TeV

lvermunt:
Update normalisation systematic D-meson analyses pp@5TeV after new VdM scan


arossi81:
+1

PR #7555, opened 2018-11-13
Fix handling of Getenv

raymondEhlers:
Try number 3 for the train helpers: `std::string` cannot be assigned to a nullptr, so we assign the
environment variable to a `const char *` and then assign that to a
string if it is not null.

The `nullptr` occurred up if the environment variable didn't exist. I don't t know I missed this in testing, but apparently I did. It is now testing and works regardless of whether environment variables are defined.

@jdmulligan , @mfasDa Could one of you please give a quick code review? The code is not quite so clean now, but I don't see a super convenient way around it.


raymondEhlers:
+test

jdmulligan:
Looks good to me, thanks for doing this!!

raymondEhlers:
+1

PR #7563, opened 2018-11-14
PWGGA/GammaConv: small fix MaterialHistos

hmurakam:


hmurakam:
Following Nico's suggestion, AOD part removed

nschmidtALICE:
+1

PR #7571, opened 2018-11-14
PWGGA/GammaConv:added THnSparse, alternative to tree output

hmurakam:
-supposed to use single train config to avoid crash


dmuhlhei:
Dear Hikari, could you please elaborate why these additions are needed? Trees are optimal for this purpose, aren't they?

hmurakam:
Hello Daniel, 

in Run2 p-Pb data sample, tree output failed to merge with DCA tree switch on,  even after Nico's improvement. in principle, I can merge by hand after downloading all sub outputs, but it took a very long time, therefore, I added them.

dmuhlhei:
I don't understand your point. Nico implemented the trees to be saved outside the TLists, so that merging is possible again. Indeed, we had problems before but the new approach does work. Could you please check?

hmurakam:
Try to explain the situation a bit ... The setting was both photon and meson DCA trees were on.
I requested train 2 times, and both of them failed to merge.  - I think it's not a temporal issue.
More precisely, merging has failed only tree part and other histograms and BrokenFiles were successfully merged. 
In sub outputs level, they contain everything and it seems only final merging were failed.

dmuhlhei:
after private discussion this ticket should be closed -1

PR #7575, opened 2018-11-15

lhusova:
finer binning in Delta Eta, Delta Phi; add mass info in correlation histograms; 
add Purity check; 
add MC Pt hard for jet events info; 
Change the way of assignment of MC reconstructed to generated


rbertens:
+1

PR #7578, opened 2018-11-15
fix task to run with no PID and allow root6 running

jaimenorman:
Some small fixes to check whether PID is set in the cut file to be able to run with no PID, and fix to add task to be able to run with root6.


fprino:
+1

PR #7581, opened 2018-11-15
updated fix for root6

haandrew:


mfasDa:
+1

@haandrew Why is this commit related to ROOT6? I don't see any line that you changed that could make trouble with ROOT6.

haandrew:
hi @mfasDa 

It's related to my previous update that was to fix for root6, the derivSubOrder was an old argument in the add task that I wasn't specifying in the trains. It was no longer needed so I removed it but I found that I hadn't quite removed every instance of it.

PR #7599, opened 2018-11-16
PWGGA/GammaConv fDoJetQA bool fix

lizlamers:
Fix to ensure that a MesonKind with 2 will not also set fDoJetQA to kTRUE


nschmidtALICE:
+1

PR #7603, opened 2018-11-16
Add flag to only save jets with positive/negative Eta

TWSman:


rbertens:
+1 @TWSman sorry for the delay!

PR #7606, opened 2018-11-16
Updates for correlation analysis

laltenka:
- bugfix when using 3D hadron efficiency map
- new method to correct for hadron efficiency after correlation signal extraction


iarsene:
+1

PR #7608, opened 2018-11-16
[EMCAL-406] add possibility to use TPC outer param instead of inner on ESDs

gconesab:


gconesab:
+test

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionClusterTrackMatcher.cxx
```diff
@@ -77,6 +78,7 @@ Bool_t AliEmcalCorrectionClusterTrackMatcher::Initialize()
   
   GetProperty("usePIDmass", fUsePIDmass);
   GetProperty("useDCA", fUseDCA);
+  GetProperty("useOuterParam", fUseOuterParam);
```
raymondEhlers: Since this is a ESD only option, if `fUseOuterParam` is true, can you please check that `fEsdMode` is true and if not, throw an AliFatal as a cross check?

gconesab: I am not sure this is needed. At some point we might decide that fUserOuterParam=true is the default. True or false, it has no effect for AODs.

raymondEhlers: Right, so I'm thinking more in terms of a user missing the comment and thinking that it does have an effect in AODs. It would just be a sanity check. We of course can't catch every issue, and it's not necessary if people read the docs, but I think it is a simple check that potentially helps out and I don't think hurts anything. But perhaps I'm missing something

gconesab:
+test

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionClusterTrackMatcher.cxx
raymondEhlers: Okay, that's fine by me - thanks Gustavo!

gconesab:
+1

PR #7613, opened 2018-11-17
PWGGA/GammaConv Jet Vector out of range second fix

lizlamers:
The first method to initialize the jet vectors was to initialize them in the ProcessJets function and then only clear them if the number of jets in the event was zero. On a very rare occasion however, the True vectors somehow end up being empty even though they shouldn't. To fix this, the vectors are now initialized only when they are used en they are cleared there shortly after


FriederikeBock:
+1

lizlamers:
Added if statements to protect the GetPdgCode function below because it keeps crashing for PHOS or PCM-PHOS in both GammaCalo en GammaConvCalo. (Added that commit to this pull request because this pull request is not merged with master yet)

FriederikeBock:
+1

PR #7617, opened 2018-11-17
Handle missing matched tracks in EpRatio task

kqf:
It seems that additional check is necessary for the matched tracks, otherwise this task fails.
Here is an example of the stack trace:

https://alimonitor.cern.ch/jobs/stackAnalysis.jsp?path=PWGGA%2FGA_pp_AOD%2F543_20181105-0923


kharlov:
+1

kharlov:
+1

PR #7620, opened 2018-11-18
PWGHF: Update Bplus for MVA studies

lvermunt:
Added a function to cut on PID D0 daughters and a D0 + Bplus invariant mass window before the reconstruction of the Bplus vertex to save CPU time. 

Needed to speed up task for MVA studies Bplus (will probably also replace similar hardcoded selections in AnalysisTask)


arossi81:
+1

PR #7635, opened 2018-11-19
add some setting for QA

sashingo:


sashingo:
+1

sashingo:
Dear expert,
I committed an updated macro which isn't necessary to compile.
And there is no information details about the fail.
So I'm not sure the reason the check wasn't successful.
regards,
Shingo

PR #7640, opened 2018-11-19
EMCal CorrFW: Fatal if the user config is missing

raymondEhlers:
A fatal error is raised early if the user config is missing (either
because it wasn't set or because it couldn't be found due to typo,
not existing, etc). Previously, this generated a warning and there would
be not histogram output, but it is better to explicitly make it a fatal
error.

cc: @jdmulligan , @mfasDa


raymondEhlers:
+1

PR #7644, opened 2018-11-20

lhusova:
unused variables deleted, 
variable size binnig at mass axis, 
less mixing tracks


akubera:
+1

PR #7672, opened 2018-11-22
Addition of the gestion of MC events.

cimordas:
+ Rewritting of the full analysis method and gestion of the TProfiles for the various correlations in order to remove some errors in their values.


rbertens:
+1

PR #7685, opened 2018-11-22
[ALIROOT-7874] Separate refit and non-refit tracks using kTRDUpdate flag

mfasDa:
Usage of kTRDUpdate flag default but made optional. If
it is off, then fall back to old method (at least 3 tracklets)


mfasDa:
+1

PR #7701, opened 2018-11-23
PWGPP/TRDPID: add new TRDPIDResponse.root containing updated referenc…

kagarner:
…e for LHC15o
- change runlists from which parameters generated to DPG runlist 'Tracklets'


chiarazampolli:
+1

PR #7733, opened 2018-11-26
Changed Non-Lin parameters. Now for the first time usable

efranzis:


raymondEhlers:
+1

mfasDa:
@efranzis As far as I can see the enum follows the GA group naming convention and not what was introduced for EMCAL previously (i.e. periods for track cuts) - we would need two definitions now. It would be better to streamline this having only one definition.

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionClusterNonLinearityMCAfterburner.cxx
```diff
@@ -375,15 +375,17 @@ void AliEmcalCorrectionClusterNonLinearityMCAfterburner::InitNonLinearityParam(T
       }
       // pp 5.02 TeV (2017) - LHC17pq
       else if(namePeriod=="k17l3b" || namePeriod=="k18j2" || namePeriod=="k17l4b" || namePeriod=="k18b8") {
```
mfasDa: This is PWGGA conversion group naming scheme - I suggest to change to period names as done in other components of the EMCAL framework in order not to have two definitions per sample.

efranzis: Yes, I agree that the naming scheme is not very excellent. I would like to wait however before investing a lot of time. The reason is that we might not actually need different parameters for each period and generator scheme. This yet has to be checked. And then if confirmed that we need it we can debate about the most elegant solution. If not needed I will anyway delete all these conditions and we just have 1 or 2 sets of parameters. -> I would hope for that case :-)

efranzis:
Thanks to both of you for your quick response!

PR #7744, opened 2018-11-27
Changed parameters to effectivley no correction for the other periods…

efranzis:
…. Fixed bug in the parameter set from Nicolas

@jdmulligan @raymondEhlers @mfasDa 

I changed now the parameters to effectively a scale-factor of 1 for all not yet available parameter sets.
And I fixed a bug in the parameter set from Nicolas that causes a reduction of cluster energy by 50%.


mfasDa:
+test

efranzis:
+1 I will approve now myself since it is urgent that this is fixed for todays tag. When @jdmulligan @raymondEhlers wake up they should just have a look but it is just numbers that have changed in the end. Nothing major to review

raymondEhlers:
Looks fine to me

PR #7762, opened 2018-11-28
PWGGA/GammaConv removed if PHOS statement because problem is fixed an…

lizlamers:
…d fixed possible typo

Removed the if PHOS statement because the PbPb/pPb mystery break in MC AOD for PHOS/PCM-PHOS is now fixed (set enableSortingMCLabels to kFALSE).
When trying to debug this function we also saw a possible typo which is also included in this commit


dmuhlhei:
+1

PR #7768, opened 2018-11-28
New set of parameters to calibrate the dE/dx vs Eta

omvazque:
The calibration of dE/dx vs Eta for the different multiplicity classes is done using two sets of calibration parameters from the different periods of Data that will be analyzed.


bdoenigu:
+1

PR #7778, opened 2018-11-29
Canges in AddTaskDiHadCorrelHighPt.C

lhusova:
new parameters to anable to change the names of output file and output container


rbertens:
+1

PR #7787, opened 2018-11-29
PWGCF: Add macro for train tasks with auto-configured subtrains

akubera:
`AddTaskFemtoSubtrains` forwards subtrain parameters to config macros


akubera:
+1

PR #7806, opened 2018-11-30
Make jetType settable in the jet container and AliAnalysisTaskEmcalJet

efranzis:
To use the JetType information in your task you have to call the AddJetContainer version with
AliAnalysisTaskEmcalJet::AddJetContainer(EJetType_t jetType, ….
or set the jet type by hand after you added the jet container to the task in another way.

@mfasDa added this setter as discussed.


mfasDa:
+1

PR #7822, opened 2018-12-03

lhusova:
Analysis mode options
Autocorrelations fixed
Add a new histogram for purity/contamination correction


akubera:
+1

PR #7827, opened 2018-12-03
Add AliExternalBDTTest.cc macro to test AliExternalBDT class

pfecchio:
AliExternalBDTTest.cc macro introduced with this commit tests the correctness of the prediction provided by the model loaded and compiled by the AliExternalBDT class.


mpuccio:
+1

PR #7838, opened 2018-12-04
PWGDQ/reducedTree: implemented trigger class name to AliReducedEventInfo

laltenka:
and possibility to select on it in AliReducedEventCut


iarsene:
+1

PR #7847, opened 2018-12-05
added new sub-task code, correlates deuterons to EMCal clusters

bschaefe:
minimum threshold is 10 GeV cluster, clusterization is from neutral meson standard


rbertens:
+1

PR #7852, opened 2018-12-05
PWGGA/GammaConv Added a MC generated in jets histogram

lizlamers:
Added a MC generated in jets histogram to determine correct acceptance for neutral mesons in jets


dmuhlhei:
+1

PR #7856, opened 2018-12-05
Fix of output container declaration

fkrizek:


PWGJE/EMCALJetTasks/macros/AddTaskEA.C
```diff
@@ -108,7 +108,7 @@ AliAnalysisTaskEA* AddTaskEA(
    task->SetDebugLevel(0); //No debug messages 0
 
    // output container
-   contHistos = manager->CreateContainer(myContName.Data(), TList::Class(), AliAnalysisManager::kOutputContainer, Form("%s:ChJetSpectra%s", AliAnalysisManager::GetCommonFileName(), myContName.Data()));
+   AliAnalysisDataContainer *contHistos = manager->CreateContainer(myContName.Data(), TList::Class(), AliAnalysisManager::kOutputContainer, Form("%s:ChJetSpectra%s", AliAnalysisManager::GetCommonFileName(), myContName.Data()));
```
mfasDa: There is no need to capture the return value of manager->CreateContainer at all in case you don't need to access it within the function.

mfasDa:
+1

PR #7865, opened 2018-12-06
PWGPP/TRD/TRDPID: add pile-up cut for LHC18q

kagarner:
- add combination of enumerate and switch-case environment to make it possible to switch between pile-up cuts for different PbPb periods


chiarazampolli:
+test
looks ok to me.

chiarazampolli:
+1

PR #7868, opened 2018-12-06
Added MC generated histogram for eta in jets and trainconfig

lizlamers:
Added two histograms for correct acceptance calculation for eta in jets
Also added one trainconfig for jet analysis without jet QA


nschmidtALICE:
+1

PR #7901, opened 2018-12-10
modify Time Calib code to handle runs with PAR/L1 splitting

gyronaut:
Time Calib code is modified to handle runs with PAR/L1 phase splitting.   Modification happened on two levels:

### Analysis:
* **AddTaskEMCALTimeCalib** - Addtask now loads in text file containing PAR information (with workaround to locally copy any remote alien:// path to the local working directory, due to .txt file not being a root file - this is similar to how YAML files are dealt with in EMCAL Correction Framework).
* **AliAnalysisTaskEMCALTimeCalib** - step 1 now produces N+1 time vs. cellAbsID histograms, with N being the number of PARs in the event (which is set by the above txt file). Step 2 and Step 3 now look for multiple L1 phase histograms for an affected run and choose the correct histogram for a given event by calculating the event global BC (calculated as: *eventBC + 3564\*eventOrbit + 59793994260\*eventPeriod* ) and comparing if against the PAR global BCs.

### Post-Processing:
* **AliAnalysisTaskEMCALTimeCalib::ProduceCalibConsts()** - modify static function that takes run-by-run Analysis output and generates the necessary 1D histograms to calculate L1 phase for all PAR regions.
* **AliAnalysisTaskEMCALTimeCalib::ProduceOffsetsForSMsV2()** - modify static function that takes output of previous function and calculates the L1 phase to now separately calculate it for all PAR histograms that were produced. Requires a local copy of the PAR info .txt file.

See also presentation at EMCAL Calibration Meeting: https://indico.cern.ch/event/761682/

### DIFF
--- a/PWGPP/EMCAL/AliAnalysisTaskEMCALTimeCalib.cxx
+++ b/PWGPP/EMCAL/AliAnalysisTaskEMCALTimeCalib.cxx
@@ -294,6 +294,51 @@ void AliAnalysisTaskEMCALTimeCalib::SetL1PhaseReferenceForGivenRun()
 }
 
 //_____________________________________________________________________
+//  Function for Setting L1 Phase reference in case of multiple phases for PAR
+
+void AliAnalysisTaskEMCALTimeCalib::SetL1PhaseReferencePAR(){
+  if(fCurrentPARs.runNumber == 0){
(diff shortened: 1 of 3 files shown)


raymondEhlers:
+test

gconesab:
+1

PR #7932, opened 2018-12-13
Adding option to use AliEventCuts

mfasDa:
With this option the AliEventCuts are used
instead of the event selection implemented
in AliAnalysisTaskEmcal. The trigger selection
from the AliEventCuts is disabled and still
done manually in the task before the event
selection.


mfasDa:
+1

PR #7938, opened 2018-12-13
Start columnar filtering (with @dberzano)

dberzano:


dberzano:
@pzhristov: this is the conversion of current ESDs to the prototype columnar format for Run 3 you did in a separate macro.

Of course this is absolutely not the final version; it is meant to allow us to collaborate by editing it.

We put it (with @mpuccio) into an analysis task and did the following relevant modifications:

* Converted to analysis task
* Added vertex information and centrality
* Defined a unique "event id" (different trees, i.e. tables, can be joined on this number for instance)
* Added a TTree flush every 1000 events (configurable)

cc: @njacazio

PR #7940, opened 2018-12-13
EMCal Trigger Sim QA

moliver813:
The beginnings of a QA task for the EMCAL trigger simulation.

Uses trigger patches from simulation to determine event trigger, build cluster spectra.
More histograms to be added.


raymondEhlers:
+test

PR #7947, opened 2018-12-14
[PWGLF] Fix MC issues: Add Y cut, Add primary cut

BongHwi:
Fix MC issue:

- Add Y cut for both injected/reconstructed particle
- Add IsPrimary() check for both MC particles
- Fix CPA default value issue


bdoenigu:
+1

PR #7993, opened 2018-12-18
Add downscale weight to spectrum task

mfasDa:
Downscale weight always taken from the CENT cluster
(weights same for the CENTNOTRD cluster for EMCAL
triggers) - CALOFAST/PLUS clusters currently not
supported.


mfasDa:
+1

PR #7994, opened 2018-12-18
Add option to use Track DCA cut also on FindMatches like in base class extrapolation method

gconesab:
In case of matching of TPC only tracks and old MCs/Data (not checked in new) the use of track DCA gives large eta residual.


gconesab:
+1

PR #7996, opened 2018-12-18
fix to pointer handling in Lc TMVAApp task

jezwilkinson:
Deletion of secondary vtx within task caused segfault when trying to run multiple wagons of the task on the same train; removing these two lines.
The memory leak these lines were meant to fix can be fixed instead using AliAnalysisTaskSECleanupVertexingHF as the final wagon on the train.


afestant:
+1

PR #7999, opened 2018-12-18
[WIP] Updated columnar filtering 

njacazio:
- Added possibility to exclude Trees
- Added track cuts
- Added information on event time 
 - for now T0Best
- Added information on TOF (TOF clusters)

### DIFF
--- a/RUN3/AliAnalysisTaskAO2Dconverter.cxx
+++ b/RUN3/AliAnalysisTaskAO2Dconverter.cxx
@@ -181,8 +181,15 @@ void AliAnalysisTaskAO2Dconverter::Prune()
     return;
   TObjArray* arr = fPruneList.Tokenize(" ");
   for (Int_t i = 0; i < arr->GetEntries(); i++)
-    for (Int_t j = 0; j < kTrees; j++)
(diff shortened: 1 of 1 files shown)


RUN3/AliAnalysisTaskAO2Dconverter.cxx
```diff
@@ -45,88 +47,143 @@ ULong64_t GetEventIdAsLong(AliVHeader *header)
...
+}
+
+void AliAnalysisTaskAO2Dconverter::Prune()
+{
+  if (fPruneList.IsNull() || fPruneList.IsWhitespace())
+    return;
+  TObjArray* arr = fPruneList.Tokenize(" ");
+  for (Int_t i = 0; i < arr->GetEntries(); i++)
+    for (Int_t j = 0; j < kTrees; j++)
+      fTree[j]->SetBranchStatus(arr->At(i)->GetName(), 0);
```
mpuccio: This will print out a lot of errors: for each tree that does not have the branch you will get an error message.
Do I get this correctly? Either we have a Prune list per tree or we should figure this out.

mpuccio:
+1

PR #8004, opened 2018-12-19
Use AliEventCuts object by default in IsEventSelected

mfasDa:
Old builtin event selection is moved to
IsEventSelectedInternal and needs to be
enabled explicitly. Event selection contains
- Trigger selection (in AliAnalysisTaskEmcal
  or inheriting - not handled by AliEventCuts)
- MC outlier removal
- Event selection by the AliEventCuts
Old optional event cuts (number of tracks in
acceptance, event plane orientation) are
only accessible in the old event selection
and are not (yet) ported to the new one based
on AliEventCuts.

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcal.cxx
+++ b/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcal.cxx
@@ -41,6 +41,7 @@
 #include "AliESDEvent.h"
 #include "AliAODInputHandler.h"
 #include "AliESDInputHandler.h"
+#include "AliEventCuts.h"
 #include "AliEventplane.h"
 #include "AliGenPythiaEventHeader.h"
 #include "AliGenHerwigEventHeader.h"
@@ -118,10 +119,12 @@ AliAnalysisTaskEmcal::AliAnalysisTaskEmcal() :
   fUseXsecFromHeader(kFALSE),
   fMCRejectFilter(kFALSE),
   fCountDownscaleCorrectedEvents(kFALSE),
+  fUseInternalEventSelection(kFALSE),
   fPtHardAndJetPtFactor(0.),
   fPtHardAndClusterPtFactor(0.),
   fPtHardAndTrackPtFactor(0.),
   fRunNumber(-1),
+  fAliEventCuts(nullptr),
   fAliAnalysisUtils(nullptr),
   fIsEsd(kFALSE),
   fGeom(nullptr),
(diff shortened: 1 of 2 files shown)


mfasDa:
Leaving pull request open for comments until 3:45 pm CET.

mfasDa:
+test

mfasDa:
+1

jdmulligan:
Dear Markus,

This commit causes a segfault for me (ROOT6). The message is not very descriptive (see below), but I've isolated it to this commit. I don't see anything obvious in the commit, but can you please take a closer look? Did it run for you?

Also, if I read this commit correctly, it changes the default behavior, potentially for many people. For example, I currently implement AliEventCuts in my own task with custom settings, and this will change the behavior (and not in an obvious way). If my reading is correct, it's absolutely crucial to announce this change to the list!

Best,
James


 *** Break *** segmentation violation
[/usr/lib/system/libsystem_platform.dylib] _sigtramp (no debug info)
[<unknown binary>] (no debug info)
[/Users/jamesmulligan/alisw/sw/osx_x86-64/AliPhysics/root6-1/lib/libPWGEMCALbase.so] AliAnalysisTaskEmcal::UserExec(char const*) /Users/jamesmulligan/alisw/sw/SOURCES/AliPhysics/root6/0/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcal.cxx:0
[/Users/jamesmulligan/alisw/sw/osx_x86-64/AliRoot/root6-1/lib/libANALYSISalice.so] AliAnalysisTaskSE::Exec(char const*) /Users/jamesmulligan/alisw/sw/SOURCES/AliRoot/root6/0/ANALYSIS/ANALYSIS/AliAnalysisTask.h:198

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcal.cxx
```diff
@@ -362,13 +367,21 @@ void AliAnalysisTaskEmcal::UserCreateOutputObjects()
...
+
   if (!fCreateHisto)
     return;
 
   OpenFile(1);
   fOutput = new AliEmcalList();
   fOutput->SetUseScaling(fUsePtHardBinScaling);
   fOutput->SetOwner();
+  
+  if(fAliEventCuts) fOutput->Add(fAliEventCuts);
```
jdmulligan: Does this work? I had always added a TList of the AliEventCuts QA plots here.

jdmulligan:
(Let me also say -- thanks for implementing this. A good improvement to the base class.)

One small additional comment: we could add the event cut QA plots to the output, which could be helpful (see e.g. AliAnalysisTaskEmcalJetPerformance::UserCreateOutputObjects).

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcal.cxx
mfasDa: I use this in my task where I overwrote the IsEventSelected method and have no problems. Adding the event cuts to the output list does not create any troubles, also since it inherits from TList.

mfasDa:
Hi @jdmulligan ,

I will have a closer look. This was basically a back-porting from my task where I overwrote the IsEventSelected method. I don't yet have a clue where the error comes from.

The implementation is necessary as the default event selection implemented in AliAnalysisTaskEmcal is not appropriate for run2. This was the reason for the normalization offset between our analyses (5 TeV and 13 TeV). However the old event selection was not removed but moved to a function IsEventSelectedInternal which the user can still require.

For the moment I guess you were anyhow overwriting the IsEventSelected method with your custom implementation. We can think about how to improve the current implementation to make the event selection more configurable. However the implementation should also focus on simplicity for the custom user.

Cheers

Markus

PR #8015, opened 2018-12-19
Updated Run3 trees prototype

njacazio:
- Added example macro to read TTrees (TOF beta vs TPC momentum)
- Added MC information of the track

### DIFF
--- a/RUN3/AliAnalysisTaskAO2Dconverter.cxx
+++ b/RUN3/AliAnalysisTaskAO2Dconverter.cxx
@@ -31,6 +31,8 @@
 #include "AliESDHeader.h"
 #include "AliESDtrack.h"
 
+#include "AliMCEvent.h"
+#include "AliMCEventHandler.h"
 #include "AliPIDResponse.h"
 
 ClassImp(AliAnalysisTaskAO2Dconverter);
@@ -144,6 +146,13 @@ void AliAnalysisTaskAO2Dconverter::UserCreateOutputObjects()
     Tracks->Branch("fTRDsignal", &fTRDsignal, "fTRDsignal/F");
     Tracks->Branch("fTOFsignal", &fTOFsignal, "fTOFsignal/F");
     Tracks->Branch("fLength", &fLength, "fLength/F");
+    if (fTaskMode) {
+      Tracks->Branch("fPDGcode", &fPDGcode, "fPDGcode/I");
+      Tracks->Branch("fLabel", &fLabel, "fLabel/I");
+      Tracks->Branch("fPDGcodeMother", &fPDGcodeMother, "fPDGcodeMother/I");
+      Tracks->Branch("fLabelMother", &fLabelMother, "fLabelMother/I");
+      Tracks->Branch("fTOFLabel", &fTOFLabel, "fTOFLabel[3]/I");
+    }
   }
   PostData(2, Tracks);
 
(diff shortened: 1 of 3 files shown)


pzhristov:
Hi @njacazio,
I see you are using TDataFrame, but since a while we have RDataFrame and it is not anymore experimental. Which Root6 version did you install?

dberzano:
Indeed, v6-14-04 has RDataFrame, and it's the one you obtain through our `next-root6` defaults (also used here for PR testing).

RUN3/AliAnalysisTaskAO2Dconverter.cxx
```diff
@@ -144,6 +146,13 @@ void AliAnalysisTaskAO2Dconverter::UserCreateOutputObjects()
     Tracks->Branch("fTRDsignal", &fTRDsignal, "fTRDsignal/F");
     Tracks->Branch("fTOFsignal", &fTOFsignal, "fTOFsignal/F");
     Tracks->Branch("fLength", &fLength, "fLength/F");
+    if (fTaskMode) {
```
mpuccio: We should explicitly check that this is equal to 1 (if we want to support more modes).
Moreover, I would then use an `enum` to specify the different task mode, just to have a cleaner code.

RUN3/AliAnalysisTaskAO2Dconverter.cxx
```diff
@@ -206,6 +215,24 @@ void AliAnalysisTaskAO2Dconverter::UserExec(Option_t *)
...
   }
 
+  // Configuration of the PID response
+  AliPIDResponse* PIDResponse = (AliPIDResponse*)((AliInputEventHandler*)(AliAnalysisManager::GetAnalysisManager()->GetInputEventHandler()))->GetPIDResponse();
+  PIDResponse->SetTOFResponse(fESD, AliPIDResponse::kBest_T0);
+  AliTOFPIDResponse TOFResponse = PIDResponse->GetTOFResponse();
+
+  // Configuration of the MC event (if needed)
+  AliMCEvent* MCEvt = nullptr;
+  if (fTaskMode) {
```
mpuccio: See previous comment on the task mode

RUN3/AliAnalysisTaskAO2Dconverter.cxx
```diff
@@ -144,6 +146,13 @@ void AliAnalysisTaskAO2Dconverter::UserCreateOutputObjects()
     Tracks->Branch("fTRDsignal", &fTRDsignal, "fTRDsignal/F");
     Tracks->Branch("fTOFsignal", &fTOFsignal, "fTOFsignal/F");
     Tracks->Branch("fLength", &fLength, "fLength/F");
+    if (fTaskMode) {
+      Tracks->Branch("fPDGcode", &fPDGcode, "fPDGcode/I");
+      Tracks->Branch("fLabel", &fLabel, "fLabel/I");
+      Tracks->Branch("fPDGcodeMother", &fPDGcodeMother, "fPDGcodeMother/I");
```
mpuccio: The PDGcodes will be redundant once we store the full kinematics tree separately. Unless there are other vital reason to keep them, I would remove them already at this stage.

njacazio:
Dear @pzhristov, indeed you are right installed aliroot with ROOT v6-10-08 I have now upgraded the ROOT version to v6.14 and included RDataFrame in the macro. Thanks for spotting this!

RUN3/AliAnalysisTaskAO2Dconverter.cxx
njacazio: I agree on this, It is now done!

RUN3/AliAnalysisTaskAO2Dconverter.cxx
njacazio: So far I added the information of TParticle in a separate TTree for Kinematics only.
I have to add now the information on the MC event.
This will be done as for the tracks, attaching more branches to the event tree.

RUN3/AliAnalysisTaskAO2Dconverter.h
```diff
@@ -37,29 +37,40 @@ class AliAnalysisTaskAO2Dconverter : public AliAnalysisTaskSE
...
     kTOF,
+    kKinematics,
     kTrees
   };
+  enum TaskModes {
+    kStandard = 0,
+    kMC
+  };
   TTree* CreateTree(TreeIndex t);
+  void PostTree(TreeIndex t) { AliAnalysisTask::PostData(t + 1, fTree[t]); };
```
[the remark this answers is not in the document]
njacazio: Agreed. I also set the MC tree as disabled when running in MC mode.

mpuccio:
+1

PR #8023, opened 2018-12-20
Adds new Mixed Triggers mode, and trigger pool saving mode

moliver813:
Adds new Mixed Triggers mode, and trigger pool saving mode.
Small style edits


e-masson:
+1

PR #8030, opened 2018-12-22
Move setup of the event selection to the ExecOnce part in Run

mfasDa:
This part of the code cannot be overwritten in
case the AliAnalysisTaskEmcal is used, but
UserCreateOutputObject can.


mfasDa:
@jdmulligan @adriansev This circumvents the problem discussed in #8004 and initializes the AliEventCuts in the ExecOnce part of UserExec - a function which should never be touched by the user. This is not the most elegant solution but guarantees that the pointer to the event cuts object is never null. Can you please check whether this fixes the problems you reported?

Cheers

Markus

adriansev:
@mfasDa I can confirm that this PR solves my problem. Thanks a lot!!

jdmulligan:
+1 I am not able to test it today, but I will let you know if I have further problems. Thanks!

PR #8012, opened 2018-12-19
Added flexible options for the occupancy ratio calculation for nomina…

efranzis:
…tor and denominator. Fix also indentation

@fkrizek please have a look if this works for you.
We can change now the options with 
SetAreaCalculationDetails
SetAreaCalculationDetails(0, 0) is what you want, this is the default

PR #8062, opened 2019-01-05
Added flexible options for the occupancy ratio calculation for nomina…

efranzis:
…tor and denominator

Total area included into fExcludeAreaExcludedJets condition

### DIFF
--- a/PWGJE/EMCALJetTasks/AliAnalysisTaskRhoSparse.cxx
+++ b/PWGJE/EMCALJetTasks/AliAnalysisTaskRhoSparse.cxx
@@ -26,6 +26,8 @@ AliAnalysisTaskRhoSparse::AliAnalysisTaskRhoSparse() :
   fNExclLeadJets(0),
   fExcludeOverlaps(0),
   fRhoCMS(0),
+  fUseTPCArea(0),
+  fExcludeAreaExcludedJets(0),
   fHistOccCorrvsCent(0)
 {
   // Constructor.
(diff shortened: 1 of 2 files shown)


efranzis:
Hi @fkrizek,

please have a look if this version fulfills all your needs. It unfortunately opened a new PR instead of adding to the old one #8012.

Kind Regards,
Eliane

PWGJE/EMCALJetTasks/AliAnalysisTaskRhoSparse.cxx
```diff
@@ -131,57 +134,91 @@ Bool_t AliAnalysisTaskRhoSparse::Run()
...
+    // but to the factor with which this rho is scaled down
+    if (fExcludeAreaExcludedJets==0 && jet->GetNumberOfTracks()>0)
+    {
+      TotaljetAreaPhys+=jet->Area();
+      // Total area of all found jets ghost+phyical jets. This is a proxy
+      // for the available detector area in which the rho could have been calculated
+      TotalAreaCovered+=jet->Area();
+    }
+    // Exclude leading background jets (could be signal)
+    if (iJets == maxJetIds[0] || iJets == maxJetIds[1])
```
fkrizek: Can these lines 161 + 162 be moved before the line 149?

fkrizek:
Hi @efranzis  and  @mfasDa 

i still see that the condition for exclusion of   two leading jets comes only after    summing up jet areas.
I have made note in the code what i think needs to be changed.

Thanks
Filip

PWGJE/EMCALJetTasks/AliAnalysisTaskRhoSparse.cxx
efranzis: Hi @fkrizek,

please look at L191. If you set fExcludeAreaExcludedJets==1 which I recommended, then the area is summed after the jets are excluded and L153-159 is ignored. I think this is what you want

fkrizek: Hi @efranzis.

yes, you are right!   I guess you can commit the code then.
Cheers
Filip

PR #8081, opened 2019-01-07
Adding event cut historgrams directly to EMCAL list

mfasDa:
Adding the event cut object introduced a segfault
in the merging in the destructor of AliEventCuts
for several EMCAL framework users. In order to
circumvent this the histograms are stored directly
in the EMCAL list. This also restores the behaviour
of the internal event selection in which the cut
histograms are stored directly in the EMCAL list.


mfasDa:
+test

mfasDa:
+1

PR #8100, opened 2019-01-09
Add option to set a user-configured event cuts object

mfasDa:
In order to allow for user-configured event cuts
objects (in the AddTask) an option to set the
event cut object is added. In case of a valid event
cuts object the object is not constructed any
more in the ExecOnce part but only the histograms
are added to the output list. Option sets usage
of internal event selection to false.

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcal.h
+++ b/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcal.h
@@ -1,7 +1,32 @@
+/************************************************************************************
+ * Copyright (C) 2017, Copyright Holders of the ALICE Collaboration                 *
+ * All rights reserved.                                                             *
+ *                                                                                  *
+ * Redistribution and use in source and binary forms, with or without               *
+ * modification, are permitted provided that the following conditions are met:      *
+ *     * Redistributions of source code must retain the above copyright             *
+ *       notice, this list of conditions and the following disclaimer.              *
+ *     * Redistributions in binary form must reproduce the above copyright          *
+ *       notice, this list of conditions and the following disclaimer in the        *
+ *       documentation and/or other materials provided with the distribution.       *
+ *     * Neither the name of the <organization> nor the                             *
+ *       names of its contributors may be used to endorse or promote products       *
+ *       derived from this software without specific prior written permission.      *
+ *                                                                                  *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND  *
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED    *
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE           *
+ * DISCLAIMED. IN NO EVENT SHALL ALICE COLLABORATION BE LIABLE FOR ANY              *
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES       *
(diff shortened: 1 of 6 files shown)


mfasDa:
@raymondEhlers @jdmulligan Please check whether this fulfils your needs for making the event cuts user-configurable.

raymondEhlers:
Hi @mfasDa ,

Thanks for this!

One question that I think could simplify the implementation: Why does the AliEventCuts object need to be allocated on the heap? Couldn't we just as easily allocate it on the stack and then just ignore it if using the internal event selection? It's not a very heavy object, and then anyone customizing it doesn't need to create the object itself, which I think is a better user experience (especially because they don't have to worry about the proper arguments to AliEventCuts). What do you think?

Best,
Raymond

mfasDa:
Hi @raymondEhlers,

Thanks for the review! So you suggest to create the object on the stack and a getter returning a non-const reference to the object? Could work - I just hope it is not raising any issue with streamers ...

If it works no objection from my side.

Cheers

Markus

raymondEhlers:
Hi Markus,

Yeah, that's what I was thinking. James and I both already stream the AliEventCuts object without any apparent issues, so I don't think there should be streamer problems.

Best,
Raymond

mfasDa:
Hi @raymondEhlers , Hi @jdmulligan ,

Please have a look at the latest change and tell me if it is what you were thinking of! Note that I had to rename fUseInternalEventSelection to fUseBuiltinEventSelection because you already defined this attribute in AliEmcalCorrectionTask.

If it is fine with you I will squash before merging.

Cheers

Markus

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcal.h
```diff
@@ -324,7 +350,23 @@ class AliAnalysisTaskEmcal : public AliAnalysisTaskSE {
   void                        SetMinBiasTriggerClassName(const char *n)             { fMinBiasRefTrigger = n                              ; }
   void                        SetTriggerTypeSel(TriggerType t)                      { fTriggerTypeSel    = t                              ; } 
   void                        SetUseAliAnaUtils(Bool_t b, Bool_t bRejPilup = kTRUE) { fUseAliAnaUtils    = b ; fRejectPileup = bRejPilup  ; }
-  void                        SetUseInternalEventSelection(Bool_t doUse)            { fUseInternalEventSelection = doUse                  ; }
+
+  /**
+   * @brief Use internal (old) event selection
+   * @param[in] doUse It true use the old internal event selection instead of AliEventCuts
+   */
+  void                        SetUseInternalEventSelection(Bool_t doUse)            { fUseBuiltinEventSelection = doUse                  ; }
```
raymondEhlers: I know that this name probably has to be maintained for backwards compatibility, but as you noted, we already used this term in the embedding helper for something rather different. Would you mind also adding a

```c++
void SetUseBuiltinEventSelection(Bool_t doUse)  { fUseBuiltinEventSelection = doUse; }
```

and marking the previous one as deprecated? I think this would help clarify the separate usage, especially since it's been renamed internally.

mfasDa:
+test

mfasDa:
+test

mfasDa:
+1

e-masson:
Hi @mfasDa,

As I try to have a working train test I obtain a fatal error which seems related to the changes discussed here, as it is new compared to train tests prior to these changes:
```
F-AliEventCuts::AutomaticSetup: Automatic period detection failed, please use the manual mode. If you need this period in AliEventCuts send an email to the DPG event selection mailing list.
F-AliEventCuts::AutomaticSetup: aborting execution due to AliFatal
```
Did I miss an e-mail saying that these changes would impact every task based on `AliAnalysisTaskEmcal` and the appropriate things to do in that case? The full error log of my train test is available [here](http://alitrain.cern.ch/train-workdir/PWGGA/GA_pp_AOD/604_20190115-1050/test/__ALL__/stdout).

Best,
Erwann

PR #8102, opened 2019-01-09
PWGGA/GammaConv Changed name of output container

lizlamers:
Changed name of output container so more jetfinders can be run alongside in a train


nschmidtALICE:
+1

PR #8112, opened 2019-01-09
Quiet down debug messages

raymondEhlers:
These messages are overwhelming other debug messages, so we set them to
a lower priority.

@mfasDa Are you okay with this minor change? At the current level, I'm finding it rather difficult to debug my task


raymondEhlers:
+test

mfasDa:
@raymondEhlers Fine with we

raymondEhlers:
+1 Thanks Markus!

PR #8165, opened 2019-01-13
yaml file for EMCal electron analysis, MC 

deepathoms:
example yaml file for configuring EMCal correction tasks for MC


deepathoms:
+1

PR #8169, opened 2019-01-14
Instructions and macros for vdM analysis

jgcn:
First set of instructions and macros used in the analyses of van der Meer scan data.
This set of macros is not yet complete, but a first basic analysis is possible using it.


rbertens:
+1

PR #8175, opened 2019-01-14
PWGGA/GammaConv Rewritten ConvJet class for mutiple jet finders in on…

lizlamers:
…e train

Almost completely rewritten ConvJet class to be able to handle mutiple jet finders in the same train.
You need to add a JetContainer name with the corresponding trainconfig. When something gets asked about the jets (inside AliAnalysisTaskGammaCalo or AliAnalaysisTaskGammaCaloIso) the corresponding trainconfig needs be given to give the correct corresponding jet characteristics.

The addTasks are also changed to load the trainconfig into the GammaCalo and GammaCaloIso class
And the ConvJet AddTask is changed to handle the input arguments for the class.


nschmidtALICE:
+1

PR #8188, opened 2019-01-15

lhusova:
Deleting of arrays, changeble number of pt bins, add eta info to purity correction histograms


rbertens:
+1

PR #8192, opened 2019-01-15
PWGGA/GammaConv Improvement of ConvJet class

lizlamers:
Improved the class so that the number of true jet containers is a global variable instead of going over the total number of containers which is a combination of rec and true containers in MC
Furthermore improved it that two Trainconfigs with the same jetfinder can also run alongside now


dmuhlhei:
+1

PR #8197, opened 2019-01-15
TRDPID: add Parameters for LHC17g to TRDPIDResponse.root

kagarner:
Warning: parameters for LHC17g based on truncated TRD signals due to sampling with larger number of time bins


fprino:
+1

PR #8213, opened 2019-01-17
Increase jet pt range to 300 GeV/c

mfasDa:
Allowing for feed down to lower det-level jet pt
in the range covered by the jet trigger.


mfasDa:
+1

PR #8222, opened 2019-01-17
update to run embedded correlation

jaimenorman:
add option to run h+jet analysis using both Pb-Pb and embedded particle containers


mfasDa:
+1

PR #8231, opened 2019-01-18
TOF pid params updated for PbPb 2018

noferini:
This is just an update of TOF PID params in OADB with a new entry for PbPb 2018


noferini:
Adding Pietro and Roberto as watcher
@preghenella 
@pantonioli

fprino:
+1

PR #8236, opened 2019-01-18
New Tasks and Classes for proton-Omega femto analysis.

otonvd:
-Added tasks and classes based in the FemtoDream code.
-Now Xi and Omega are treated independently and can be directed into the otuput.
-Config File added so the AddTask can be set via alien.
-Tree for Xi and Omega selection added as additional output.


bhohlweger:
+test

bhohlweger:
+1

a-mathis:
+1

PR #8237, opened 2019-01-18
Update for p-Sigma0 femto

a-mathis:
o Use global tracks instead of TPConly for the protons
o Further improvements for memory handling
o Further tests


a-mathis:
+1

PR #8243, opened 2019-01-18
PWGGA/GammaConv: Adding more kinenmatic histograms for MC truth

mfasDa:
- Adding more kinematic histograms (pt, eta, phi) for the
  heavy meson and its decay products at various steps
  (all, in channel, reconstrutible, reconstructed)
- Fix indentation in various places
- Modernize: Use nullptr instead of NULL


mfasDa:
Some histograms mainly needed for the D0: Look at the kinematics of the particle and the decay products but not only for reconstructed but also for all and reconstructible D0s.This is needed in order to check whether the decay products are in a kinematic range so that they can be reconstructed via the channel with Pi0 at all. Might also be helpful for the other heavy mesons.

mfasDa:
Merge conflicts solved - review can start!

mfasDa:
Hi @FriederikeBock , Hi @nschmidtALICE ,

I added the AOD MC case and resolved the merge conflicts. Can it be merged now?

Cheers

Markus

nschmidtALICE:
+1

PR #8259, opened 2019-01-21
cleanup and mem leak fixes

loizides:
Cleaned now no warning is left, all ptrs are set properly to zero.
Memory leak should be fixed as well.


loizides:
+1

PR #8263, opened 2019-01-21
Fixing root6 support for PhotonIsolation task.

rqwa:
To achieve root6 support the option for individual config files is
removed for root5 and root6.


e-masson:
+test

e-masson:
+1

PR #8278, opened 2019-01-22
Vertex selection on original events implemented in EmbeddingHelper

kimbyumkyu:
A functionality added in AliAnalysisTaskEmcalEmbeddingHelper.* concerning the vertex cut on original events. When embeddingHelper->SetMaxVertexDistance(2); setter is enabled (distance between the primary vertices of embedding and embedded events), it takes infinite time to find a correct embedded event. 
For example, when an original event has a z vertex like 27 cm, then it fails to find an MC event having z vtx in 25 to 29 cm, and it goes to an infinite loop.

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
+++ b/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
@@ -116,6 +116,7 @@ AliAnalysisTaskEmcalEmbeddingHelper::AliAnalysisTaskEmcalEmbeddingHelper() :
   fMCRejectOutliers(false),
   fPtHardJetPtRejectionFactor(4),
   fZVertexCut(10),
+  fInternalZVertexCut(999),
   fMaxVertexDist(999),
   fInitializedConfiguration(false),
   fInitializedNewFile(false),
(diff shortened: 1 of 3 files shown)


kimbyumkyu:
Hi Raymond, 
Thank you for information :-).  Then I will retreat the commit. By the way, isn't this cut needed? For example, if an original Pb-Pb data has a z_vtx at -5cm and the embedded event at 5 cm, then a jet is reconstructed with respect to -5cm and isn't the pythia jet also reconstructed with the zvtx at -5cm and doesn't it give a different jet pt than what we input?

So, I was testing by requiring the max distance. I'd like to know if I am wrong.

raymondEhlers:
Hi Beomkyu,

To briefly address your pre-edit question (which is quite a reasonable question), applying a max vertex difference distance cut would be preferred. However, I know of no reasonable way to implement it given the grid requirements. Every possible solution that I've thought up would require so much file IO as to be impractical. The reason that it is so slow now is that a vertex which satisfies the requirements shows up infrequently enough that the embedded file is changed every few events (at best), which is way too expensive, as opening additional files for embedding is generally slow.

Further, since we combine the tracks and clusters at the reconstructed level, I think it's okay. I believe that the track pt should be right regardless, although the cluster E does depends on the vertex. However, I don't think that is going to be the dominant effect, especially since the clusters are likely to contain many data cells given the high occupancy in (central) PbPb.

Best,
Raymond

kimbyumkyu:
Hi Raymond, 
you are right. I just removed my question because I realized that pts are merged when I was writing. :-) Thanks a lot for your nice comment. 

Best regards,
Beomkyu

PR #8279, opened 2019-01-22
PWGGA/GammaConv Get rid of compiler warning and make ConvJetReader mo…

lizlamers:
…re efficient

the compiler warning in AliAnalysisTaskGammaCalo and AliAnalysisTaskGammaCaloIso is now removed
I added that the ConvJetReader class first checks if a jet container is already added with the same name before it adds it to the class. That way, it doesn't need to loop over more jet containers which are the same anyway


nschmidtALICE:
+1

PR #8285, opened 2019-01-22
Several cross-checks for Femto

a-mathis:
o Track splitting/merging at high multiplicities
o Do we see Lambda or photon clones?


a-mathis:
+1

PR #8295, opened 2019-01-23
PWGGA/GammaConv: Added AddTask

hmurakam:


dmuhlhei:
Dear Hikari, could you please rework this request and implement your configs into the existing AddTasks since we do not support such kind of AddTasks. Thanks, Daniel

FriederikeBock:
Hi Hikari,
could you put back the 5 TeV configs into the standard AddTasks, as also others might be using them please. Otherwise the new AddTask in this user specific folder is ok. However, we do not guarantee any maintenance of addtasks in that folder if arguments to the task are changed, this is your responsibility and only yours.
After the configs are back in the nominal addtasks I will approve the commit.
Regards,
Friederike

hmurakam:
put back 5TeV configs  ... & added AddTask_hikari.C

PR #8299, opened 2019-01-23
PWGGA/GammaConv Added public constructor without arguments

lizlamers:
Added public constructor without arguments to hopefully make it able to work on the grid


nschmidtALICE:
+1

PR #8315, opened 2019-01-23
Adding task filling softdrop response

mfasDa:
The task relies on functionality from RooUnfold. It will be
built only in case RooUnfold is available.


mfasDa:
+test

Will rely on RooUnfold (ROOT6-builds only).

mfasDa:
+1

PR #8326, opened 2019-01-24
Changing the way the TMVA interface/library is committed/compiled/used.

chiarazampolli:
In this way we can compile for several pT bins at the same time.
The common class has also been streamlined in a separate file.


fprino:
+test

fprino:
+1

PR #8333, opened 2019-01-24
Add TPC splines and maps for 16l.2, 17k.1, 18b.1, 18d.1, update 17p.1

wiechula:
The update for 17p.pass1 is trying to improve the electron splines at
high momenta, this may influence all analyses. Analysers using the splines of this period before are asked to check the influence on their analysis.


fprino:
+1

PR #8344, opened 2019-01-25
PWGHF: Fix pp/pPb bug for Dstar introduced by PreSelect

lvermunt:
The commit o f the PreSelect function for Dstar, broke the code for pp and pPb. Added two possibilities now to get the D0 candidate, for when candidate is filled vs not filled. Tested for all three systems


afestant:
+1

PR #8348, opened 2019-01-25
PWGGA/GammaConv: further fixes for heavy meson task on AODs

fjonasALICE:
- added correct handling of true pure- and mixed calo information
- added relabeling of true AOD conversion photons


nschmidtALICE:
+1

PR #8356, opened 2019-01-27
Small improvements for Sigma0 analysis

a-mathis:
o Fantastic CINT needs NULL instead of nullptr
o Additional debug histos for the track clones in ESDs


a-mathis:
+1

akubera:
Just start running everything on `vAN-*_ROOT6` so you can use c++11 features

bhohlweger:
@akubera how to do that with the trains, can I somehow specify it when submitting them? Cheers

akubera:
You could ask us, but honestly it'd be difficult for us to remember which ones to select.

I think the best method would be to agree on a prefix to the wagon name, like ROOT6_... or just R6_, and then the operators will know to group those.

The more involved solution would be a change to the train system/website to include a "ROOT6" checkmark in the wagon setup, and some other identification (like background or font color) to indicate to the operators which root to use.

I'll send an email and we'll discuss it this week.

PR #8368, opened 2019-01-28
Adapt build cases for Fastjet 3.3.2 and 3.2.1

mfasDa:
Interface change in Recluster for the new fastjet
version, which moved from the namespace fastjet::contrib
to the namespace fastjet. Affected are only classes
directly accessing the SoftDrop.


mfasDa:
+test

mfasDa:
+1

PR #8377, opened 2019-01-29
PWGGA/GammaConv Adjusted includes for ConvJetTask

lizlamers:
Added more includes and removed using namespace std; and added std::vector everywhere in the hope that that will fix the grid problems.
Furthermore, I do not actually want to push the root streamers commit but I can't find how to delete it. I am hoping that by my second commit that discards those changes


nschmidtALICE:
+1

PR #8384, opened 2019-01-29
Changing static void  MakeMissingChambersAliases() to be public

AMechler:


AMechler:
This change is needed because the automatic QA crashes for some detectorsystems.
This commit should hopefully fix that issue

miranov25:
+1

PR #8420, opened 2019-01-31
PWGGA/GammaConv: fixed missing histograms & wrong checks in omega task

fjonasALICE:
- fixed missing histograms & wrong checks in omega task
- introduced test configs for pp 5 TeV and pp 13 TeV for omega


nschmidtALICE:
+1

PR #8435, opened 2019-02-01
PWGGA/GammaConv: option to require ITS hits

hmurakam:
any ideas ?


hmurakam:
once close this PR, and re-implement according to Friederike's suggestion

PR #8438, opened 2019-02-01
PWGGA/GammaConv added configs for peripheral PCMEMC PbPb systematics …

mikesas:
…and PCMPHOS nonlin LHC17pq correction


FriederikeBock:
+1

PR #8453, opened 2019-02-02
PWGHF: Check for memory leak LEGO train PbPb

lvermunt:
We observe a memory leak (only) when running PbPb on the train system, probably related to the writing of our big TTrees. Unfortunately, we can't reproduce the leak locally. Earlier commit didn't fix it, hopefully this one will.


afestant:
+1

PR #8457, opened 2019-02-04
Changed default to false in AliFemtoDreamCollection

bhohlweger:
Sets default to false for all pairs in order not to cause a segmentation violation


bhohlweger:
+1

PR #8463, opened 2019-02-04
pi/K/p spherocity analysis

omvazque:
I have included the mid-rapidity multiplicity estimator in order to contrast both spherocity using V0M and mid-rapidity estimator.


eperezle:
+1

PR #8468, opened 2019-02-04
PWGHF: MVA, Fix for running mesons separate

lvermunt:
Fixing two bugs:
* When no particle TTrees were enabled, the task crashed because the event selection cuts were not initialised.
* Because of a copy-paste mistake, the Ds, D+, and Dstar couldn't be run standalone.


afestant:
+1

PR #8473, opened 2019-02-04
fix PID selection bugs

jaimenorman:
Change PID selection check to use the IsSelected function , rather than the more specific IsSelectedPID which incorrectly uses Nsigma PID
Switch off PID before checking topological cuts


afestant:
+1

PR #8475, opened 2019-02-04
D0 in pPb for 1020,2040,4060 added

cterrevo:
cuts for D0 in finer pt bins in pPb 0-100  and for centrality ranges 1020, 2040, 4060.


afestant:
+1

PR #8482, opened 2019-02-05
[RooUnfold] Improve check for scale object support

mfasDa:
Check for scale object support outsourced to new function.
This allows to add RooUnfoldResponse to the supported
object types including the check whether AliPhysics was
built with RooUnfold support.


mfasDa:
+1

PR #8484, opened 2019-02-05
PWGHF: MVA, Fix memory leak event cuts

lvermunt:
* Fixing memory leak of event cuts (mainly) in centrality selection
* Moving filling of 'IsEventRejectedDueTo...' to correct position (after first call IsEventSelected(aod))


arossi81:
+1

PR #8491, opened 2019-02-05
PWGHF: MVA, Adding evCuts as class variable

lvermunt:
Full fix of the memory leak. Just adding 'delete evCuts' was not enough, as some of the destructors in the AliRDHFCuts derived classes are missing


afestant:
+1

PR #8497, opened 2019-02-05
Clean up task and add new features

a-mathis:
o Use FemtoDreamEventCuts
o Choose automatically the right filter bit
o Slightly modified cuts and tests
o Sigma0 yield analysis in separate class


a-mathis:
+1

PR #8528, opened 2019-02-07
PWGGA/GammaConv: Fixing small memory leaks

mfasDa:
AliVCluster pointer handled by std::unique_ptr. In addition
remove background handlers and not just the array containing
them in the destructor.


dmuhlhei:
I'm fine with these changes, @nschmidtALICE or @FriederikeBock could you please have a look as well?

nschmidtALICE:
+test

nschmidtALICE:
Also looks okay from my side. I'll first test and then approve it.

nschmidtALICE:
+1

PR #8564, opened 2019-02-08
Some histograms are not optional (debug mode)

chiarazampolli:
The concerned histograms are the 2D with the BDT variables vs BDT response.
They are needed for debugging purposes, and take a lot of memory.
In this commit, also the number of bins was decreased.


afestant:
+1

PR #8578, opened 2019-02-11
Use vector as internal storage for combined track selection results

mfasDa:
TClonesArray is a very inefficient object for the underlying
user storage. Usage of std::vector improves performance by almost
factor 2.


mfasDa:
Work in progress - don't yet merge! It works under ROOT6 but not sure it does under ROOT5.

mfasDa:
+test

mfasDa:
+test

mfasDa:
+1

PR #8623, opened 2019-02-13
Add splitting of the response matrix for MC closure test

mfasDa:
Splitting done as additional dimension in the THnSparse
filled with 0 or 1,  where the tag is assigned based on
uniform random sampling. By default 80% of the jets
are used for the response matrix and 20% for the prior
spectrum.


mfasDa:
+1

PR #8636, opened 2019-02-14
Initial draft of jet tree for HFtree framework

jdmulligan:
WIP: Please do not approve yet.

@ginnocen, @nzardosh, @fgrosa, @lvermunt,

Please have a look at this first draft of the jet tree. The jet info itself is minimal at this point (pT, eta, phi, etc.), but the infrastructure is there to easily add more.

The basic idea is as follows: One should run a `JetFinder` task before the `TreeCreator` task (which will write a jet array to the InputEvent), and then in `TreeCreator::AddTask` one should add a line to attach a jet container corresponding to that jet array. This mimics the standard workflow in ALICE jet framework -- the idea of the jet container is to automatically handle access to the jet array, and to provide a configurable set of cuts. The `JetTreeHandler` will then construct a std::vector of all of the jets (subject to whatever criteria one wants to configure via the jet container) in each event. The jet info is encapsulated in a small class `AliJetInfoCompressed`, which takes advantage of the `Double32_t` type to minimize the footprint.  One can add as many jet finders as you like, one branch will be written for each.

Some notes:
* There is just a single JetTreeHandler at the moment, rather than an inheritance structure, but this can easily change depending what direction we decide to go. 
* I haven't added yet the MC tree info
* Event selection will inherit whatever is used for the HF trees
* Track selection should be configured independently of the HF track selection, using track containers (hybrid tracks)
* I had to add a few dependencies to the treeHF compilation, I think it should not cause any problem but it would be good if someone more familiar with HF can verify.

I have only run this locally on a small number of events, so it has not been tested super rigorously. But I wanted to get this draft out there so we can discuss. Please let me know what you think, and we can decide how to proceed. 

Best,
James

### DIFF
--- a/PWGHF/treeHF/AliAnalysisTaskSEHFTreeCreator.cxx
+++ b/PWGHF/treeHF/AliAnalysisTaskSEHFTreeCreator.cxx
@@ -74,6 +74,8 @@
 #include "AliHFTreeHandlerBplustoD0pi.h"
 #include "AliHFTreeHandlerDstartoKpipi.h"
 #include "AliHFTreeHandlerLc2V0bachelor.h"
+#include "AliEmcalJet.h"
+#include "AliRhoParameter.h"
 #include "AliAnalysisTaskSEHFTreeCreator.h"
 
 using std::cout;
@@ -171,14 +173,32 @@ fnV0A(0),
 fFillMCGenTrees(kTRUE),
 fDsMassKKOpt(1),
 fLc2V0bachelorCalcSecoVtx(0),
-fTreeSingleTrackVarsOpt(AliHFTreeHandler::kRedSingleTrackVars)
+fTreeSingleTrackVarsOpt(AliHFTreeHandler::kRedSingleTrackVars),
+fWriteNJetTrees(0),
+fVariablesTreeJet(0),
+fTreeHandlerJet(0),
+fLocalInitialized(kFALSE),
+fJetCollArray(),
+fRhoName(),
+fRho(0),
+fRhoVal(0),
+fFillPtUncorr(false),
+fFillArea(true),
+fFillNConstituents(true),
+fFillZLeading(true),
+fFillRadialMoment(true),
+fFillpTD(true),
+fFillMass(true),
+fFillMatchingJetID(false)
 {
 
 /// Default constructor
+  
+  fJetCollArray.SetOwner(kTRUE);
 
 }
 //________________________________________________________________________
-AliAnalysisTaskSEHFTreeCreator::AliAnalysisTaskSEHFTreeCreator(const char *name, TList *cutsList):
+AliAnalysisTaskSEHFTreeCreator::AliAnalysisTaskSEHFTreeCreator(const char *name, TList *cutsList, int fillNJetTrees):
 AliAnalysisTaskSE(name),
 fEventNumber(0),
 fNentries(0x0),
@@ -265,11 +285,28 @@ fnV0A(0),
 fFillMCGenTrees(kTRUE),
 fDsMassKKOpt(1),
 fLc2V0bachelorCalcSecoVtx(0),
-fTreeSingleTrackVarsOpt(AliHFTreeHandler::kRedSingleTrackVars)
+fTreeSingleTrackVarsOpt(AliHFTreeHandler::kRedSingleTrackVars),
+fWriteNJetTrees(fillNJetTrees),
+fVariablesTreeJet(0),
+fTreeHandlerJet(0),
+fLocalInitialized(kFALSE),
+fJetCollArray(),
+fRhoName(),
+fRho(0),
+fRhoVal(0),
+fFillPtUncorr(false),
+fFillArea(true),
+fFillNConstituents(true),
+fFillZLeading(true),
+fFillRadialMoment(true),
+fFillpTD(true),
+fFillMass(true),
+fFillMatchingJetID(false)
 {
     /// Standard constructor
-    
-    
+  
+    fJetCollArray.SetOwner(kTRUE);
+  
     if(fFiltCutsD0toKpi){
     delete fFiltCutsD0toKpi;fFiltCutsD0toKpi=NULL;
     }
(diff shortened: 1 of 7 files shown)


lvermunt:
Hi @jdmulligan,

Thanks a lot! I had only a quick look (will go in more detail later this week). Two quick comments already though:
 * Is there a reason you don't set the MaxVirtualSize for the jet TTree? ("->SetMaxVirtualSize(1.e+8/nEnabledTrees);")
* Please put an if-statement before calling "FillJetTree()", to check if it was enabled in the AddTask. 

And can you also put WIP in the title of the PR, to be sure it is not merged by accident :)?

jdmulligan:
Hi @lvermunt 

Thanks for taking a look! I implemented your suggestions.
(The first was not enabled when I started developing this, so I missed it -- the second was purely an oversight!)

jdmulligan:
Hi @ginnocen, @nzardosh, @fgrosa, @lvermunt,

I added the two features that we discussed yesterday:
* A separate tree is now written for each jet finder / container
* The jet variables are now stored in a flat vector structure (like the HF trees), instead of inside a class (which may pose problems for analysis of the tree)

Please let me know if you have other thoughts/suggestions.
Once we agree on the basic picture I'll go ahead an implement more details.

Cheers,
James

jdmulligan:
Hi @ginnocen @lvermunt 

I added a commit which allows one to fill truth-det matching information into the jet tree. It works like the following: If jets in tree A are matched to jets in tree B, then each jet entry in tree A stores the index of the jet in tree B that it is matched to (and vice versa). Then one can compute deltaR, deltaPt, compare jet substructure observables, etc in the analysis stage. Please let me know if you see any issue with this approach. It is an efficient way to store the information and allows full flexibility (as opposed to directly storing deltaR, deltaPt, etc.), but one has to do a little work at the analysis stage to get the information (which I haven't thought in detail about).

The jet matching itself is done in a separate task. 

I also realized there is no need to separate "Gen" trees in the case of jets, since we just manually add all the jet containers that we desire; all jet containers are treated the same from the perspective of the TreeCreator and TreeHandler.

jdmulligan:
Hi @ginnocen @lvermunt @nzardosh,

I added another commit with a few more quick jet observables: Z_leading, jet mass, radial moment, pT,D. I am hesitant to spend more time in this direction (e.g. reclustered observables such as SoftDrop zg, or implementing background subtraction for the jet shape observables) until we have a more clear idea of what we will want. That being said, please share suggestions!

From my side it would be good to get this PR approved soon so that we can test on the train -- I think the set of features for this first iteration is complete.

Best,
James

lvermunt:
Hi @jdmulligan

I went through your commit in detail, see my comments below. I'm not an expert on the jet framework, so about that part there was not a lot I could say. Maybe someone else can have a look there? Also @fgrosa and @afestant might want to have a look before it is merged (also because I will be offline from tomorrow on).

Three general comments:
 * I see you created 8 'fFill*' variables for 8 branches (although not yet used). Is is possible to reduce this number? Because they should all be set in the arguments of the AddTask I guess, where we already have 22 parameters. For the HF TTrees we for example use an enum for different PID options and another one for single track variables. So we have several options to enable/disable branches with only two variables.
 * I remember a discussion about the EventCuts on Slack. What did you decide for this? It seems you are using the HF objects right? Fabrizio had a point here that he wanted to discuss with you.
 * Please make sure that you don't introduce memory leaks. As this is a heavy task, a big effort end January was needed to remove all (small) leaks so the task was able to run on PbPb (with a high success rate on the LEGO trains). We have not yet started this production for the HF mesons on PbPb, so it would be a pity if new memory leaks are introduced. I think I spotted a small one in your PR (you don't delete fRho), but I would suggest you run a check with valgrind before merging.

And a few smaller ones:
 * You defined the functions SetFillPtUncorr till SetFillMatchingJetID twice, both in the JetTreeHandler and the JetTreeCreator. Is this intended? Or a leftover?
 * As this will be the first PR to AliPhysics, the ClassDefs should be just +1 right? Now you do +3 for the TreeCreator, and you start with 4 for the JetTreeHandler
 * You use Pt and pT in your variables/branch names, please use the same ('PtCorr' and 'pTD' for example). Or is this intended?

Cheers,
Luuk

ginnocen:
Hi Luuk, I will have a look at the "physics" content. I think most important at this stage is to check that the technical details are ok, so that we can take it from there. 
Thanks to all, Cheers GM

fgrosa:
Hi @jdmulligan, I went through your code (not yet to the last commit) and I agree with the comments of @lvermunt and @ginnocen. I would say that if you implement the suggestions of @lvermunt and you make sure that the code works, we could go ahead with the approval of the PR. Thanks!

Since @lvermunt mentioned the normalisation of the events with AliEventCuts, I thought that the vertex efficiency correction with the TH1 would not be optimal for us, because if you want to apply a selection on the events offline on the tree (e.g. on the centrality), then you don't have any information for the correction of the events that you selected. Therefore the optimal solution would be to find a way to get a flag from AliEventCuts to store in the tree the reason why the events were rejected (as we do now with the HF cut object). I think that the introduction of AliEventCuts can be added in a second PR, after we find how to do that properly.

jdmulligan:
Hi @lvermunt @fgrosa ,

Thanks for a thorough review!

To address your points:

- The "fFill" options don't need to go into the AddTask argument, they rather can just go in the AddTask customization. The variables we will use are still under discussion, for now I just put all of these setters there for now for maximum flexibility, in general probably only one or two may be needed. 

- Regarding AliEventCuts, yes let's discuss this separately from this PR. I would still advocate that we migrate at some point in the near future, since it gives the benefit of standard, well-studied cuts. I will take a look, I think it should be no problem to store the rejection reason event-by-event. Centrality of course should be stored regardless.

- Okay I will take a look at valgrind. Why do you say fRho is leaking? (maybe I overlook something but I don't see any dynamic allocation). 

- The Setters are defined separately in the TreeCreator and TreeHandler, since there are separate private members in each class.

- The ClassDef incrementation is because there are multiple commits, and in principle I think one should keep the commits correct (if anything I may have forgotten to increment some of them). If you really want I can squash the commits, but personally I prefer to have separated commits to keep a more detailed code history. 

- I think the only case of this is pTD, I would argue you should just view it like a completely unrelated observable to Pt.

Best,
James

jdmulligan:
(I squashed the commits while fixing a conflict that got merged on master)

jdmulligan:
Hi All,

Can someone please approve this PR?

Thanks,
James

arossi81:
+test

afestant:
+1

PR #8642, opened 2019-02-14
Implemented comparison operator

rhaake:


rhaake:
@mfasDa, what do you think?
I would like to use the operator for easily using std::sort.

rhaake:
+test

mfasDa:
Absolutely in favour of this. Not only this might be useful for std::sort but maybe also for other algorithms from the stl. Eventually an operator== should be added as well.

rhaake:
+1

rhaake:
+1

PR #8643, opened 2019-02-14
Add separation of signal and bkg in MC for  XiC+ -> Xi pi pi 

gluparel:
Add possibility to separate signal and bkg in MC for XiC+ -> Xi pi pi and minor modification of the AddTask


arossi81:
+1

PR #8650, opened 2019-02-14
Fix copy constructor in AliEmcalParticleConstituents

mfasDa:
- Call Copy constructor of AliEmcalJetConstituent
- Fix unitialized values


mfasDa:
+1

PR #8653, opened 2019-02-14
Recognize LHC11h in MultSelection

raymondEhlers:
I only found this out accidentally, but despite the [twiki](https://twiki.cern.ch/twiki/bin/viewauth/ALICE/AliMultSelectionCalibStatus) marking LHC11h as calibrated for use with MultSelectionTask, the period auto detection doesn't appear to work. I've added the run number so that it's properly identified. The run numbers are from the RCT

### DIFF
--- a/OADB/COMMON/MULTIPLICITY/AliMultSelectionTask.cxx
+++ b/OADB/COMMON/MULTIPLICITY/AliMultSelectionTask.cxx
@@ -2717,6 +2717,7 @@ TString AliMultSelectionTask::GetPeriodNameByRunNumber() const
     
     //Registered Productions : Run 1 Pb-Pb
     if ( fCurrentRun >= 136851 && fCurrentRun <= 139517 ) lProductionName = "LHC10h";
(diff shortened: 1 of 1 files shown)


ddobrigk:
Hi @raymondEhlers, yes, yes, that was on purpose... the autodetection was disabled deliberately because the LHC11h calibration hasn't been completely validated so far and was accessible only via bypasses. The point is that because this period used triggers, the min-bias sample used for calibration is exceptionally small and runs had to be grouped together to get a more reliable calibration. 

I would still suggest not to go through with making this fully available in the framework before running a few analysis checks to make sure that the results from AliMultSelection are identical to the ones from AliCentrality... This because once this is deployed there is a large chance that people will use it and never check this equivalency.

raymondEhlers:
Hi @ddobrigk , Ah, okay, thanks for the explanation! As I said, I found this on accident, so when I noticed that it didn't auto detect it, I glanced through a bit of the history and couldn't find an explanation. I figured that it may have been accidentally missed and I decided to lend a hand (In my experience, I find it easy to accidentally overlook a period when maintaining those type of functions). I'm happy to close :-)

PR #8655, opened 2019-02-15
PWGGA/GammaConv: fixed missing bracket in AddTask

fjonasALICE:
sorry, I don't know how I missed this. I think my local testing environment uses a different AliPhysics version that I thought it is using ...


dmuhlhei:
seems your test ran with root5 +1

dmuhlhei:
+1

PR #8664, opened 2019-02-15
PWGGA GammaConv Added Jet Pt mixing

lizlamers:


dmuhlhei:
+test

PWGGA/GammaConv/AliAnalysisTaskGammaCalo.cxx
```diff
@@ -6520,11 +6520,16 @@ void AliAnalysisTaskGammaCalo::CalculateBackground(){
             }
           }
           if(JetNearEMCal){
-            Double_t zbin = fBGHandler[fiCut]->GetZBinIndex(2);
```
dmuhlhei: Do you really want to get rid of this?

lizlamers: Yes because I now I want to mix with the actual z bin instead of trowing all background candidates in the same z bin. And the variable definition is also not a problem because it is defined at the top of the CalculateBackground function

dmuhlhei: Ok, just wanted to make sure this is not an unwanted removal

dmuhlhei:
+1

PR #8683, opened 2019-02-15
Removing hardcoded setting for centrality selection (leftover...)

chiarazampolli:
...and also setting the default value of the argument to decide
whether to store or not the debug histograms


afestant:
+1

PR #8698, opened 2019-02-18

laltenka:
- AliReducedAnalysisJpsi2eeCorrelations: added possibility to mix between triggered and MB events
- AliAnalysisTaskReducedTreeMaker, AliReducedTrackInfo: added momentum on EMCal surface for E/p calculation


iarsene:
+1

PR #8699, opened 2019-02-18
changes the close pair rejection criterium to the radius

bhohlweger:
Fixes a bug. Instead of using ((dPhiMin)^2 + (dEtaMin)^2)^2 it uses as intended (dPhiMin)^2 + (dEtaMin)^2


bhohlweger:
+1

PR #8715, opened 2019-02-18
 fix library dependencies to correctly include AliAnalysisUtils

zwound40:
... which is in library OADB. ALICEanalysis can be removed from dependencies


akubera:
+test

akubera:
+1

PR #8721, opened 2019-02-19
changes bins of matching histo so that it can be used as a flexible RM

efranzis:
@mfasDa or @raymondEhlers could you please approve it for today's tag? @jdmulligan is informed but still sleeping ;-)


mfasDa:
+1

efranzis:
Thank you @mfasDa

PR #8726, opened 2019-02-19
Fix exclusion of Dalitz decays

mfasDa:
Exlcusion was formulated in a postive way, so
operator must be "==" instead of "!="


nschmidtALICE:
+1

PR #8733, opened 2019-02-20
Separates event trigger into EMCal and DCal triggers.

moliver813:
Separates event trigger into EMCal and DCal triggers.
Corrects a swap between EG1 and EG2, etc


raymondEhlers:
+1

PR #8746, opened 2019-02-20
Added AliAnalysisTaskUPCforward and changed CMakeLists and LinkDef.

siragoni:
Added a new class for the analysis of charmonium states decaying to opposite sign dimuon for the forward region. Added $Class.{h, cxx} and the Add{$Class}.C macro. Added the source file to the CMakeLists.txt and added the pragma statement for the linker (with the corresponding +). INITIAL COMMIT.

### DIFF
--- a/PWGUD/UPC/AddTaskUPCforward.C
+++ b/PWGUD/UPC/AddTaskUPCforward.C
@@ -1,0 +1,31 @@
+AliAnalysisTaskUPCforward* AddTaskUPCforward(TString name = "name")
+{
+    // get the manager via the static access member. since it's static, you don't need
+    // to create an instance of the class here to call the function
+    AliAnalysisManager *mgr = AliAnalysisManager::GetAnalysisManager();
+    if (!mgr) {
+        return 0x0;
+    }
+    // get the input event handler, again via a static method.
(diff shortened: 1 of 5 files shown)


siragoni:
Dear All,
may I know whether this code is suitable for merging or if further work on my part is needed?
This AnalysisTask is only the first I need for the LEGO trains.
Many thanks.
Best regards.

siragoni:
A further comment: I am regularly using this task on local, gridTest and GRID basis. As such, it should be safe...

mbroz84:
+1

siragoni:
Dear All,
many thanks for the approval.

I did not expect it to fail however...
I normally run it without problems giving the expected result.
The command is explained here:
https://root.cern.ch/doc/master/hlabels1_8C.html

I have found a workaround here:
https://root.cern.ch/content/main-histogram-changes-root-6
In this case where the labels are ALL alphanumeric, it seems this option is automatically set.
Hence my only change has been to comment the faulty line out, and the output is exactly the same...
I am not sure how to feel about this.

Please, forgive me for the inconvenience.
Many thanks.
Best regards.

mbroz84:
+1

PR #8771, opened 2019-02-21
Add task investigating outliers in the response matrix

mfasDa:
Outliers are defined as part jets with pt >> event
pt-hard and small det jet pt.


mfasDa:
+1

mfasDa:
+1

PR #8774, opened 2019-02-21
Add gaussian fit to hResidualTrend

strogolo:
Add gaussian fit to the ResidualTrend distribution after background subtraction.


fprino:
+1

PR #8782, opened 2019-02-22

laltenka:
- momentum on calo for AOD tracks (forgot to add in previous commit)
- momentum on calo for E/p if available
- calo M02 for matched tracks


iarsene:
+1

PR #8797, opened 2019-02-25
JetQA: Fix memory leak in pt-hard scaling script

jdmulligan:
(update to post-processing script)

@efranzis With this change I am able to successfully scale pt-hard bins in alidock


mfasDa:
+1

PR #8800, opened 2019-02-25
fix for AOD path, and fixing order of initialization

btrzecia:
Fix for AOD path string, and fixing order of variable initialization in constructors.


fcolamar:
+1

PR #8810, opened 2019-02-25
Fix crash when recycling embedded events

raymondEhlers:
Regardless of the state of the builtin event selection, we need to be able to track when embedded events are recycled. Therefore, the rejection histogram must always be defined.

@mfasDa - Are you okay with this modification?


raymondEhlers:
+test

raymondEhlers:
+1 I'm interpreting @mfasDa's thumbs up as agreement, and I've now fully tested the changes, so I'm approving

PR #8815, opened 2019-02-26
Use built in AliEventCuts

raymondEhlers:
Move from my custom implemented version to the centralized version in
`AliAnalysisTaskEmcal`


raymondEhlers:
+1

PR #8820, opened 2019-02-26
Uses the correct number of pairs for Close pair rejection

bhohlweger:
Before fWhichQA was wrongly used to do so. Additionally removes unnecessary  arguments from the Systematics task


bhohlweger:
+

bhohlweger:
+1

PR #8825, opened 2019-02-26
Change histogram for jet finding efficiency

mfasDa:
- One histogram with ptpart, ptdet and status code
  (0 - all, 1 - tagged, 2 - accepted)
- Loop over all part. jets in acceptance
Adding also the information whether the particle
level jet was in the fiducial acceptance of the
detector level jet to the response matrix.


mfasDa:
+1

PR #8829, opened 2019-02-26
PWGLF: Update analysis code for systematic study

BongHwi:
Add Systematic option.
Modify AddTaskMacro to configure by using text input


BongHwi:
Could you approve this commit so that I can test with today's AliPhysics tag ?

BongHwi:
Got it... hot fix applied, this should be no problem.

PR #8835, opened 2019-02-27
By default use the Armenteros-Podolandski cut for the Sigma0

a-mathis:
o Plus some systematic variations to check the effect on the purity


a-mathis:
+1

PR #8837, opened 2019-02-27
PWGGA/GammaConv: added new MC Cluster energy corrections for EDC and …

jokonig:
…PCMEDC for the 13TeV datasets from 2016 and 2017


nschmidtALICE:
+1

PR #8846, opened 2019-02-28
[WIP] Workaround for photon double counting issue in PYTHIA8

mfasDa:
Use AliEmcalMCTrackSelector in order to reject photons
that are the mothers of other photons. The track selector
task needs to run in this case also for AOD analyses
as otherwise it cannot be guaranteed that the mother
photons are removed before.

### DIFF
--- a/PWG/EMCAL/EMCALtasks/AliEmcalMCTrackSelector.cxx
+++ b/PWG/EMCAL/EMCALtasks/AliEmcalMCTrackSelector.cxx
@@ -1,8 +1,29 @@
-//
-// Class to select particles in MC events.
-//
-// Author: S. Aiola
-
+/**************************************************************************************
+ * Copyright (C) 2012, Copyright Holders of the ALICE Collaboration                   *
+ * All rights reserved.                                                               *
+ *                                                                                    *
+ * Redistribution and use in source and binary forms, with or without                 *
+ * modification, are permitted provided that the following conditions are met:        *
+ *     * Redistributions of source code must retain the above copyright               *
+ *       notice, this list of conditions and the following disclaimer.                *
+ *     * Redistributions in binary form must reproduce the above copyright            *
+ *       notice, this list of conditions and the following disclaimer in the          *
+ *       documentation and/or other materials provided with the distribution.         *
+ *     * Neither the name of the <organization> nor the                               *
+ *       names of its contributors may be used to endorse or promote products         *
+ *       derived from this software without specific prior written permission.        *
+ *                                                                                    *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND    *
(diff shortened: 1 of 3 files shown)


mfasDa:
Work in progress - don't merge yet!

mfasDa:
Workaround for double-counting issue for photons at particle level seen in PYTHIA8 jet-jet productions (see ALIROOT-8172). More tests needed. In particular for AODs this might create problems as the original tree structure became modified.

mfasDa:
+test

mfasDa:
@efranzis @raymondEhlers @jdmulligan What do you think? Could this raise potential issues for the embedding?

mfasDa:
+test

efranzis:
> @efranzis @raymondEhlers @jdmulligan What do you think? Could this raise potential issues for the embedding?

Well I am not an embedding expert. But at leas this will fix the strange outlier problem in the response. Thank you!

raymondEhlers:
Hi @mfasDa,

At a quick glance, I think it's fine. I don't think `AliEmcalMCTrackSelector` is adapted to work fully with the embedding (line 92 where it grabs the `InputEvent()` would need to be modified), but I don't think your change has any impact on that compatibility one way or another. Is there something in particular that you're worried about?

Best,
Raymond

mfasDa:
Hi @raymondEhlers ,

First of all if you take one of the existing PYTHIA8 productions for embedding you will need to use the MC particle also in the embedding case because otherwise you count the photons twice or multiple times in the jet finder at particle level, so `InputEvent()` will not be sufficient. What I am then worried is that in the new list we remove particles but we don't manipulate the labels, the indices in the container do not reflect the position of particles in hierarchy as for the standard branch - this still will need to be used in case of access to mothers / daughters.

Cheers

Markus

raymondEhlers:
Hi @mfasDa,

Okay, I think I better understand the issue now. I thought you were asking if this would interfere with the embedding, but now I see that you're asking whether the embedding will work properly using this task.

I haven't read through the task in great detail, but I think for this to work properly with the embedding framework, the idea would be to retrieve the right input event (either internal or external) depending on the `IsEmbedded()` status of the attach particle containers. I could imagine doing this by hand, but we may also be able to leverage some other code that is used to handle this already: [`AliEmcalCorrectionEventManager`](https://github.com/alisw/AliPhysics/blob/master/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionEventManager.h) (which we can rename - it doesn't really rely on the correction FW - it's just used there). The setup of the input is implemented in `AliEmcalCorrectionTask`. It's definitely a hack, but it let us seamlessly drop in as a replacement for `InputEvent()` - basically just `InputEvent()` -> `fEventManager.InputEvent()`.

I'm still not very familiar with how we perform the MC indexing (I'm assuming that you're not talking about the global index used in `AliEmcalContainerIndexMap`), so I'm not sure I can comment on how well that will work. But embedding doesn't touch that part, so if it works in the standard case, then it should work with embedding.

Best,
Raymond

mfasDa:
+1

Small scale tests were OK, large scale tests now to be done

PR #8851, opened 2019-02-28
PWGGA/GammaConv Fixed memory leak and undefined rapidity cout

lizlamers:
Added a delete Cluster0 and delete Cluster1 to fix memory leak
Changed IsSignal boolean in IsMesonSelected to stop "undefined rapidity" cout


nschmidtALICE:
+1

PR #8856, opened 2019-02-28
Improve debug information for physics selection

raymondEhlers:
We clearly report the problematic key if an invalid key is passed.


raymondEhlers:
+1

PR #8874, opened 2019-03-01
Added AliAnalysisTaskUPCforwardMC. Added histos to AliAnalysisTaskUPCf*.

siragoni:
Created a new AnalysisTask, AliAnalysisTaskUPCforwardMC, with the sole purpose to analyze the MC for the PbPb AODs. The class is more or less based on AliAnalysisTaskUPCforward but the large difference in histogram content and code in the UserExec() justifies the need for a new class... The CMakeLists.txt and Linker Definitions have been changed consistently. New histograms have been added to the AliAnalysisTaskUPCforward reflecting the need to investigate new properties of the J/Psi (plus the indentation has been changed here and there). 
IMPORTANT: the two codes have been tested on local and they can analyse data properly. They both should be pretty safe to merge... 
Many thanks. 
Regards.


mbroz84:
+1

PR #8881, opened 2019-03-01
Add non-standard MC particle container

mfasDa:
Needed when running with MC track selector
task in order to remove double counting
of photons


mfasDa:
+1

PR #8886, opened 2019-03-02
EMCal Embed: Validate physics selection

raymondEhlers:
If other analysis tasks have broader physics selections than the embedding helper, then those other tasks can access an already used embedding event when the task is selected and the embedding helper isn't. To address this, the embedding helper will checks the other tasks connected to the analysis manager and throws an error if:

- The internal physics selection is broader than the embedding helper physics selection.
- Any other task has a broader physics selection than the embedding helper.

This should ensure that the are no events should are selected for other tasks, but not the embedding helper. It will now also require explicit configuration of the internal physics selection if not using the automatic physics selection from AliEventCuts. This will ensure that it isn't changed unexpectedly.

I tested this in a variety of scenarios, and it seems to work as expected.

cc: @jdmulligan


raymondEhlers:
+test

raymondEhlers:
+1

PR #8901, opened 2019-03-04
Modified the computation of the multiparticle correlations.

cimordas:
* Made the code able to compute all 2-p and 4-p correlations and the needed 6-p correlations between the harmonics v1 to v6 without input of the harmonics from the user.
* Add a new cut for the track selection.
* Remove old computations of 3-p correlations, now useless.


naghmehmohammadi:
+1

vpacik:
+1

PR #8914, opened 2019-03-05
Added way to scale hMatching distance histogram and min/max pT hard bin

efranzis:
@jdmulligan, here is the commit that adds the scaling of the hMatchingDistance histogram and adds the possibility to only scale certain pT hard bins


efranzis:
@raymondEhlers could you approve this please?

raymondEhlers:
+1

PR #8939, opened 2019-03-06
PWGUD/UPC: Updated the MC macro and changed the ZDC part of the AliAnalysisTaskUPCforward.

siragoni:
Following the UPC meeting of 5/3/2019, accordingly changed the ZDC part of AliAnalysisTaskUPCforward to check in inconsistencies for the fZNAenergyH and fZNCenergyH. Based on these changes, the differential neutron emission part of the analysis has seen new changes. The AliAnalysisTaskUPCforwardMC has seen changes too, with the addition of the Pt plot for the Dimuon pairs at Generation level.
IMPORTANT: both changes have been checked on local. 
Regards.


mbroz84:
+1

PR #8941, opened 2019-03-06
Reverted selection of protons: Back to all protons selected (instead …

otonvd:
…of requiring the presence of at least one omega candidate)


a-mathis:
+1

PR #8964, opened 2019-03-07
Updated TOF QA

njacazio:
- Computing TOF-T0 once per event -> CPU performance improved
- Added T0 mask vs TOF hit multiplicity TH2
- Added feature to used AliTOFT0v1 in calib. mode and fill the corresponding histograms


fbellini:
+1

PR #8974, opened 2019-03-08
AliAnalysisTaskHFJetIPQA: open up different probabilities of jet flav…

kagarner:
…our definition:
-add possibility to ask for matching of flavour that defines the jet with mother of particles that have been clustered by the jet finder
	* save all jet constitutens in  fJetCont
	* loop over MC particles, and ask whether daughters correspond to jet constituents
	* if not store daughter label and original flavour in map daughtermothers and compare all following particles
	* if matching found, continue as with former method
- add flags for setting minimal numbers of jet constituents (fNoJetConstituents), cut on difference between daughter particle and jet axis (fDaughtersRadius), use matching of flavour to particles within jet cluster (fDoFlavourMatching)
- consider IP distributions seperately for strangeness and u-/d- quarks, add corresponding histograms
	* fill strange and light quarks in vectors
	* sort for highest pt
	* chose parton with largest pt as jet flavour


fcolamar:
+1

PR #8978, opened 2019-03-08
PWGUD/UPC: Added bin migration matrix and code for the helicity analysis of J/Psi.

siragoni:
Added bin migration matrix and code for the helicity analysis of J/Psi. The concerned files are those of the MC study, hence AliAnalysisTaskUPCforwardMC.{h, cxx}. 
IMPORTANT: the code has been tested on local. The bin migration trend is linear as expected, possibly very mild bin migration effect in Coherent J/Psi to dimuon.


mbroz84:
+1

PR #8981, opened 2019-03-08
[EMCAL-264,EMCAL-270,EMCAL-271,EMCAL-265] OADB/EMCAL: Upload of bad channel maps LHC18jkl and range adjustment LHC18m

nschmidtALICE:


FriederikeBock:
+1

PR #8991, opened 2019-03-09
Add one input for subwagon

jikim1290:
Add one input for subwagon of Legotrain


vpacik:
+1

vpacik:
Hi, @jikim1290 after looking at the code before merging I just realised the following issue now. Even though you added suffix argument into you AddTask macro, you forget to modify the name of you output container accordingly. Without appending suffix to your output container name, running with sub-wagons will not work. So if you intent to run with subwagons, make the corresponding modifications as described [on TWiki](https://twiki.cern.ch/twiki/bin/viewauth/ALICE/AnalysisTrains#Subwagons)

jikim1290:
Dear @vpacik, Thanks you very much for your comments! I will correct it.

PR #8997, opened 2019-03-09
New Analysis task to study leading hadron angular correlation for dif…

anmishra0:
…ferent multiplicity classes


anmishra0:
New Analysis task to study leading hadron angular correlation for different multiplicity classes

aortizve:
Hello Aditya. Before approve the code, could you please confirm whether you need all histograms defined in your analysis task? I can see that your analysis task is very similar to AliAnalysisTaskUePid, actually all histograms defined in UePid are also defined in your analysis task.  I suggest to debug the code.

PR #9003, opened 2019-03-11

lhusova:
add the possibility to change the multiplicity estimator and to enable the pile-up cuts for tracks andV0 separately
change the pseudorapidity to rapidity cut for generated V0


vpacik:
+1

PR #9005, opened 2019-03-11
Measurement of masses and asymmetry parameters of cascades

gskorodu:
New observables for the measurement of parity violation in cascade decays are included.


eperezle:
+1

PR #9006, opened 2019-03-11
Removing functionality for kResponseMatrix

btrzecia:
Removing functionality for the old kResponseMatrix correlation method.


fcolamar:
+1

PR #9009, opened 2019-03-11
MC weights from charged pions at 5TeV produced from 2.76TeV measurement, rescaling it by the ratio of charged particle spectrum at the two energies

mfaggin:
(HFE analysis in PbPb, 5 TeV, mfaggin)


adubla:
+1

PR #9019, opened 2019-03-11
Add switch to use centrality-correlations cut (Pb-Pb)

strogolo:
Add the possibility to activate and apply the centrality-correlations cuts from AliEventCuts in the analysis of the Pb-Pb data samples. The default is to keep these cuts off.
In particular it was specified to used SetupPbPb2018 (for 18q and 18r) and SetLHC15o (for 15o).


afestant:
+test

afestant:
+1

PR #9021, opened 2019-03-11
Store previous event selection and embedded status

raymondEhlers:
Allows us to keep track of rare doubly corrected embedded events


raymondEhlers:
+1

PR #9022, opened 2019-03-11
Fully move AliEmcalJetTask AddTask to static method

raymondEhlers:
Previously, we had two copies (one as a static method in the class, and one
as a .C macro), which then diverged when one got updated. This now
adopts the same approach as used in a number of other tasks, where the
macro just wraps the static method defined in the class.

@mfasDa -  This commit merges the changes that were made to the .C macro
into the static method and then wraps the static method. This way we only have
one AddTask to maintain. What do you think?


raymondEhlers:
+test

mfasDa:
Hi @raymondEhlers ,

Thanks for the implementation! I am fully in favour of moving the framework add macros to static add functions! Implementation looks OK - should be merged!

Cheers

Markus

raymondEhlers:
+1

PR #9051, opened 2019-03-13
PWGUD/UPC: Added rapidity study for helicity J/Psi and templates for Pt-fit.

siragoni:
Changes involve my analysis tasks, so AliAnalysisTaskUPCforward{MC}.{ h. cxx }. Added the rapidity study for the helicity of the J/Psi. This is to check whether there are scarcely populated regions in acceptance and potentially exclude them for the analysis. Preliminary studies on a small sample on local actually show that this is the case... Added the properly set code enabling the creation of the templates for the Pt-distribution fits.
IMPORTANT: the code has been tested on local, both tasks compile and yield the expected results. The entries for the rapidity study properly add up to the rapidity integrated case...


mbroz84:
+1

siragoni:
Dear All,
would you be able to check what the problem is?
I can see even through the logbook that there should be no problem at all with the code itself, but it may be possible that the problem lies elsewhere...?
I am fairly sure because there are no big differences compared to before, and the code works well on local.
Regards.

PR #9053, opened 2019-03-13
Increase pt-range at detector level up to 300 GeV

mfasDa:
Needs increase of the pt-range at particle level up to
600 GeV as well.


mfasDa:
+1

PR #9074, opened 2019-03-14
Implementation VZERO vs TPC clusters pile-up cut

strogolo:


fprino:
+1
NOTE: I approve this commit, so that you can start using the feature in the D0 analysis.
I was planning to add these cuts in AliRDHFCuts, in order to have them available for all mesons and all tasks. I will commit the modified version of AliRDHFCuts in the next days.
Once we have the correlation cuts there, I suggest to use the cut from AliRDHFCuts.

PR #9082, opened 2019-03-15
Adding refitted parametrisation file for Pb-Pb 5 TeV (LMee)

miweberSMI:


nschmidtALICE:
Dear @miweberSMI ,
the file contains several severely broken fits. Could you please recheck that the base fits are fine?
brokenfits

miweberSMI:
hi @nschmidtALICE,

thanks for checking. Actually I need just one centrality, which seems to be broken in the initial file (the one without "refit"). Could you check that one as well?
I just blindly called the same macro that was used for the preliminary cocktail and didn't check too much.

Cheers,
Michael

nschmidtALICE:
The original file also contains broken fits... someone might have missed that before. If you really only need one centrality, then could you either create a file containing only this centrality or update the existing file with just a new parametrization for this centrality.
It seems wasteful to upload a 0.5MB file if only a fraction of its content is needed.

miweberSMI:
Hi @nschmidtALICE 

fully agree with you (I wanted to have the possibility to use any parametrization file from GRID on Lego trains, but this seems not possible [not sure if you are on the train operators list]).
Now I tried to get only one centrality (which works for the fit), but still the macro creates directories for all centralities, how can I avoid this?

thanks and cheers,
Michael

miweberSMI:
Hi @nschmidtALICE,

in the updated file only the two ranges that I found problematic in the original file are kept. 

Thanks and cheers,
Michael

miweberSMI:
Hi @nschmidtALICE 

concerning the file size I am wondering in general if we should not put the files outside of ALIPHYSICS? It is usually suggested to put them to EOS:
http://alisw.github.io/git-advanced/#how-to-use-large-data-files-for-analysis

Cheers,
Michael

nschmidtALICE:
+1

nschmidtALICE:
@miweberSMI,
thank you for making a new file that only contains your required parametrizations.
From my side nothing would speak against putting the files on EOS. We then just need very strict modification access rules in the respective directory.
The code would also need slight adjustments to be able to load the EOS-based files.
Cheers, Nicolas

PR #9084, opened 2019-03-15
PWGUD/UPC: Increased streamers. Eliminated the usage of problematic vectors.

siragoni:
Due to issues with the LEGO trains, increased all the streamers, and eliminated the usage of the incriminated vectors. Only kept those related to good quality runs, considering their modest size (vectors of ~200 doubles). 
IMPORTANT: this code has been tested on local. The MC class has the exact same output as the version before on the small data sample I am using for tests. It is possible that the issues were due to the massive size of the vectors when analysing MC, considering that ~10^7 reconstructed UPC events are expected each time on the full LHC18l7 simulations...


mbroz84:
+1

PR #9091, opened 2019-03-15
Possibility to compile with C++17

pzhristov:
These are in fact only two changes:
- replacing std::random_shuffle by std::shuffle
- replacing auto_ptr by unique_ptr


dberzano:
+1: does not break the current case

PR #9116, opened 2019-03-19
MC supercalibration option (nonlinear, etc) added

ddobrigk:
Calibrator changes to enable an advanced MC centrality calibration


ddobrigk:
+1

PR #9130, opened 2019-03-19
Bug fixes

chiarazampolli:
- AliCFVertexingHFCascade: The prong for the bachelor is the 0 (both for D* and Lc).
- In the CF task, if one uses AliCFVertexingHFCascade for the Lc (and not
AliCFVertexingHFLctoV0bachelor), the order of the track cuts need to be different
(the proton is the third in the array)


afestant:
+1

PR #9134, opened 2019-03-19
PWGUD/UPC: Added Collins-Soper and Helicity Frames for helicity J/Psi.

siragoni:
Now both the Collins-Soper and the Helicity Frame are added together, both cos(theta) and phi. This way we could possibly investigate changes for the two different parametrizations...
In addition, added calibration for the ZDC energies for the LHC15O period. Getting ready to analyze even the 2015 data! This way it would be possible to verify the pedestal peak in the ZNC/A energy plots... Incremented streamers in both analyses tasks.
IMPORTANT: this code has been tested on local. Results as expected. No vectors this time, so it should work on the first try even in the LEGO trains...
Thanks.
Regards.


mbroz84:
+1

PR #9135, opened 2019-03-19
Proper access to MC particle

victor-gonzalez:
According to the recommendation of not using neither ``Stack`` nor ``TParticle``


shahor02:
+1

Hi @victor-gonzalez 
Thanks! Small comment: in principle, one can still use ``TParticle* AliMCEvent::Particle(i)`` (which is just an alias to  
````
((AliMCParticle*)AliMCEvent->GetTrack(i))->Particle();
````
but one should not use the doughter/mother labels of this TParticle. But in any case, the way you implemented (accessing kinematics via AliVParticle::P... interface) is better since allows to handle uniformly both AOD and ESD cases.

PR #9138, opened 2019-03-19
 Add methods to introduce eta gaps, and remove tracks from qn

fgrosa:
1) add method to introduce eta gap between resolution sub-events
2) add method to remove daughter tracks from qn (1: remove daughter tracks of selected candidate, 2: remove daughter tracks of all the candidates in the selected event)
3) add method to perform random downsampling of tracks from qn

The three methods can be called together.


afestant:
+1

PR #9139, opened 2019-03-19
Add variable-width mass binning for vn vs. mass

fgrosa:
1) add variable-width mass binning for vn vs. mass
2) add parameters in config file for vn vs. mass:
    - flag to fix mean in vn vs. mass
    - flag to fix sigma in vn vs. mass
    - limits of the variable mass binning for vn vs. mass


fprino:
+1

PR #9146, opened 2019-03-20
PWGUD/UPC: Added run selection for 2015 data. Added ZDC plots before timing selec.

siragoni:
As asked by the UPC PAG, I am getting ready to analyze 2015 data too. The run selection for 2015 data has been taken from the analysis note of the coherent J/Psi. 
Given the great concern related to the ZDC plots, histograms related to the ZDC energy without the timing selection have been applied. These changes involve only the AliAnalysisTaskUPCforward.{h, cxx}.
IMPORTANT: this code has been tested on local. It works fine but now the vector containing the run numbers is starting to be big (~400 integers). Hopefully this won't be a problem on the LEGO trains. This is not scalable but again, it is not like I am expecting to add more runs...


mbroz84:
+1

PR #9152, opened 2019-03-20
Add flag to select pile-up events only

strogolo:
Add flag to select out-of-bunch pile-up events only. This flag will be used to cross-check the pile-up events for the D-meson analyses.


fprino:
+1

arossi81:
Hi @strogolo,
I do not think that the following are the reasons for breaking the AliPhysics compilation but:
- can you increase the ClassDef number?
- are you sure that you want to avoid the writing of the fRejectOutOfBunchPileUp value? (I mean right now the value is not streamed since you have //! as comment). 

Cheers,
Andrea

PR #9159, opened 2019-03-21
Add 3D efficiency correction method.

jikim1290:


dcaffarr:
+test

dcaffarr:
@jikim1290 I noticed you have modified two tasks belonging to different PWGs. Was it intentional? Or you pushed all the modifications by mistake? If you confirm it is ok, I can proceed with the approval (if mine is enough). For next time, better to keep separate the two commits.

PR #9161, opened 2019-03-21
Adding information (and histograms) on number of tracklets

chiarazampolli:
The existing variable was counting in [-1, 1], now we have also
the info in [-999, 999].
Additionally, some formatting was changed, and some compilation
warnings were removed (some are left on unused variables, kept there
in case we will ever use them).
The AddTask was fixed for the output slot n. 7


chiarazampolli:
Pining @andreaalici for his information.

afestant:
+1

PR #9188, opened 2019-03-23
Modify structure of trees for HF particles to flat trees

fgrosa:
- restore flat tree structure for HF particles trees (not jet)
- fix rapidity of Lc (using correct mass)


afestant:
+1

PR #9212, opened 2019-03-25
Update refit parametrisation: use softHard instead of qcdRatio

miweberSMI:


miweberSMI:
Hi @nschmidtALICE, all,

I updated the refitted parametrization file, since the problem when running the cocktail on the train stayed unfortunately (it seems there were still problematic regions of the fit). Now I use softhard instead of qcdratio for these two centrality ranges that are included in tsi file and at least the local train test works. I hope this is fine now...
Including also @sebaleh and @carstooon here.

Cheers,
Michael

nschmidtALICE:
+1

PR #9217, opened 2019-03-25
Improves on memory leak on Dalitz

eddoming:
Dear AliPhysics experts,
On the next commit, I implement the change of new* classes for auto_ptr, this works like 2 on 1, using new* on the beginning and after closing the cycle where the class lives use the delete*. I recheck my systematics and the need some changes plus new cases.
Regards Edgar

### DIFF
--- a/PWGGA/GammaConv/AliConversionMesonCuts.cxx
+++ b/PWGGA/GammaConv/AliConversionMesonCuts.cxx
@@ -504,7 +504,7 @@ Bool_t AliConversionMesonCuts::MesonIsSelectedAODMC(AliAODMCParticle *MCMother,T
 }
 
 //________________________________________________________________________
-Bool_t AliConversionMesonCuts::MesonIsSelectedMCAODESD(AliDalitzAODESDMC *fMCMother,AliDalitzEventMC *mcEvent, Double_t fRapidityShift){
+Bool_t AliConversionMesonCuts::MesonIsSelectedMCAODESD(AliDalitzAODESDMC *fMCMother,AliDalitzEventMC *mcEvent, Double_t fRapidityShift) const{
 // Returns true for all pions within acceptance cuts for decay into 2 photons
 // If bMCDaughtersInAcceptance is selected, it requires in addition that both daughter photons are within acceptance cuts
     if(!mcEvent)return kFALSE;
@@ -527,8 +527,8 @@ Bool_t AliConversionMesonCuts::MesonIsSelectedMCAODESD(AliDalitzAODESDMC *fMCMot
         // Select only -> 2y decay channel
         if(fMCMother->GetNDaughtersG()!=2)return kFALSE;
         if((fMCMother->GetLastDaughterG()<0) || (fMCMother->GetFirstDaughterG()<0)) return kFALSE;
-            AliDalitzAODESDMC *MDaughter0=mcEvent->Particle(fMCMother->GetFirstDaughterG());
-            AliDalitzAODESDMC *MDaughter1=mcEvent->Particle(fMCMother->GetLastDaughterG());
+            std::unique_ptr<AliDalitzAODESDMC> MDaughter0=std::unique_ptr<AliDalitzAODESDMC>(mcEvent->Particle(fMCMother->GetFirstDaughterG()));
+            std::unique_ptr<AliDalitzAODESDMC> MDaughter1=std::unique_ptr<AliDalitzAODESDMC>(mcEvent->Particle(fMCMother->GetLastDaughterG()));
         if ((MDaughter0->GetPdgCodeG()!=22) || (MDaughter1->GetPdgCodeG()!=22)) return kFALSE;
             return kTRUE;
     }
@@ -653,7 +653,7 @@ Bool_t AliConversionMesonCuts::MesonIsSelectedAODMCDalitz(AliAODMCParticle *fMCM
 }
 
 //________________________________________________________________________
- Bool_t AliConversionMesonCuts::MesonIsSelectedMCDalitzAODESD(AliDalitzAODESDMC *fMCMother,AliDalitzEventMC *mcEvent, Int_t &labelelectron, Int_t &labelpositron, Int_t &labelgamma, Double_t fRapidityShift){
+ Bool_t AliConversionMesonCuts::MesonIsSelectedMCDalitzAODESD(AliDalitzAODESDMC* fMCMother,AliDalitzEventMC *mcEvent, Int_t &labelelectron, Int_t &labelpositron, Int_t &labelgamma, Double_t fRapidityShift) const{
 // Returns true for all pions within acceptance cuts for decay into 2 photons
 // If bMCDaughtersInAcceptance is selected, it requires in addition that both daughter photons are within acceptance cuts
     if( !mcEvent )return kFALSE;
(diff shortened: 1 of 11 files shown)


akubera:
`auto_ptr` has been deprecated in favor of `unique_ptr`, is there a reason you chose the former? I only mention this because there was a pull request a week ago that actually replaced all instances of auto_ptr with unique_ptr (#9091), and I think this would disrupt the goals of that commit.

It should be as easy as find-and-replace "auto_ptr"/"unique_ptr", the interface is mostly the same, as far as I know.

eddoming:
@akubera Dear Andrew,
I think my book of C++ is a little bit old, I will make the changes and ask for another commit or can I edit a commit?

akubera:
if you just make a new commit it *should* be a simple `git push` and it will update this pull request. No need for separate pull request.

(or  `git push <github-remote-name>`, whatever you did to push this to github initially)

akubera:
For reference see here https://en.cppreference.com/w/cpp/memory/auto_ptr. It looks like it's completely removed in C++17, so to save the pain of changing it later (if indeed C++17 is targeted by AliPhysics the future) it's probably wise to change. Maybe @pzhristov knows more about such plans?

PWGGA/GammaConv/AliConversionMesonCuts.cxx
```diff
@@ -527,8 +527,8 @@ Bool_t AliConversionMesonCuts::MesonIsSelectedMCAODESD(AliDalitzAODESDMC *fMCMot
         // Select only -> 2y decay channel
         if(fMCMother->GetNDaughtersG()!=2)return kFALSE;
         if((fMCMother->GetLastDaughterG()<0) || (fMCMother->GetFirstDaughterG()<0)) return kFALSE;
-            AliDalitzAODESDMC *MDaughter0=mcEvent->Particle(fMCMother->GetFirstDaughterG());
-            AliDalitzAODESDMC *MDaughter1=mcEvent->Particle(fMCMother->GetLastDaughterG());
+            auto_ptr<AliDalitzAODESDMC> MDaughter0=mcEvent->Particle(fMCMother->GetFirstDaughterG());
```
akubera: One more thing, it seems that some of these are kind-of overkill; in this case we're create a whole "new" particle to just get the PDGCode and then delete the particle immediately afterwords. It seems like it would be much more efficient to add a method to `AliDalitzEventMC`, something like:
```cpp
Int_t GetPgdCodeOfParticle(Int_t i)
{
  if (fIsESDMC) {
    return fESDEventMC->Particle(i)->GetPdgCode();
  } else {
    auto branch = AliAODMCParticle::StdBranchName();
    auto *array = static_cast<TClonesArray*>(fAODEvent->FindListObject(branch));
    auto *particle = static_cast<AliAODMCParticle*>(array->At(i));
    return particle->GetPDGCode();
  }
}
```

No `new`, no `auto_ptr`, no copying, just a lookup.

Then, these lines become:

```cpp
if (mcEvent->GetPgdCodeOfParticle(fMCMother->GetFirstDaughterG()) != 22 ||
    mcEvent->GetPgdCodeOfParticle(fMCMother->GetLastDaughterG()) != 22) {
   return kFALSE;
}
return kTRUE;
```

or more succinctly:

```cpp
// returns true if both pdg-codes are 22
return mcEvent->GetPgdCodeOfParticle(fMCMother->GetFirstDaughterG()) == 22 &&
       mcEvent->GetPgdCodeOfParticle(fMCMother->GetLastDaughterG()) == 22;
```

I'm PWGCF, not PWGGA, so I'll stop bothering you, just some suggestions.

eddoming: Dear Andrew,
Thanks a lot for the feedback, I really enjoy learning and see things I didn't see on the beginning, this last part of the call of the Daughter I use that way because I keep the same idea from the principle to create a switch class between AOD or ESD. So I create this class AliDalitzAODESDMC to get the info, the problem you just point me, I use for everything when some times is just better to make this lockout. I will see how all this work on the code and if is work -> make a commit 
Regards Edgar
(Copy and paste code sometimes never work, most of all the times... ;>)

PWGGA/GammaConv/AliDalitzData.cxx
```diff
@@ -32,16 +32,18 @@ ClassImp( AliDalitzData )
     
         AliDalitzData::~AliDalitzData(){
         }
-    
-     AliDalitzAODESD* AliDalitzData::GetTrack(Int_t i){
+
+    auto_ptr<AliDalitzAODESD> AliDalitzData::GetTrack(Int_t i){
```
eddoming: Hello Andrew,
reading and coding, unique_ptr is not the same that auto_ptr, not event shared_ptr, shared and unique they are very well defined in terms on where they live and the pointers, meanwhile, auto_ptr is more flexible in that way, for not said bad defined. For example here, on this part I define this switch tracks for AOD and ESD with auto, changing with unique didnt, permit to keep using the Particle the same way.
`    unique_ptr<AliDalitzAODESD> AliDalitzData::GetTrack(Int_t i){
        
         if (fIsESD==kTRUE){
                unique_ptr<AliDalitzAODESD> esdtrack( new AliDalitzAODESD((AliESDtrack*)fESDEvent->GetTrack(i)));
                esdtrack->ComputeImpactParameter();
                return esdtrack;}
         else { unique_ptr<AliDalitzAODESD> aodtrack( new AliDalitzAODESD((AliAODTrack*)fAODEvent->GetTrack(i)));
                aodtrack->ComputeImpactParameter(fAODEvent->GetPrimaryVertex(),fAODEvent->GetMagneticField());
                return aodtrack;}
     } `

PWGGA/GammaConv/AliConversionMesonCuts.cxx
```diff
@@ -1349,16 +1349,16 @@ Bool_t AliConversionMesonCuts::MesonIsSelectedMCChiCAODESD(AliDalitzAODESDMC *fM
...
 
         //Int_t labeljpsiChiC = -1;
 
         for(Int_t index= fMCMother->GetFirstDaughterG();index<= fMCMother->GetLastDaughterG();index++){
             if(index < 0) continue;
-            AliDalitzAODESDMC* temp = mcEvent->Particle(index);
+            auto_ptr<AliDalitzAODESDMC> temp = mcEvent->Particle(index);
             //TParticle* temp = (TParticle*)mcEvent->Particle( index );
 
             switch( temp->GetPdgCodeG() ) {
```
eddoming: Here one of the big problems using unique_ptr, after defining the AliDalitzAODESDMC track for the temp, and using the switch, you can not copy a unique_ptr one to another very easily
`    //std::unique_ptr<AliDalitzAODESDMC> positron;
   // std::unique_ptr<AliDalitzAODESDMC> electron;
   // std::unique_ptr<AliDalitzAODESDMC> gamma;
    for(Int_t index= fMCMother->GetFirstDaughterG();index<= fMCMother->GetLastDaughterG();index++){
        if(index < 0) continue;
        std::shared_ptr<AliDalitzAODESDMC> temp = mcEvent->Particle(index);
        //TParticle* temp = (TParticle*)mcEvent->Particle( index );
        switch( temp->GetPdgCodeG() ) {
        case ::kPositron:
            std::shared_ptr<AliDalitzAODESDMC> positron = std::make_shared<AliDalitzAODESDMC>(*temp);
          //  auto positron      =  std::make_shared(temp);
            labelpositron = index;
            break;
        case ::kElectron:
          //  electron      =  temp;
            std::shared_ptr<AliDalitzAODESDMC> electron = std::make_shared<AliDalitzAODESDMC>(*temp);
         //auto electron      =  std::move(temp);
            labelelectron = index;
            break;`
One can only change one case at the time because you erase the temp previously with the "move" command, this is the only way you can completely pass the inputs to another unique_ptr, so the changes will rebuild all thinks of classes that share something and pass or copy some information, with respect the GetPdgCode function to lookout, the problem is the input you will give it to search for the info, and again you need something to define?

pzhristov:
Please do not use auto_ptr, it is deprecated with -std=c++17 and the code will not compile. We are trying to have uniform compilation options for AliRoot/AliPhysics and O2, ant the later uses C++ 17 standard.

PWGGA/GammaConv/AliConversionMesonCuts.cxx
akubera: Use three backticks  (```) so the code is readable:

\```
`code goes here`
\```

But yeah, the difficulty of copying is a feature, keeping the code as efficient as possible by avoiding accidental copying the whole object the pointer is pointing to.

In this case you're not creating any new `TParticles`, I don't see any need to change this code to use `std::unique_ptr` or `std::shared_ptr`.

eddoming: Dear Andrew,
yes on here, when I summon a temp 
`std::shared_ptr<AliDalitzAODESDMC>` temp = mcEvent->Particle(index);`
I wish to use on some case, and if is the type of particle I want, I will save all the TParticle information, the problem that this gives me is, one the way you can transfer from one `unique_ptr` to another `unique_ptr`,  after passing the first case,
```
        case ::kPositron:
            std::shared_ptr<AliDalitzAODESDMC> positron = std::make_shared<AliDalitzAODESDMC>(*temp);
          //  auto positron      =  std::make_shared(temp);
            labelpositron = index;
            break;
```
On the second case, I can use `temp` to compare on the next case.

eddoming:
Dear Andrew and Peter (@akubera @pzhristov ),
I finish the transition from `auto_ptr` to `unique_ptr,` yes I admit `auto_ptr`is a little dirty command, Is the same idea, just I have to re-think the way I cast the class to get the best result and I realize that I didn't need on every part of the code.
Regards Edgar

eddoming:
Dear Andrew,
is everything ok, or they are some problems?
Regards
Edgar

pzhristov:
+1

PR #9218, opened 2019-03-25
Creating BDT reader only when needed.

chiarazampolli:
And a minor fix (float to int in number of bins of a histogram)


chiarazampolli:
Pinging @andreaalici for information

afestant:
+1

PR #9224, opened 2019-03-26
Cent supercalib fix, better defaults for weak dec finder

ddobrigk:
- Supercalib fix applied to MultSelection MC calibrator task (mundane bug) 
- Changed defaults for weak decay finding - these were set to loose after a PWG-LF discussion but if this finder is used systematically then better set to a compromise between loose and (too) tight to conserve computing power


ddobrigk:
+1

PR #9225, opened 2019-03-26
PWGUD/UPC: Y-dependent helicity analysis + neutron emission analysis

siragoni:
Added the Y-dependent helicity analysis both for the Helicity frame and the Colin-Soper helicity frame. Changed the ZNC and ZNA energy threshold values for the differential neutron emission analysis. This way the 0N0N, XN0N, 0NXN, XNXN invariant mass plots would become available.
TESTED


mbroz84:
+1

PR #9226, opened 2019-03-26
 Removed FMD info from output and added ITS/TOF info to output

ewillsher:
- FMD info that required AOD handler has been removed from output
- Additional ITS/TOF info has been added to output


dcolell:
+1

PR #9229, opened 2019-03-26
Triggers update

mbroz84:


mbroz84:
The files to be stored in the OADB are maps of bad TOF trigger maxipads. Previously these maps were used from OCDB (TRIGGER/TOF/TriggerMask) using objects created at start-of-run, but we (PWGUD/UPC) found that there are often severe issues with these objects what is affecting reliability of MC and even making some runs unusable. Thus we decided to created the maps offline and store in OADB to use them at analysis level. The OCDB object were in the past used during MC production.

dberzano:
+1

PR #9248, opened 2019-03-27
PWGUD/UPC: Added ZDC checks. Fixed bug neutron emission analysis.

siragoni:
Added ZDC plots to check for potential problems in either my analysis or the data themselves. Fixed the bug in the neutron emission analysis.
IMPORTANT: this has been tested on local and it compiles properly and it runs too.


siragoni:
Did I make a mistake somewhere...?
I had some problems on my local branch, so I pulled my remote master and pushed back after a rebase. Did this trigger something I was not supposed to touch?
Sorry for the problem.
Regards.

mbroz84:
+1

PR #9261, opened 2019-03-28
New version of AliAnalysisTaskMinijet using AliDirList instead of TList.

mgheata:
This branch clones AliAnalysisTaskMinijet, using AliDirList instead of TList for the list of outputs. This is temporary in order to compare the use of memory in merging for the two approaches.

### DIFF
--- a/PWGCF/Correlations/DPhi/AliAnalysisTaskMinijetV2.cxx
+++ b/PWGCF/Correlations/DPhi/AliAnalysisTaskMinijetV2.cxx
@@ -1,0 +1,2154 @@
+#include <TChain.h>
+#include <AliDirList.h>
+
+#include <TTree.h>
+#include <TH1F.h>
+#include <TH2F.h>
+#include <THnSparse.h>
+#include <TProfile.h>
+#include <TCanvas.h>
+#include "TRandom.h"
+
+#include "AliVEvent.h"
+#include "AliVParticle.h"
+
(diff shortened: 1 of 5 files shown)


vpacik:
+1

naghmehmohammadi:
+1

akubera:
I had issues with memory usage during merging. I ended up with writing a new class: [AliFemtoResultStorage](https://github.com/alisw/AliPhysics/blob/master/PWGCF/FEMTOSCOPY/AliFemto/AliFemtoResultStorage.h) and an analysis task [AliAnalysisFemtoNu](https://github.com/alisw/AliPhysics/blob/master/PWGCF/FEMTOSCOPY/AliFemto/AliAnalysisTaskFemtoNu.h) to go along with that.

I had to adopt the conventions of the AliFemto framework, but it may serve as some inspiration.

mgheata:
Hi,
Indeed looks like AliFemtoResultStorage has part of AliDirList functionality allowing to write the content to a folder. In addition, AliDirList can read content from the folder in a lazy way (just read object names and classes, the actual reading in memory being done upon object access. It also provides TList/TObjArray accessors to make easier manipulation as a collection. 
It appears that AliPhysics is not build against AliRoot master that contains this class, I am trying to test this new utility.

vpacik:
+1

vpacik:
+1

PR #9265, opened 2019-03-28
TENDER/TenderSupplies: Re-added the MC energy fractions per label to the output

nschmidtALICE:
Some time ago, there was a commit that removed the MC energy fractions from the standard output of the Tender, see https://github.com/alisw/AliPhysics/commit/2f73f2c9d567be9d01552e3c6c52090622fa8f22#diff-ce4c5e3ee3cdb974ac13f17ca80cbf21 
As this is rather crucial information and the boolean fSetCellMCLabelFromEdepFrac also modifies the information stored in these fractions, i would highly recommend to add the fractions back to the output.

Maybe @gconesab knows why this was put behind the if() condition.


lcunquei:
+1

nschmidtALICE:
@loizides and I asked @lcunquei to approve this for now. In case there is a very good reason to not write out the fractions per default, then the re-addition of the if() condition can be discussed again.

PR #9276, opened 2019-03-29
 PWGGA: Add Config for PCM PHOS for Weights

jluehder:
Added a two new Configs for GammaConvCalo_pp; Needed for Checking of Weights with PCM-PHOS


FriederikeBock:
+1

PR #9294, opened 2019-03-31
Fix narrowing conversion error in new  MacOS

ycorrales:
Just a simple explicit cast in array initialisation to avoid compilation error in MacOS 10.14.4 with Xcode 10.2 due to narrowing conversion....


akubera:
+1

vpacik:
Hi, @FrejaThoresen since this is primarily your task, perhaps take a look

FrejaThoresen:
I don't think static_cast makes sense in this case, will change it to Double_t(x).

ycorrales:
using static_cast is a safe C++-style way, double(x) - unsafe old C-style way. Btw, it's exactly the same in this case, so use as you preference.

PR #9303, opened 2019-04-01
[PWGCF] Modify gestion of the track selection.

cimordas:
* Add selection on the charge of the tracks.
* Add Bool_t to select if a cut on a variable is applied or not for every cut in the track selection.


vpacik:
+1

PR #9305, opened 2019-04-01
PWGGA/GammaConv: Added new function for SM wise energy correction of …

jokonig:
…EMCal and DCal

This was previously implemented in the function ApplyNonLinearity. To make this correction independent of the NonLin cutstring a new function was implemented.

Additionally a cutstring in GammaCalo_pp Add task was changed and the train config 2002 was added to the pi0 only output to save memory on the grid.


FriederikeBock:
+1

PR #9315, opened 2019-04-01
PWGUD/UPC: Added further checks for the timing.

siragoni:
Added further checks for the separation main-satellites collisions in UPC. 
IMPORTANT: the code has been tested, and an abort due to too large of a binning has already been solved on local.


mbroz84:
+1

PR #9335, opened 2019-04-03
Provisions to supercalibrate old datasets

ddobrigk:
Implements a map that allows the MC calibrator steering macro to define a highest-threshold for the V0M amplitudes so that one can reject too-high-multiplicity events roughly. This is not a replacement for generating tuned MC but will serve as a stopgap solution that allows us to use older, already generated MC using the same supercalibration used in 2018 MC to almost the same effect in terms of <Ntracklets> vs V0M matchup.


ddobrigk:
+1

PR #9336, opened 2019-04-03
Create lightweight output for normal running

mfasDa:
Fill THnSparse only when selected (for
debugging), otherwise use TH1/2 for spectrum,
response matrix, energy scale and resolution.
Needed for ultra high pt studies


mfasDa:
+1

PR #9338, opened 2019-04-03
PWGUD/UPC: Added analysis in 10 rapidity bins. Fixed bugs.

siragoni:
Added analysis in 10 rapidity bins for both data and MC. Fixed some nasty bugs in the MC code, that were not so easy to find... Eliminated unused code chunks, that had been just commented out.
IMPORTANT: the code has been tested. Processes were killed on local due to bugs while instantiating the histograms. Now this has been fixed.


mbroz84:
+1

PR #9340, opened 2019-04-03
Revised code + new class added  

amastros:
The class can create v0s according to 3 different cases: opposite sign pairs, like sign pairs and creating a pair having the pion track rotated by pi. The class is necessary to compute the LNN background from data.


bdoenigu:
+1

PR #9355, opened 2019-04-03
PWGUD/UPC: Added the trigger conditions for 2015 data.

siragoni:
Added the trigger conditions for 2015 data. This means CMUP10, CMUP11, and CMUP13. CMUP11 is in common with 2018 data. This explains why this code could partially analyse 2015 data too...
TESTED.


mbroz84:
+1

PR #9359, opened 2019-04-03
AliAnalysisCuts interface + MCGEN train support

mpuccio:
@fprino this changes should be transparent for the everyday use of AliEventCuts.
MCGEN trains and Kine only production have run number equal -1, however I wonder if there are some other corner cases in MC that can hurt us badly...

@mfasDa @jgrosseo fyi


chiarazampolli:
+1

PR #9370, opened 2019-04-04
PWG/EMCAL: Re-addition of stored MC label energy fraction to cluster

nschmidtALICE:
As already re-added for the Tender, i also propose to re-add the cluster energy fraction per MC label back to the clusters per default. (see https://github.com/alisw/AliPhysics/pull/9265 for the Tender PR.)
This is necessary as the boolean "fSetCellMCLabelFromEdepFrac" also modifies and limits the MC cluster label list to 4 labels, which is way too little.
Let me know if you have any objections. If not, then please give it a +1 as this additional cluster information is crucial for ongoing analyses.


FriederikeBock:
+1

PR #9378, opened 2019-04-05
PWGUD/UPC: Added AliAnalysisTaskCTrue.

siragoni:
Added the AliAnalysisTaskCTrue.{h, cxx} with the steering macro AddTaskCTrue.C, plus the task has been added to the UPC library. It runs on the same dataset as AliAnalysisTaskUPCforward, so it should lower my consumption on resources... This task is actually the initial merging of other code of the UPC group. The hope for the future is to develop it further, analysing CTRUE B/E/A/C at the same time.
IMPORTANT: while this is more like a initial commit, the code has been tested on local and this seems to work. Now just a consistency check is required...


mbroz84:
+1

PR #9379, opened 2019-04-05
replacing custom PID strings by getting function with enum

jgrosseo:


jgrosseo:
+1

jgrosseo:
@mpuccio There is also one new protection in AliEventCuts. Let me know in case you disagree

mpuccio:
That protection was indeed missing, thanks!

PR #9387, opened 2019-04-05
PWGUD/UPC: New part of helicity analysis. Possible bug fix too. Streamers indicators.

siragoni:
This pull request is divided in many parts. On one hand, I have added a new part for the helicity analysis. In this new piece of code I try to obtain the relative abundance of GammaGamma and J/Psi to the decay angular distribution, simply by obtaining the invariant mass distribution in a specific CosTheta bin. On the other hand, I have noticed a possible bug that was not easy to spot. When trying to fill, it seems like the program implicitly converted Double_t to Int_t. This could possibly the reason why the acceptance distributions for the angular distributions showed such a strange shape... Now an explicit conversion to Double_t is requested every time. In addition, I have added the streamer indicators for the private class members that I had previously missed...
IMPORTANT: all these changes have been tested on local for both Analysis Tasks.


mbroz84:
+1

PR #9415, opened 2019-04-08
Determine combinatorial backgrounds

gskorodu:
For the analysis of the background shape and further precise signal extraction: combinatorial backgrounds of the cascade invariant mass spectra are determined and implemented.
The parameters needed to rerun cascade vertexer can be now set as variables in the addtask macro.
The output of the addtask macro is set to default 'AnalysisResults.root'


bdoenigu:
+1

PR #9416, opened 2019-04-08
Completes the V0 preselection and adds Setters for the criteria

bhohlweger:


bhohlweger:
Needs more work, test with local run script fails due to the fact that 

 const AliAODEvent* evt = static_cast<const AliAODEvent*>(v0->GetEvent());
  if (!evt->GetPrimaryVertex()) {
    return false;
  }

doesn't return a proper event in some of the cases

bhohlweger:
Correction: GetEvent always returns 0 and therefore event related information cannot be accessed. How can we fix this? This is needed for Cosine Pointing Angle & Transverse Radius selection.

jgrosseo:
I will commit this from my branch

PR #9420, opened 2019-04-08
PWGUD/UPC: Merging almost complete. Adding histos for veto counting.

siragoni:
The merging between the original TTree producing analysis task and the offline macros is almost complete. It has now reached the stage where the two versions have to differ greatly due to the different methodology... I reckon that by the end of the week the whole code will be completely merged and fully functioning online.
I have now added the histograms for the VETO counting and I am about to fill them so as to finally compute the required efficiencies for the CTRUE analysis.
IMPORTANT: this has been tested locally.


mbroz84:
+1

PR #9428, opened 2019-04-08
PWGUD/UPC: Finished merging. Only efficiency extrapolation missing.

siragoni:
I have finished merging the original analysis task and the offline macros... Now I have to check for consistency. The information is however recorded as histograms with the run numbers as string labels. Only the efficiency extrapolation is missing. The aim would be to have it done on the fly... I am still exploring this possibility. However, to do so would require TGraphErrors with the full statistics from each run. The way I have imagined it to be would be to invoke the function for filling the graphs right before the end. Maybe the Terminate() function. 
IMPORTANT: this has been tested.
Sorry for the double pull request today!!

### DIFF
--- a/PWGUD/UPC/AliAnalysisTaskCTrue.cxx
+++ b/PWGUD/UPC/AliAnalysisTaskCTrue.cxx
@@ -133,6 +133,8 @@ AliAnalysisTaskCTrue::AliAnalysisTaskCTrue()
       fVectorEfficiency(0),
       // plots for CTRUE
       fEntriesAgainstRunNumberProperlyH(0),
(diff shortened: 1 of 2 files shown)


mbroz84:
+1

siragoni:
Dear All,
I do not fully understand this problem.
I guess if I take away the TGraphErrors everything would be fine, but why is there a problem here?
Could it be that it is not possible to use TGraphs or TGraphErrors in AliPhysics? 
I am asking mainly because this task compiles and runs on local...
Many thanks.
Regards.

mbroz84:
I guess its because you are missing #include TGraphErrors or class TGraphErrors?

PR #9432, opened 2019-04-08
Fix compilation for MacOS clang

mpuccio:


mpuccio:
@gskorodu your commit from today was breaking the compilation on osX. This fixes the issue without changing the behaviour.

PR #9433, opened 2019-04-09
Remove wrong cut on gen Zvtx in case of MC

fgrosa:
This cut is already in FillMCGenAccHistos for the gen Ds selection, but if applied also outside the function it is also applied to the reco Ds


fprino:
+1

PR #9442, opened 2019-04-09
PWGUD/UPC: commenting out TGraphErrors from the CTRUE analysis task...

siragoni:
It seems like the interaction between TGraphErrors and AliPhysics is more troublesome than expected... The code related to this is temporarily left out. The TGraphErrors production will be on local after this.
IMPORTANT: tested.


mbroz84:
+1

PR #9456, opened 2019-04-09
AliEventCuts for pp run1 data

mlknichel:
Event selection extended to pp run1 data 
- only run ranges from LHC10d, LHC11are active for now
- since cut on SPD vertex dispersion does not work with AOD, this cut only works for ESD


jgrosseo:
+1

PR #9460, opened 2019-04-10
 Add flag to set fine pT bins in sparses 

fgrosa:
Method mainly useful for offline pT weights application (only in case of few candidates, i.e. signal only)


fprino:
+1

PR #9462, opened 2019-04-10
PWGUD/UPC: Changed binning for the Phi Helicity.

siragoni:
Changed the binning for all the plots involving the Phi angle of the Helicity analysis. It is a waste of resources to push further than (-4, 4)...
IMPORTANT: tested.


mbroz84:
+1

PR #9470, opened 2019-04-10
[EMCAL-315,EMCAL-318,EMCAL-319,EMCAL-321,EMCAL-322,EMCAL-442] OADB/EMCAL: Updated BC maps LHC18befhim and TC for LHC18gn with default 600ns shift

nschmidtALICE:
Analysis QA level bad channel map updates were provided for LHC18befhim.
In addition, a default 600ns has been applied for the time calibration of LHC18gn where especially the low gain could not be calibrated properly.


raymondEhlers:
+1

PR #9475, opened 2019-04-10
Changes for CF and main task for Lc --> K0s+p for PbPb

chiarazampolli:


chiarazampolli:
Pinging @andreaalici .

@agrelli: I had to modify again the CF for the LC as the ITS/TPCrefit MC checks were not ok otherwise. I am telling you in case also the D* is affected by the same problem. Note that we dont' use these steps, so maybe it is not urgent, but I wanted to have it fixed.

afestant:
+1

PR #9480, opened 2019-04-10
Correction to Dalitz framework

eddoming:
On the next commit, I correct many warnings generate for the new macros from Dalitz. I add the trains inputs, plus the corrections on the  PWGGA/GammaConv/AliConversionMesonCuts.cxx where never I get a kTRUE, plus optimized changing from AODESDtracks a just `Bool_t` flags


dmuhlhei:
+1

PR #9482, opened 2019-04-10
treeHF/JetTree: uniformize evt variable names

jdmulligan:
(trivial update to allow easier merging of event tree with jet tree)


afestant:
+1

PR #9494, opened 2019-04-11
Improver ask update for LHC18q,r periods

mfaggin:
----------------------------------------
 - correction templates for dca sigma, mean and ratio pulls developed separately for tracks requiring kFirst or kOnlySecond condition in the SPD
 - correction templates for dca mean developed in 24 bins of phi, instead of 4
The correction is performed in this way only for PbPb2018 period. This is ruled by a boolean variable (fIsPbPb2018), which is switched to the kTRUE value only
wheter the files, containing the correction templates, called from the AddTask are stored in a path including a "LHC18r" or "LHC18q" directory.
Backward compatibility tested on a LHC18d period (pp collisions)


afestant:
+test

afestant:
+1

PR #9504, opened 2019-04-11
PWGUD/UPC: Added heat map helicity analysis.

siragoni:
Added heat map for the helicity analysis.
This would show CosThetaHelicityFrame VS PhiHelicityFrame.
IMPORTANT: tested.


mbroz84:
+1

PR #9513, opened 2019-04-11
Pos Swap Fix

moliver813:
Optimize Pi0Cand bins
Fixes a part of pos-swap bkg method
Fixes a small eta phi error


moliver813:
Good catch! Fixing it now. I think that also applies to the references to CaloClusterVec in the next few lines

efranzis:
+test

efranzis:
+1

PR #9522, opened 2019-04-12
Fix in AddTask

gskorodu:
The parameters of the constructor are now provided in Addtask macro when creating an instance.
Names of variables are consistent now.


mkruegerGitHub:
+1

PR #9526, opened 2019-04-12
Change AliVEvent::EOfflineTriggerTypes to ULong64 

fgrosa:
Modification that allows the selection of more trigger masks at the same time (i.e. kINT7 | kCentral)

### DIFF
--- a/PWGHF/vertexingHF/macros/AddTaskHFSystPID.C
+++ b/PWGHF/vertexingHF/macros/AddTaskHFSystPID.C
@@ -1,7 +1,7 @@
 AliAnalysisTaskSEHFSystPID *AddTaskHFSystPID(int system = 0,
                                             bool readMC = false,
                                             TString trigClass = "",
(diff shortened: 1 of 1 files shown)


afestant:
+1

jezwilkinson:
Hi @dberzano ,  the test seems to be failing with this odd " fatal error: AliNanoAODTrackMapping.h: No such file or directory" error that was reported on the operations list yesterday, although tests on other PRs seem to have been working. Is there any chance that this can be worked around so that this commit can still make it onto today's tag?
Thanks,
--Jeremy

ktf:
It should eventually catch up with the changes in AliRoot.

jgrosseo:
Are we sure the field in the aod has 64 bit?

fgrosa:
Yes it should be 64 but. It is defined as a ULongt64_t also in the other HF code (see e.g. https://github.com/alisw/AliPhysics/blob/339bb675bd92127c1777306c10c4f0e1b9000752/PWGHF/vertexingHF/AliRDHFCuts.h#L439)

PR #9530, opened 2019-04-12
PWGUD/UPC: Fixed binning heat map helicity (now CosTheta=x, Phi=y).

siragoni:
It is more comfortable to have CosTheta in the x-axis and Phi as the y-axis...


mbroz84:
+1

PR #9551, opened 2019-04-15
PWGUD/UPC: Added first steps towards OCDB implementation. Still commented though.

siragoni:
Added the first steps for the OCDB implementation. In the future it should be possible to retrieve at runtime level the position in the L0 bit string of each trigger. It is still commented out as of yet. I still have to understand exactly how to work it out, but it is a huge step compared to before...


mbroz84:
+1

siragoni:
Added truth tables for fV0A/CDecision and fADA/CDecision.
I had forgotten to add them in the previous commit...
IMPORTANT: tested (even if the binning is still not too clear to me...).

mbroz84:
+1

PR #9555, opened 2019-04-15
PWGGA: adding config for the Dalitz, splitting of ITS region, removin…

amaringarcia:
…g the gamma asymmetry cut for nomB


nschmidtALICE:
+1

PR #9560, opened 2019-04-15
PWGDQ: Updates for calorimeter cluster handling in dielectron and reducedTree frameworks

laltenka:
- introduced calorimeter cluster cut classes to dielectron and reducedTree frameworks
- added possibility to select on cluster-track matches in  dielectron/core/AliDielectronTrackCuts
- added possibility to select on calorimeter clusters in AliAnalysisTaskReducedTreeMaker
- updated AliReducedAnalysisSingleTrack with calorimeter cluster selection and histograms

### DIFF
--- a/PWGDQ/dielectron/CMakeLists.txt
+++ b/PWGDQ/dielectron/CMakeLists.txt
@@ -48,6 +48,7 @@ set(SRCS
       BtoJPSI/AliDielectronBtoJPSItoEle.cxx
       core/AliDielectronCF.cxx
       core/AliDielectronCFdraw.cxx
+      core/AliDielectronClusterCuts.cxx
       core/AliDielectronCutGroup.cxx
       core/AliDielectronCutQA.cxx
       core/AliDielectron.cxx
--- a/PWGDQ/dielectron/PWGDQdielectronLinkDef.h
+++ b/PWGDQ/dielectron/PWGDQdielectronLinkDef.h
@@ -11,6 +11,7 @@
 #pragma link C++ class AliDielectronHistos+;
 #pragma link C++ class AliDielectronCF+;
 #pragma link C++ class AliDielectronCFdraw+;
+#pragma link C++ class AliDielectronClusterCuts+;
 #pragma link C++ class AliDielectronHF+;
 #pragma link C++ class AliDielectronHFhelper+;
 #pragma link C++ class AliDielectronMC+;
--- a/PWGDQ/dielectron/core/AliDielectronClusterCuts.cxx
+++ b/PWGDQ/dielectron/core/AliDielectronClusterCuts.cxx
@@ -1,0 +1,127 @@
+/*************************************************************************
+ * Copyright(c) 1998-2009, ALICE Experiment at CERN, All rights reserved. *
+ *                                                                        *
+ * Author: The ALICE Off-line Project.                                    *
+ * Contributors are mentioned in the code where appropriate.              *
+ *                                                                        *
+ * Permission to use, copy, modify and distribute this software and its   *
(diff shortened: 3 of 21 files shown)


iarsene:
Hi Lucas,
I just started looking over the modified files and I will start listing here my comments.
Cheers, Ionut

iarsene:
For the AliDielectronClusterCuts class, we should also have geters for all the data members.

iarsene:
For AliDielectronTrackCuts I suggest to also add getters for the calo related data members.

iarsene:
In the tree maker, in the function IsCaloClusterSelected(), you apply multiple filters and store the individual decisions in an array of bool, but this does not seem to be used when you decide which cluster to write in the tree. You write all clusters which fulfill at least one of the conditions, so this is an OR between all the filter the user configures in the AddTask. Is this intended like that?

iarsene:
In the jpsi2ee analysis class, you changed the range where the calo related variables are reset. I guess this is an update due to the fact that you added more variables. I think you could had just added those new variables in the VarManager in a different order so that these changes are not needed. Or, even better, we should probably have an extra variable in the VarManager which is called something like kNEMCALvars. This variable would never be used, but we should make sure that all of the calo related vars are defined before this one, such that we do not need to make this kind of updates in the future (or forget to do them).

iarsene:
In AliReducedEventInfo::GetCaloClusterFromID() function you iterate over the list of clusters to find the one with a given ID. I do not know how often you will call this function, but you should improve performance if you use a TIterator to parse the clusters in the TClonesArray.
In addition, if this is going to be really heavy CPU wise, you might want to make a (cluster ID, cluster pointer) map outside of the track loop so that you can go faster between track->clusterID->cluster.

laltenka:
> In the tree maker, in the function IsCaloClusterSelected(), you apply multiple filters and store the individual decisions in an array of bool, but this does not seem to be used when you decide which cluster to write in the tree. You write all clusters which fulfill at least one of the conditions, so this is an OR between all the filter the user configures in the AddTask. Is this intended like that?

Yes, this is intended that way. It's set up just like the track selection in IsTrackSelected(), where the individual filter decisions are also stored but not used at the moment.

laltenka:
> In the jpsi2ee analysis class, you changed the range where the calo related variables are reset. I guess this is an update due to the fact that you added more variables. I think you could had just added those new variables in the VarManager in a different order so that these changes are not needed. Or, even better, we should probably have an extra variable in the VarManager which is called something like kNEMCALvars. This variable would never be used, but we should make sure that all of the calo related vars are defined before this one, such that we do not need to make this kind of updates in the future (or forget to do them).

The variable you refer to in the jpsi task is a "track" quantity and not a calorimeter one since it's only filled after the track matching. I agree that I could have added a flag but in this case, it would not have been kNEMCALvars but one that refers to the track variables. We could implement such a thing but for now I'll leave it be.

laltenka:
> For the AliDielectronClusterCuts class, we should also have geters for all the data members.

> For AliDielectronTrackCuts I suggest to also add getters for the calo related data members.

I added getters.

laltenka:
> In AliReducedEventInfo::GetCaloClusterFromID() function you iterate over the list of clusters to find the one with a given ID. I do not know how often you will call this function, but you should improve performance if you use a TIterator to parse the clusters in the TClonesArray.
> In addition, if this is going to be really heavy CPU wise, you might want to make a (cluster ID, cluster pointer) map outside of the track loop so that you can go faster between track->clusterID->cluster.

I will not touch this on this PR but will try to implement a more efficient method on the next one.

iarsene:
+1

PR #9571, opened 2019-04-16
Makes the Cluster Min energy apply in more of the task

moliver813:
Makes the Cluster Min energy apply in more of the task
Enables run-time to be improved with tighter cuts


e-masson:
+test

e-masson:
+1

PR #9573, opened 2019-04-16
 Implement method to set NsigmaTPC correction for LHC18qr

fgrosa:
Implement a method to apply a correction to the NsigmaTPC for pion, kaon and proton (independently) in LHC18q and LHC18r data. The correction is implemented as:

NsigmaCorr = (Nsigma - mean(Nsigma)) / sigma(Nsigma)

where mean(Nsigma) and sigma(Nsigma) are the mean and sigma values of the Nsigma distributions extracted from the data using tagged samples of pions, kaons and protons

### DIFF
--- a/PWGHF/treeHF/AliAnalysisTaskSEHFTreeCreator.cxx
+++ b/PWGHF/treeHF/AliAnalysisTaskSEHFTreeCreator.cxx
@@ -194,7 +194,9 @@ fFillZLeading(false),
 fFillRadialMoment(false),
 fFillpTD(false),
 fFillMass(false),
-fFillMatchingJetID(false)
+fFillMatchingJetID(false),
+fEnableNsigmaTPCDataCorr(false),
+fSystemForNsigmaTPCDataCorr(AliHFTreeHandler::kNone)
 {
 
 /// Default constructor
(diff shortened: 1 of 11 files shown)


fprino:
Hi Fabrizio,

I saw that there is a bunch of code (containing in particular the numerical values with the corrections) that is triplicated: in AliHFTreeHandler, AliAODPidHF and AliAnalysisTaskSEHFSystPID.
I would suggest, if possible to avoid this, because it can cause mistakes in case we need to update the numerical values.
Can we use a root file as input? Or a common method in AliVertexingHFUtils?

Thanks!

Ciao, Francesco

fgrosa:
Hi Francesco,

you are right indeed. Do you have preferences for one or the other solution? I would prefer the AliVertexingHFUtils one because it allows me to keep essentially the same code, but just move it.

Thanks!
Fabrizio

fprino:
Hi,

yes, AliVertexingHFUtils is fine.
Alternatively, you can consider to define a (static) method in AliAODPidHF that you can call from the other classes to initialize the arrays.

Cheers, Francesco

fgrosa:
Ciao Francesco,

indeed this last option that you mentioned is my preferred one, so I implemented it.

Thanks again!
Cheers,
Fabrizio

fprino:
+test

fprino:
Thanks Fabrizio!

fprino:
+1

PR #9578, opened 2019-04-16
pp jet matching implemented

jaimenorman:
Detector level to particle level jet matching implemented, needed for pp analysis


jaimenorman:
hello, did anyone get a chance to look at this?

PR #9597, opened 2019-04-17
Update code for HF PID syst

fgrosa:
1) add possibility to use daughter-track p instead of daughter-track pT for systematic propagation to the D mesons

2) add additional option for track downsampling in order to have independent samples (e.g. one for the calibration of Nsigma and another one to estimate the residual systematic uncertainty)


arossi81:
+1

PR #9599, opened 2019-04-17
PWGUD/UPC: Changed binning of 2D distributions for helicity analysis in the MC.

siragoni:
The LEGO train results showed that I had set the binning wrongly.
Now Phi is ~-4 to 4 and CosTheta ~-1 to 1.


mbroz84:
+1

PR #9601, opened 2019-04-17
Added event extractor

rhaake:
This development is similar to the jet extractor and is currently able to extract events for ML studies, especially charged jets.


rhaake:
+test

rhaake:
+1

PR #9611, opened 2019-04-17
Update: add method for D0 2018 Pb-Pb syst. unc.

strogolo:
Update AliHFSystErr with two methods for systematic uncertainties of D0 analysis in 0-10 and 30-50 centrality classes.


jezwilkinson:
Hi @strogolo,
I think that since this adds new class members to AliHFSystErr, the ClassDef in the header file should be incremented to 11. Could you add this please?
Cheers,
--Jeremy

strogolo:
Hi Jeremy,
my fault, I changed it automatically while it should not be the case.
I'll fix it.
-- Stefano

jezwilkinson:
+1

PR #9659, opened 2019-04-19
Fixed AliGFW where subprofiling would crash on new AliPhysics

vvislavi:
Revamped AliProfileSubset class (after last update, AliGFWFlowContainer would randomly crash). Also, removed the redundant AliGFWHelper.h


aohlson:
+1

PR #9660, opened 2019-04-19
Weights for central events (PbPb 2018)

chiarazampolli:


chiarazampolli:
Dear @dberzano , @pzhristov ,

This PR allows to run the ML on the grid. Since these files are to be compiled, and they contain the weights, they are heavy in compilation, but we think that it should anyway be fine as they don't get recompiled often. We hope this is fine also with you. 

Chiara

jezwilkinson:
+test

jezwilkinson:
+1

PR #9663, opened 2019-04-19
Use clusterizerv2 for embedding

raymondEhlers:
Due to modifications in the clusterizerv3 algorithm, it won't properly
clusterize embedded cells. So we use clusterizerv2 for embedding. Note
that the clusterization computation time is quite small compared to
cluster-track matching, so the speed up if clusterizerv3 was made
compatible won't make much difference.


raymondEhlers:
+1

PR #9666, opened 2019-04-20
Add possibility to calibrate TPC Q-vector

fgrosa:
If TPC calibration factors found in the input file applied, otherwise not applied (but TPC Q-vector computed anyway)


fprino:
+1

PR #9694, opened 2019-04-24
Support estimator value in the NanoAOD header

mpuccio:


mpuccio:
Ciao @ktf , could you please have a look at the ROOT6 builds? It looks like it is stuck (not only for this PR)

ktf:
I am checking.

ktf:
one of the builders was stuck. Should be back to normal now.

mpuccio:
Grazie!

PR #9699, opened 2019-04-24
Add support for data from 2018

raymondEhlers:
@mfasDa - I don't modify the track selection very often, but I think this is what's appropriate for supporting the 2018 data (I'm not sure if we need a corresponding modification for ESDs). Can you confirm?


raymondEhlers:
+test

mfasDa:
Hi @raymondEhlers , exactly.  ESD would go analogously ( in AliEmcalESDTrackCutsGenerator), but it’s not urgent.

Thanks for implanting!

Cheers 

Markus

raymondEhlers:
+1 Thanks @mfasDa! I'll leave the ESD implementation for another day

PR #9700, opened 2019-04-24
Use clusterizerv3 in examples, samples

raymondEhlers:
Update examples, samples to use clusterizerv3, as this is the new standard


raymondEhlers:
+1

PR #9701, opened 2019-04-24
Changing variables to be stored in the tree for ML training

chiarazampolli:
- in extended mode: added nSigmaTPCpi, nSigmaTPCka, momentum of the bachelor
at the inner wall of the TPC;
- in reduced more: added nSigmaTPCpi, nSigmaTPCka, momentum of the bachelor
at the inner wall of the TPC, momentum of the bachelor; removed (to compensate):
alphaArm, ptArm, V0negativeEta, massGamma (all not used)

Additionally, now the nSigmaTPC and nSigmaTOF are taken from the AliAODPidHF,
so that we can use the recentered value (there is actually a flag to decide
whether to use AliAODPidHF, or the PIDresponse).


chiarazampolli:
Pinging @andreaalici , for his information.

Chiara

fprino:
+1

PR #9702, opened 2019-04-24
[WIP] Reducing compilation time by factor 10 (at least)

mpuccio:
Ciao @chiarazampolli ,

this is one possible way to reduce drastically the compilation time. Even without the -O1 I get the things compiled in 24 minutes (too long anyway, but at least I see an end).
With -O1 I get it compiled in ~1minute. Can you have a look if this has negative impact on the inference performance in your task?


chiarazampolli:
Ciao Max,

I tested your changes, and the BDT gives exactly the same output, so I would merge them, and we will follow your approach.

Thanks for the suggestion! I will also check (maybe with @ktf , since he seemed to have some ideas) if there is something more we can do.

We can proceed with the Pull Request.

Cheers,

Chiara

jezwilkinson:
+test

jezwilkinson:
+1

chiarazampolli:
Ciao Jeremy,

I think we should remove the [WIP] in the title of the PR... 

Chiara

PR #9706, opened 2019-04-25
Two new PID cases for Lc --> pK0s (A. Alici)

chiarazampolli:
- case 13: use TOF only when available, and apply a 3 sigma cut on both TOF
and TPC
- case 14: 3 sigma cut on both TOF and TPCm with TOF always required but
only up to low momentum


chiarazampolli:
Pinging @andreaalici , for information.

jezwilkinson:
+test

jezwilkinson:
+1

PR #9721, opened 2019-04-25
Introduce AliEventCuts in AliAnalysisTaskEmcalLight

mfasDa:
- Same code handling as for AliAnalysisTaskEmcal:
  + Standard event selection: AliEventCuts
  + Bypass: Builtin event selection (for run1 or MCgen train)
- Object constructed in the constructor and can be accessed
  by a getter returning a reference to the event cuts object

In addition:
- Move doxygen documentation from the cxx to the header file
- Introduce EMCAL_STRINGVIEW macro using std::string_view
  in case of ROOT6 builds from ROOT for non-owning strings
  and const std::string & in case of ROO5 builds


mfasDa:
+test

mfasDa:
+1

PR #9722, opened 2019-04-25
Added possibility to set a bound sigma (+/- some range) of gaus signal func

MycrofD:
Added possibility to set a sigma bounded by user specified region (+/- some fraction of the specified sigma) for fitting the signal func of a gaus


jezwilkinson:
Hello @MycrofD,
I see that this changes the default fit option in the constructor to include "B" - are you sure this does not affect the default behaviour of the fit class? It would probably be better to leave the default option as it is and to alter it only if SetBoundGaussianSigma() is called (as in the commented-out part in that function in your commit).
The ClassDef will in any case need to be incremented in the header as you're introducing new class members.
Cheers,
--Jeremy

MycrofD:
That was good suggestion, @jezwilkinson . Implemented following edits as suggested.
1. Removed "B" from default fit option. 
2. And incremented the ClassVersionID in ClassDef.

jezwilkinson:
+test
Thank you for the edits, i think this should be OK now.
--J

jezwilkinson:
+1

PR #9725, opened 2019-04-25
Fix macro for ROOT6

BongHwi:
1. Just add a function name on the top.
2. Modify warning lines below:
ReweightEfficiency.C:77:34: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  char rname[500]; sprintf(rname,R->GetName());


BongHwi:
+test

PR #9732, opened 2019-04-26
Final correction on PsiPair cut, and correction on ESD data, (removin…

eddoming:
Dear AliPhysics,
On the next commit, I add correction on the calculation on PsiPair cut, calculating the AOD constrained param with the primary vertex and magnetic field, like on ESD, plus some corrections on the parameters.
Regards
Ed


dmuhlhei:
+1

PR #9733, opened 2019-04-26
Enable the mT histograms (hardwired)

a-mathis:
@bhohlweger we should revert this when you're back from vacation ;)


a-mathis:
+1

PR #9747, opened 2019-04-26

lhusova:
enable to run only primary V0 for closure test, 
enable to accept/reject lambdas from cascades in MC, 
add pseudorapidity cut for generated V0 daughters


vpacik:
+1

PR #9748, opened 2019-04-26
new task for the hypertriton 3 body becomes the default

pfecchio:
With this commit the old code for the analysis of the hypertriton 3 body decay is removed while the "new" task becomes the default. In the Hyp3Body folder a README.md is added with the reference of the last Tag with the old code.


mpuccio:
+1

PR #9751, opened 2019-04-27
Options for cluster fractional energy

mfasDa:
- Add max cell and fraction of the max cell to the leading cluster
  THnSparses
- Add Setters to switch on/off histograms when not needed
- Remove request for analysis utils as the task is based on AliEventCuts


mfasDa:
+1

PR #9752, opened 2019-04-27
Removed pT check from AliGFW cuts and moved them into AnalysisTaskGFW…

vvislavi:
…. This way, we can select different pT range for RF and POI without increasing the CPU time


vpacik:
+1

PR #9755, opened 2019-04-27
Porting EMCAL_STRINGVIEW also to AliAnalysisTaskEmcal

mfasDa:
Allows transparent usage of std::string_view from ROOT in case of ROOT6 builds for non-owning strings (i.e. in interfaces).


mfasDa:
+1

PR #9758, opened 2019-04-28
Fix gettting EMCAL leading cell

mfasDa:
Fraction of the cell amplitudes all set to 1,
therefore leading cell has to be obtained by
looping over all cells and finding the max
element.


mfasDa:
+1

PR #9761, opened 2019-04-28
Possibility to determine event multiplicity with ITSrefit tracks

mfaggin:


PWGPP/ITS/AliAnalysisTaskSEImpParResSparse.cxx
```diff
@@ -598,8 +642,24 @@ void AliAnalysisTaskSEImpParResSparse::UserExec(Option_t */*option*/)
...
+    highMult = (nTrks>500 ? kTRUE : kFALSE);
+    fNumTracksperEv->Fill(nTrks);
+    }
+    else{    // Consider the number of tracks with ITS refit only, avoiding the TPC pile-up
+        Int_t nTrks_ITSrefit=0;
+        AliVTrack* trk = 0;
+        for (Int_t it=0; it<nTrks; it++){ //start loop over tracks
+            trk = (AliVTrack*)event->GetTrack(it);
+            if(!trk) {continue;}
+            if(fIsAOD)  {if(((AliAODTrack*)trk)->GetStatus()&AliESDtrack::kITSrefit)    nTrks_ITSrefit++;}
```
fprino: Hi Mattia,

this if is not needed: you can call GetSTatus on the AliVTrack directly, no need to cast, this should speed up the execution.

mfaggin: Hi Francesco,

thanks for the comment. I implemented it in the new commit that I pushed. 
Apart from the piece of code that you highlighted yesterday, I modified another couple of lines where the same typecast was performed. Without it, the running time seems to be significantly reduced and the output (local test) does not seem to show differences with respect to the previous configuration.

fprino:
+test

fprino:
+1

PR #9764, opened 2019-04-28
MC study revisit and finished

vpacik:


vpacik:
CINT issue with having member array initialiser in header file (C++14+ feature)

PR #9768, opened 2019-04-29
Update AliAnalysisTaskNetLambdaMCTrad.cxx

eumaka:


aohlson:
+1

vpacik:
Hi @eumaka, looking at the commits in the past, as a suggestion did you consider instead of having the fix value for these cuts hard-coded in the task, implementing them as variables together with some setters? By doing so, you could change the value whenever needed without actually committing them every time.

eumaka:
@vpacik, indeed. This will be done in the next major upgrade of the analysis task.

PR #9779, opened 2019-04-29
PWGGA/GammaConv: trigger mimicking for heavy meson task & more

fjonasALICE:
trigger mimicking for heavy meson task & preparations for running over LHC18l6b1 + c1


nschmidtALICE:
+1

PR #9790, opened 2019-04-29
fixed handling of case when no efficiency file is loaded/applied

gyronaut:
Fixed crash when running with no efficiency file loaded in AddTask/train. Instead, code now correctly runs with no efficiencies applied.


vpacik:
+1

PR #9793, opened 2019-04-30
A change in AliAnalysisTaskDiHadCorrelHighPt.cxx

lhusova:
add a line to be sure there are no crashes


vpacik:
+1

lhusova:
I don't understand the crash here and see some other commits receive the same problem. How could it be solved?

PR #9794, opened 2019-04-30
Fix generated photon from pi0 or eta pT weights estimation

gconesab:


gconesab:
I guess this is a glitch in the system?

PR #9798, opened 2019-04-30
Increase pt-range of the tracking QA task up to 350 GeV

mfasDa:
Add bins(<250 GeV - 5 GeV width, < 350 GeV - 10 GeV width)
in the pt-binning in order to address the tracking performance
in the ultra high pt region.


mfasDa:
+1

mfasDa:
+1

PR #9801, opened 2019-04-30
PWGGA/GammaConv: fixed heavy track matching & more

fjonasALICE:
- fixed track matching of heavy meson task which wasn't applied correctly
- added weights for LHC18l6b1+c1


nschmidtALICE:
+1

PR #9813, opened 2019-05-01
Add missing TGrid::Connect

fgrosa:


jezwilkinson:
Hi Fabrizio,
Just to be sure about this, is this loading with grid connectiin done at a step that is still at job creation stage, or on the worker node? If I remember right UserCreateOutputObjects is on the node..
Generally speaking it is discouraged to have the individual worker nodes connecting to the grid at runtime to collect external files, so if possible it may be better to access those files during the AddTask and attach them to the job then, similar to what is done for the cutfiles with other tasks, unless there is a very good reason for downloading on each individual worker node.

Cheers,
--Jeremy

fgrosa:
Hi Jeremy, 

yes you are right, I modified it so to load the input file in the addtask. Moreover, since I was anyway too late for today's tag, I added the preselection to speed up a bit the task in PbPb.

Cheers,
Fabrizio

jezwilkinson:
Hi,
Sorry, I'm a little confused by the new changes here. Did you test this locally? I wasn't aware that ROOT allowed you to import a TFile object as a member within another object that is then stored inside a TFile.. 
If it does work as expected then I suppose it's fine, but my suggestion was more of adding the various "fHistEff*" to the task object during the AddTask step (so, doing the stage of LoadEffSystFile() either during the constructor or within the AddTask before it is added to the manager). Putting the file as a class member and then pulling the objects out at runtime seems like it just adds unnecessary steps to the processing, unless there is something that might change which objects need to be accessed per job.
Sorry if this wasn't clear before..
Cheers,
--J

fgrosa:
Hi Jeremy, 
I tested it locally and it worked. Indeed you are right about the pulling the objects in the usercreateoutputobjects (even if I believe that it is done only once per job), so I changed it to load the histos in the addtask. I just did not add it in the task creator because in the creator it's not possible to return if the file does not contain the correct histos.

Cheers,
Fabrizio

jezwilkinson:
+test

jezwilkinson:
+1

PR #9828, opened 2019-05-01
Logarithmic histogram bins for PosSwap modification

moliver813:
Adds logarithmic histogram bins for PosSwap modification
Adds option to use linear bins


e-masson:
+test

e-masson:
+1

PR #9844, opened 2019-05-02
Update Ds syst unc for cut variation in pPb vs mult

fgrosa:
Modify systematic uncertainties from cut variation of Ds vs. mult in pPb to match those evaluated for the MB measurement @cbedda


jezwilkinson:
+1

PR #9865, opened 2019-05-03
PWGUD/UPC: Added code for Invariant Mass distr J/Psi polarisation 2D.

siragoni:
Added code to obtain the invariant mass distributions in bins of CosTheta and Phi in the helicity frame of the J/Psi to extract the signal before computing the polarisation parameters.
The histograms have been created as a TH1F***, possibly not the best option, but solid enough.
IMPORTANT: tested on local, there doesn't seem to be any problem at all. The logic seems to be solid enough. On the small test run there is not enough statistics to see results... I can only see there are a few entries where the J/Psi's peak is supposed to be.


mbroz84:
+1

PR #9897, opened 2019-05-06
Adding different statistical methods to the AliGFWFlowContainer

vvislavi:
Single subsample, jackknife, proper bootstrapping


vpacik:
+1

PR #9912, opened 2019-05-07
PWGGA/GammaConv: added OmegaMCStudies task

fjonasALICE:
introduced OmegaMCStudies task that in the future can be used to study various omega properties with MCGen trains on the GRID. For now, it will be only used to study pT of pi0 in dependence of pT of omega


dmuhlhei:
+test

dmuhlhei:
+1

PR #9931, opened 2019-05-08
Force explicit cast to EMCAL_STRINGVIEW to circumvent problems on old…

mfasDa:
…er platforms


mfasDa:
+1 

This fix is needed for users still using the old ROOT6 version (v6-10-08) from the root6/user-root6 defaults which is still shipped with the buggy version of the std::string_view.

PR #9949, opened 2019-05-08
PWGUD/UPC: Trying to fix.

siragoni:
Dear @mbroz84, this macro is actually the same as version1, just I am commenting out the lines we discussed about to check if those very same lines were creating problems in there...
IMPORTANT: tested.


mbroz84:
+1

PR #9963, opened 2019-05-09
PWGHF: MVA, add generated pT for reco cand

lvermunt:
Add generated pT (matched to reco candidates) to evaluate MC pT shape systematic


afestant:
+1

PR #9964, opened 2019-05-09
PWGGA/GammaConv: Modifications on the trigger simulation width for pp…

jokonig:
… 13 TeV dataset; included LHC18x to trigger mimicking


nschmidtALICE:
+1

PR #9970, opened 2019-05-10
Fix compilation issue when fastjet is disabled

dstocco:
The class AliAnalysisTaskHFJetIPQA.cxx was added twice in the SRCS of the CMakeLists.txt:
once in the SRCS that are always compilend and another time in the SRCS that are compiled only when fastjet is found.
Since the task depends on fastjet, it has to be added only in the second case.


dstocco:
Hi all,
can anyone trigger a test please?
Thanks in advance,
best regards,
Diego

rhaake:
+test

rhaake:
Hi @kagarner, I think you are actively working on this task. Can I approve this change?

rhaake:
+1 (just got the OK from Katharina per mail)

PR #9976, opened 2019-05-11
Add option for the Q/pt

mfasDa:
Q/pt shift is added (meaning 1/pt shift added for positive charged particles and subtracted for negative charged particles)


mfasDa:
+test

mfasDa:
+1

PR #9982, opened 2019-05-13
PWGUD/UPC: Added pt-distr analysis for neutron emission classes.

siragoni:
Added the analysis of the pt-distribution divided for neutron emission classes.
IMPORTANT: tested.
IMPORTANT: still using vectors... Awaiting tests to switch to new code.


mbroz84:
+1

PR #9991, opened 2019-05-13
Sphericity of events containing strange particles

jpriskin:
Study of sphericity of events containing strange particles of purpose of first fast check.


mmeres:
+1

PR #9995, opened 2019-05-13
PWGUD/UPC: Uncommented event selection in test macro.

siragoni:
Dear @mbroz84 , 
this commit is what we have just discussed... 
I have checked and it works fine on local. 
Best wishes,
Simone


mbroz84:
+1

PR #10002, opened 2019-05-13
Maintenance on AliAnalysisPIDtrack and AliGFWFlowContainers

vvislavi:
1) fgTOFPIDResponse in AliAnalysisPIDtrack.h is now public static
2) AliGFWFlowContainers now also store the pT-axis of the preset binning I also add the default pT binning, in case older files are used


vpacik:
+1

PR #10018, opened 2019-05-14
PWGUD/UPC: Switched to arrays for run selection...

siragoni:
After experiencing some issues due to the usage of c++11's vectors on the LEGO framework, we have decided to go back to normal arrays... Only the tests on the datasets will prove if this works fine.
IMPORTANT: tested on local => same results as before (not sure about the total execution time though).


mbroz84:
+1

PR #10029, opened 2019-05-14
Update usage of D0 reflections in MultiTrial

strogolo:
Usage of D0 reflections showed some problems when using fBkg>kPol2. The issue was related to how the background functions was passed to the MassFitter. This fix works and is currently used for the D0 raw yield systematic evaluation on Pb-Pb 2018 data sample.

### DIFF
--- a/PWGHF/vertexingHF/AliHFMultiTrials.cxx
+++ b/PWGHF/vertexingHF/AliHFMultiTrials.cxx
@@ -32,7 +32,7 @@ ClassImp(AliHFMultiTrials);
 
 
 //_________________________________________________________________________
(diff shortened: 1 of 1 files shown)


fprino:
Ciao Stefano,
why do you use the old class of multitrials?
In the one the uses the new fitter (AliHFInvMassMultiTrialFit) the reflection functionalities are there.
Ciao, Francesco

strogolo:
Ciao Francesco,
sorry, my mistake. I was pointed to the old one and thus I started doing the systematic with the old AliHFMultiTrial. I'll update my macro and this PR can be rejected or closed.
Ciao,
Stefano

PR #10041, opened 2019-05-14
Adds an energy swap map for determining same-position PS distribution

moliver813:
Adds an energy swap map for determining same-position PS distribution
Also cleans up a few couts.


moliver813:
Found a potential error I need to fix first; I think the SetE option isn't what I want, as it only changes the E entry of the TLorentzVector

moliver813:
Should be correct now

e-masson:
+test

e-masson:
+1

PR #10042, opened 2019-05-14
Use same SPD acceptance in 2015-2018 Pb-Pb

strogolo:
Add the flag and the method to use the same SPD acceptance in 2015 and 2018 Pb-Pb analysis.
The two regions excluded are defined in order to avoid to use candidates which could have passed in SPD HSs 1A0 or 5C0, which were out during 2018 Pb-Pb data taking period.


afestant:
+1

PR #10046, opened 2019-05-15
PWGUD/UPC: Finer binning for pt-distributions.

siragoni:
Increased the binning for all pt-distributions in data.
This is needed to get to the same binning as presented in the last paper...
IMPORTANT: tested.


mbroz84:
+1

PR #10048, opened 2019-05-15
Recover filling of the tree only when track is tagged

fgrosa:
Modify code to fill tree only when track is tagged from V0 / kinks / tight selection on TOF or TPC (for kaons), otherwise now filled always because of introduction of charge in bit map


afestant:
+1

PR #10049, opened 2019-05-15
PWGUD/UPC: Fixed the run selection. Adjusted the binning of the pt-distributions.

siragoni:
Switched to arrays for the run selection even in the MC macro...
Fixed the binning of the pt-distributions to be able to reproduce the same results as the paper...
IMPORTANT: tested.


mbroz84:
+1

PR #10075, opened 2019-05-16
fixed bug

ysekiguc:


vpacik:
+1

akubera:
@ysekiguc Note: C++ supports auto-sizing static c-arrays:

```cpp
Double_t binning_pt_fmdtpc[]={0.2,3.};
Int_t count = sizeof(binning_pt_fmdtpc) / sizeof(Double_t); // returns 2
```

makes for one less thing to think about.

PR #10077, opened 2019-05-16
Added gestion of eta gaps for the 2-p correlations

cimordas:
* Rewriten the TLists inside the output file.
* Added a new cut and TH2I correlation plot to try and remove high multiplicity outliers.
* Modified a bit the already existing cuts.


vpacik:
+1

PR #10083, opened 2019-05-16
Adding a new descriminator for the templates

hahassan7:
We decided to add a new discriminator to be used as a template for the data-driven efficiency determination.


btrzecia:
+1

PR #10084, opened 2019-05-16
PWGGA/GammaConv: omega cut config changes

fjonasALICE:
change of DCA track cut. It was found that pT dependent DCA cut causes non-uniform phi distribution for tracks for LHC10


nschmidtALICE:
+1

PR #10091, opened 2019-05-17
Removes unnessessary function and equalizes to nano to AODs

bhohlweger:
Invariant mass of pairs was used for a short investigation, however it is not used anymore and just lingers around. 
Additionally a redifinition of the VEvent was removed, which was already a member of the analysis task. Further pile up rejection & dcas were taken from the 128 track rather than the global track - fixed.


bhohlweger:
+1

PR #10095, opened 2019-05-17
Add macro for DPMJET

qgp:


FriederikeBock:
@qgp maybe you want to call this macro PWG/MCLEGO/AddMCGenDPMJET_pPb.C as its not setup to run anything else.

PR #10111, opened 2019-05-18
Update Dzero systematics binnings for 2018 Pb-Pb

strogolo:
Updating the binnings for 2018 Pb-Pb systematic uncertainties for the Dzero in 0-10 and 30-50. The values reported are still those of 2015 analysis and will be updated with the finals as soon as they will be finalized.


fprino:
+1

PR #10127, opened 2019-05-20
Minor improvements for Sigma0 and NanoAOD analysis

a-mathis:
- The v0 of the photons is not saved in the NanoAOD -> protection implemented
- Enable Addv0InESDs for the photons
- Check gamma+K0 inv. mass for weird structures (that hopefully won't show up)


mpuccio:
+1

PR #10130, opened 2019-05-20
Added a correlation histogram

cimordas:
* Modified the filling of the TH2I for the correlations between types of tracks to see the HM outliers.
* Added a second TH2I after the cuts to see their effects on the correlations.


vpacik:
+1

PR #10139, opened 2019-05-20
Adding new weights

chiarazampolli:
- central: noP, noPCts
- semicentral: noP

note that we do not compile the noPCts version, but we have the weights.


chiarazampolli:
Pinging @andreaalici for information.

shahor02:
+1

jezwilkinson:
Hello @chiarazampolli ,
on looking at this it seems it adds some "noPCts" files (e.g. LHC19c2a_TMVAClassification_BDT_12_25_noPCts.class.h) that are not included in the new version of the linkdef/CMakelist, is this intended?

chiarazampolli:
Ciao @jezwilkinson , 

Yes, see comment in the PR. We prepared the files, but since compilation is long, and since today we discussed that the cosThetaStar should be included, for now we go like this. If we want to change, the files are there, we just need to enable them.

Cheers,

Chiara

PR #10141, opened 2019-05-21
Last Test on PsiPairCut like Dieletrons like everyone else

eddoming:
I add a new way to calculate PsiPair cut on Dalitz, with the propagation of the momentum on plane xy plus 50.


nschmidtALICE:
+1

PR #10152, opened 2019-05-21
Implement variable FD hypo to be consistent with RAA

fgrosa:
@arossi81 @strogolo @RenuBala123 @syaefudinjaelani note that the variable hypo is needed also for the yield. Could you please double check that it is consistent with what is done in HFPtSpectrumRAA.C? Thanks!


strogolo:
Ciao @fgrosa, I checked the changes you did and they are consistent with what I implemented in the HFPtSpectrumRaa.C. Sorry for missing to change also this macro and thanks a lot for the update.
I'll re-run the analysis with these updates.
--Stefano

jezwilkinson:
+1

PR #10166, opened 2019-05-22
Fix option for variable FD hypothesis

fgrosa:
@arossi81 @strogolo @RenuBala123 @syaefudinjaelani note that the version of yesterday (PR #10152) does not work properly because it modifies the FD hypo in the full pT range


jezwilkinson:
+1

PR #10212, opened 2019-05-24
PWGGA: Copied 5TeV MaterialBudget Cuts for 13TeV and changed NonLin Cut

jluehder:
The Non Linearity Correction of the 5TeV Cutstrings is not available for 13TeV.
Therefore i have created own configs for 13TeV instead of changing the 5TeV Configs


nschmidtALICE:
+1

PR #10217, opened 2019-05-25
 Add bit map for track quality info in ttree 

fgrosa:
Add bit map to store SPD information (kFirst, kAny), ITS refit, TPC refit, and geometrical cut in output tree


arossi81:
+1

PR #10232, opened 2019-05-27
 PWGGA/GammaConv: Implemented TPC dE/dx recalibration OADB for conversion tasks

nschmidtALICE:
By using "enableElecDeDxPostCalibration = kTRUE" and not setting a dEdx recalib file name one can now acess the OADB file stored in OADB/PWGGA which contains period-dependent recalibrations.
If one also sets a filename, then the recalibration parameters will be loaded from the file instead of the OADB (this is the old standard procedure).


nschmidtALICE:
The required approval from the people listed above is only needed for the readme file:  OADB/PWGGA/ReadmeRecalibOADB.txt
(I would also like to have approval rights in the OADB/PWGGA folder as the maintainer of the new OADB objects)

nschmidtALICE:
+test

dmuhlhei:
+1

PR #10249, opened 2019-05-28
Add systematic uncertainty from geom cut correction for Ds and D+

fgrosa:
Added systematic uncertainty from geometrical cut correction for Ds and D+ summed in quadrature to tracking and ME systematic uncertainty:
4% in 30-50% centrality class
6% in 0-10% centrality class


cterrevo:
+1

PR #10253, opened 2019-05-28
Update D0 and Dstar systematic uncertainties

strogolo:
Update of D0 systematic uncertainties in 0-10 and 30-50:
- MC pt-shape (new)
- Cut efficiencies (new)
- Tracking efficiency (include syst. from correction factor)

Update of Dstar systematic uncertainties in 0-10 and 30-50:
- Tracking efficiency (include syst. from correction factor)


cterrevo:
+1

PR #10256, opened 2019-05-29
Added ability to test results with HFe paper cuts and settings

efgauger:
No able to turn each cut on/off independently. Also added histograms to test effect of E/p shift in MC, event cuts on the nEvents, and cluster time cuts.


deepathoms:
+1

PR #10259, opened 2019-05-29
PWGUD/UPC: Added triggers' histograms for luminosity computation.

siragoni:
Added triggers' histograms for luminosity computation.
Fixed run numbers for LHC15o.
IMPORTANT: tested.


mbroz84:
+1

PR #10274, opened 2019-05-30
Adapt event selection to CALOFAST cluster

mfasDa:
For CALOFAST cluster the event selection must
change from kINT7/kEMCEJE to kMuonCalo. In addition
the downscale factors must be used from the CALO cluster.


mfasDa:
+1

PR #10318, opened 2019-06-04
Adding artificial scale shift to det. level pt

mfasDa:
The scale shift if apply should be roughly compatible
to the observed scale shift in simulations. It is added
to the measured pt value.


mfasDa:
+1

PR #10329, opened 2019-06-04
Added Sumw2() calls for TProfiles in AliGFWFlowContainer (default in …

vvislavi:
…root6, but apparently not in root5)


vpacik:
+1

PR #10340, opened 2019-06-05
[Doxygen] Add group for classes of the EMCAL correction framework

mfasDa:
For easier navigation move all classes related to the EMCAL correction
framework to a dedicated group EMCALCORRECTIONFW in doxygen.

### DIFF
--- a/PWG/EMCAL/Doxymodules.h
+++ b/PWG/EMCAL/Doxymodules.h
@@ -49,6 +49,14 @@
  */
 
 /**
+ * \defgroup EMCALCORRECTIONFW EMCAL correction framework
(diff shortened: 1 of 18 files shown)


mfasDa:
+test

mfasDa:
@raymondEhlers What do you think? I think it makes it easier to find the class documentation of any correction framework class if it is in a separate group, and the correction framework can be considered as a package separate from the core EMCAL framework. Please merge if you agree! The group could get a short description (in doxymodules.h) which can happen at a later stage.

raymondEhlers:
+1 @mfasDa Thanks for this nice improvement! It looks good to me

PR #10350, opened 2019-06-06
Possibility to cut time ranges based on global event id

wiechula:
The commit contains
* Time range class with mask
* Container class to keep a list of time ranges
* Cut class to read time ranges from OADB
* Macro to set up time range OADB


chiarazampolli:
+1

chiarazampolli:
+1

mfasDa:
Hi @wiechula, Hi @mpuccio,

When running on MC with disabled time cut we get 100% if the events rejected. Looks like the flag for the time cut has to be enabled in case the cut is disabled (which is currently not the case), otherwise the event get rejected due to missing time cut.

Could you please have a look?

Thanks in advance!

Cheers

Markus

PR #10355, opened 2019-06-06
fixing the filling of TOF values

amaringarcia:
Dear all,
indeed without the change the TOF values were not filled in for the PhotonQA.
now they are.
The condition I am using to select causes were TOF signal is there , I have seen for example in dielectrons.

Ana


nschmidtALICE:
+1

PR #10356, opened 2019-06-06
Fix initialization of histogram containers

mfasDa:
Histogram arrays no longer created in the constructor
but in the UserCreateOutputObjects as they are non-streamable,
and memory allocation in the dummy constructor should be
prevented.


mfasDa:
+1

PR #10366, opened 2019-06-07
Extract trigger selection tools to new class

mfasDa:
Use multiple inheritance in order to share common code
for trigger section among several classes: The implementation
code is extracted into an implementation class
(AliAnalysisEmcalTriggerSelectionHelperImp) which can be
derived from in analysis tasks (no inheritance from TObject).
For classes which need streamer functionality from TObject
i.e. for encapsulation the interface class
AliAnalysisEmcalTriggerSelectionHelper should be used.


mfasDa:
+test

Not sure this works under ROOT5, but it it works it is appealing in order to reduce code duplication.

mfasDa:
+test

mfasDa:
+1

PR #10375, opened 2019-06-07
Do time cut initialisation only once per run

mpuccio:


fprino:
+1
Pinging @wiechula for info

PR #10376, opened 2019-06-08
PsiPair correction with a new train to test ITS option kBoth

eddoming:
Remove extra variables to propagate the momentum like dielectrons, and adding a TH3F for extra study on PsiPair and theta in terms of the momentum. Plus an extra train ID for the study of ITS.


dmuhlhei:
+1

PR #10381, opened 2019-06-09
PWGCF: AliFemto - Simplify & standardize Report() methods + trim trailing whitespace

akubera:


akubera:
This replaces many calls to `snprintf` with ROOT's `Form`, and replaces all uses of an awkward un-namespaced `string` (`std::string`) with the typedef'd `AliFemtoString` (which ends up being std::string anyways). Trailing whitespace was automatically removed by my editor.

PR #1035, opened 2017-05-22
PWG/CALO: fix missing return value

dberzano:

PR #10388, opened 2019-06-10
Set default flag in case time cut is not applied

mfasDa:
Fixing alisw/AliPhysics#10350: In case the time cut is not
applied the flag has to be set, otherwise all events
are rejected.


mfasDa:
@mpuccio Please review! This fixes an issue introduced in alisw/AliPhysics#1035 that all events are rejected in case the newly introduced time cut is not applied.

mpuccio:
Thanks Markus! This actually was missing...

chiarazampolli:
+1

PR #10410, opened 2019-06-12
Add task for EMCAL trigger normalization based on downscale method

mfasDa:
Calculate the donnscale-corrected number of min. bias events
(cluster luminosity) and the corrected number of events for
the various EMCAL triggers. The event counts are organized in
centrality percentiles.


mfasDa:
+1

PR #10415, opened 2019-06-12
PWGGA/GammaConv: omega task light output fixes & more

fjonasALICE:
- fixed some wrong histo filling in omega task when running with light output
- added pp13TeVanch17 with decay photon trigger to use NonLin of pp13TeV Jet-Jet


nschmidtALICE:
+1

PR #10425, opened 2019-06-12
the function should be passed to the task

chengtt0406:


chengtt0406:
the ! exclamation mark was removed because we pass this function from the add-task
can someone please approve it?

thanks in advance

jezwilkinson:
+1

PR #10432, opened 2019-06-13
Require L0 trigger used for L1 in trigger string

mfasDa:
In order to differenciate between the different L0
triggers used for EMCAL L1 (i.e. INT7 and 0TVX) the
L0 trigger is required when matching the trigger
class for the determination of the downscale weight.

In addition: Couple of trivial bug fixes (forgotten
exit condition, missing general histograms)


mfasDa:
+1

PR #10444, opened 2019-06-13
Add new vertexer to hypertriton task

fmazzasc:


PWGLF/NUCLEX/Hypernuclei/Hyp2Body/AliAnalysisTaskHyperTriton2He3piML.cxx
```diff
@@ -310,11 +320,17 @@ void AliAnalysisTaskHyperTriton2He3piML::UserExec(Option_t *)
     return (sig - expS) / (fCustomResolution * expS);
   };
 
-  for (int iV0 = 0; iV0 < esdEvent->GetNumberOfV0s();
+  esdEvent->ResetV0s();
+  AliVertexerHyperTriton2Body *V0Vertexer = new AliVertexerHyperTriton2Body();
```
mpuccio: Why do you create a new vertexer for every event? Do you delete it? If not you will have a memory leak... Put a `AliVertexerHyperTriton2Body` object as public data member instead: this will also allow the user to configure it.

mpuccio:
+1

PR #10490, opened 2019-06-19
PWGGA: Added bad channels for LHC18 to existing PHOS bad channel map …

jluehder:
…(only periodwise, selected by hand)


nschmidtALICE:
+1

PR #10498, opened 2019-06-19
PWGGA/GammaConv: omega analysis: changed default PHOS track matching …

fjonasALICE:
…and added EMCal E/p matching


nschmidtALICE:
+1

PR #10502, opened 2019-06-19
Update AliAnalysisTaskSEHFTreeCreator destr

benedikt-voelkel:
Make the destructor more coherent by also deleting the dynamically allocated AliJetTreeHandler objects.


afestant:
+1

PR #10523, opened 2019-06-20
PWGHF/jetsHF: Reconstructing 4,5,6 prong SV for templates

hahassan7:
The SV fE shape was not so discriminative between the different jet flavors, so we reconstructed SVs with more prongs to a shape which similar to the one for the ATLAS.


btrzecia:
+test

btrzecia:
+1

PR #10524, opened 2019-06-20
PWGUD/UPC: Added maps 2D ETA-PHI with and without matching. Single Muon Triggering.

siragoni:
Added 2D maps in Eta vs Phi for single muons at forward rapidities. Single Efficiencies have been added too. The analysis is done with and without Single Muon Triggering.
IMPORTANT: tested.


mbroz84:
+1

PR #10534, opened 2019-06-21
PWGUD/UPC: Added cross-check for polarisation.

siragoni:
This commit adds checks for the polarisation analysis.
It is basically a confirmation that the solid I had implemented was solid enough...
IMPORTANT: tested.


mbroz84:
+1

PR #10545, opened 2019-06-22
PWGHF: Fix TreeCreator part2

lvermunt:
Continuation of PR10542, where I forgot to take care of a continue

(Generated MC TTree was not filled for Lc->pK0s when both Lc decays were enabled. Fixed.)


afestant:
+1

PR #10556, opened 2019-06-24
PWGUD/UPC: Added acceptance plots for the 2D polarisation

siragoni:
Added acceptance plots for 2D polarisation with my variable binning.
IMPORTANT: tested.


mbroz84:
+1

PR #10560, opened 2019-06-25
Update the methods to analyse AOD and MC events.

cimordas:
* Add QA histograms for the multiplicity distribution to check its evolution with the application of the selection criteria.
* Update the methods to cut the outliers in the MC events.
* Increase the readability of the code by diving the selection criteria in two methods: one for the event selection, one for the track selection.


vpacik:
+1

PR #10584, opened 2019-06-26
Remove FATAL condition in case ALIEN is not set

mfasDa:
FATAL condition makes it impossible to compile AliPhysics
with the o2 defaults which deactivate alien. Configuration
of ALIEN should be handled by the build recipe.

### DIFF
--- a/cmake/FindROOT.cmake
+++ b/cmake/FindROOT.cmake
@@ -182,29 +182,29 @@ if(ROOTSYS)
 
     # Assume AliEn support as external ROOT plugin.
     if(ALIEN)
-    add_definitions(-DWITHALIEN)
+        add_definitions(-DWITHALIEN)
 
-    # AliEn might bring some system libraries, we need to use them
-    if(EXISTS "${ALIEN}/lib")
-        link_directories(${ALIEN}/lib)
-    endif()
+        # AliEn might bring some system libraries, we need to use them
+        if(EXISTS "${ALIEN}/lib")
+            link_directories(${ALIEN}/lib)
+        endif()
 
-    # api/lib should always exists
-    if(EXISTS "${ALIEN}/api/lib")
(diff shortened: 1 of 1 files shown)


mfasDa:
@ktf @Atlantic777 Please review! The FATAL_ERROR condition makes it impossible to compile AliPhysics with o2 defaults as alien is not built in this case. It will also make compilation without Alien in general impossible, where building without alien might be a desirable feature on certain systems without outgoing network support. The FATAL is also not really needed as the alien settings are handled in any case by the build recipes, and for the release or root6 defaults the AliEn-Legacy-ROOT plugin is built in any case.

sawenzel:
I agree. I think it is beneficial to be able to compile AliPhysics together with O2.

mfasDa:
Any update on this issue?

sawenzel:
+test

sawenzel:
@mfasDa : The git diff shows many changes lines not related to this PR. Could you just change the one particular line?

mfasDa:
@sawenzel I also fixed the code indentation in order to make the cmake file more readable. I would still prefer to fix this as well, but I could try to split the change to two separate commits, handled in the same pull request - is this fine with you?

sawenzel:
Ok, if you wish.

sawenzel:
+test

mfasDa:
@sawenzel, OK, I extracted the code indentation in the ALIEN if-block into a separate commit. The original one (https://github.com/alisw/AliPhysics/pull/10584/commits/2c6ebae8d9b1fc62840e44546db68a0810b608ec) now only contains the FATAL condition blocking builds without ALIEN (i.e. using the o2 defaults).

sawenzel:
Merging this (there is no opposing argument and it seems to be a useful change).

sawenzel:
+1

PR #10589, opened 2019-06-26
More V0M diagnostic histgrams added

fkrizek:


mfasDa:
+1

Please avoid commits such a huge amount of changes as they are close to impossible to check. In addition the hard coded list of VZERO signal properties per run can be significantly simplified, as the VZERO properties are the same for all runs in a certain range: You just need a struct with 5 values (run min, run max and the 3 properties) and you initialize 1 value per range. It will also be more efficient as instead of O(2000) tests to find the run you need only O(10), if done with binary search only O(LOG(10)).

PR #10622, opened 2019-06-30
Moved stack loop after event cuts, added TPC+ITS refit.

efgauger:
(Trying again - last request failed with empty merge error.)


deepathoms:
+1

PR #10627, opened 2019-07-01
Add vertex-Z selection skipper for EvSelCode

ddobrigk:
This feature is meant to cover a specific use case in which event losses due to vertex-Z selections are studied. It does not change the default behavior of the task - it only introduces a switch so that, if that switch is enabled, the EvSelCode from the framework will reflect all event selections used at calibration time except the vertex-Z position check.


fprino:
+test

ddobrigk:
+1

PR #10637, opened 2019-07-01
delthaPhi with constrained parameters and train for kAny and kBoth

eddoming:
Dear AliPhysics,
On the next commit, I add a function for deltaPhi on terms of constrained param and the vertex selection for the events.
Regard
Ed


FriederikeBock:
+1

PR #10647, opened 2019-07-02
Update of run by run mean VZERO signal values

fkrizek:


mfasDa:
+1

For further commits I request that the hard coded db of VZERO signal as function of the run number is outsourced to a different class.

PR #10654, opened 2019-07-02
PWGUD/UPC: Increased dynamical range of the ZDC.

siragoni:
Increased dynamical range of the ZDC after having a look at the:

https://alice-notes.web.cern.ch/system/files/notes/analysis/481/2017-Jul-14-analysis_note-ALICE_analysis_note.pdf

This is needed for the XNXN analysis to check for the QA of the muoncalo_pass3.
IMPORTANT: tested.


mbroz84:
+1

PR #10667, opened 2019-07-03
PWGUD/UPC: Improvement on ZDC for MB + new 1D variable binning pol analysis.

siragoni:
Improved the ZDC calibration stuff for MB by resetting each time the ZDC values and better default value. 

Added new variable binning analysis for 1D polarisation.

IMPORTANT: tested.


mbroz84:
+1

PR #10669, opened 2019-07-03
Further checks for p-Sigma0 femto

a-mathis:
- Check the impact of the track cleaner
- Run some checks for the transverse radius of the photon (a lot of background when 70 < r < 80 cm, which corresponds to the inner part of the TPC)


a-mathis:
+1

PR #10672, opened 2019-07-03
changes to LctopKpi Syst values for pPb 2016 and pp 2017

chills-ALICE:
changes to Syst values for LctopKpi2016pPb and addition of LctopKpi2017pp values to AliHFSystErr


afestant:
+1

PR #10675, opened 2019-07-03
PWGGA/GammaConv & OADB: fixing handling of TRD triggers for run2 pp

FriederikeBock:
- correct the defaultPP object for run 2 regarding the TRD triggers
- adapted TRD trigger selection for run2 data to the correct trigger strings


FriederikeBock:
@iarsene this now should fix the OADB object for the TRD

ddobrigk:
+1

PR #10703, opened 2019-07-05
Minor fixes to get the Lambda-Xi analysis running

a-mathis:
- Fix numbering of output containers
- Fill the global track array to use the daughter tracks


a-mathis:
+1

PR #10704, opened 2019-07-05
Add option for multiplicity-dependent normalization in p-Pb

mfasDa:
If requested normalization histograms are filled as function
of centrality. Must be enabled explicitly by the user as
the task will crash in case the AliMultiplicitySelection is
not found.


mfasDa:
+1

PR #10738, opened 2019-07-09
PWGGA/GammaConv Additional Sphericity calc for more MC particles

mikesas:


PWGGA/GammaConv/AliAnalysisTaskGammaCalo.cxx
```diff
@@ -1298,6 +1304,9 @@ void AliAnalysisTaskGammaCalo::UserCreateOutputObjects(){
     fHistoPionSpectrum                      = new TH1F*[fnCuts];
     fHistoProtonSpectrum                    = new TH1F*[fnCuts];
     fHistoKaonSpectrum                      = new TH1F*[fnCuts];
+    fHistoNPionSpectrum                     = new TH1F*[fnCuts];
```
nschmidtALICE: could you encapsulate all sphericity histograms also here in the if() condition? The "new" command will already here allocate memory which we would like to avoid in non-sphericity running of the task.

mikesas: Done

nschmidtALICE:
+1

PR #10759, opened 2019-07-10
MC true and det level can be analyzed in parallel.

osaarima:
The fIsMC in AliJCDijetTask will decide if additional MC is ran. It can be set on and off in a macro. The detector level MC has its own histograms and analysis object. Next addition will be the creation of response matrices in a new function.


vpacik:
+1

PR #10773, opened 2019-07-10
Remove obsolete members

raymondEhlers:
They were moved to the JetHUtils class, but I forgot to remove them here


raymondEhlers:
+1

PR #10784, opened 2019-07-11
adding 2D profiles for SPD and V0 corrections - J/psi EMCal analysis

crisjahnke:
SPD-vtx correction changed to 2D correction, based on group of runs.
V0-aging correction added as 2D profile, where vertex and agind are corrected on the same time.

Files of profiles updated.


zwound40:
+1

PR #10785, opened 2019-07-11
Fix compiler warnings

mfasDa:
Comparison of non-zero value with constant in operator&&
always returns true, bitwise operator& expected.


mfasDa:
During recompilation I came across these compiler warnings which are rather trivial to fix. Warnings are reasonable and the behaviour before was not intended like this. @adubla for review!

adubla:
+test

adubla:
+1

ktf:
+1

PR #10803, opened 2019-07-12
PWGUD/UPC: XNXN systematics + new binning 1D.

siragoni:
Added systematics plots for the neutron emission analysis. 
The plan is the following:

- Coh   Inv Mass Spectra 
- Coh   Inv Mass Spectra XNXN
- Incoh Inv Mass Spectra 
- Incoh Inv Mass Spectra XNXN

x5 due to the different Pt-thresholds (0.200, 0.225, 0.250, 0.275, 0.300). In addition, 

- Pt-distr (normal + XNXN) with 20 MeV/c shift

Furthermore, new binning for 1D pol + 1D acceptance.

IMPORTANT: tested.


siragoni:
The last commit is to implement the signal extraction in this new approach.
I have already checked the histograms, and it seems to work nicely.
The problem is that I am not really selecting J/Psi in this way and this last commit would fix that.

mbroz84:
+1

PR #10821, opened 2019-07-15
Modifying the correction framework so it takes one histogram instead of four for all BCs

hahassan7:
The correction framework was modified by adding a switch whether someone want to use one histogram for the time calibration instead of four for all the BCs. Also in case it is one histogram, this histogram will be TH1S and not TH1F.

### DIFF
--- a/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellTimeCalib.cxx
+++ b/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellTimeCalib.cxx
@@ -56,6 +56,10 @@ Bool_t AliEmcalCorrectionCellTimeCalib::Initialize()
   if (!fRecoUtils)
     fRecoUtils  = new AliEMCALRecoUtils;
     
+
+  if (fDoMergedBCs)
+    fRecoUtils->SetUseOneHistForAllBCs(fDoMergedBCs);
+
   fRecoUtils->SetPositionAlgorithm(AliEMCALRecoUtils::kPosTowerGlobal);
 
   return kTRUE;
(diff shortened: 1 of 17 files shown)


PWG/EMCAL/EMCALtasks/AliEmcalCorrectionComponent.h
```diff
@@ -104,6 +104,8 @@ class AliEmcalCorrectionComponent : public TNamed {
   void SetIsESD(Bool_t isESD) {fEsdMode = isESD; }
   void SetCustomBadChannels(TString customBC) {fCustomBadChannelFilePath = customBC; }
 
+  void SetUseMergedBCs(Bool_t useOneHist) {fDoMergedBCs = useOneHist; }
```
raymondEhlers: Please don't add a setter - it should be set via YAML.

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionTask.cxx
```diff
@@ -1321,6 +1325,7 @@ Bool_t AliEmcalCorrectionTask::Run()
     component->SetCentralityBin(fCentBin);
     component->SetCentrality(fCent);
     component->SetVertex(fVertex);
+    component->SetUseMergedBCs(fDoUseMergedBCs);
```
raymondEhlers: This attribute doesn't belong in the correction task. Each one of the other values that are set here vary event-by-event, but this one doesn't. It should be set individually for each component using YAML.

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionComponent.h
```diff
@@ -136,12 +138,14 @@ class AliEmcalCorrectionComponent : public TNamed {
   TString                fBasePath;                       ///< Base folder path to get root files
   TString                fCustomBadChannelFilePath;       ///< Custom path to bad channel map OADB file
 
+  Bool_t 		 fDoMergedBCs;			  ///< flag to use one histo for all BCs
```
raymondEhlers: This is only used in the time calibration, so it should be defined in the time calibration component

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellTimeCalib.cxx
```diff
@@ -56,6 +56,10 @@ Bool_t AliEmcalCorrectionCellTimeCalib::Initialize()
   if (!fRecoUtils)
     fRecoUtils  = new AliEMCALRecoUtils;
     
+
+  if (fDoMergedBCs)
```
raymondEhlers: The value should be initialized from YAML in this function before here. See the clusterizer component to see how and where to do this.

Remember to define the value in the default YAML file

hahassan7:
Hi Raymond,

Thanks for the comments, I will implement them ASAP.

amatyja:
Hi,
Just remember that before 2015 all 4 BCs were used and it has to be like that unless a large modification of constants will be present. During 2015-2018 4 BCs are used (plus additional 4 for Low Gain could be used). No modification that only one BC has been presented. Only one BC calibration constants cannot be a default unless this kind of calibration is finished and accepted. For future developments it is fine but do not introduce too much user free settings. It causes mistakes.
There are my additional 3 coins.
Regards

hahassan7:
Hi Adam,
Yes you are right. The flags are set to false by default, so by default four histograms will be used.

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellTimeCalib.h
```diff
@@ -43,6 +43,7 @@ class AliEmcalCorrectionCellTimeCalib : public AliEmcalCorrectionComponent {
   
   Bool_t                 fCalibrateTime;          ///< flag cell time calibration
   Bool_t                 fCalibrateTimeL1Phase;   ///< flag cell time calibration with L1phase shift
+  Bool_t		 fDoMergedBCs;		  ///< flag to use one histogram for all BCs
```
raymondEhlers: Need to increment the `ClassDef`

PWG/EMCAL/config/EMCalSampleConfig.yaml
```diff
@@ -6,6 +6,7 @@ CellBadChannel:
     enabled: true
 CellTimeCalib:
     enabled: true
+    doMergedBCs: false # flag to use one histogram for the time calibration instead of four for all BCs
```
raymondEhlers: This needs to go in the default YAML file, `AliEmcalCorrectionConfiguration.yaml`. Please also match the comment indentation in that file.

Have you tested the code since making these changes? Because as it stands, if settings are not in the mentioned file, the CorrFW should fail to configure

hahassan7: Hi Raymond,

I did all the modifications. I didn't make any test after removing the old setters. But you are right, it has to be tested before merging.

raymondEhlers: Okay, thanks. Everything now looks fine from my perspective, up to the need for testing

FriederikeBock:
+1

raymondEhlers:
Per the conversation above, I thought that this still needed to be actually tested?

hahassan7:
Hi All,

I made all the necessary tests with the switch On and Off and it is working. So we can merge now.

Best Regards
Hadi

raymondEhlers:
+1

FriederikeBock:
@gconesab could you approve please, we'd like to test it in production mode on the grid.
Regards,
Friederike

mfasDa:
+1

shahor02:
+1

lcunquei:
+1

PR #10822, opened 2019-07-15
Automatic storage selection for muon track refit

hupereir:
Added ::ReAlignFromDefaultStorage to use CDB settings from the ESD as old alignment storage and latest entry in OCDB for new alignment storage


pillot:
+1

PR #10831, opened 2019-07-16
adding option to not fill THnSparse, due to large output files

crisjahnke:
A flag was added to fill or not the THnSparse.
Few new histograms were added.
Some bins were changed, to reduce the size of the outputs.


iarsene:
+1

PR #10842, opened 2019-07-16
Fix for total number of TPC clusters wrong pointer when ESD events

victor-gonzalez:
The usual copy and paste mistake


vpacik:
+1

PR #10856, opened 2019-07-17
Fix efficiency implementation

raymondEhlers:
There was a bug in the eta parametrization, which required a
recalculation of the normalization


raymondEhlers:
+1

PR #10862, opened 2019-07-17
AliFemto: config to run pi-K femto on NanoAODs

lgraczykCern:
adding Config file for pion-kaon femto to run on NanoAODs in LHC11h data


lgraczykCern:
+1

PR #10865, opened 2019-07-17
[ALIROOT-8287] Reduce verbosity related to missing centrality

mfasDa:
Print warning in case centrality is missing only when requested.
Default: Always print warnings. Disabled for EMCAL trigger QA
in the QA train.


mfasDa:
+1

PR #10869, opened 2019-07-17
Added a DeltaR matching check for dijet components

osaarima:
The detector level leading and subleading jets are compared to MC truth
counterparts in phi-eta plane. The maximum was added as an argument in
the macro.


vpacik:
+1

PR #10900, opened 2019-07-19
Fix OCDB access with no AliEn (ALIROOT-8290)

dberzano:
When using the OCDB from CVMFS, the default storage is set to `raw://` and an
environment variable is set too (`OCDB_PATH`). In case the variable is set, we
should not attempt to connect to AliEn as we will not need it.

This covers the cases where AliEn access is deliberately disabled (e.g. the
release validation).

The fix is the same adopted in alisw/AliDPG#31.


dberzano:
+1

mconcas:
LGTM

PR #10913, opened 2019-07-19
[ALIROOT-8287] Swap order of internal event- and triggerselection

mfasDa:
For AliAnalysisTaskEmcalLight the trigger selection
is not called within the internal event selection,
consequently it has to be called before. Otherwise
the trigger selection is not applied.


mfasDa:
+1

mfasDa:
Needed because IsEventSelectionInternalI() shadows IsTriggerSelected().

PR #10867, opened 2019-07-17
Add D meson - HFe correlation task

hzanoli:
Dear all,

I have added a new task to generate trees to analyse D meson - HFe correlations. The yaml file was added to the macros folder. Please let me know in case you have any comments.

Best
Henrique

PR #10917, opened 2019-07-19
D-meson -- HFe correlations task

hzanoli:
Hi all,

I have corrected the previous pull request (#10867). The main problem is that I have used some C++11 features that ROOT5 cannot handle. I tried to port the conflicting parts, but this would involve doing quite some changes to the code. So I have preferred to keep the C++11 and disable the build for ROOT 5. The task works with no problem in ROOT 6. This is done by some changes in the PWGHFcorrelationHFLinkDef.h and CMakeLists.txt and it does not affect other tasks.
Please let me know if you have any comments.

Best,
Henrique


btrzecia:
+test

btrzecia:
+1

PR #10942, opened 2019-07-22
Fixes in obtaining lund map parameters

mfasDa:
- Don't apply hard cutoff, with current settings
  will reject all subjets
- Correctly use pt of the lower subjet
- Fix constructor of the splitting structure
- Fix iteration over histograms in manager
  when filling the output list


mfasDa:
+1

PR #10947, opened 2019-07-22
Allow usage of the jet task in subwagons

mfasDa:
Add necessary suffix parameter required for
the usage as subwagon. Attention: Suffix
parameter not yet added to the jets collection
name - this currently prevents to run more
than one jetfinder with the same jet definition
at the time.


mfasDa:
+1

PR #10958, opened 2019-07-23
fix to be able to run with and without MBW in the same train

amaringarcia:


nschmidtALICE:
Please also add this new feature to the pPb and PbPb AddTasks.

nschmidtALICE:
+1

PR #10962, opened 2019-07-23
More modification on the correction framework to take one histo for all BCs

hahassan7:
I made few more modification on the correction framework to take one histogram instead of four for all BCs.

### DIFF
--- a/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionComponent.h
+++ b/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionComponent.h
@@ -141,7 +141,7 @@ class AliEmcalCorrectionComponent : public TNamed {
   AliEmcalCorrectionComponent &operator=(const AliEmcalCorrectionComponent &);    // Not implemented
   
   /// \cond CLASSIMP
-  ClassDef(AliEmcalCorrectionComponent, 6); // EMCal correction component
+  ClassDef(AliEmcalCorrectionComponent, 8); // EMCal correction component
   /// \endcond
 };
 
--- a/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionComponent.h
+++ b/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionComponent.h
@@ -104,8 +104,6 @@ class AliEmcalCorrectionComponent : public TNamed {
   void SetIsESD(Bool_t isESD) {fEsdMode = isESD; }
   void SetCustomBadChannels(TString customBC) {fCustomBadChannelFilePath = customBC; }
 
-  void SetUseMergedBCs(Bool_t useOneHist) {fDoMergedBCs = useOneHist; }
-
   /// Set %YAML Configuration
   void SetYAMLConfiguration(PWG::Tools::AliYAMLConfiguration config) { fYAMLConfig = config; }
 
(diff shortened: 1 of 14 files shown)


gconesab:
+test

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionComponent.h
```diff
@@ -138,14 +136,12 @@ class AliEmcalCorrectionComponent : public TNamed {
...
 
-  Bool_t 		 fDoMergedBCs;			  ///< flag to use one histo for all BCs
-
  private:
   AliEmcalCorrectionComponent(const AliEmcalCorrectionComponent &);               // Not implemented
   AliEmcalCorrectionComponent &operator=(const AliEmcalCorrectionComponent &);    // Not implemented
   
   /// \cond CLASSIMP
-  ClassDef(AliEmcalCorrectionComponent, 7); // EMCal correction component
+  ClassDef(AliEmcalCorrectionComponent, 6); // EMCal correction component
```
raymondEhlers: I think this needs to be bumped to 8. If someone were to run a train with today's tag, I'm fairly certain that this could cause a problem in the future (although I'm not 100% sure). However, I'm sure that it's safe to bump it to 8, so that's what I would suggest

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionTask.h
```diff
@@ -244,16 +243,14 @@ class AliEmcalCorrectionTask : public AliAnalysisTaskSE {
...
-
   TObjArray                   fParticleCollArray;          ///< Particle/track collection array
   TObjArray                   fClusterCollArray;           ///< Cluster collection array
   std::vector <AliEmcalCorrectionCellContainer *> fCellCollArray; ///< Cells collection array
   
   TList *                     fOutput;                     //!<! Output for histograms
 
   /// \cond CLASSIMP
-  ClassDef(AliEmcalCorrectionTask, 7); // EMCal correction task
+  ClassDef(AliEmcalCorrectionTask, 6); // EMCal correction task
```
raymondEhlers: Same deal as with the CorrectionComponent. Please bump this to 8.

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionComponent.h
hahassan7: Hi Raymond,

Thank you for checking. Since we decided not to use setters instead use the yaml configuration, the number of parameters was reduced by one. The "fDoMergedBCs" was removed so we reduced the number of streamed parameters from 7 to 6.

raymondEhlers: Hi Hadi,

I understand that you're removing the member - I think I need to explain in more detail. Imagine the following:

- Someone uses today's tag, which will write a `train.root` file with `AliEmcalCorrectionComponent` at 7.
- We revert to 6 in tomorrow's tag.
- Someone makes a change later, and the ClassDef is bumped to 7.
- Someone opens the `train.root` from today's tag. The streamers don't match and opening the file fails.

Admittedly, it's not super likely, but it's possible. And we can avoid the issue entirely by bumping to 8. So that's why I suggested moving to 8

hahassan7: Ah OK, I see. Yes you are right. I will increment it to 8.

raymondEhlers:
+1 LGTM, thanks

gconesab:
+1

PR #10967, opened 2019-07-23
Fix AddTask (for trains) and minor changes in the variables

hzanoli:
Hi,

I have implemented minor changes to make it possible to run the analysis on trains. There was a small problem in the AddTrask file.

Best
Henrique


btrzecia:
+1

PR #10983, opened 2019-07-24
Fix container access in case of AliAnalysisTaskEmcalLight

mfasDa:
AliAnalysisTaskEmcalLight stores containers in associative arrays,
access via index meaningless. The default implementation of
CheckMCOutliers takes each the first container it finds in the
container map, but it is not guaranteed that this is indeed the
first container inserted by the user. Users should better
overwrite the function.


mfasDa:
+1

PR #10987, opened 2019-07-24
new class inheriting from AliAnalysisTaskEmcalEmbeddingHelper to add …

jaimenorman:
…further selection to embedded data


PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelperData.cxx
```diff
@@ -0,0 +1,97 @@
...
+
+AliAnalysisTaskEmcalEmbeddingHelperData::~AliAnalysisTaskEmcalEmbeddingHelperData()
+{
+}
+
+void AliAnalysisTaskEmcalEmbeddingHelperData::UserCreateOutputObjects() 
+{
+
+  // run function from embedding helper base class
+  AliAnalysisTaskEmcalEmbeddingHelper::UserCreateOutputObjects();
```
raymondEhlers: I'm not 100% sure if this would work, but created the histogram below duplicates a bunch of code. It might work to define the hist, add it to the hist manager, and then call the base class? Then you could remove adding the histograms, as well as the `PostData`

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelperData.cxx
```diff
@@ -0,0 +1,97 @@
...
+#include "AliVAODHeader.h"
+#include "AliEmcalList.h"
+
+/// \cond CLASSIMP
+ClassImp(AliAnalysisTaskEmcalEmbeddingHelperData);
+/// \endcond
+
+AliAnalysisTaskEmcalEmbeddingHelperData::AliAnalysisTaskEmcalEmbeddingHelperData():
+  AliAnalysisTaskEmcalEmbeddingHelper(),
+  fHistManager(0),
```
raymondEhlers: I don't think you need to initialize this here. It's done in the base class

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelperData.cxx
```diff
@@ -0,0 +1,97 @@
...
+  fPtSelectionMin(),
+  fPtSelectionMax()
+{
+}
+
+
+AliAnalysisTaskEmcalEmbeddingHelperData::AliAnalysisTaskEmcalEmbeddingHelperData(const char *name):
+  AliAnalysisTaskEmcalEmbeddingHelper(name),
+  fHistManager(name),
```
raymondEhlers: Same deal here

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelperData.cxx
```diff
@@ -0,0 +1,97 @@
...
+  if(nTracksInRange==0) {
+    fHistManager.FillTH2(histName, 1.5, ptMax );
+    return kFALSE;
+  }
+  else {
+    fHistManager.FillTH2(histName, 0.5, ptMax );
+  }
+
+
+  if(!AliAnalysisTaskEmcalEmbeddingHelper::CheckIsEmbeddedEventSelected()) return kFALSE;
```
raymondEhlers: I haven't checked this closely, but I would suggest moving this to the top of this function. No need to check the tracks if the event isn't selected anyway

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelperData.cxx
```diff
@@ -0,0 +1,97 @@
...
+
+Bool_t AliAnalysisTaskEmcalEmbeddingHelperData::CheckIsEmbeddedEventSelected()
+{
+
+
+  Int_t nTracks = fExternalEvent->GetNumberOfTracks();
+  Int_t nTracksInRange = 0;
+  Double_t ptMax = 0;
+  for(Int_t i=0;i<nTracks;i++) {
+    Double_t pt = fExternalEvent->GetTrack(i)->Pt();
```
raymondEhlers: Does this approach apply proper track selections when accessing tracks this way? Perhaps this doesn't matter for your purposes?

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelperData.h
```diff
@@ -0,0 +1,43 @@
...
+  virtual ~AliAnalysisTaskEmcalEmbeddingHelperData()                 ;
+
+  void            UserCreateOutputObjects() ;
+
+  void            AddPtSelection(Double_t min, Double_t max) { fPtSelectionMin.push_back(min); fPtSelectionMax.push_back(max);}
+
+  protected:
+  Bool_t          CheckIsEmbeddedEventSelected();
+
+  THistManager                fHistManager;///< Histogram manager
```
raymondEhlers: This shadows the hist manager in the base class. I don't think you need to redefine it

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelperData.h
```diff
@@ -0,0 +1,43 @@
...
+
+  std::vector<Double_t>       fPtSelectionMin; ///< min pt
+  std::vector<Double_t>       fPtSelectionMax; ///< max pt
```
raymondEhlers: Just a suggestion: you could use just one variable of `std::vector<std::pair<double, double>> fPtSelection`. This may be a bit more intuitive, and this enforces the length being the same. But the way that you've designed it also works perfectly fine

PWG/EMCAL/macros/AddTaskEmcalEmbeddingHelperData.C
```diff
@@ -0,0 +1,59 @@
+/**
+ * \file AddTaskEmcalEmbeddingHelperData.C
+ * \brief AddTask for AliAnalysisTaskEmcalEmbeddingHelperData
+ *
+ * The main implementation of the AddTask. 
+ * Same structure as AddTaskEmbeddingHelper
+ *
+ */
+AliAnalysisTaskEmcalEmbeddingHelperData *AddTaskEmcalEmbeddingHelperData()
```
raymondEhlers: I'd suggest that you make this a static method in the class and then just make the .C macro a wrapper. Or we could make the static method in the base class into a template class, and then we could use it for both classes

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelperData.h
```diff
@@ -0,0 +1,43 @@
...
+class AliAnalysisTaskEmcalEmbeddingHelperData : public AliAnalysisTaskEmcalEmbeddingHelper{
+ public:
+
+  AliAnalysisTaskEmcalEmbeddingHelperData()                          ;
+  AliAnalysisTaskEmcalEmbeddingHelperData(const char *name)          ;
+  virtual ~AliAnalysisTaskEmcalEmbeddingHelperData()                 ;
+
+  void            UserCreateOutputObjects() ;
+
+  void            AddPtSelection(Double_t min, Double_t max) { fPtSelectionMin.push_back(min); fPtSelectionMax.push_back(max);}
```
raymondEhlers: It might also be nice to override `RetrieveTaskPropertiesFromYAMLConfig()` and allow these to be set there. You'll need to make it virtual

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelperData.cxx
```diff
@@ -0,0 +1,97 @@
...
+
+}
+
+/**
+ * Override inherited function and add check on track with max pT.
+ * Rejects event if track within set pT range is not found
+ */
+
+Bool_t AliAnalysisTaskEmcalEmbeddingHelperData::CheckIsEmbeddedEventSelected()
+{
```
raymondEhlers: You may not need this at the moment, but you might also consider adding an AliEventCuts member so that we can perform event selection on the pp data. I don't know if this is super critical, but something to keep in mind. You can see this is implemented for the internal event in the embedding helper base class

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelperData.cxx
```diff
@@ -0,0 +1,175 @@
...
+
+
+void AliAnalysisTaskEmcalEmbeddingHelperData::RetrieveTaskPropertiesFromYAMLConfig() {
+
+  AliAnalysisTaskEmcalEmbeddingHelper::RetrieveTaskPropertiesFromYAMLConfig();
+
+  // get pt selection from YAML file
+  // defined in 2 arrays, pt
+
+  std::vector <double> ptSelectionMin;
```
raymondEhlers: That all looks good. To simplify, you might try loading directly into the object (I didn't test this here, but I'm done similar in other places):

```yaml
ptSelection: [[1, 2], [3,4]]
```

and then load it with

```c++
bool res = fYAMLConfig.GetProperty("ptSelection", fPtSelection);
```

But feel free to take it or leave it

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelperData.cxx
```diff
@@ -0,0 +1,199 @@
...
+
+
+/// \cond CLASSIMP
+ClassImp(AliAnalysisTaskEmcalEmbeddingHelperData);
+/// \endcond
+
+AliAnalysisTaskEmcalEmbeddingHelperData::AliAnalysisTaskEmcalEmbeddingHelperData():
+  AliAnalysisTaskEmcalEmbeddingHelper(),
+  fPtSelection(),
+  fTrackContainer(0x0)
```
raymondEhlers: Trivial, but better to use `nullptr` now that we have c++11

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelperData.cxx
```diff
@@ -0,0 +1,199 @@
...
+{
+}
+
+
+AliAnalysisTaskEmcalEmbeddingHelperData::AliAnalysisTaskEmcalEmbeddingHelperData(const char *name):
+  AliAnalysisTaskEmcalEmbeddingHelper(name),
+  fPtSelection(),
+  fTrackContainer(0x0)
```
raymondEhlers: Trivial, but better to use `nullptr` now that we have c++11

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelperData.cxx
```diff
@@ -0,0 +1,199 @@
...
+
+  AliAnalysisTaskEmcalEmbeddingHelperData * mgrTask = static_cast<AliAnalysisTaskEmcalEmbeddingHelperData *>(mgr->GetTask(name.Data()));
+  if (mgrTask) return mgrTask;
+
+  // Create the task that manages
+  AliAnalysisTaskEmcalEmbeddingHelperData * embeddingHelper = new AliAnalysisTaskEmcalEmbeddingHelperData(name.Data());
+
+  // set embedded track container
+  AliTrackContainer *contEmb = new AliTrackContainer("tracks");
+  contEmb->SetTrackCutsPeriod("LHC17p");
```
raymondEhlers: I wouldn't recommend leaving this as the default, but it's of course fine for now. In my opinion, I think it would be best to set it via YAML.

raymondEhlers:
+test

raymondEhlers:
+1

PR #10991, opened 2019-07-24
added misID and secondary check in closure test

panjinjin:
added misID and secondary check in closure test


panjinjin:
dear approvers, could you please +1 at your earliest convenience? The change is urgent for the paper proposal. Thanks a lot.

dcaffarr:
+1

PR #10994, opened 2019-07-25
AliFemto: modifications to NanoAOD reader

lgraczykCern:
changes to the NanoAOD reader - possibility to run if covariance matrix not set


lgraczykCern:
+1

PR #11011, opened 2019-07-26
Det and true MC can be run during the same run.

osaarima:
Later this will be used to create response matrices. The root5 compiling
error was fixed.


aohlson:
+test

aohlson:
+1

PR #11018, opened 2019-07-26
PWGGA/GammaConv: Changed trigger mimicking. The thresholds and width …

jokonig:
…are now stored in an OADB file and are loaded from there. Also implemented a Supermodule-wise trigger mimicking.

Except for the pp 13TeV triggered datasets, there were no changes made to the threshold and width of the turn-on curves


FriederikeBock:
+1

PR #11035, opened 2019-07-29
PWGGA: Some Omega EMCal / DCal Configs added

jluehder:


nschmidtALICE:
Hi Jens, please check if there are already some MC fine tunings for EMCal available and change your new configs accordingly.
If they are not available, then at least use the 01 nonlinearity.

jluehder:
sry there is a doubled config

nschmidtALICE:
+1

PR #11052, opened 2019-07-30
PWGPP-538, PWGPP-529 - updated pileup rejection algorithm

miranov25:
-  adding option to reset histogram
- make ITS filter more effective
- increase outlier cut
- adding LTM estimator + disable tgl cut


miranov25:
+1

PR #11054, opened 2019-07-30
add possibility to embed pp data events

jaimenorman:
- add embedding option for embedding pp data
- fix cut on fraction of shared pt, and QA histogram for shared pt


raymondEhlers:
+1

PR #11060, opened 2019-07-30
AddTask macro to set-up AliAnalysisTaskPtMothFromPtDaugh task

ffionda:
AddTask macro to set-up AliAnalysisTaskPtMothFromPtDaugh task (which runs on kinematics only)  for B->J/psi analysis.


iarsene:
+1

PR #11097, opened 2019-08-01
Fix selection of trigger clusters

mfasDa:
Trigger clusters were in case of data not propagated to the array trgclusters, therefore the relevant histogram was filled only with "ANY"


mfasDa:
+1

PR #11100, opened 2019-08-01
PWGHF: Fix TreeCreator Lc->pK0s

lvermunt:
TreeCreator was combining both offline and on-the-fly v0's for pp. Fixed, now only selecting offline ones as is done for Lc analyses.

(Didn't show up in validation TreeCreator with PbPb because there are no on-the-fly v0's. SQM '19 preliminaries not affected)


adubla:
+1

PR #11110, opened 2019-08-02
PWGGA/GammaConv: fixed missing delete in heavy meson task

fjonasALICE:
* fixed two missing pointer deletions
* Cluster histos that are needed for calculation of rejection factors will now be filled even if light output is activated


nschmidtALICE:
+1

PR #11112, opened 2019-08-02
Changes in a readerAOD class.

wrzesaCERN:
Delete an "AliFemtoTrack" object before the "continue" function in a loop if a cut rejects this track. In femtoscopic analysis (heavy ions collisions) such objects which are not deleted when we are going back to the beginning of the loop lead to a large memory leak.


aohlson:
+1

akubera:
One last point: I'd recommend putting a *little* more info into the git message. Instead of "Changes in a readerAOD class", I'd say something more like: "Fix memory leak in AliFemtoEventReaderAOD". Still brief, but describes the commit with more information than just "changes"

PR #11122, opened 2019-08-04
[PWGLF] Fix bug in switch

jikim1290:
Fix bug in pile up selection.
Analyze event even if the number of events for mixing is insufficient.


BongHwi:
+1

PR #11131, opened 2019-08-05
PWGHF: Small fixes TreeCreator

lvermunt:
* Fix compilation warnings: Missing variables in constructor
* Fix removal daughters PV calculation: Previously it was done after call to IsSelected(). Moved to right place (minor bug, 0.01% effect)
* Fix indentations to same convention


qgp:
+1

PR #11137, opened 2019-08-05
Added response matrices for jets and dijets.

osaarima:
Also arranged the dijet calculation to be more neat and easier to use
with response matrix calculations.


aohlson:
+1

PR #11149, opened 2019-08-06
Calculating E/p with non-linearity cluster energy

deepathoms:
E/p obtained with EMCal cluster energy calculated after non-linearity correction


deepathoms:
+1

PR #11156, opened 2019-08-06
Add MC OADBs for p-Pb 2013

ddobrigk:


ddobrigk:
This commit adds standard calibs for DPMJet MC anchored to 2013 p-Pb: should now work for any standard analysis. Supercalibs being investigated as an option too...

ddobrigk:
+1

PR #11157, opened 2019-08-06
Changes that allow to change magnetic field value 

annakawecka:
(0.5 T is default but in some runs, for example Xe-Xe run we have 0.2 T)


aohlson:
+1

PR #11173, opened 2019-08-07
AliFemto: adding possibility to read config file from grid

lgraczykCern:
adding possibility to read config file stored on the GRID


lgraczykCern:
+1

PR #11180, opened 2019-08-07
dca to be in abs in the custom dca of V0 daughters

pganoti:
1. dca to be in abs in the custom dca of V0 daughters
2. minor change in sigma star config macro for PbPb 2018


mkruegerGitHub:
+1

PR #11188, opened 2019-08-07
[PWGCF] Change the streamer of AliJJetTask

jikim1290:
Change the streamer of AliJJetTask for legotrain operation.
Stop copying efficiency file in the current directory


aohlson:
+1

PR #11192, opened 2019-08-08
Changes that allow to change the value of magnetic field (default it …

annakawecka:
…is 0.5, but in some cases, like Xe-Xe run, we have different value so DPhiStar is not calculated correctly)


lgraczykCern:
+1

PR #11198, opened 2019-08-08
Hypertriton Binding energy

alcaliva:
Analysis task for the measurement of Hypertriton Binding energy


mpuccio:
+1, even though I think we already have the code for this (and analysers looking at it...)

mpuccio:
+1

PR #11200, opened 2019-08-08
PWGGA/GammaConv: fixed mem leak in trigger mimicking

fjonasALICE:
AliAODBContainer was never deleted and therefore created for every event


FriederikeBock:
+1

PR #11206, opened 2019-08-08
PWGUD/UPC: Change ZNA energy threshold.

siragoni:
This should fix the differences between the 0NXN and the XN0N classes...
Maybe.


mbroz84:
+1

PR #11208, opened 2019-08-08
Fixing a problem with the code

hahassan7:
The code crashing when setting fFillHeavyHisto to kFALSE due to a mistake. Now it is fixed.


raymondEhlers:
+1

PR #11218, opened 2019-08-09
PWGGA/GammaConv: major change in heavy meson back calc

fjonasALICE:
in this PR the event mixing for heavy mesons was rewritten to increase speed. Furthermore, instead of calculating four individual background histograms depending on which pions are allowed to be from the same even, only a single event mixing background is calculated, where the user can specify in AliConversionMesonCuts a mode from 1 to 4, where 4 is no pions coming from the same event (default).

Furthermore a bug in the BGHandler was fixed, that caused only every second event mixing pool to be filled.


nschmidtALICE:
+1

PR #11221, opened 2019-08-09
PWGUD/UPC: New calibration for 2015 data sample.

siragoni:
Discovered bug in the calibration.
Possibly fixed now...?
IMPORTANT: tested.


mbroz84:
+1

PR #11224, opened 2019-08-09
PWGHF: Fix TTreeCreator

lvermunt:
Fix in RemovalDaughtersPV in PR #11131 was only needed for Lc's. For the D's it was correct, and the reposition actually introduced a small inconsistency. Reverted + a fix for Dstar (not removing soft pion as is done in AliRDHFCutsDStartoKpipi)


fprino:
+1

PR #11229, opened 2019-08-09
Fix filling of histos with MC truth and disable time-range cut for MC

fgrosa:
1) add protection to check whether histos filled with MC truth are defined for each particle specie
2) add protection to always skip time-range cut when running on MC


fprino:
+1

PR #11258, opened 2019-08-13
Changes in configs

gmorycCern:


dcaffarr:
Two out of three config files, seem to be heavily changed. Is this wanted @gmorycCern ?

vpacik:
Hi @dcaffarr, I am not a Femto insider, but looking at the file structure, it seems each analysis/user have its own folder with config files. Which is the case for the two new files. For the third/modified one, looking at the history of previous commits, it was only modified by @gmorycCern. 
So I think it is fine to approve such changes.

vpacik:
+1

PR #11262, opened 2019-08-13
Readmyedits

gmandaglio75:
Please update asap the macro AddTaskKStarPlusMinusMult.C, as I did in my branch.
The previous update creates a mismatch between the name of macro and the defined class,
this make the macro useless.


mkruegerGitHub:
+1

PR #11263, opened 2019-08-13
Add the possibility of kaon identification via kinks in the RSN package

pganoti:
-Add a class for kaon identification via kinks in the RSN package.
-Modified the CMakeLists.txt and PWGLFresonancesLinkDef.h (add the class)
-Add task and Config file for L star analysis with kaons ID via kinks


mkruegerGitHub:
+1

PR #11264, opened 2019-08-13
PWGGA/GammaConv/  for PCM-EDC pp13TeV

AMechler:
+add new NonLins using testbeam param
+updating cutstrings to use best smearing


dmuhlhei:
+1

PR #11268, opened 2019-08-13
PWGHF: Fix Lc->pKpi nSigma PID selection

lvermunt:
Bug found when validating HFTreeCreator for the nSigma and nSigmaMin PID options (the default case). Fix reduces the number of selected candidates with ~30%, using filtering cuts + 3sigma TPC+TOF.

NB: For our published Lc physics papers, more advanced PID selection strategies were used that are not affected by this bug.


afestant:
+1

PR #11281, opened 2019-08-14
added new cuts for the systematic studies of track selection and eID

crisjahnke:
- Config file modified to have several configuration cuts.

- AddTask modified to support several outputs, given a configuration cut.


iarsene:
+1

PR #11284, opened 2019-08-14
PWGGA/GammaConvBase/

AMechler:
fixing bug, EDC NonLin was applied to PHOS clusters
bug was introduced just yesturday


nschmidtALICE:
+1

PR #11298, opened 2019-08-15
Add Qn Corrections AddTask with configurable outputs

raymondEhlers:
The AddTask doesn't allow for configuration of the calibration output,
which is large enough to cause merging issues. Unfortunately, to make
this configurable, I had to copy almost all of the code from the existing
AddTask (located at PWGPP/EVCHAR/FlowVectorCorrections/QnCorrectionsInterface/macros/AddTaskFlowQnVectorCorr ectionsNewDetConfig.C).
The benefits are that the AddTask is more configurable, including the
outputs (I switched to using YAML for the config because it's familiar),
and the code is compiled instead of interpreted.


raymondEhlers:
+1

PR #11300, opened 2019-08-15
AliAnalysisTaskHFJetIPQA: fix streamer error caused by fHLundIterative

kagarner:


PWGJE/FlavourJetTasks/AliAnalysisTaskHFJetIPQA.h
```diff
@@ -3,6 +3,7 @@
 #include "AliAnalysisTaskEmcalJet.h"
 #include "TGraph.h"
 #include "THistManager.h"
+#include "THnSparse.h"
```
mfasDa: This is not necessary, you still use a pointer type, it could be handled with a forward declaration. Better stay with forward declarations in the header file unless including is absolutely needed (i.e. because you used member functions in the header or you use objects as value types).

PWGJE/FlavourJetTasks/AliAnalysisTaskHFJetIPQA.h
```diff
@@ -223,7 +223,7 @@ class AliAnalysisTaskHFJetIPQA: public AliAnalysisTaskEmcalJet
     TH2D *fh2dCharmDeta; //!
     TH2D *fh2dBottomDeta; //!
 
-    THnSparse *fHLundIterative;//       iterative declustering
+    THnSparseF *fHLundIterative;//!       iterative declustering
```
mfasDa: Also here the specification of the derived type in member declaration is not necessary, the problem was more the missing "!".

PWGJE/FlavourJetTasks/AliAnalysisTaskHFJetIPQA.h
```diff
@@ -399,7 +399,7 @@ class AliAnalysisTaskHFJetIPQA: public AliAnalysisTaskEmcalJet
 
 
 
-    ClassDef(AliAnalysisTaskHFJetIPQA, 36)
+    ClassDef(AliAnalysisTaskHFJetIPQA, 37)
```
mfasDa: ClassDef change would only be needed if in a previous commit data members were added and the ClassDef was not increased. Just change of the pointer type from derived to base class shouldn't require a ClassDef change.

mfasDa:
+test

mfasDa:
+1

PR #11304, opened 2019-08-15
OADB/COMMON/MULTIPLICITY: bug-fix in macro OADB-creation for pPb

FriederikeBock:
- bug fix for the OADB creation macro in pPb to correctly read the parameters of the Z-vertex correction from the root-input file
- inverting axis of the debug output per run for the MC calibration in order to ease comparisons for the supercalib


ddobrigk:
+1

PR #11311, opened 2019-08-15
Changing the name of the OADB file in case of running on Merged BCs

hahassan7:
The OADB file that has merged BCs has a different name from the one with 4 BCs. The correction frame was changed such that in case of running on merged BCs the OADB file "EMCALTimeCalibMergedBCs.root" will be read.


FriederikeBock:
+1

PR #11314, opened 2019-08-15
PWGGA: Change PHOS CutStrings for 13TeV

jluehder:
NCells changed to 2 in multiple configs
Different timing cut
Get rid of unused triggers


dmuhlhei:
+1

PR #11321, opened 2019-08-16
[EMCAL-589] Determine FastOR multiplicity above threhold

mfasDa:
Determining the FastOR mulitplicity as function of the
event centrality for different thresholds at ADC level.
Event selection is always min. bias to get unbiased
distributions, might accept also EMCAL triggers in the
future.


mfasDa:
+test

mfasDa:
+1

PR #11324, opened 2019-08-16
Bug fixes to jT analysis

TWSman:
Correct crash when using other than V0 for centrality
Remove unnecessary warnings from vector indices


vpacik:
+1

PR #11327, opened 2019-08-16
PWGGA: Add 13TeV PCMPHOS and PHOS cfgs for different checks

jluehder:
I added some Train Cfgs to check Smearing, Mult., and Sphericity
I also changed a cutstring which i forgot to change for MBW


dmuhlhei:
+1

PR #11337, opened 2019-08-17
PWGGA/GammaConv: fixed rising memory in trigger mimicing.

jokonig:
If the trigger specified in the cutstring does not match any trigger specified in the trigger mimicking function, return true. This avoids reloading the OADB root file for every event where the trigger values are stored


nschmidtALICE:
+1

PR #11356, opened 2019-08-19
PWGGA/GammaConv: introduced pT dependend mass cut for HNM task

fjonasALICE:
- introduced pT dependent mass cut for HNM task, so far only for EMC only
- increased pT range of inv mas histos


nschmidtALICE:
+1

PR #11373, opened 2019-08-20
PWGHF: TTreeCreator, update ITS3 particles

lvermunt:
* Fix memory leak Lb and Bs
* Improve Lb and Bs
* Small general fixes


afestant:
+1

PR #11374, opened 2019-08-20
remove duplicate functions

pillot:
Hi,

I imagine this is a leftover from a previous commit. I remove it as it makes the compilation fail on macos.

Cheers,
Philippe


dcaffarr:
@lgraczykCern do u confirm is it ok?

lgraczykCern:
+1

lgraczykCern:
Yes, it's correct, those lines of course should be removed

PR #11388, opened 2019-08-21
PWGUD/UPC: Fix construction list. Add v2 ZDC analysis.

siragoni:
Fixed construction list a bit.

Added v2 of the neutron emission analysis. This time I am using the inherent ZDC activity through the isZNAfired() method and the corresponding ZNC method, instead of the energy threshold...

IMPORTANT: tested.


mbroz84:
+1

PR #11391, opened 2019-08-21
[Doxygen] Fix documenation of the iterative declustering algorithm (d…

mfasDa:
…eclustering done on harder and not softer subjet)


mfasDa:
+1

PR #11393, opened 2019-08-21
PWGGA Add Bad Channels of LHC18 by runwise macro to the online OADB

jluehder:
Add local BC maps of LHC18 to the PWGGA OADB files


nschmidtALICE:
+1

PR #11396, opened 2019-08-21
PWGHF: Improvements TTreeCreator

lvermunt:
* Changed defaults AddTask to less heavy options (number of arguments still to be reduced)
* Improved Hijing background function for Lb and Bs, switching to AliVertexingHFUtils functions that properly account for negative label tracks


afestant:
+1

PR #11403, opened 2019-08-22
PWGHF: Fixes for Lb and Bs for ITS3 productions

lvermunt:
* Fix in TTreeCreator to run the task twice in same train (needed for TTree and TTree+ITS3 improver comparison)
* Temporary fixes for the no-momentum conservation in the upgrade productions for beauty particles (per mille difference between sum daughters and mother, but big enough to get rejected)
* Temporary fix for the "unphysical" Bs and Lb decays in terms of charge of Ds/Lc and pi


arossi81:
+!

arossi81:
+1

PR #11412, opened 2019-08-23
AliAnalysisTaskHFJetIPQA: Add IP based track counting tagger

kagarner:
- read in histograms with jetpt dependent threshold variables
- perform tagging on basis of single track, two or three tracks; add switch to exclude tagging via a single or two tracks
- fill histograms with number of correctly and falsely tagged jets for each method

-added switch to turn on and off Lund Plane investigations
-changed printf error outputs to AliError


mfasDa:
+1

PR #11426, opened 2019-08-23
Incorporate MC spherocity into alimultselection (experimental)

ddobrigk:


ddobrigk:
(N.B. this will have no effect on the regular operation of the framework and is merely meant for MC-only exercises for the time being)

ddobrigk:
+1

PR #11428, opened 2019-08-23
PWGHF: Fix for Lb&Bs in ITS upgrade prods part2

lvermunt:
The not-conserving of momentum in ITS upgrade productions for B-decays is also rejecting signal candidates due to the mom. conservation in the general MatchToMC function (see also #11403). Added the function in the dedicated MatchToMCB3Prong, loosening the momentum check from 0.001% to 0.5%.


fprino:
+1

PR #11446, opened 2019-08-26
Xic->pKpi task updated

mfaggin:
1. info about pKpi or piKp case for generated Lc (Xic) added
 2. tree filling only with generated particles implemented (when running on MC)
 3. bug fix in AliRDHFCutsXictopKpi in the mass cut


fgrosa:
+test

fgrosa:
+1

PR #11450, opened 2019-08-27
PWGUD/UPC: Last ZDC check.

siragoni:
Last ZDC check.


siragoni:
NB: based on what I have seen until now, this way we should be able to obtain the neutron emission classes' plots even with the new LHC15o_AODs!!

mbroz84:
+1

PR #11452, opened 2019-08-27
Removal of the upper-bound of the monte-carlo rapidity cut.

smaff92:
Just a super quick edit where I removed the upper bound on the MC rapidity distribution (file size got too large).


mkruegerGitHub:
+1

PR #11459, opened 2019-08-27
PWGUD/UPC: TildePhi fix.

siragoni:
Fixed TildePhi. Now it wraps continuously.
This should help in keeping it flat without the previous step structure...

IMPORTANT: tested.


mbroz84:
+1

PR #11463, opened 2019-08-28
Multiplicity class compatible only to MultSelection

kimbyumkyu:
LHC13 b c d e now calls MultSelection to get multiplicity information. The old method GetCentrality was removed.


aohlson:
+1

PR #11464, opened 2019-08-28
AddTask-s amended

klapidus:
apparently the absence of a newline at the end of an AddTask macro yields an interpreter mistake


mfasDa:
+1

PR #11474, opened 2019-08-28
Comment of TH3F and splitting trains

eddoming:
Systematics trains for MC were too heavy, so I believe that splitting them and remove one TH3F histogram will do the job.
Regards
Ed


nschmidtALICE:
@eddoming , i am fairly certain that this TH3 was not the root problem of the memory consumption. In your task are several TH2 with 4000 or 1000 bins which is far too much! Reducing the bin count to something reasonable here will save large amounts of memory.
Also slicing the pT axis of your histograms into 0.025 GeV bins is not necessary. As in your final analysis you'll likely only slice in 0.1 GeV bins, you can decrease the bin count by a factor 4 here already for all pT-dependent histograms.

I recommend you take a closer look at all histograms and figure out the largest possible binning for them without loosing information.

Also a light output mode for your task would be good. This mode would then only create and fill the histograms needed to run the afterburners and not contain QA plots.

eddoming:
@nschmidtALICE, Thanks for the feedback, I check some of the histograms and change the binning for a bigger one, the part of running a lighter version of the code will take more time to be implemented.
Regards
Ed

nschmidtALICE:
+1

PR #11496, opened 2019-08-30
[PWGCF] change of the location of event mixing

jikim1290:
change of the location of event mixing
switch for systematic uncertainty study


majanik:
+1

PR #11508, opened 2019-08-30
AliFemto: removing calling alien-token-init

lgraczykCern:
alien-token-init cannot be called since it causes problems with the services


lgraczykCern:
+1

PR #11513, opened 2019-08-31
Commit to add track cuts for V0 candidates

smaff92:
Added track qualitiy cuts to get a more consistent MB spectra for Lambas and K0s


bdoenigu:
+1

PR #11528, opened 2019-09-02
Info about some PID standard cuts added to THnSparses

mfaggin:
CutObject PID strategy stored in bin centered around 0


fgrosa:
+test

fgrosa:
+1

PR #11533, opened 2019-09-02
Fix the path of the efficiency in my grid 

MustafaNajiAnaam:
the path of Efficiency was wrong and i have been fixed


aohlson:
+1

PR #11541, opened 2019-09-02
Early execution break when recycling an external event

raymondEhlers:
This takes advantage of a new feature in `AliAnalysisManager`. This way,
the users don't need to enable embedded event recycling. However, this
does slightly change procedure because the later user task internal event
counting won't be accurate - the user task won't be executed.


raymondEhlers:
+test

raymondEhlers:
@lcunquei @lhavener This new method would allow us to skip setting `AliAnalysisTaskEmcal::SetRecycleUnusedEmbeddedEventsMode(kTRUE)` for each of the user tasks. However, it's a WIP because I haven't yet tested it fully

raymondEhlers:
+test I'll squash after everything is confirmed to be working

raymondEhlers:
+test

raymondEhlers:
+1 Thanks to @hbossi and @lhavener for testing

PR #11544, opened 2019-09-03
Reduced TH3D bins to save memory usage / Added ptB-based weight

cmengke:
Reduced TH3D bins to save memory usage
Added ptB-based weight


afestant:
+1

afestant:
+1

PR #11545, opened 2019-09-03
PWGHF: Add filling cand in ITS3Improver

lvermunt:
ITS3Improver crashes when trying to run over standard PbPb MC's (for example GP PbPb '18). Candidates needs to be filled first.


afestant:
+1

PR #11554, opened 2019-09-03
Add flags to apply physics selection and event downsampling

fgrosa:
1) add flag to apply physics selection at the task level to avoid the storage of unwanted events (false by default, so nothing changed unless enabled explicitly)

2) add flag to enable random event downsampling (false by default, so nothing changed unless enabled explicitly). Needed in case of a necessity of a homogeneous fraction of the data sample


afestant:
+1

PR #11557, opened 2019-09-04
Implement jet matching in user task

raymondEhlers:
Moving the matching to the user task is a work around for issues with LHC15o embedding when using full jets


raymondEhlers:
+1

PR #11569, opened 2019-09-05
Forgot to increment classdef

ddobrigk:
(This isn't relevant for the vast majority of people, but sometimes, under specific circumstances, people use stored copies of AliMultSelectionTask - in which case this of course matters!)


ddobrigk:
+1

PR #11576, opened 2019-09-05
Fixes to jet matching

raymondEhlers:
The indexing wasn't quite write in a number of places. Now it agrees
with the jet tagger


raymondEhlers:
+1

PR #11581, opened 2019-09-05
Changing the EMCal clusterizer from v1 to v2

deepathoms:
Changing the EMCal clusterizer from v1 to v2 for pp data


deepathoms:
+1

PR #11582, opened 2019-09-05
Changing the EMCal clusterizer from v1 to v2-pPb

deepathoms:
- Adding yaml config file for MC for pPb data
- Changing the EMCal clusterizer from v1 to v2 for pPb data


deepathoms:
+1

PR #11593, opened 2019-09-06
Include IDs in the trees

hzanoli:
Dear all,

I have included the ID of the electrons and D meson particles in the trees to perform some checks. I have also moved some code to make it more clear, separating the electron and the D meson part.   

Best,
Henrique


btrzecia:
+1

PR #11600, opened 2019-09-06
Fix out-of-bounds write, add validation warning

raymondEhlers:
If NcentBins was set after constructing the object to anything greater
than 4, it would write the histogram out of bounds.

Also added a warning that the task isn't yet validated and that the
standard tagger should be used.


raymondEhlers:
+test

raymondEhlers:
@mfasDa I noticed that the fast tagger writes out of bounds when setting n cent bins > 4. It's not super urgent since the task isn't validated, but I figured I should fix it now so that it's not an issue later (there are certainly better fixes than this, but it's trivial and should get the job done). I also added a warning that the task isn't validated so there's no confusion. Let me know if you're okay with merging

mfasDa:
+1 @raymondEhlers Thanks for spotting!

PR #11603, opened 2019-09-07
Adding setter to force correcting for nonlinearity

deepathoms:
Adding setter to force correcting for nonlinearity of EMCal clusters


deepathoms:
+1

PR #11606, opened 2019-09-08
Expand delta eta range

raymondEhlers:
Allows for the possibility of expanding the background dominated region.
Note that to take advantage and fully fill out to +/- 1.4, the user must
extend the eta acceptance out to +/-0.9. Otherwise, it can't populate
the correlations out to +/-1.4.


raymondEhlers:
+1

PR #11609, opened 2019-09-09
Modify the ESD Lambda and Photon cuts consistently with the NanoAOD analysis

a-mathis:
…nalysis


a-mathis:
@makor I set the Lambda cuts now by default to the new values

a-mathis:
+1

PR #11612, opened 2019-09-09
PWGUD/UPC: Added templates for yield extraction both Helicity and CS, 1D and 2D

siragoni:
Added templates for yield extraction both Helicity and CS, both 1D and 2D.

IMPORTANT: tested.


mbroz84:
+1

PR #11695, opened 2019-09-13
PWGUD/UPC: Pt templates in the exact invariant mass distribution. Rapidity wise too.

siragoni:
Changed from  2.8 < M < 3.3  to   2.85 < M < 3.35.
Added rapidity wise templates.

IMPORTANT: tested


siragoni:
IMPORTANT: this is needed to compute the $f_I$ coefficients in the neutron emission analysis.

mbroz84:
+1

PR #11706, opened 2019-09-16
Changed histo binning for more general purposes.

osaarima:
2nd try. Original commit was overwritten by mistake before pull request was accepted.


dongjokim:
+1

PR #11720, opened 2019-09-17
PWGUD/UPC: Fix invariant mass range in templates pol.

siragoni:
The code for the template was using an invariant mass selection a priori...
I had not realised this when running on the CohJPsi MC because they are already in that range.
Considering that I am dealing with rapidity integrated invariant mass plots, I can use them for the fits, but this has to be amended...

IMPORTANT: tested.


mbroz84:
+1

PR #11722, opened 2019-09-17
PWGGA/GammaConv: HNM implementation of new mass cuts for 5 TeV

fjonasALICE:
- added new functions describing pi0 peak position and width for pp@5TeV with correct nonlin
- adapted cuts to use correct NonLin
- implemented dEdx recalib functionality already present in other add tasks


nschmidtALICE:
+1

PR #11733, opened 2019-09-18
PWGHF: Temporary fixes for ITS3 Lb analysis

lvermunt:
* Added temporary "fix-flags" (related to selected background) to compare old ITS2 analysis with  new ITS3 one. Fixes to be made final after checks+discussions.
* Fixed some memory leaks.
* Only require PID for pion track when PID is also required for the Lc candidate.
* Added possibility to skip the rotational background part, so task is less heavy for running tests.


afestant:
+1

PR #11736, opened 2019-09-18
Files to create a tree containing dimuon information.

rarnaldi:
Code allows the creation of trees containing information on the event and on dimuons. Dimuon information are stored only for masses larger than a given mass threshold.
The code refers to pp collisions at 13TeV


tdahms:
+1

PR #11741, opened 2019-09-19
PWGHF: Loosen mom. consv. check for beauty ITS upgrade studies

lvermunt:
* Several correct beauty candidates in the ITS Upgrade MC's were still rejected by the momentum conservation check. Loosened the check further, and added a strategy to tag and use them add task level (with negative labels).
* Store the still rejected decays in TTreeCreator for offline study (tagging them as "feed-down", as this flag wasn't used for beauty decays)
* Removed resonance check in MatchToMCB3Prong, didn't work as expected


afestant:
+1

PR #11742, opened 2019-09-19
OADB/COMMON/MULTIPLICITY: Fixing OADB objects for 16q, 16r, 13f to include defaults

FriederikeBock:
The OADB object for LHC16q, 16r & 13f didn't include all runs after the last update, this pull request adds all runs back again and add a default for safety measures. Also the corresponding MC objects have been fixed accordingly.


ddobrigk:
+1

PR #11746, opened 2019-09-19
Major Overhaul of TPCTOFCascade analysis task

smaff92:
1. Change of cuts for both V0 and Cascade candidates
2. Added AliMultSelection for RefMult and SPDTracklets
3. Changed MC functionality.
4. Made changes for Cascades to check whether the associated V0s are primary or not.


ddobrigk:
+1

PR #11755, opened 2019-09-20
PWGUD/UPC: Pol 1D and 2D with 0N0N class.

siragoni:
I am doing now the polarisation analysis both in 1D and 2D, both with HE and CS frames, with the 0N0N neutron emission class. This suppresses completely the incoherent component and the feed-down contribution, e.g. the pt-distributions!! This means that once the values are extracted, this different analysis would be able to switch immediately to the systematics...

IMPORTANT: tested.


mbroz84:
+1

PR #11758, opened 2019-09-20
PWGHF: Fix storing of reflections in TTree

lvermunt:
* Reflections were not stored in TTreeCreator (neither when storing signal only, nor signal+background). Needed for D0 invariant mass fit


afestant:
+1

PR #11760, opened 2019-09-20
Small fixes + addition of CMake and LinkDef files

rarnaldi:
I've corrected few compilation warnings and I've added the CMake and LinkDef files for the PWGDQ/dimuons directory


miweberSMI:
+1

PR #11761, opened 2019-09-20
PWGHF, Lb: Memory leak fix + cleaning code

lvermunt:
* Fixed the biggest and most obvious memory leak in the Lb rotation procedure by merging the rotate function (that returned an array full off "hidden" pointers) and the histogram filling function (that used the array).
* Cleaned the code

Tested on 1 AOD locally: exact same result as before fix+cleaning.


ginnocen:
+1

PR #11805, opened 2019-09-25
Update weight factor, test reweighting efficiency

chengtt0406:
After normalization, then update the new weighting parameters to get the new weight factor. At the same time, also using the new weight factor to reweight the reconstruction effciency.


fprino:
+test

fprino:
+1

PR #11812, opened 2019-09-26
PWGGA small fix: AliFatal -> AliWarning

hmurakam:
AliFatal cause error while test. Revert to AliWarning which works before.


nschmidtALICE:
You know that if it ran into the AliFatal during a test, that your recalibration does not work correctly?
With your "fix" you will now simply allow undefined behaviour. Meaning it could also run the code without application of the recalibration despite you having requested it. I don't think we can approve this. @FriederikeBock do you agree?

PR #11817, opened 2019-09-26
Converting the Energy Recalib histogram to 1D histogram

hahassan7:
We converted the 2D histogram of the energy recalibration into a 1D histogram. Also the correction framework has been modified so it can read 1D histogram. By default this option is disabled and it be enabled in the yaml file by setting "load1DRecalibFactors" to "true".


mfasDa:
+1

shahor02:
+1

PR #11819, opened 2019-09-26
Important fixes for Xic->pKpi task

mfaggin:
1) AddTask modified - container name can be set by the user
 2) fix in AliRDHFCutsXictopKpi::IsSelectedPID about the kaon PID
 3) in AliAnalysisTaskSEXicTopKpi::UserExec() the fCutsXic->IsEventSelected has been added
    ---> mandatory to correctly set the Bayes PID
    ---> we'll remove the AliRDHFCutsDtoKpi object soon (still temporary solution)


arossi81:
+1

PR #11829, opened 2019-09-27
Add the option of decoupling SigmaC analysis from XiC, looping on fil…

arossi81:
…tered Lc candidates, w/o building candidates on the flight. Add sparse for single-track efficiency (still rough). Start to clean-up (still a lot to clean)


arossi81:
+1

PR #11850, opened 2019-10-01
added FONLL weight for B decay electron DCA

sashingo:


btrzecia:
+1

sashingo:
Dear expert,

the compile was failed, and looks the reason is due to KFParticle.
Could you take a look ?

regards,
Shingo

PR #11858, opened 2019-10-01
minor changes to modify the directory where the tree is written and i…

rarnaldi:
…ncrease of the size for few vectors


miweberSMI:
+1

PR #11882, opened 2019-10-03
Read and access L1 phases for PAR runs in EMCal

amatyja:
L1 phase in EMCal time calibration macro modified to store object with PAR Global IDs.
AliCalorimeterUtils modified to read OADB object with PAR Global IDs for EMCal.
AliEMCALRecoUtils modified to access L1 phases for PAR runs.
Default is still like for the run without PAR.


gconesab:
+test

gconesab:
+1

mfasDa:
PLEASE avoid c-style dynamic arrays, in particular in class headers, and use TArray instead!

PR #11884, opened 2019-10-03
Halve the deltaM bin width in SigmaC sparse. Default DeltaM cut reduc…

arossi81:
…ed to avoid too large increase of sparse size.


arossi81:
+1

PR #11889, opened 2019-10-03
Fix super slow compilation

mpuccio:


mpuccio:
@pzhristov we were talking about this... 
@mfasDa not sure if it is the best fix, but at least with this the compile time is order of seconds

raymondEhlers:
+1

PR #11896, opened 2019-10-03
Apply L1 phase calibration for PAR runs in correction framework

amatyja:
Apply L1 phase calibration for PAR runs in correction framework in AliEmcalCorrectionComponent.cxx.

### DIFF
--- a/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionComponent.cxx
+++ b/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionComponent.cxx
@@ -210,18 +210,18 @@ void AliEmcalCorrectionComponent::UpdateCells()
   
   if (fRecoUtils){
     //In case of PAR run check global event ID
-    Short_t currentParIndex = 0;
     if(fRecoUtils->IsParRun()){
+      Short_t currentParIndex = 0;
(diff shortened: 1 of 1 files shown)


PWG/EMCAL/EMCALtasks/AliEmcalCorrectionComponent.cxx
```diff
@@ -208,9 +208,22 @@ void AliEmcalCorrectionComponent::UpdateCells()
...
+    Short_t currentParIndex = 0;
+    if(fRecoUtils->IsParRun()){
+      ULong64_t globalEventID = (ULong64_t)bunchCrossNo + (ULong64_t)fEventManager.InputEvent()->GetOrbitNumber() * (ULong64_t)3564 + (ULong64_t)fEventManager.InputEvent()->GetPeriodNumber() * (ULong64_t)59793994260;
+      for(Short_t ipar=0;ipar<fRecoUtils->GetNPars();ipar++){
+	if(globalEventID >= fRecoUtils->GetGlobalIDPar(ipar)) {
+	  currentParIndex++;
+	}
+      }
+    }
+    fRecoUtils->SetCurrentParNumber(currentParIndex);
```
loizides: This was not here before so even if par is off why do we set it here?

amatyja: currentParIndex is set to 0. And 0 means that the regular calibration before any PAR is taken.

amatyja: In priniple it is safe to go with this setting into the condition statement.

loizides: I would prefer the latter...

loizides:
+1

PR #11898, opened 2019-10-04
Updating task for BBar on Nano with inclusion of Sphericity/Spherocit…

sartozza:
…y cuts

### DIFF
--- a/PWGCF/FEMTOSCOPY/macros/AddTaskFemtoNanoGrandma.C
+++ b/PWGCF/FEMTOSCOPY/macros/AddTaskFemtoNanoGrandma.C
@@ -7,17 +7,25 @@
 #include "AliFemtoDreamCascadeCuts.h"
 #include "AliFemtoDreamCollConfig.h"
 
-AliAnalysisTaskSE *AddTaskFemtoNanoGrandma(bool fullBlastQA = false, int phiSpinning =
-                                         0,
-                                     int nSpins = 1, double corrRange = 0.1,
-                                     bool Systematic = false,
+AliAnalysisTaskSE *AddTaskFemtoNanoGrandma(bool fullBlastQA = false,//1
+									 bool isMC = false,				//2
+									 int phiSpinning =0,			//3
+                                     int nSpins = 1,				//4
+									 double corrRange = 0.1,		//5
+									 TString triggerData = "kInt7",	//6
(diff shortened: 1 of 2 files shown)


PWGCF/FEMTOSCOPY/macros/AddTaskFemtoNanoGrandma.C
```diff
@@ -1165,7 +1204,7 @@ AliAnalysisTaskSE *AddTaskFemtoNanoGrandma(bool fullBlastQA = false, int phiSpin
   task->SetCorrelationConfig(config);
   mgr->AddTask(task);
 
-  TString addon = "PL";
+  TString addon = "HM";
```
a-mathis: this is not going to work. Here make a unique specifier for your analysis ("BBar" or so), otherwise we won't be able to run multiple analyses at the same time

PWGCF/FEMTOSCOPY/macros/AddTaskFemtoNanoGrandma.C
```diff
@@ -7,13 +7,16 @@
...
-                                         0,
-                                     int nSpins = 1, double corrRange = 0.1,
+AliAnalysisTaskSE *AddTaskFemtoNanoGrandma(bool fullBlastQA = false,
+									 int phiSpinning =0,
+                                     int nSpins = 1,
+									 double corrRange = 0.1,
                                      bool Systematic = false,
-                                     const char *cutVariation = "0") {
+                                     const char *cutVariation = "0",
+									 const char *sTcut = "8") {
```
a-mathis: I am not so sure about this - I would rather make this the second-to-last argument (since there are only 8 variations) and use the wagon configuration to take care of all the systematic variations.
Also, below you compare "8" == "08" which I don't think is going to work

PWGCF/FEMTOSCOPY/macros/AddTaskFemtoNanoGrandma.C
a-mathis: ... and maybe for the same reason change the name of the task
  AliAnalysisTaskNanoLoton* task = new AliAnalysisTaskNanoLoton("**femtoLoton**");

a-mathis:
+1

PR #11899, opened 2019-10-04
Allow the Skim tasl to select events with a track within a given pt bin

fkrizek:


PWG/EMCAL/EMCALtasks/AliAodSkimTask.cxx
```diff
@@ -157,7 +157,7 @@ void AliAodSkimTask::UserExec(Option_t *)
 
   // Accept event only if an track with a minimum pT of fTrackMinPt has been found in the event
   Bool_t storePt = kFALSE;
-  if (fTrackMinPt > 0 ){
+  if (fTrackMinPt > 0 && fTrackMaxPt < 0){
```
loizides: Not sure about this upper check, should we not check the "or" if any of the two are turned on (-1 means not used)

fkrizek: In principle the default situation is fTrackMaxPt=-1    so  if fTrackMaxPt is not set by user  and you give just fTrackMinPt > 0   the macro should behave as before,  you select all events with pT> fTrackMinPt .

PWG/EMCAL/EMCALtasks/AliAodSkimTask.cxx
```diff
@@ -171,6 +171,23 @@ void AliAodSkimTask::UserExec(Option_t *)
     storePt = kTRUE;
   }
 
+  // Accept event only if an track within a pT interval (fTrackMinPt, fTrackMaxPt) has been found in the event
```
loizides: I would have preferred a solution where we have one loop that decides all cases depending on the value of fTrackMinPt and fTrackMaxPt (ie only check the dependence if they are different from -1)

fkrizek: OK i will change the code

PR #11903, opened 2019-10-04
Calibrating the low gain cells

hahassan7:
The correction framework has been modified to it calibrates the low gain cells. As a first trial we will try the calibrate the low gain cells by reading the LG histogram for each period.


raymondEhlers:
+1

PR #11930, opened 2019-10-07
New histos and dijet matching improved.

osaarima:
Added delta-pt and other histograms. Dijet invariant mass matching
improved so that a dijet is a match even if leading and subleading jets
change places.


aohlson:
+1

PR #11944, opened 2019-10-08
Added min jet pt as an argument + fix

osaarima:
Added min jet pt as an argument. In addition changed the dijet matching
which was actually planned already for the previous commit.


aohlson:
+1

PR #11949, opened 2019-10-08
PWGGA/GammaConvBase: Added function, only track selection

hmurakam:


nschmidtALICE:
@hmurakam, i do not see your new function being used in any task. Will this be a separate commit?

hmurakam:
yes, it will be.

nschmidtALICE:
+1

PR #11965, opened 2019-10-09
PWGGA: Edit Cutstrings for new Timing Eff and new Nonlin

jluehder:
1. Edit Cutstrings for new Timing Eff and new Nonlin
2. Adding the new NonLins to AliCaloPhotonCuts


nschmidtALICE:
+1

PR #11967, opened 2019-10-09
PWGUD/UPC: Added new analysis task for pPb with 2016 data.

siragoni:
Added new analysis task for pPb with 2016 data.
Periods LHC16r and LHC16s.

IMPORTANT: tested only for compilation and minor bugs.


mbroz84:
+1

PR #11991, opened 2019-10-11
PWGGA: Added option For Timing Cut=> Open for cases >5GeV

jluehder:
PWGGA: Added option For Timing Cut=> Open for cases >5GeV
Cut Variation, to open up the Cuts and increase statistics when timing efficiency drops


dmuhlhei:
+1

PR #11993, opened 2019-10-11
Tuning of L1 phase extraction for PAR runs in EMCal

amatyja:
Main changes: 
- fCurrentPARIndex=0 means events before any PAR. Changed histogram filling according to this scheme. 
- when low number of entries after PAR for some SM -> this SM is not treated as empty if before PAR it was filled
- added clearer printouts in case of problems


gconesab:
+test

amatyja:
+1

PR #11994, opened 2019-10-11
Setting up the pion analysis.

makor:
Not yet ready to be used on the grid for the intended purpose. Needs additional testing (most certainly contains some mistake) + tuning of some cuts.


a-mathis:
+1

PR #11997, opened 2019-10-12
[PWGCF] Change the trigger of AliJJetTask

jikim1290:
Change the trigger of AliJJetTask when kHighMultV0 is used for analysis.


dongjokim:
+1

PR #12013, opened 2019-10-14
Better way to calculate the varibles on Constrained Param

eddoming:
With this new commit only a few block are lost for part of AliV0reader, and Dalitz reaming constant.
Regards
Edgar


nschmidtALICE:
+1

PR #12021, opened 2019-10-15
AliFemto: reducing number of bins in monitors

lgraczykCern:
reducing bin numbers in monitors of Minv and CPA vs pT for V0 and cascades


lgraczykCern:
+1

PR #12022, opened 2019-10-15
AliFemto: adding Config for Xi MC Closure Test

lgraczykCern:
adding config to run Xi analysis on MC reconstructed AODs


lgraczykCern:
+1

PR #12029, opened 2019-10-15
PWGDQ/reducedTree: updates for correlation analysis

laltenka:
- AliReducedVarManager : added trigger pair decay length and SPD pair type (also had to update AliReducedAnalysisJpsi2ee accordingly)
- AliReducedVarCut : added 2nd dependent variable to function cuts
- AliReducedVarManager: added flag for EMCal triggered event


iarsene:
+1

PR #12049, opened 2019-10-16
return statment on GetParamG

eddoming:
As the name said, return for GetParamG, thanks to Florian for the help and Nico as always.
Regards
Edgar


nschmidtALICE:
+1

PR #12062, opened 2019-10-17
EMCal low gain cells calibration

hahassan7:
I modified the correction framework so it calibrates the LG cells. The correction framework reads from OADB a file "EMCALTimeCalibMergedBCLG.root" which contains the histograms containing the LG time calibration coefficients for all the periods merged in one histogram.


raymondEhlers:
+1

PR #12063, opened 2019-10-17
PWGDQ/reducedTree/AliCorrelationExtraction: bugfix in bin width norm

laltenka:
normalization was sensitive to equal/not equal bin sizes, it's not anymore


iarsene:
+1

PR #12064, opened 2019-10-17
Lc (pKpi and pK0s) added to SysPropagation

chills-ALICE:
Changes to TrackingSysPropagation.cxx, .h and the corresponding AddTask macro to include Lc as an option.


cterrevo:
+test

cterrevo:
+1

PR #12071, opened 2019-10-18
PWGUD/UPC: Add subwagon feature.

siragoni:
IMPORTANT: tested.

IMPORTANT: it works, but I am not too sure if this is enough to access the subwagon feature...


mbroz84:
+1

PR #12079, opened 2019-10-18
PWGGA/GammaConv: implementation of gamma from pi0 energy cut

fjonasALICE:
Implemented new cut that allows to select only neutral meson candidates where at least one (or two) gammas from neutral meson fulfill energy threshold. 
Can be set via minPt cut in AliConversionMeson, but has to be treated properly in analysis task itself.


nschmidtALICE:
+1

PR #12085, opened 2019-10-19
[PWGLF] Implementation of AliEventCuts for 2018PbPb

jikim1290:
Implementation of AliEventCuts for 2018PbPb.
Making initial configuration work one time.


BongHwi:
+1

PR #12108, opened 2019-10-22
Fixed event counting bug + added more arguments.

osaarima:
Now only events which are accepted by aliJCatalystTask are counted. Also
added an argument which can be used to choose additional event rejection
criteria.


PWGCF/Correlations/JCORRAN/Pro/AliJCDijetTask.h
```diff
@@ -89,6 +95,8 @@ class AliJCDijetTask : public AliAnalysisTaskSE {
         int fCBin;
         int fCBinDetMC;
         TDirectory     *fOutput; // Output directory
+        UInt_t flags; //
+        AliAnalysisUtils *fUtils=nullptr;//!
```
akubera: Last I knew, initializing members like this is not yet supported, due to old non-c++11 compatibility in some environments. Maybe it's ok now?

osaarima: Hi,
well noticed. I compiled this succesfully in alidock with aliroot5 and aliroot6, but I suppose there might be still problems with other environments. As it is not actually necessary to initialize here (it is done in .cxx as well), I suggest that I would just remove the initialization here just in case.
```suggestion
        AliAnalysisUtils *fUtils; //!
```

akubera:
+1

PR #12115, opened 2019-10-22
A new class that allows you to set different cuts on DCA for different ranges of pt

wrzesaCERN:
A new class that allows you to set in a config file different cuts on DCA for different ranges of pt -- to kanon-proton femto analysis. This class is an extension of AliFemtoMJTrackCut.


lgraczykCern:
+1

lgraczykCern:
+1

PR #12138, opened 2019-10-24
Updated version of the Run3 convertor.

pzhristov:
- The use of event cuts is optional
- The collision time is used from the best method without momentium dependency
- MUON, MUON clusters, ZDC, VZERO as proxy for FIT added
- V0s and cascades added as pairs of indexes
- MC information excluded for the moment, to be replaced
- TOF clusters excluded for the moment, to be clarified


pzhristov:
+test

pzhristov:
+1

PR #12143, opened 2019-10-24
[EMCAL-625] Adaptions of CALO tables for EMCAL

mfasDa:
- Add gain type to cell table
- Add table for calorimeter trigger information


mfasDa:
@pzhristov @mpuccio Please review!

For the PHOS trigger information this should be implemented by the PHOS experts (@kharlov).

How do you intent to store the trigger class? This is needed for the EMCAL trigger (ideally should also be configurable as by default I think the AliEventCuts selects only INT7 events). We will probably need two entries for the event-wise table for the trigger (EMCAL/DCAL median).

pzhristov:
+1

PR #12150, opened 2019-10-25
PWGUD/UPC: Update with AD trigger bits.

siragoni:
Check for CMUP6.
Using the AD HW trigger bits.
Comparison with SW decisions too.

IMPORTANT: tested.
IMPORTANT: commonly used on GRID. Updating it on the LEGO trains for bookkeeping and to allow 
                       others to check the code I am normally using on GRID.


mbroz84:
+1

PR #12164, opened 2019-10-26
[PWGCF] Implementation of the pt efficiency correction.

cimordas:
* Correction hard-coded for TPC-only tracks and LHC10h datasets.


dongjokim:
+1

PR #12184, opened 2019-10-29
Add HF cut-object like classes for application of ML python models

fgrosa:
Co-authored-by: fcatalan92 <fabio.catalano@cern.ch>
1) add mother class for application of ML models trained offline with python on the grid for a generic HF particle that loads the inputs (model paths, pt bins, preselection, list of features on ML output) from a yaml configuration file
2) add classes for Ds and D+ mesons
3) update ML application in Ds task using the object
4) implement application of ML models in D+ and vn (for Ds and D+) tasks


adubla:
+1

PR #12186, opened 2019-10-29
Added TPC active length variable in VarManager; Fill quality flags histograms in the jpsi2ee task; Updated CalibratePeriodPP macro to allow running with both ROOT5 and ROOT6 

iarsene:


iarsene:
+1

fprino:
+1

PR #12189, opened 2019-10-29
Correction on configs

eddoming:
Extra config, plus a correction on low Rejection cut and momentum cut for pions on primary vertex selection


nschmidtALICE:
+1

PR #12215, opened 2019-10-31
Remove hardcoded mult. estimators and recalculate missing info for D

hzanoli:
Dear all,

I have solved a bugs in the code, it affected the cases where the AOD information for the D mesons needed to be filled on the fly (Pb-Pb). It does not affect pp or p-Pb.

Best,
Henrique


btrzecia:
+1

PR #12242, opened 2019-11-05
AliAnalysisTaskHFJetIPQA: adapt cut settings

kagarner:
- introduce individual functions for cutting on DCA, vertex/event, track
- adapt cut default values
- combine all cut values in fAnalysisCuts


lcunquei:
+1

PR #12244, opened 2019-11-05
Added my analysis files to CMakeList and LinkDef files so that they get compiled.

sadkoeni:
- Commented out any print (cout) statements
- Fixed a minor bug in the header.
- Added my Analysis to the CMakeList and LinkDef files.


bdoenigu:
+1

PR #12265, opened 2019-11-07
PWGGA/GaammaConv: in Calibration Task: removed unused histos causing …

jokonig:
…seg fault when running a JJ MC


nschmidtALICE:
+1

PR #12271, opened 2019-11-07
Updates in Make2Prong for V0+bachelor treatement

fprino:
Skip second calculation of V0+bachelor invariant mass
Recompute primary vertex for 2prong candidates only when needed
Do not propagate neutral tracks to DCA


fprino:
Hi @ktf 
from the log of the error it doesn't seem an issue with my modification, which locally I could compile successfully with root6.
Can you please have a look and let me know how to move on?

Thanks!

Best regards, Francesco

ktf:
Looks like the compiler went out of memory. Can you push some minor change to try again (or wait it reprocesses it). I am away right now and I cannot investigate what is going on.

fprino:
Hi @ktf,

thanks for the hint, I added a commit to trigger the reprocessing.

Ciao, Francesco

PR #12276, opened 2019-11-08
Clean dependencies of ML package

fcatalan92:
Investigating some problems related to the number of libraries loaded in some PWGHF trains I founded that the ML library has unneeded dependencies. 
In my opinion this package should be as lightweight as possible, therefore I removed all the dependencies that seemed not necessary.

### DIFF
--- a/ML/CMakeLists.txt
+++ b/ML/CMakeLists.txt
@@ -23,9 +23,8 @@ include_directories(${AliPhysics_SOURCE_DIR}/ML
 )
 
 # Additional includes - alphabetical order except ROOT
-include_directories(${ROOT_INCLUDE_DIRS}
-                    ${TREELITE_ROOT}/include
-  )
+include_directories(${TREELITE_ROOT}/include
+)
 set(SRCS
     AliExternalBDT.cxx
 )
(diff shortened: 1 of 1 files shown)


mpuccio:
@pzhristov this looks fine to me! Can you also give to both me and Fabio the permissions to commit here? We are working and maintaining this...

chiarazampolli:
Ciao @mpuccio , 

I think that this is a file for which special previleges are needed, so even if @fgrosa has the rights to approve in PWGHFvertexingHF, he does not have here - only those with "admin rights" over the whole AliPhysics have. I believe this is intentional, I would keep it this way. Of course, @pzhristov is the one with the final word :-)
For now, I approve. 

Chiara

chiarazampolli:
+1

chiarazampolli:
> Ciao @mpuccio ,
> 
> I think that this is a file for which special previleges are needed, so even if @fgrosa has the rights to approve in PWGHFvertexingHF, he does not have here - only those with "admin rights" over the whole AliPhysics have. I believe this is intentional, I would keep it this way. Of course, @pzhristov is the one with the final word :-)
> For now, I approve.
> 
> Chiara

I mentioned @fgrosa , but in fact it was @fcatalan92  who opened the PR, sorry. Anyway, the reasoning is the same: only admins of the whole AliPhysics have the rights to approve PR on these files. Usually I don't approve myself such things but consult with others before, but this is what we discussed in a separate thread, and looks fine. 

Chiara

mpuccio:
Eheh but this is a directory I created 😛

chiarazampolli:
I think this "rule" applies to all CMake files and LinkDef files: only the admins have the rights to approve changes there. I think, at least.

mpuccio:
Ah no, this rule in the end was not applied. For instance, if I change the pwglf cmakefile or linkdefs I get the PR through without admin approval (and this is also why some libraries are growing indefinitely)

PR #12297, opened 2019-11-11
Update PID for reflections

hzanoli:
Dear all,

I updated how the PID is saved to better handle the reflections.

Best
Henrique


btrzecia:
+1

PR #12310, opened 2019-11-12
PWGDQ/reducedTree/AliCorrelationExtraction: efficiency correction

laltenka:
correction for trigger particle normalization


iarsene:
+1

PR #12317, opened 2019-11-12
AliAnalysisTaskHFJetIPQA: add probability tagging

kagarner:
- enable reading of track probability threshold histograms
- move setting of thresholds to dedicated function ReadThresholdHists 
- introduce DoProbTagging for track probability tagging, same efficiency histograms as for TC tagging are utilised


mfasDa:
+1

PR #12326, opened 2019-11-13
 Add angle between sum and difference of momentum vectors of the two …

sebaleh:
…dielectron legs


aaroncapon:
+1

miweberSMI:
Hi @sebaleh 

i think you need to add the new varoable also in the var manager cxx file (otherwise the namings/titles get confused).

Cheers,
Michael

miweberSMI:
+1

miweberSMI:
+1

PR #12331, opened 2019-11-14
small change in the event cuts affecting the triggers

pganoti:
small change in the event cuts affecting the triggers


BongHwi:
+1

PR #12333, opened 2019-11-14
Changing the correction framewrok to read by default the 1D OADB objects

hahassan7:
As we discussed in the EMCal meeting, we changed the correction framework to read by default the 1D bad channel maps and the 1D recalibration factors. Also now by default it will read the merged the BC histograms. Later, we will change the correction framework to calibrate the LG cells by default.


raymondEhlers:
+1

PR #12348, opened 2019-11-15
Update to He3 Absorption ratio analysis

sadkoeni:
- Edited the task so other particles (p,d) don't get analyzed unless explicitly called.
    - Added Ntuple (should have of Order 5000 entries over LHC161718 pp-HM) for more versatile post analysis.
    - fixed a normalization issue due to different filterbits.


mpuccio:
+1

PR #12349, opened 2019-11-15
Include new feature: kT vs k* 2D-hists for set multBins.

makor:
Plus changes to the AddTaskMacros for additional QA and the new feature.


bhohlweger:
+1

PR #12377, opened 2019-11-18
PWGUD/UPC: several bug fixes + tentative solution to MC problem...?

siragoni:
Pb-p analysis:
- fixed run list
- fixed pt-distr (missing inv mass sel)
- Increased streamer in MC as a possible solution to the LEGO problem (not observed on GRID)...

IMPORTANT: tested a.d already fully run on GRID.


mbroz84:
+1

PR #12378, opened 2019-11-18
Reverting the default setting back to the original value

hahassan7:
After some problems were reported with the time calibration, we decided to revert back default settings to the original default values.


raymondEhlers:
+1

PR #12385, opened 2019-11-19
Standard ESD track cuts configuration set from outside

mfaggin:
The default setting is kDefault (compatibility with previous versions)

fESDtrackCuts = AliESDtrackCuts::GetStandardITSTPCTrackCuts2010(kFALSE,0);
fESDtrackCuts->SetMinRatioCrossedRowsOverFindableClustersTPC(fCrossRowsOverFndCltTPC);

Now, the fESDtrackCuts object can be set to one of the standard configurations through the function SetStandardESDtrkCuts

### DIFF
--- a/PWGPP/EvTrkSelection/AliAnalysisTrackingUncertaintiesAOT.cxx
+++ b/PWGPP/EvTrkSelection/AliAnalysisTrackingUncertaintiesAOT.cxx
@@ -94,8 +94,8 @@ AliAnalysisTrackingUncertaintiesAOT::AliAnalysisTrackingUncertaintiesAOT()
   fCutGeoNcrNclGeom1Pt(1.5),
   fCutGeoNcrNclFractionNcr(0.9),
   fCutGeoNcrNclFractionNcl(0.7),
-  fWhichCuts(kDefault),
-  fTPCclstCut(0)
+  fWhichCuts(kStdITSTPCTrkCuts2011),
+  fTPCclstCut(1)
 {
 
 }
@@ -141,8 +141,8 @@ AliAnalysisTrackingUncertaintiesAOT::AliAnalysisTrackingUncertaintiesAOT(const c
   fCutGeoNcrNclGeom1Pt(1.5),
   fCutGeoNcrNclFractionNcr(0.9),
   fCutGeoNcrNclFractionNcl(0.7),
-  fWhichCuts(kDefault),
-  fTPCclstCut(0)
+  fWhichCuts(kStdITSTPCTrkCuts2011),
+  fTPCclstCut(1)
 {
   //
   // standard constructur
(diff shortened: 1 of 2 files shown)


PWGPP/EvTrkSelection/AliAnalysisTrackingUncertaintiesAOT.cxx
```diff
@@ -138,7 +140,9 @@ AliAnalysisTrackingUncertaintiesAOT::AliAnalysisTrackingUncertaintiesAOT(const c
   fCutGeoNcrNclLength(130.),
   fCutGeoNcrNclGeom1Pt(1.5),
   fCutGeoNcrNclFractionNcr(0.9),
-  fCutGeoNcrNclFractionNcl(0.7)
+  fCutGeoNcrNclFractionNcl(0.7),
+  fWhichCuts(kDefault),
+  fTPCclstCut(0)
```
fprino: Ciao Mattia,
shouldn't we use as default fTPCclstCut(1)?
In this case, when using the standard TPCITS 2011 cuts we use the selection:
     esdTrackCuts->SetMinNCrossedRowsTPC(70);
    esdTrackCuts->SetMinRatioCrossedRowsOverFindableClustersTPC(0.8);
(see below).

Ciao, Francesco

PWGPP/EvTrkSelection/AliAnalysisTrackingUncertaintiesAOT.cxx
```diff
@@ -175,9 +179,48 @@ void AliAnalysisTrackingUncertaintiesAOT::UserCreateOutputObjects()
...
+    if(fTPCclstCut==0)  printf("   ---> cut on TPC # clusters\n\n");
+    else if(fTPCclstCut==1) printf("   ---> cuts on the number of crossed rows and on the ration crossed rows/findable clusters\n\n");
+    fESDtrackCuts = AliESDtrackCuts::GetStandardITSTPCTrackCuts2010(kFALSE,fTPCclstCut);
+    break;
+
+  case kStdITSTPCTrkCuts2011:
+    printf("\n### kStdITSTPCTrkCuts2011 case for ESD track cuts\n   fESDtrackCuts = AliESDtrackCuts::GetStandardITSTPCTrackCuts2011(kFALSE,%d)\n",fTPCclstCut);
+    if(fTPCclstCut==0)  printf("   ---> cut on TPC # clusters\n\n");
+    else if(fTPCclstCut==1) printf("   ---> cuts on the number of crossed rows and on the ration crossed rows/findable clusters\n\n");
+    fESDtrackCuts = AliESDtrackCuts::GetStandardITSTPCTrackCuts2011(kFALSE,fTPCclstCut);
```
fprino: Ciao,

this is the line in which by default I would call GetStandardITSTPCTrackCuts2011(kFALSE,1) and for this reason I suggested to set the default value of fTPCclstCut to 1.

Francesco

fprino:
Hi Mattia,

thanks for implementing this feature.
I added two technical comments in the code, please have a look and let me know if you agree with my proposed modification.
Once this is clarified, we can approve the PR.

Ciao, Francesco

mfaggin:
Ciao Francesco,
I chose to set as "kDefault" what was set so far.
But I can modify as you suggest, for me is totally fine.
I will ping you as soon as I have finished.
Ciao. Mattia.

miranov25:
Hello Francesco.

Please hold on. In reconstruction productions done before 2012 (not sure when we implemented the fix) we did not store cluster map assigned to track. 
````
(2012-01-23 15:46:03 +0000 2118) Float_t AliESDtrack::GetTPCClusterInfo(Int_t nNeighbours/*=3*/, Int_t type/*=0*/, Int_t row0, Int_t row1, Int_t bitType ) const)
````
PbPb 2011 was reconscturted with already corrected version.

For the reconstruction before the fix, it could happen, that we declared many "crossed rows" but too few clusters. Therefore in my recomendation presentation for RAA (~ 6 years ago) I suggested  - for past data also minimal cut on Ncl. 

I'm not sure how this information propagated. For data<2012 we should keep also minimal Ncl cut.

Marian

fprino:
Hi,

@mattia: I agree to keep kDefault to what was used so far (hence with the cut on the clusters) in order to have soem backward compatibility. My comment was about the case kStdITSTPCTrkCuts2011 that I would call with 1.

@marian: the idea here is to have different sets of cuts implemented in the task of the matching efficiency, so that the used can select (from the AddTask) whether to use the 2011 standard cuts (with the crossed rows approach) or the 2010 ones.


Ciao, Francesco

mfaggin:
Ciao @fprino ,
I have modified the code putting GetStandardITSTPCTrackCuts2011(kFALSE,1) as default.
Reading again your comments above, maybe you preferred as default
GetStandardITSTPCTrackCuts2010(kFALSE,1)
Probably I misunderstood.
Let me know and I case I will fix it.
Caio. Mattia.

mfaggin:
Ciao @fprino ,
before I misunderstood your comments. Now the code should be ok according to your suggestions.
Ciao.
Mattia

fprino:
+1

PR #12394, opened 2019-11-20
Include MC handling for Sigma0 Nanos

a-mathis:


jgrosseo:
@ktf do we know why it is not enough that @a-mathis approves this? He says he has rights for the CF and GA folders...

ktf:
+test

Not sure. It looks like the system thinks no one is in charge of it... I will check.

ktf:
+1

PR #12399, opened 2019-11-20
add class and task to analysis strongly intensive quantities and FB corr

isputows:
add class and task to analysis strongly intensive quantities and FB corr


aohlson:
+test

aohlson:
+1

PR #12401, opened 2019-11-20
PID requirement

hzanoli:
I have changed part of the code to ensure that the PID selection is applied i(f requested) and that the PID information for the default PID is saved in any case.


fcolamar:
+1

PR #12403, opened 2019-11-20
Re-introduce basic jet QA plots running all the time

mfasDa:
Observables were part of the THnSparse, but when
moving back to ND-histograms in order to reduce
the memory consumption for the large datasets (pp
triggered and jj anchored to pp 13 TeV) they were
dropped. Added as 2D plots (always versus jet pt)
as QA plots always running.


mfasDa:
+1

PR #12410, opened 2019-11-21
Make EPOSLHC detection working in cocktails

mpuccio:


ddobrigk:
+1 -> approving as this is a copy-paste of a cocktail detection that already works for DPMJet but not for EPOS (discussed with Maximiliano - thanks for the implementation, @mpuccio!)

ddobrigk:
+1

PR #12415, opened 2019-11-21
PWGHF: Fix for missing phi in Bs decay (ITS2 MC)

lvermunt:
Phi meson not written in MC stack in new ITS2 production, so decay is read as non-resonant. Checking also invariant mass of KK-pair to tag now


arossi81:
+1

PR #12427, opened 2019-11-22
Added Task to CMakeLists.txt and LinkDef.h

SHornung1:


lramona:
+1

SHornung1:
The issue is linked to the dependency on the "external" package KFParticle which is not set as possible dependence in the CMake procedure. I have contacted Dario Berzano and Andreas Morsch to ask for support.

PR #12431, opened 2019-11-22
 PWGDQ: LMEE: event variables for different Spherocity calculations

jjungIKF:
PWGDQ: LMee: event variables for spherocity calculation


pdillens:
+1

PR #12452, opened 2019-11-25
PWGDQ: LMEE: spherocity calculation update

jjungIKF:
PWGDQ: LMee: spherocity calculation update


pdillens:
Hi Jerome,
I think it would help the readability of the code a lot if you would use more meaningful names for the iterator variables in your loops, instead of just i, j, etc.

pdillens:
+1

PR #12509, opened 2019-11-29
Jet unfolding threshold changed

mtakamur:
Threshold for jet unfolding process in neutral meson measurement inside jets has been changed from 5 GeV/c to 10 GeV/c.

5 GeV/c was for temporary test use, so it's been back to the original value.


dmuhlhei:
+1

PR #12511, opened 2019-11-29
PWGUD/UPC: Fix Alec's task. Adding my subwagon feature...

siragoni:
IMPORTANT: tested, but I don't know if mine works on LEGO as intended...


mbroz84:
+1

PR #12518, opened 2019-11-29
Avoid using TParticle::GetMother/Daughter...

shahor02:
Code using these calls will be suppressed soon as not compatible with MC embedding.


shahor02:
+1

PR #12522, opened 2019-11-30
add tagging as exotic to clusters with no cells in diff TCard

gconesab:


gconesab:
+1
I have other commits pending so I approve it. In any case, I can revert or change the code in another commit if something is not right.

PR #12524, opened 2019-11-30
add MC variables for secondary electrons

dsekihat:
add MC variables (secondary from material, secondary from weak decay)


miweberSMI:
+1

PR #12555, opened 2019-12-03
modifications for analyses of rare resonances

aknospe:
Introduction of Pseudorapidity cuts on resonance daughters, method for extracting generated long-lived particles (pi±, K±, p) from MC, modification for Lambda+Lambda analysis


BongHwi:
+1

PR #12564, opened 2019-12-04
PWGGA/GammaConv: adapted pi0 masscut for 7TeV + final nonlin & more

fjonasALICE:
-- adapted inv mass cut to better fit with 7 TeV data
-- added configs for sys variations LHC11


nschmidtALICE:
+1

PR #12567, opened 2019-12-04
PWGHF: Option to only process already filled candidates in ITS3Improver

lvermunt:
ITS3Improver takes too long because it fills all HF candidates by default, flag to only process already filled ones


fgrosa:
+1

PR #12569, opened 2019-12-04
including DCAL trigger

crisjahnke:
including possibility to use DCAL trigger together with EMCal trigger, as a single detector


miweberSMI:
+1

PR #12572, opened 2019-12-04
PWGHF: Speed up ITS3Improver

lvermunt:
ITS3ImproverTask was very slow due to GetEntries() call in loop (in combination with the huge amount of 3prongs stored in the new ITS Upgrade MC)

Processing 1 event: 5 minutes -> few seconds


fprino:
+1

fgrosa:
+1

PR #12573, opened 2019-12-05
Calibrator now able to select fired trigger classes from string check

ddobrigk:


ddobrigk:
@iarsene, this should do what you wanted. You should invoke the function AliMultSelectionCalibrator::SetFiredTriggerString("blabla") - please let me know if it all works fine :-)

ddobrigk:
+1

PR #12574, opened 2019-12-05
PWGUD/UPC: Suffix to list, not to folder.

siragoni:
Switched to output list for the LEGO train suffix.

IMPORTANT: tested.


mbroz84:
+1

PR #12583, opened 2019-12-05
corrected paircleaner loop

magicep0t:
In CleanDecayAndDecay the UseParticle check was done twice on particle1 instead of both


bhohlweger:
+1

PR #12584, opened 2019-12-05
Out-of-Bunch Pileup cut for resonances

aknospe:
Introduced out-of-bunch pileup cut for resonance analyses. At least one final-state decay product of the resonance must have an ITS or TOF hit.


BongHwi:
+1

PR #12595, opened 2019-12-06
Fixed physics selection OADB for LHC18b

ekryshen:
Introduced special conditions for isolated bunch fills and special online-vs-offline correlation cut in V0 due to non-standard V0 settings.

This fix should solve the issue with ubnormal average number of tracklets.


ekryshen:
+1

PR #12607, opened 2019-12-08
AliEventCuts:: corrected run range for run1 pp 8TeV

mkruegerGitHub:


chiarazampolli:
This corresponds to PHYSICS runs, with data migrated, correct @mkruegerGitHub ?
I am approving as checking in the logbook it seems fine. Before, almost all 12i was excluded, if I am not wrong.

Chiara

chiarazampolli:
+1

mkruegerGitHub:
Hello Chiara,
the runs I specified now are corresponding to the first one in LHC12a and the last LHC12i available in our file system.
The run range in the version before did not include LHC12i because I accidentaly took the first instead of the last run in this period.
Thanks for approving.
Cheers,
  Mario

chiarazampolli:
Thanks @mkruegerGitHub for explaining!

File system --> alien, right? So my query to the logbook was correct.

mkruegerGitHub:
Yes

PR #12613, opened 2019-12-09
PWGHF: Adding PreSelect for pK0s for Pb-Pb

lvermunt:
Attempt to speed-up running the Lc->pK0s analyses in Pb-Pb 2018 by adding a similar PreSelect function as used for the D mesons.

Note that we don't call the ReconstructSecondaryVertex() for the pK0s candidate, so the FillRecoCasc function is not as heavy as for the D mesons.


fprino:
+1

PR #12614, opened 2019-12-09
Add notes on unfolding normalization

raymondEhlers:
As per our discussion, I briefly wrote up these questions so that we we would have a reference in AliDoc. Let me know if you'd like any additions or changes

@mfasDa @lcunquei


raymondEhlers:
+1 We can always update it further if necessary

PR #12616, opened 2019-12-09
Reduction of number of output histograms

fkrizek:
the ouput has been too large  i  reduced number of output plots
  (i willl fill histogram only for one rho estimator,  and i will not fill centrality histograms)


mfasDa:
+1

PR #12623, opened 2019-12-09
fix stat error calculation after scaling

gconesab:


gconesab:
+1
Just to wake up the process

PR #12624, opened 2019-12-09
modified macros for rare resonances and phi

aknospe:
modifications to my AddTask and Config macros, needed for studies of rare resonances, pentaquarks, and phi


BongHwi:
+1

PR #12643, opened 2019-12-11
PWGCF: Remove uncessary methods

cimordas:
Major cleaning of the code with:
* Changed the Double_t and TH1D into Float_t and TH1F where it was possible
* Gathered the TH1 into different arrays to ease maintenance and future possible additions of control histograms
* Added support to compute correlators with v7 and v8 (previously up to v6)
* Removed unnecessary methods like nested loops to ease the readability


yzhouCERN:
+1

PR #12655, opened 2019-12-11
ML-based background correction grid task

rhaake:
Make sure the tasks works with current version of ROOT/AliPhysics/python-modules.


rhaake:
+1

PR #12673, opened 2019-12-12
Adding OADB with run-by-run TOF trigger efficiency for Run 2

ekryshen:
Covering pPb2016, pp2017-2018, XeXe2017 and PbPb2018


ekryshen:
+1

PR #12676, opened 2019-12-12
Add TMVA-BDT selection in vn framework

cmengke:
TMVA:BDT-based selection in vn framework by xinye


PWGHF/vertexingHF/vHFML/CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ include_directories(${ROOT_INCLUDE_DIRS}
                     ${AliPhysics_SOURCE_DIR}/OADB
                     ${AliPhysics_SOURCE_DIR}/PWGHF/vertexingHF
                     ${AliPhysics_SOURCE_DIR}/PWGHF/vertexingHF/charmFlow
+                    ${AliPhysics_SOURCE_DIR}/PWGHF/vertexingHF/vHFBDT
```
fgrosa: You should include the task in vertexingHF and not vHFML for the moment, since you don't need what is in vHFML and we have the issue with treelite still unsolved.

PWGHF/vertexingHF/vHFML/PWGHFvertexingHFMLLinkDef.h
```diff
@@ -8,11 +8,12 @@
 #pragma link C++ class AliAnalysisTaskSEDplus+;
 #pragma link C++ class AliAnalysisTaskSEDs+;
 #pragma link C++ class AliAnalysisTaskSECharmHadronvn+;
+#pragma link C++ class AliAnalysisTaskSECharmHadronvnTMVA+;
```
fgrosa: Same comment as for the CMakeList file

cmengke:
@fgrosa @xinyepeng 
Hi Fabrizio
We can try to move these out from ML directory, but thet still rely on some ML dependencies scattered in the code. Do you think it possible to suppress any potential problem in treelite by simply turning the flag fApplyML false? Since at least I don't see any problem compiling for now.

fgrosa:
Hi @cmengke, 

the problem with treelite is not giving errors in the compilation, but when running trains in specific grid sites, most probably due to openmp (which is currently used by treelite if found on the machine). In particular, when jobs run on these sites, they all fail because it fails in loading the libraries. Hence, unfortunately setting fApplyML to false does not solve the issue because what matters are the libraries that are loaded, and the libraries are loaded independently of the flag, since they are loaded at compile time. So unfortunately it is necessary to remove all the lines of code where AliHFMLResponse is used to get rid of the issue. 

Alternatively you can of course accept the higher rate of failures and live with it, but in this case there is no need to duplicate the code and you can include your developments in the original task.

Ciao,
Fabrizio

fgrosa:
Hi @cmengke, thanks a lot! 
+1

fgrosa:
+1

PR #12700, opened 2019-12-13
Make eta limits configurable

raymondEhlers:
Also some minor modernizations to make iteration of particle and track
containers safer


raymondEhlers:
+1

PR #12732, opened 2019-12-17
Fix sparse binning

raymondEhlers:
@mfasDa Whether the hist is a closure is currently applied as a weight for the sparse. I think this should be an axis (I think this was your intent based on how you define the binning). I'll leave this to you to approve or close


mfasDa:
+1 @raymondEhlers Thanks for spotting!

PR #12748, opened 2019-12-18
PWGGA/GammaConv: Added new train configs for EDC pp 13 TeV for pi0 and eta. 

jokonig:
Also modified EDC NonLin settings for LowB p 13 TeV analysis.
Added two new train configs for eta->pi+pi-pi0 in pp 13 TeV


dmuhlhei:
+1

PR #12761, opened 2019-12-19
Possibility to use xml files loaded on CVMFS for BDT application (F. Jonkman)

dcaffarr:
@chiarazampolli @andreaalici FYI


cterrevo:
+1

PR #12763, opened 2019-12-19
PWGGA/GammaConv: omega task, implemented correct trigger handling for MC

fjonasALICE:
Changes fix wrong efficiency and acceptance when using trigger mimicking with MC and the heavy meson task. Before, MC particles on generator lever were never processed for a rejected event, however, this is needed in order to account correctly for trigger efficiency which is not 100% near threshold.


nschmidtALICE:
+1

PR #12777, opened 2019-12-20
Automatic configuration of the energy smearing

mfasDa:
Users no longer need to configure a smear model
by hand if they want to use a smeared patch energy
calculated from the FEE energy with the same
resolution as in data. The smearing is automatically
activated in data and MC. Currently the default
smearing is used - custom smear models would still
work.


mfasDa:
+1

PR #12785, opened 2019-12-20
Adding possibility to estimate background via soft-pion track rotatio…

arossi81:
…n for SigmaC; adding possibility to set pdg code different from Lc one at reco level.


arossi81:
+1

PR #12837, opened 2020-01-05
bug fix and removed unnecessary variables

hmurakam:
this needs to be fixed (other wise task will crash): AliAnalysisTaskMaterialHistos.cxx line 334


FriederikeBock:
+1

PR #12842, opened 2020-01-06
PWGHF: Apply ML decision on GRID for TTreeCreator

lvermunt:
Using the newly developed tools in the vHFML directory to apply a filtering XGBoost probability cut on the GRID before filling the TTrees.

**Note:**
Because of the treelite dependency of PWGHF/vertexingHF/vHFML dir, and the fastjet and EMCALjet dependencies of the PWGHF/treeHF dir, it was not (easily) possible to integrate the AliHFMLResponse class in the existing TTreeCreator task. I therefore made a stripped version of the TTreeCreator and TTreeHandler tasks and located them in the vHFML directory. Hopefully this is a temporary solution, as this duplicates code.


fgrosa:
+test

fgrosa:
+1

PR #12860, opened 2020-01-07
PWGGA/GammaConv: Added new JJ MCs with high energy gamma in calo acceptance for pp 13 TeV (2016, 2017, 2018))

jokonig:


nschmidtALICE:
+1

PR #12871, opened 2020-01-08
Updated task V0ch for pp and Pb_Pb

MustafaNajiAnaam:
I updated the task  V0-Ch for pp and Pb_Pb to make it easy for sys and i changed the cuts for Vtx and mix


aohlson:
+1

PR #12881, opened 2020-01-08
New method for correcting the IP in MC

hahassan7:
We are trying to investigate a new method for correcting the IP in MC which follows the same procedure as [https://www.sciencedirect.com/science/article/pii/0168900295012877](url). These codes extract the necessary histograms for the fitting procedure.


btrzecia:
+test

btrzecia:
+1

PR #12888, opened 2020-01-09
New train config for injet analysis in pPb at 5 TeV added

mtakamur:
train config 1600 with 3 set of cut numbers has been added for injet analysis.
1 for MB analysis
2 for injet analysis without jetQA
3 for injet analysis with jetQA


dmuhlhei:
+1

PR #12889, opened 2020-01-09
PWGGA/GammaConv: pt dep mass cut fix FWHM to sigma

fjonasALICE:
after a closer look by @jluehder it was discovered that not the FWHM of the pi0 has been used to determine the mass cut parametrisations but instead sigma. Therefore the additional conversion from FWHM to sigma in the task is wrong


dmuhlhei:
+1

PR #12911, opened 2020-01-10
Some development of treelite ML interface 

fcatalan92:
This PR introduces some improvements to the treelite interface:
- it is no more needed to add the treelite and yaml-cpp include path in the makefile of the package that use AliMLResponse
- it is possible to pass directly a vector of variables to the model predict (can be useful to debug)
- the case in which the config file is in the current directory is now supported

@mpuccio @pfecchio What do you think?

### DIFF
--- a/ML/AliMLModelHandler.cxx
+++ b/ML/AliMLModelHandler.cxx
@@ -13,16 +13,20 @@
 #include "AliMLModelHandler.h"
 
 #include <map>
-#include "assert.h"
 #include "yaml-cpp/yaml.h"
 
 #include <TFile.h>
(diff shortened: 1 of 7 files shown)


fgrosa:
+1

mpuccio:
Ciao @fcatalan92 , with the update of treelite there is a conflict: can you fix it? Then we can merge this as soon as @pfecchio gives green light

fcatalan92:
Ciao @mpuccio, the conflict should be fixed

mpuccio:
+test

fgrosa:
+1

PR #12929, opened 2020-01-13
Revert change to make sure dailies can compile.

ktf:


ktf:
@mpuccio we seem to have a problem with the version of treelite used by the daily builds. Was it updated as well?

ktf:
+1

mpuccio:
I thought it was, but does it mean that the pr checker runs different defaults wrt the release?

mpuccio:
Indeed there is a mismatch between PR checker and release...

PR #12947, opened 2020-01-15
Readme edits

gmandaglio75:
Macros modified to include the possibility to use the Out-Of-Bunch check in the analysis


gmandaglio75:
Please update these two Macros, to include the possibility to use the OOB check for KStarCharged analysis

BongHwi:
+1

PR #12955, opened 2020-01-16
PWGHF: Update for ITS3 background shape studies

lvermunt:
* Storing injected Ds (prompt or FD) with random HIJING pion for background shape studies
* Use Ds PreSelect function to speed up
* Add option for basic topological selection on Bs candidate to reduce output size TTrees


arossi81:
+1

PR #12973, opened 2020-01-17
PWGGA/GammaConv: added AliConversionAodSkimTask

fjonasALICE:
added AodSkimTask to GammaConv which is based on AliAodSkimTask and offers new functionality to skim for events with conversion photons with minPt and in given acceptance. I decided to create this "new" version in the GammaConv directory due to library linking issues when trying to access conversion photons from within the EMCal libraries. I think in the long run it would be best to use this macro by default.


nschmidtALICE:
Addtask is missing. Please add that, too.

nschmidtALICE:
+1

PR #12990, opened 2020-01-20
Add option for Trigger Alias

mfasDa:
Trigger aliases allow to have multiple names for the
same trigger condition. This is of relevance for the
Data period from 2012: The trigger classes were called
EGA and EJE, however in order to port code adapted to
data periods with EG1/EG2 and EJ1/EJ2, it is easier
to identify the trigger with either of the names (EGA/EJE
for backward compatibility, EG1/EG2/EJ1/EJ2 for new
code)


mfasDa:
+test

mfasDa:
+1

PR #13002, opened 2020-01-21
Fix pt binning at part. level, add theta_g

mfasDa:
- Fix copy/paste error at particle level leading
  to det. level binning used at part. level
- Add histograms and response matrices for
  theta_g (=R_g/R)


mfasDa:
+1

PR #13007, opened 2020-01-22
PWGUD/UPC: Reweighting procedure to change MC input shape.

siragoni:
Reweighting procedure to change MC input shape from STARlight.

IMPORTANT: tested, functionalities also ok.


mbroz84:
+1

PR #13023, opened 2020-01-22
PWGHF: Fix to run upgrade analyses on data

lvermunt:
Fix segmentation violation from AliVertexingHFUtils::IsTrackInjected when running on actual data


arossi81:
+1

PR #13033, opened 2020-01-23
PWG/Cocktail/parametrisations: added parametrisations for sec. cocktail for 13TeV

FriederikeBock:
- parametrisations only for secondary cocktail, not to be used for dir gamma or di-electron analysis yet


nschmidtALICE:
+1

PR #13035, opened 2020-01-23
Include MC information and minor updates

hzanoli:
The following things were included:
- Save MC information at generated level for HFe and D mesons.
- Added possibility to turn on/off the processing of D, electrons and events
- Added more information for the Events
- Removed TPC N clusters in favor of TPC crossed rows for electrons
- Change event selection to use AliEventCuts
- Added variables for systematics of NHFe subtraction
- Optimised some of the default values to reduce size


btrzecia:
+1

PR #13048, opened 2020-01-26
remove pile-up cut for MC

panjinjin:
remove pile-up cut for MC


panjinjin:
dear approvers, could you please +1 for this pull request? so that the change can be included in the today's AliPhysics tag. Thanks a lot

altsybee:
+1

PR #13056, opened 2020-01-27
Use fine pt binning for data

mfasDa:
Use fine pt-binning for data and do not truncate in pt-dimension.
Apply realistic binning and tructation offline. This is of
relevance in order to combine triggers at data level and unfold
the over the combined pt range.


mfasDa:
+1

PR #13063, opened 2020-01-27
PWGHF: ITS3 preselect for standard D mesons

lvermunt:
To be able to run tests/for QA'ing purposes on the new ITS3 MC productions


cterrevo:
+1

PR #13093, opened 2020-01-29
oob pileup rejection flag and AliEventCuts for run-2 analyses - in ge…

tuvarichert:
…neral preparing for run 2 analyses


bdoenigu:
+1

PR #13096, opened 2020-01-29
Moved trigger/V0M check before event cuts

vvislavi:
Moved trigger/V0M check before event cuts so that event selection QA gets properly filled


vvislavi:
+1

PR #13098, opened 2020-01-29
AO2d links from event array to lower hierarchy arrays

mgheata:
To use AO2D format in current analysis tasks, the event data needs to be aggregated from separate arrays. Now there is no connection allowing to see which segment of the arrays belong to a given event. The PR adds 2 arrays of indices to the O2events tree for each entry. The first array contains the start indices of the event data in the trees of tracks, v0s, cascades, ..., while the second array contains the number of entries to be read. Adding these arrays only increase the total file size by 0.015%.


pzhristov:
+1

PR #13111, opened 2020-01-30
PWGHF: Split ITS3 Upgrade Analysis flag

lvermunt:
Splitting ITS3 analysis flag into three options for more freedom:
 * Flag to say you run on ITS3 production for differences in CheckOrigin
 * Flag to enable ITS3 PreSelect function
 * Flag to store HIJING only background


jezwilkinson:
+1

PR #13143, opened 2020-02-03
PWGDQ EtaReconstruction Implement pairing with V0s

feisenhu:
I implemented the selection of V0s in order to use them in secondary pairing and use them as well in the four pairing. Also implemented cuts on AliDielectronPairs.


miweberSMI:
+1

PR #13151, opened 2020-02-03
PWGHF: Container name fix for ITS2-ITS3 setup

lvermunt:
The TTreeCreator -> ITS3Improver -> TTreeCreator setup we use on trains only worked for beauty hadrons. Added an extra argument in AddTask, so we can use same setup also for the other particles.


arossi81:
+1

PR #13164, opened 2020-02-04
Add new substructure analysis task

raymondEhlers:
Lots of cleanup to be done, but this is just the first pass


raymondEhlers:
+1

PR #13176, opened 2020-02-05
[PWGCF] Add support for particle weights.

cimordas:
> Add support to get the pT, eta and phi distributions at kine and reco levels for MC-Generator AODs
> Add support to use root file with weights on the local machine and the Grid.
> Major cleaning of the data members and the comments.


dcaffarr:
+1

PR #13179, opened 2020-02-05
Move dynamical before hard cutoff, add # of splittings

raymondEhlers:
Dynamical doesn't need the hard cutoff. This is one of the benefits of
the method


raymondEhlers:
+1

PR #13192, opened 2020-02-06
Variety of minor bug fixes, task improvements

raymondEhlers:
Including:

- Fix labeling for det level quantities. A missing "," lead the last two
values to be truncated (but the data isn't corrupted - just disabled).
- Reduce verbosity of code.
- Remove unused code. Still more to be done.


raymondEhlers:
+1

PR #13196, opened 2020-02-06
PWGGA/GammaConvBase/ 

AMechler:
update Nonlin for PCM-EDC pp 13 TeV
Adding plot to investigate trackPID vs pt

### DIFF
--- a/PWGGA/GammaConvBase/AliConversionPhotonCuts.cxx
+++ b/PWGGA/GammaConvBase/AliConversionPhotonCuts.cxx
@@ -102,6 +102,7 @@ AliConversionPhotonCuts::AliConversionPhotonCuts(const char *name,const char *ti
   fHistograms(NULL),
   fPIDResponse(NULL),
   fDoLightOutput(kFALSE),
+  fDoPlotTrackPID(kFALSE),
   fV0ReaderName("V0ReaderV1"),
   fMaxR(200),
   fMinR(0),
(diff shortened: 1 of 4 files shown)


PWGGA/GammaConvBase/AliConversionPhotonCuts.cxx
```diff
@@ -595,6 +597,24 @@ void AliConversionPhotonCuts::InitCutHistograms(TString name, Bool_t preCut){
       //    }
   }
 
+  fHistoTrackPID=new TH2F(Form("TrackPID %s",GetCutNumber().Data()),"TrackPID",15,-0.5,14.5,250,0,25);
```
dmuhlhei: Could you please make this histogram being optional using a switch? We are more on the upside of memory usage, so that adding another 2D histogram should be carefully considered

AMechler: ok sure
i will change this

AMechler: > Could you please make this histogram being optional using a switch? We are more on the upside of memory usage, so that adding another 2D histogram should be carefully considered

Done

AMechler: i will do another commit sorry where i add the switch to the addtask

FriederikeBock:
+1

PR #13208, opened 2020-02-06
Include correct handling of RUN1 evts to the AddTask for the CC inves…

makor:
…tigation.

Fix in StandardCutSelection.


a-mathis:
+1

PR #13218, opened 2020-02-06
adding in studies to match the subjets

lhavener:
forgot to increment class def

fixing momory leak

Adding in the studies to match subjets


raymondEhlers:
+1

PR #13240, opened 2020-02-07
add AddTaskAccContForWeights.C for the purpose of calculating weights

ShiQiu1:
Add a modified AddTaskAccContForWeights.C based on PWGCF/EBYE/macros/AddTaskAccCont.C so that the distribution of e.g. pt, eta for each centrality bin can be plotted at one time properly


aohlson:
+1

PR #13255, opened 2020-02-10
Defining variables for ML when 7 were used for training

chiarazampolli:


chiarazampolli:
Pinging @andreaalici , @dcaffarr , @florisjonkman

PR #13264, opened 2020-02-11
[PWGCF] Adding new task.

lauraser:


aohlson:
Hi @lauraser, 
It seems like you are adding a task to do femtoscopy in the PWG/FLOW directory.  Wouldn't it make more sense to add it to PWGCF/FEMTOSCOPY?
Cheers, 
Alice

aohlson:
Also, I may be missing something, but it seems like your task does not really do anything yet (just filling a pt spectrum).  In that case, it may be easier and faster to develop your task locally, before committing.  Once a task with this name is in AliPhysics, it can be more challenging to run an updated version in "local" or "test" mode because of naming conflicts.  
In this period, when you're making a lot of significant updates and want to be able to develop and change and test things quickly, it might actually be a hindrance to have your task in AliPhysics.  My advice would be to wait to commit the developed task until you want to run over a large dataset.  
Just my two cents, 
Alice

aohlson:
+test

aohlson:
+1

PR #13268, opened 2020-02-11
PWGHF: Add ITS3 mass PreSelect also for Lc pKpi

lvermunt:
Using extra PreSelect function for Lb -> Lc pi, also for Lc -> pKpi itself


arossi81:
+1

PR #13269, opened 2020-02-11
 PWGGA/GammaConv changing cuts for pp 13 TeV pcm-edc and edc-phos

AMechler:
- PWGGA/GammaConv changing cuts for pp 13 TeV pcm-edc and edc-phos
- adding LHC16+17+18 to the reject LED events periods


dmuhlhei:
+1

PR #13271, opened 2020-02-11
Major overhaul of tree creation for S0/RT Analyses

smaff92:
Major changes to the Lund Trees used in the pp@13 TeV S0/RT analyses.
New classes are added (modifications from old classes) that contain only the relevant information needed for these analyses. More complete trees can be found at TPCTOFfits (albeit without cascades).

An alternative version of the AT will be created in the near future that contains all of the previous capabilities.


ddobrigk:
+1

PR #13286, opened 2020-02-12
Possibility to separate direct decays from resonant ones in MC

mfaggin:
Modifications:
 1. TH3 transformed in THn objects, with n=4
 2. Axis added in THnSparse at reco level
 3. Axis added in THnSparse for Lc(<-Sc) at gen level

 These new axes store the return value of
  - AliAnalysisTaskSEXicTopKpi::CheckXicpKpiDecay for Xic
  - AliVertexingUtils::CheckLcpKpiDecay                   for Lc

   (4. cosThetaStar calculation fixed, using the fPdgFiducialYreco mass hypothesis)


arossi81:
+1

PR #13295, opened 2020-02-12
 adding combined trigger for EMCal and DCal

crisjahnke:
combined trigger for J/psi analysis using EMCal and DCal.


miweberSMI:
+1

PR #13305, opened 2020-02-13
 PWGGA/GammaConv/   adding low b-field and exotic cluster variation for triggers

AMechler:
- adding low b-field and exotic cluster variation for triggers

_I'm sorry fore the many small pull requests over the last few days_


nschmidtALICE:
+1

PR #13308, opened 2020-02-13
PWGGA/GammaConv: added 0.1% multiplicity bins for high mult. triggered data & added respective train configs

jokonig:


nschmidtALICE:
+1

PR #13336, opened 2020-02-14
Fix for compilation issue with emplace_back

raymondEhlers:
Apparently emplace_back isn't supported for std::bool until c++14


raymondEhlers:
+1

PR #13345, opened 2020-02-16
Keep subjects from the same splitting together

raymondEhlers:
This should just affect the ordering of the subjets output - not
the actual stored information.


raymondEhlers:
+1

PR #13348, opened 2020-02-17
Add THnsparse and histos for MC analysis for XicPlus

gluparel:
Hello,
could you please commit these modifications to the task for the XiCPlus analysis?
Many thanks
Grazia


cterrevo:
+1

PR #13352, opened 2020-02-17
Add Hijing BIT for Hijing signal check

stefanopolitano:
-BIT for Hijing signal and his setter are added in AliHFMLVarHandler.

- Conditions to switch on/turn off Hijing BIT are added in AliAnalysisTaskSEDs.


fgrosa:
+1

PR #13359, opened 2020-02-18
PWGDQ EtaReconstruction: fixed condition in order to use the prefilter

feisenhu:
Small changes to the PreFilter condition.


miweberSMI:
+1

PR #13364, opened 2020-02-18
AliAnalysisTaskHFJetIPQA: add possibility to use fixed thresh cut

kagarner:
- enable usage of one single IP theshold value for all track types 
- enable cuts on IP and IP significance
- add histograms for threshold effects from IP>0, IP>IPthresh


lcunquei:
+1

PR #13367, opened 2020-02-18
Changing the method of selecting b->e for MC DCA

deepathoms:
- Changing the method of selecting b->e and c->e for MC DCA templates by going back to the first B hadron

- Extending the electron THnsparse to 60 GeV/c


deepathoms:
+1

PR #13375, opened 2020-02-19
1. Dumpsted added to proton -deuteron  NanoAOD task and AODtask, 2.  Output containers for proton-deuteron AOD task redefined such that GentleFemto can be used easily.

sbhawani:


a-mathis:
+1

PR #13382, opened 2020-02-19
Fix for untagged jets for Rg, Thetag and Nsd

mfasDa:
For untagged jets (zg < zcut) the values for Rg,
Thetag and Nsd are set to negative values ~0.
Dedicated bins in the histograms for the observables
handle untagged jets.


mfasDa:
+1

PR #13414, opened 2020-02-21
Changing the Track Counting Tagger to tag on the IP significance

hahassan7:
I changed the analysis code so the track counting tagger can the b-jet candidates using the IP significance instead of the IP. I added a flag to turn off/on this feature.


fcolamar:
+1

btrzecia:
+1

PR #13419, opened 2020-02-22
Add pt hard information for pp

raymondEhlers:
Also added note about harmless calls for setting the rho name in
the case of pythia. I left it because I'm not entirely sure that it
never plays a role in PbPb.


raymondEhlers:
+1

PR #13447, opened 2020-02-25
Allow Individual V0RadiusMin/Max cut and V0CPA cut

BongHwi:
Previously, if we put the V0CPA cut, it returned always false.


mpuccio:
+1

PR #13462, opened 2020-02-26
add Cocktail/parametrisations/PbPb_5020GeV_hadronicOnly.root

sstiefel19:


nschmidtALICE:
Please check/confirm that the cocktail framework can actually work with such a set of parametrizations. We have never checked that it can work without the pi0 parametrization.

nschmidtALICE:
+1

PR #13468, opened 2020-02-27
Silence error message of failed grooming in case of purely charged/ne…

mfasDa:
…utral jets

In case only using charged / neutral constituents for the
SoftDrop jets with NEF=1/0 lead to jets with 0
constituent. Those jets should be discarded in the
SoftDrop, and the error message should be skipped.


mfasDa:
+1

PR #13471, opened 2020-02-27
Update of OmegaToPi0Gamma Task

mhemmer-cern:
+Added more AOD support regarding MC histograms
+changed pT binning (to get to 100 GeV/c) and minv binning (less bins)
+Added photon selection as extension for the selection window cut
+Added train configs for photon selection

### DIFF
--- a/PWGGA/GammaConvBase/AliConversionMesonCuts.cxx
+++ b/PWGGA/GammaConvBase/AliConversionMesonCuts.cxx
@@ -2161,6 +2161,13 @@ Bool_t AliConversionMesonCuts::SetSelectionWindowCut(Int_t selectionCut){
       fSelectionNSigmaHigh = 2.;
       fMassParamFunction   = 4;
       break;
+    case 29: // t // EMC-EMC  gamma selection
+      fSelectionLow       = 0.1;
+      fSelectionHigh      = 0.145;
+      fAcceptMesonMass    = kTRUE;
+      fUseGammaSelection  = kTRUE;
+
+      break;
     default:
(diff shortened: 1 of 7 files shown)


PWGGA/GammaConvBase/AliConversionMesonCuts.cxx
```diff
@@ -126,6 +126,7 @@ AliConversionMesonCuts::AliConversionMesonCuts(const char *name,const char *titl
   fMesonKind(0),
   fIsMergedClusterCut(0),
   fUsePtDepSelectionWindow(kFALSE),
+  fUseGammaSelection(kFALSE),
```
nschmidtALICE: Why did you add this variable? It is not being used...

mhemmer-cern: It should be used in SetSelectionWindowCutPtDep for multiple cases

nschmidtALICE: Maybe i'm overlooking something, but it is only set to kTRUE there but the variable is not used for anything.

mhemmer-cern: there is also the function 
```C++
Bool_t   UseGammaSelection() const{return fUseGammaSelection;}
```
in AliConversionMesonCuts.cxx
which is called in the OmegaToPiZeroGamma Task to check if the flag is set or not.

nschmidtALICE: I figured it out :)

nschmidtALICE:
+1

PR #13486, opened 2020-02-28
PWGGA/GammaConv: Added new in event background calculation method

jokonig:
This new method is based on swapping the position of gamma pairs thus keeping the 4-momentum of the hypothetical particle the gamma pair origins from. 
This can be realized with a simple rotation of both gammas around the momentum axis of their hypothetical mother particle. A fixed 90 degree rotation is implemented as well as a random rotation (which excludes roatations that are too close to 0 or 180 degree)
Furthermore this method cn be realized with TGenPhaseSpace which generates a completly new decay. Using TGenPhaseSpace is for now only implemented in the GammaCalo task.
For the random rotations and TGenPhaseSpace multiple different rotations/decays for the same gamma pair can be applied to increase statistics (still has to be checked how many independant rotations are possible).
To check if the newly created gammas are within the detector acceptance and not placed on a bad cell of the calorimeters, the cell Id is calculated from the eta phi coordiantes which is then checked if the cell is a bad cell.


PWGGA/GammaConvBase/AliConversionMesonCuts.h
```diff
@@ -302,6 +306,10 @@ class AliConversionMesonCuts : public AliAnalysisCuts {
     Bool_t      fDoJetPtMixing;                 ///< flag to enbale mixing by jet pt bins
     Bool_t      fDoSphericityMixing;            ///< flag to enable Sphericitymixing for meson bg estimation
     Bool_t      fUseTrackMultiplicityForBG;     ///< flag to use track multiplicity for meson bg estimation (else V0 mult)
+    Bool_t      fDoGammaSwappForBg;             ///< flag to use cluster swapping for background estimation
+    Bool_t      fDoWeightingInSwappBg;          ///< flag to use multiplicity weighting for cluster swapping for background estimation
+    Int_t       fGammaSwappMethodBg;            ///< flag to switch between different methods for cluster swapping: 0= 90 degree; 1=random angle
+    Int_t       fNumberOfSwappsForBg;           ///< flag to enable multiple rotations for 1 photon pair for cluster swapping Bg
     Bool_t      fEnableMinOpeningAngleCut;      ///< flag to enable min opening angle cut
     Bool_t      fEnableOneCellDistCut;          ///< flag to enable 1 cell dist cut
     Bool_t      fAllowCombOnlyInSameRecMethod;  ///< flag to disable inv mass pairing among different calo's
```
dmuhlhei: ClassDef increment is missing here

jokonig: done

dmuhlhei:
+test

PWGGA/GammaConvBase/AliCaloPhotonCuts.cxx
```diff
@@ -677,6 +681,10 @@ void AliCaloPhotonCuts::InitCutHistograms(TString name){
       fHistClusterEtavsPhiAfterAcc->GetXaxis()->SetTitle("#varphi (rad)");
       fHistClusterEtavsPhiAfterAcc->GetYaxis()->SetTitle("#eta");
       fHistograms->Add(fHistClusterEtavsPhiAfterAcc);
+      fHistClusterEtavsPhiBackground  = new TH2F(Form("EtaPhi_Background %s",GetCutNumber().Data()),"EtaPhi_Background",nEmcalPhiBins,EmcalPhiBins,nEmcalEtaBins,EmcalEtaBins);
```
dmuhlhei: What about creating these histograms only if ")->DoGammaSwappForBg()" is running?

jokonig: I added a flag and a setter in the photon cuts and modified the AddTasks for Calo and ConvCalo accordingly. Now the histogram is only initialized if one uses the cluster swapping background. Is this solution okay?

FriederikeBock:
+1

PR #13496, opened 2020-02-28
PWGDQ EtaReconstruction: implemented MCSignal check for generated and…

feisenhu:
implemented functions to check the full MC Signal of four particles. 
Similar to the changes made in the last commit.


miweberSMI:
+1

PR #13506, opened 2020-02-29
Updates on AnalyzeScOutput.C macro

mfaggin:
- explicit selection of prompt reconstructed Lc(<-Sc) or Sc
 - fixed the DeltapT normalization in cross section calculation
 - add the possibility to perform the Sc analysis (ie: as a function of Sc pT)
 - correct branching ratio values of Lc decay channels inserted


arossi81:
+1

PR #13511, opened 2020-03-02
PWGGA: Change Limits for Timing Efficiency Functions to Variables

jluehder:
Instead of using 4GeV and 6GeV as fixed values, it can now be adjusted, when which timing efficiency is applied


dmuhlhei:
+1

PR #13527, opened 2020-03-03
PWGDQ EtaReconstruction Implemented loose secondary cuts before PreFi…

feisenhu:
…lter and apply standard secondary cuts after PreFilter

Implemented a method to optimise the PreFilter.
Use loose secondary cuts to not reject to much particles that can be reconstructed as a pion in the PreFilter. then apply PreFilter and afterwards apply standard cuts to reduce the combinatoric for the four pairing.


miweberSMI:
+test

miweberSMI:
+1

PR #13542, opened 2020-03-04
Fix strategy to identify univocally an event in a run for both MC and Data

ginnocen:
@qgp


ginnocen:
@shahor02 Ciao Ruben, we added the event identification as you recommended. In general it works just fine. But I see some very rare case (~2 MC events in a pretty big test dataset) where I have duplicated indices. I will do some extra checks and let you know. Thanks for your help.

qgp:
+test

qgp:
+test

qgp:
+1

PR #13566, opened 2020-03-05
PreSelectMass method fix

mmazzilli:


PWGHF/vertexingHF/AliRDHFCutsLctopKpi.cxx
```diff
@@ -1533,7 +1531,9 @@ Bool_t AliRDHFCutsLctopKpi::PreSelectMass(TObjArray aodTracks){
       }
     }
   }
-    
+ 
+    if (ptLc > 2.0) vv=true;
```
fgrosa: If (ptLc < 2.0) maybe you could already return false after the pTLc calculation, to avoid the calculation of the masses if you anyway would return false (it would speed up the execution). Do you agree or I missed something?

mmazzilli: Ciao @fgrosa,
I totally agree! Should I make a new commit?

fgrosa: Yes please, it will directly enter in this PR. I would say that you can also avoid the usage of the "vv" flag. Thanks!

fgrosa:
+1

PR #13598, opened 2020-03-07
-Inlcude also Sigma+/- as feeddown candidates

makor:
-Change from Form to more robust TString::Format
-Adapt AddTask in order to accommodate cut variations


bhohlweger:
+1

PR #13605, opened 2020-03-09
Flag to reduce PID binning in reco sparses introduced

mfaggin:
Conservative and "dirty" solution to make the sparse merging easier


arossi81:
+1

PR #13628, opened 2020-03-10
fixed error in physicsSelection for pp TRD triggers

mhartung71:


ekryshen:
Dear Michael, could you please explain what was wrong? It is not obvious to me. As far as I understand, you excluded T-mask and WU-like classes. Why? Was it discussed with TRD community? Do you have any link to some slides on this issue?
Are you sure that you don't need T mask for TRD triggers?
Are you sure you don't need WU-like classes?

ekryshen:
Sorry, Michael, I've just noticed that you added CINT7[HJT|HSE|HQU|HNU]-[B|I|S|T] as a separate group. So the list of trigger classes seems to be the same as before. But I don't understand what was wrong with the previous definition.

mhartung71:
Dear Evgeny,

In 2017 the TRD hardware triggers were redefined and the additional
trigger HNU was introduced.
WU was only used in Run1, since Run2 the TRD was using the LM instead.
So the WU classes are only valid in Run1 data.
With the current version of physic selction, all HNU and most other TRD
triggered events are rejected for Run2 data. This is because
AliTriggerAnalysis.cxx (line 1075 - 1107) tests for Run1 trigger
values.

So wrong fixes from previous touch of the TRD trigger part were
inverted to be compatible to Run1 trigger.
The new implementation in lines 240-244 is needed to use the TRD
triggers ran in 2017 and 2018.

The changes are discussed with my supervisor Benjamin Dönigus, who was
the TRD Trigger Coordinator for Run2 and is also the current TRD
Technical Coordinator.

ekryshen:
Dear Michael, thanks a lot for clarification. You mentioned that lines 1075 - 1107 in AliTriggerAnalysis.cxx are not compatible with Run2 selection. As far as I undestand you simply remove TRDtrigger checks for Run2 classes. Does it mean that you plan to replay these conditions in your user tasks instead of physics selection?

mhartung71:
Dear Evgeny,

yes, this is correct.

Thanks and kind regards,
Michael

ekryshen:
Ok, thanks.

ekryshen:
+1

PR #13632, opened 2020-03-10
PWGDQ EtaReconstruction New structure of support histogram and implemented primary pair cuts

feisenhu:
I made some changes to the output of support histogram lists
and added a new primary pair cut to obeserve and cut on primary pairs observables.


miweberSMI:
+1

PR #13633, opened 2020-03-10
PWG/EMCAL: optimized trigger maker values for pPb8TeV

nschmidtALICE:
From analysis output i found that the values were set slightly too high. I've reduced them by a few hundred MeV.


raymondEhlers:
+1

PR #13638, opened 2020-03-10
Apply NonLinearity correction from YAML only

deepathoms:
Applying non-linearity correction from YAML setting instead of from the task


deepathoms:
+1

PR #13640, opened 2020-03-10
Link against fastjet

qgp:
- link properly against fastjet instead of using ROOT auto-loading
- ensures correct loading of dependencies by linker, not depending on LD_LIBRARY_PATH

NB: still testing ..


qgp:
+test

ktf:
+1

PR #13661, opened 2020-03-12
Mt mult binning

bhohlweger:
Adds the distribution of multiplicity and kstar for several mT bins. Can be activated via the event collection config with SetDomTMultBinning. The bins can be defined in a std::vector<float> via SetmTBins.


bhohlweger:
+test

bhohlweger:
+1

PR #13668, opened 2020-03-12
PWGGA/GammaConv: Modified NonLin settings for EDC pp 13TeV nominal an…

jokonig:
…d low B-field and changed cutstings respectively


dmuhlhei:
+1

PR #13680, opened 2020-03-13
PWGGA/GammaConv: properly derived AliConversionAodSkimTask 

fjonasALICE:
- AliConversionAodSkimTask is now properly derived from the original AliAodSkimTask, including the user exec. From now on, tasks can override the SelectEvent() to influence the selection of events selected for the skim and all copying is handeled by AliAodSkimTask
- intorduced a check that ensures that the V0Reader is not running in the analysis manager during skimming by accident. When relabelling by default with the V0Reader while skimming, the trains running on the skimmed dataset will fail to read the V0 candidates when attempting to relabel again.


nschmidtALICE:
+1

PR #13684, opened 2020-03-13
Activate the built-in AliEventcuts event selection

BongHwi:
The event will be selected by using the built-in AliEventCuts if we turn on "fUseAliEventCuts" flag as true.


mpuccio:
+1

PR #13726, opened 2020-03-18
Modified version of the AliTPCcalibResidualPID task.

mciupek:
Modified version of the AliTPCcalbiResidualPID task, which generate the output for the TPCspline creation framework. A reduced Tree was introduced to the output with the necessary variables to speed up the spline creation.


wiechula:
+1

PR #13729, opened 2020-03-18
 changing the efficiency correction, applying weight normalized. 

crisjahnke:
changing the efficiency correction, applying weight normalized.
Flags added for emcal trigger studies.


miweberSMI:
+1

PR #13744, opened 2020-03-19
PWGGA/GammaConv: Added new decay gamma JJ production for EMCal and DC…

jokonig:
…AL/PHOS (LHC20b1*)


nschmidtALICE:
Please determine the correct pTh weights correctly in a local test. Otherwise the first train will not give meaningful output.

nschmidtALICE:
+1

PR #13754, opened 2020-03-19
New background scheme (swapping method)

mhemmer-cern:
+Added new background scheme (swapping method by Joshua)
+Added new train configs for 13 TeV
+Reordered and redone the 13 TeV train configs
+Fixed mass calculation for a rare case for neutral pions
+Added and reordered some deletes to be in the proper position
+Added lightweight option (still needs work, currently only used in the new background scheme)


PWGGA/GammaConv/AliAnalysisTaskOmegaToPiZeroGamma.cxx
```diff
@@ -2889,6 +2927,8 @@ void AliAnalysisTaskOmegaToPiZeroGamma::CalculateOmegaCandidates()
                 dropOutGammas_CALO.insert(secondGammaIndex);
               }
             }
+            delete pi0cand;
```
dmuhlhei: This line and following ones: if a "new" operation is really not needed try to avoid it -> then we also don't have to worry about proper delete statements

mhemmer-cern: Tweaked the part added by me, to not need new and delete.

dmuhlhei:
+1

PR #13773, opened 2020-03-20
PWGGA/GammaConv: Simplification of bad channel rejection for backgrou…

jokonig:
…nd rotation method

### DIFF
--- a/PWGGA/GammaConvBase/AliCaloPhotonCuts.cxx
+++ b/PWGGA/GammaConvBase/AliCaloPhotonCuts.cxx
@@ -3535,7 +3535,7 @@ Bool_t AliCaloPhotonCuts::CheckDistanceToBadChannelSwapping(const Int_t CellID, 
   }
 
 
-  if(iBadCell != 0) return kTRUE;
+  if(iBadCell > 0) return kTRUE;
   return kFALSE;
 }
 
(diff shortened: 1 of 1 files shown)


PWGGA/GammaConvBase/AliCaloPhotonCuts.cxx
```diff
@@ -3509,153 +3509,33 @@ Bool_t AliCaloPhotonCuts::CheckDistanceToBadChannel(AliVCluster* cluster, AliVEv
...
-        rowdiff = TMath::Abs( largestCellirow - irow ) ;
-        coldiff = TMath::Abs( largestCellicol - icol ) ;
-        //cout << "rowdiff: " << rowdiff << ",coldiff: " << coldiff << endl;
-        if ((coldiff + rowdiff <= fMinDistanceToBadChannelSwapping )) return kTRUE;
-        //cout << "not within distanceToBadChannel!" << endl;
-      }
-    }
-  }
 
+  if(iBadCell != 0) return kTRUE;
```
dmuhlhei: Shouldn't we have iBadCell > 0 here to protect against the initialization of iBadCell = -1?

jokonig: returning true actually means that the cell is bad (I implemented that similar to the existing DistanceToBadChannel function). So this means that if the cell is good iBadCell == 0. And otherwise its bad (so iBadCell !=0 covers the cases where the cell is bad including the initialization of -1).

dmuhlhei:
+1

PR #13776, opened 2020-03-20
Added alpha cut variation train configs

mhemmer-cern:
+Added alpha cut variation train configs for pi0n and omega meson
-Removed some news & delete which were not necessary


dmuhlhei:
+1

PR #13779, opened 2020-03-20
Add fix for eventid in the Run2->Run3 converter for MC and add rejection of spd vertices

ginnocen:
Please do not merge this yet because it is still being tested. 
@pzhristov @jgrosseo. This PR should include the event id calculation for MC where the BC information are not stored and reject the events with only SPD vertex, which as we discuss should not be converted given that they will not be present in Run3. 
I will do a few checks and then remove the WIP. Please let me know if you have any suggestions. 
Thanks Cheers GM

### DIFF
--- a/RUN3/AliAnalysisTaskAO2Dconverter.cxx
+++ b/RUN3/AliAnalysisTaskAO2Dconverter.cxx
@@ -480,6 +480,9 @@ void AliAnalysisTaskAO2Dconverter::UserExec(Option_t *)
   if (!pvtx) {
     ::Fatal("AliAnalysisTaskAO2Dconverter::UserExec", "Vertex not defined");
   }
+  TString title=pvtx->GetTitle();
+  if(pvtx->IsFromVertexer3D() || pvtx->IsFromVertexerZ()) return;
+  if(pvtx->GetNContributors()<2) return;
 
   //---------------------------------------------------------------------------
   // Collision data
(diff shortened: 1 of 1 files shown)


ginnocen:
Hello @jgrosseo @pzhristov. tests are good. It is ready to be merged from my point of view.

RUN3/AliAnalysisTaskAO2Dconverter.cxx
```diff
@@ -490,7 +493,11 @@ void AliAnalysisTaskAO2Dconverter::UserExec(Option_t *)
 
   vtx.fNentries[kEvents] = 1;  // one entry per vertex
   vtx.fRunNumber = fESD->GetRunNumber();
-  vtx.fGlobalBC = GetEventIdAsLong(fESD->GetHeader());
+  ULong64_t evtid = GetEventIdAsLong(fESD->GetHeader());
+  if(!evtid){
+    evtid = (ULong64_t(fESD->GetTimeStamp())<<32) + ULong64_t((fESD->GetNumberOfTPCClusters()<<5)|(fESD->GetNumberOfTPCTracks()));
```
pzhristov: Do we really have to encode there the number of TPC clusters and tracks? How is the time stamp created in MC?

jgrosseo: This is Ruben's idea. The idea is to "randomize" it for the case that two events are produced at the same time on different Grid nodes.

jgrosseo:
+test

ginnocen:
@jgrosseo @pzhristov please don't merge yet. I am observing some discrepancy in the framework, which is almost 100% unrelated to this one. But I want to make sure this is indeed not the case.

ginnocen:
@jgrosseo @pzhristov. As expected the problem was unrelated. Please go ahead with the merging if you think it is ok. Sorry for the false alarm.

RUN3/AliAnalysisTaskAO2Dconverter.cxx
pzhristov: It seems that in the MC simulation we are using all the time stamps are 0. To get different behaviour, one has to set the flag fUseTimeStampFromCDB>0. What if we set it simply to Random covering some period in 2021?

ginnocen:
Hi @pzhristov (I report our discussion here for the others). On the MC productions that I have tested this on, everything was fine but probably cause they were associated to real data (?). There is no problem for me if you have another solution that is safer with these new Run3 simulation. Thanks GM

pzhristov:
+1

RUN3/AliAnalysisTaskAO2Dconverter.cxx
jgrosseo: Or use the current time when the conversion job is running? We could use fESD->GetTimeStamp and fall back to the current time if that is 0...

PR #13785, opened 2020-03-21
update for the lastes optimization

BongHwi:
- Update vertexer configuration
 - Add INELgt0 parameter
 - Add default V0 cut for Nano filter


mpuccio:
+1

PR #13787, opened 2020-03-21
PWGHF: Several small fixes + improvements

lvermunt:
Small fixes to several HF tasks after a direct comparison of the output between the dedicated task, the DvsMultiplicity task, and HFTTreeCreator task. 

Should have no, or a negligible, impact on the ongoing/published analyses. Only when certain flags are enabled in cutobject (fRemoveDaughtersFromPrimary, fUseCutGeoNcrNcl) for Lc tasks, but these are disabled by default.


jezwilkinson:
+test

fprino:
+1

PR #13797, opened 2020-03-23
DJetCorrelations developer copy created: task.cxx, task.h, macros/addtask.c.

MycrofD:
Substructure functionalities added for D-jets


MycrofD:
Hi all, 

just to clarify, there is nothing to be worried about with this PR.There are two changes here. 
1. Minor- This doesn't affect the stable code we have been using. It contains only a few comments, indentation etc. to improve readability and understanding of the code. I did this as I went through the code.
2. Substructure functionality- This functionality has been introduced only on a copy of the code we have been using. 
So, the stable code we use on trains/wagons isn't affected at all. 

Best, Auro

fcolamar:
+1

PR #13798, opened 2020-03-23
Macro for Lc(<-Sc) and Sc analysis updated

mfaggin:
- possibility to study different fit possibilities in data and MC
 - comparison plots with other hadrons (eg: Lc, D0) and PYTHIA updated
 - ingredients to study the yield, efficiency and cross section variation with cuts added


cterrevo:
+1

PR #13804, opened 2020-03-23
  PWGDQ EtaReconstruction Implemented looser primary track cuts before PreFilter

feisenhu:
I added a new cutsetting that can be used before the PreFilter in order to select more particles that can be paired in the PreFilter and reject false identified particles afterwards with an other cutsetting.


rbailhac:
+1

PR #13806, opened 2020-03-23
PWGGA/GammaConv: fixed eta phi calculation for rotation background cl…

jokonig:
…usters
(TLorentz vectors have a differnet phi range than function expects)


dmuhlhei:
+1

PR #13811, opened 2020-03-23
Incorporated the rejection of injected signals on MC datasets

victor-gonzalez:
Also
- conditional execution of the momentum ordered pair track swaps
- discarding MC reco tracks with negative label
- fixes for anomalous int arithmetic
- lambda veto for MC generator level


aohlson:
+1

PR #13824, opened 2020-03-24

laltenka:
- AliReducedCompositeCut: new method using values (like other cut classes)
- AliReducedAnalysisJpsi2ee: methods to return cuts (e.g. useful when setting up mixing handler)
- AliReducedTrackInfo: new data member to hold EMCal cluster energy and corresponding setters/getters
- AliReducedVarManager: assigning EMCal cluster energy to track after cluster-track matching, new variable to store EMCal cluster energy per leg for pair


iarsene:
+1

PR #13846, opened 2020-03-25
Add TimeRangeMask for 18r pass3, add missing run 297481

wiechula:
LHC18r pass3 requires the same time masking for PID as pass1.
Run 297481 was missing, this is now added.
The OADB file as well as the macro to set up the file are added.


chiarazampolli:
+1

PR #13849, opened 2020-03-26
OmegaToPiZeroGamma added PHOS configs

mhemmer-cern:
Added PHOS configs (changed train configs to 4 digits just to be safe)
Fixed bug which disabled the EtaPhiMaps for the swapping background method
Modified standard photon selection range to be what currently looks best


dmuhlhei:
+1

dmuhlhei:
+1

PR #13852, opened 2020-03-26
OmegaToPiZeroGamma added PHOS configs

mhemmer-cern:
Added PHOS train configs
Fixed bug with not filling EtaPhi histogram for swapping background
Moved the include for std::set into the header


dmuhlhei:
+1

dmuhlhei:
+1

PR #13854, opened 2020-03-26
Add dir to make unique IDs

hzanoli:
In some cases (~3% of data), the old scheme to generate the unique event numbers was not working, because there were too many events in the AOD. To fix that, I have added the number of the directory as a separate variable, and then the unique ID can be generated by combining the RunNumber + Directory + Event Number.


hzanoli:
@fcolamar or @btrzecia could you please have a look? I would like to try to get the tag of today.

Best
Henrique

btrzecia:
+1

PR #13866, opened 2020-03-28
fix-Reverting the "fArrayDStartoD0pi"-renaming in DFilter task

MycrofD:
Hi all,

"fArrayDStartoD0pi" was renamed recently in DFilter task to check how it affected. And this affected the Container input provided to AliEmcalJetTask. So fixing this mistake I did few days back. Another renaming in Correlations task is also reverted just to be safe.

Best, Auro


MycrofD:
+1

btrzecia:
+1

PR #13882, opened 2020-03-30
excluding EG2 events from EG1 trigger -- nothing done for EG2, where …

crisjahnke:
excluding EG2 events from EG1 trigger since it has 10% of overlap -- nothing done for EG2, where we use all events. 
1) EG2 we use all events
2) EG1 we need to remove EG2 (since it was already counted in case 1))


iarsene:
+1

PR #13891, opened 2020-03-31
PWGDQ EtaReconstruction Implemented PreFilter to reject pions from tw…

feisenhu:
…o photons

Option to apply a prefilter in order to reject secondary pairs that originate from a decay of a pion into two photons.


rbailhac:
+1

PR #13906, opened 2020-04-01
PWGGA/GammaConv: Adjustments for pp 13TeV pT binning and light output

jokonig:
Additionally, a check if the JJ weight is nan was implemented.


dmuhlhei:
+1

PR #13916, opened 2020-04-02
Including variables related to the DCA resolution for XY plane and Z

hfranzde:
Including the variables related to the DCA resolution for XY plane and Z


hfranzde:
Dear all,

Please, find attached the histogram of the " kImpactParXYres" variable I
had yesterday when testing with PARfiles including those committed
modifications. Everything looks fine to me, but please, let me know if
everything is ok about this way it is written or if I forgot something.

Thanks in advance!
Hermann

Em qui., 2 de abr. de 2020 às 00:30, ALICE Builder <notifications@github.com>
escreveu:

PR #13925, opened 2020-04-03
Including new variable related to DCA resolution and updating AddTask and Config files

hfranzde:
Including new variable related to DCA resolution in XY plane and Z and updating my own AddTask and Config files to run with these new variables


miweberSMI:
+1

PR #13929, opened 2020-04-03
pWGGA/GammaConv: Added new train configs for pp 13 TeV, added Test Be…

jokonig:
…am NonLin without fine tuning, implemented AliFatal in case of JJ weight equals to nan or zero for debuging


dmuhlhei:
+1

PR #13930, opened 2020-04-03
Including the background for the analysis

gmantzar:
Opening the Cuts to include the background of the Lambdas Xis. Also correct renaming of some variables to avoid overwriting.


a-mathis:
+1

PR #13933, opened 2020-04-03
Fix for electron MC 

hzanoli:
Restore MC information for all HFE and selected tracks. Adds dir number to the events an improves a few parts of the code.


btrzecia:
+1

PR #13947, opened 2020-04-05
Fix negative EZNA EZNC bug and add two related histograms to record t…

ShiQiu1:
Fix negative EZNA EZNC bug which cause problem in Math::Power(EZNA/EZNC, fZDCGainAlpha) and add two related histograms to record the # of occurrence of negative EZNA and EZNC.


aohlson:
+1

PR #13968, opened 2020-04-07
PWGPP-550, ATO-465 - phi,tgl switch - BUG FIX

miranov25:
### PWGPP-550, ATO-465 - adding barrel multiplicity and sum pt counters - bug fix
phi and tgl were switched in commits before


miranov25:
Hello @ktf, @chiarazampolli 

Looks like a problem with build server. Could somebody check?
Thank you

ktf:
Looks like it solved itself.

PR #13977, opened 2020-04-08
Update of the Config and AddTask files

hfranzde:


otonvd:
+1

hfranzde:
Dear all, I am sorry but I can't find where the error is by looking at this output log. In this log we have errors that seems not to be related to my task, but I may not completely understand it. I did several tests on grid with this new task before commiting it and everything worked as supposed to. Does anyone have any suggestions?

PR #13983, opened 2020-04-08
DRAFT: Reject AOD events for which the relabeling of photon candidate…

sstiefel19:
…s fails (DATA & MC)

This change impacts only scenarios where the relabeling of AOD photon
candidates is performed within the AliV0Reader and NOT within the actual
AliAnalysisTask.

General info: After an event has passed the V0Reader's event selection,
the photon candidates are getting imported from the delta AOD file and
relabeled. This means that for every photon (skipping the candidate from now on)
from the delta AOD file its daughter tracks are tried to match with
two tracks from the set of all tracks within the event and relabeled on
success.

Before this change: If for a photon the relabeling fails (i.e. that
 the photon's daughter tracks can't be found within the event), then
its daughter tracks are set to negative and the photon will create
an entry under 'no tracks' in the
GammaConvV1_...->ConvCuts_...->IsPhotonSelected histogram and there
will be an error message. The event will not get rejected. The log
can get big since the error is shown for each photon where it fails.

With this change: If for a photon the relabeling fails, it will create
an entry in the same histogram as above. In addition, the quality of the
event will be set to 2 such that all AnalysisTasks which check the event
quality (I guess all) will reject this event. This will become visible
in the GammaConvV1_...->Cut Number...->...ESD Histograms->NEvents
histogram under 'Miss MC or inc. ev.' An error is shown at maximum once
per event (even less since it would be identical AliErrors)
Moreover, the filename of the current AOD file be added to the tree
of broken files. (Only for the GammaConvV1 task with this change)

The problem with failing relabeling was observed at GA_PbPb_AOD_MC
trainrun 1544, file /alice/sim/2019/LHC19h2a/297481/AOD/027/
also see: https://alice.its.cern.ch/jira/browse/ALIROOT-8304

### DIFF
--- a/PWGGA/GammaConv/AliAnalysisTaskGammaConvV1.h
+++ b/PWGGA/GammaConv/AliAnalysisTaskGammaConvV1.h
@@ -408,7 +408,7 @@ class AliAnalysisTaskGammaConvV1 : public AliAnalysisTaskSE {
 
     AliAnalysisTaskGammaConvV1(const AliAnalysisTaskGammaConvV1&); // Prevent copy-construction
     AliAnalysisTaskGammaConvV1 &operator=(const AliAnalysisTaskGammaConvV1&); // Prevent assignment
-    ClassDef(AliAnalysisTaskGammaConvV1, 48);
+    ClassDef(AliAnalysisTaskGammaConvV1, 49);
 };
 
 #endif
--- a/PWGGA/GammaConv/AliAnalysisTaskGammaConvV1.h
+++ b/PWGGA/GammaConv/AliAnalysisTaskGammaConvV1.h
@@ -402,6 +402,7 @@ class AliAnalysisTaskGammaConvV1 : public AliAnalysisTaskSE {
     Bool_t                            fDoMaterialBudgetWeightingOfGammasForTrueMesons;
     TTree*                            tBrokenFiles;                               // tree for keeping track of broken files
     TObjString*                       fFileNameBroken;                            // string object for broken file name
+    Bool_t                            fFileWasAlreadyReported;                    // to store if the current file was already marked broken 
 
   private:
 
--- a/PWGGA/GammaConvBase/AliV0ReaderV1.cxx
+++ b/PWGGA/GammaConvBase/AliV0ReaderV1.cxx
@@ -1244,45 +1244,48 @@ Bool_t AliV0ReaderV1::GetAODConversionGammas(){
 
   AliAODEvent *fAODEvent=dynamic_cast<AliAODEvent*>(fInputEvent);
 
-  if(fAODEvent){
-
-    if(fConversionGammas == NULL){
-      fConversionGammas = new TClonesArray("AliAODConversionPhoton",100);
-    }
-    fConversionGammas->Delete();//Reset the TClonesArray
-
-    //Get Gammas from satellite AOD gamma branch
-
-    AliAODConversionPhoton *gamma=0x0;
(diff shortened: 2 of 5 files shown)


amaringarcia:
Hi,
this change for relabeling may need be be also ported to any other task using converted gammas. or? like convCalo and Dalitz, or Omega,.... It just came to my mind when seeing the list of at the committed files.
Ana

sstiefel19:
Hi Ana,
that part, that adds the filenames to the tree with broken filenames, I guess we would want to port to other tasks too, yes. The rejection of events itself would already be enabled with this change for any other task which checks the eventQuality  from the V0reader (which I guess are all)

sstiefel19:
Another issue would scenarios, where the relabeling happens inside the AnalysisTask and not within the V0reader. But I guess that should be addressed in an extra effort.

amaringarcia:
If the removal of event happens at the V0Reader , it should be enough....

PWGGA/GammaConvBase/AliV0ReaderV1.cxx
```diff
@@ -1262,14 +1263,20 @@ Bool_t AliV0ReaderV1::GetAODConversionGammas(){
...
+          if(fRelabelAODs) {
+            relabelingWorkedForAll &= RelabelAODPhotonCandidates(gamma);}
+          if(fConversionCuts->PhotonIsSelected(gamma,fInputEvent)){
+            new((*fConversionGammas)[fConversionGammas->GetEntriesFast()]) AliAODConversionPhoton(*gamma);}
         }
       }
+      if (!relabelingWorkedForAll){
+        AliError("For one ore more photon candidate the AOD daughters could not be found. The labels of those were set to -999999 and the event will get rejected.");
+        return kFALSE;
+      }
```
dmuhlhei, line 1279: What if  "GetEntries()" in L 1276 returns 0 - meaning we have no V0 candidates? In your fix, this case will also be flagged with event type 2, which must not be the case in my opinion. Could you please check this? We need to be really careful here...

sstiefel19: Good catch, thanks :) I agree, if we have no V0 candidates (or none of them is passing the cut) we do not necessarily want to flag the event with type 2

PWGGA/GammaConv/AliAnalysisTaskGammaConvV1.h
```diff
@@ -408,7 +408,7 @@ class AliAnalysisTaskGammaConvV1 : public AliAnalysisTaskSE {
 
     AliAnalysisTaskGammaConvV1(const AliAnalysisTaskGammaConvV1&); // Prevent copy-construction
     AliAnalysisTaskGammaConvV1 &operator=(const AliAnalysisTaskGammaConvV1&); // Prevent assignment
-    ClassDef(AliAnalysisTaskGammaConvV1, 48);
+    ClassDef(AliAnalysisTaskGammaConvV1, 49);
```
sstiefel19, line 411: I just see this has increased in the meantime, so I guess I need to increase it once more?

dmuhlhei:
Ok, I'm fine with the PR. 

@FriederikeBock and/or @nschmidtALICE  could you please have a look as well and approve this request? We should double check these changes

nschmidtALICE:
PR also looks okay from my side.

nschmidtALICE:
+test

nschmidtALICE:
+1

PR #13989, opened 2020-04-09
Start moving Hypertriton analysis to NanoAODs

fmazzasc:


mpuccio:
+1

mpuccio:
Ciao @ktf do you have any idea of what happens here?

PR #13992, opened 2020-04-09
SPD tracklets RUN3 conversion

valzacc:
Adding SPD tracklets to AO2D RUN3 converter


valzacc:
Hello, can somebody with approving rights +test this?  @jgrosseo,  @qgp? Thanks!

pzhristov:
Hi @valzacc, do we really need the tracklets in their present form? If I remember correctly, the idea was to convert them to the external track parameter form (which probably means some fake 1/PT) and to add field for the track typ. In this way we emulate the ITS standalone tracks in Run3.

valzacc:
Hello @pzhristov,
sorry the PR is a bit dynamic in this moment as it is connected to my fork, if you prefer I can close it and reopen it on a later stage.

PR #13995, opened 2020-04-09
Adding the possibility to remove kaons coming from phi decay (MC pdg …

zkhabano:
…check) within the data-driven invariant mass selection


altsybee:
+1

PR #14003, opened 2020-04-10
Add THnSparse also for Pseudodata

mfasDa:
Add THnSparse in order to be able to get the truth level
distribution with detector level cuts. Necessary to
distinguish between effects of unfolding and kinematic
efficiency in the MC closure test


mfasDa:
+1

PR #14012, opened 2020-04-11
Revert "theta, E added to THnSparse"

MycrofD:
Reverts alisw/AliPhysics#13976  - (1/3)

Current D-jets code is broken. And the issue cannot be located. So going back to the point of stable code. This is the first reversal of the three PRs created that supposedly broke the code.


btrzecia:
+1

PR #14014, opened 2020-04-11
Revert "fix-Reverting the "fArrayDStartoD0pi"-renaming in DFilter task"

MycrofD:
Reverts alisw/AliPhysics#13866  - (2/3)

Current D-jets code is broken. And the issue cannot be located. So going back to the point of stable code. This is the second reversal of three PRs that supposedly broke the code.


btrzecia:
+1

PR #14030, opened 2020-04-13
Define fGlobalBC consistently for collision and trigger table

ginnocen:
@pzhristov @jgrosseo. Shouldn't we use for consistency the new definition of the fGlobalBC also for the trigger table? I don't think we emulate the trigger decision for MC at this moment so I dont think it currently makes any difference. But I was wondering if for consistency we should update the definition also in this case. Thanks for your feedback and sorry in advance if I misunderstood the problem here. GM


ktf:
This PR is from 8 months ago and does not merge. Closing.

PR #14050, opened 2020-04-15
Modified Rotation Background Method for OmegaToPiZeroGamma

mhemmer-cern:
- Split rotation background method up into Calo and PCM part
so it should work with both properly now
- Added photon selection to the background calculation
Therefore the sets which contained the preselected gamma indices
are now a variable of the AliAnalysisTask.
- Added more trainconfig to now be able to run EMCal or DCal alone
- Added one more trainconfig to be able to run normal and strong photon
selection for EDC-EDC at once


dmuhlhei:
+1

PR #14060, opened 2020-04-16
PWGDQ EtaReconstruction: Fixed issue with V0 selection, now only OnTh…

feisenhu:
…eFly V0 are selected

I corrected a comment in the AliDielectronMC.cxx file.

Fixed Issue in selection of V0. Status of V0 Finder was not set. Now it only selects OnTheFly V0's.


miweberSMI:
+1

PR #14074, opened 2020-04-17
Adding Cent and SemiCent trigger option

deepathoms:
Adding Central and SemiCentral trigger option for 2018 Pb-Pb data


deepathoms:
+1

PR #14077, opened 2020-04-17
 PWGDQ EtaReconstruction, Implemented rejection of tracks after prefilter

feisenhu:
I implemented  the rejection of tracks in the particle vectors, inside of the prefilter.
Some further clean up of the code.


miweberSMI:
+1

PR #14090, opened 2020-04-19
FemtoNanoLD: change cuts, FB 256, ITS PID

stheckel:
o Change default cuts including strict DCA
o 1: with FB 256 instead of FB 128
o 2: in addition to 1 with ITS PID up to 1.4 GeV/c


a-mathis:
+1

PR #14097, opened 2020-04-20
Updated framework for track eff reweighting for D-h correls

fcolamar:


fcolamar:
Hi all, do you know if the list of people with commit rights has changed?
Up to now, as PAG coordinator I had rights to commit in PWGHF/correlationHF folder, but it seems that now an external approval is required for this commit (and I also don't see the other HF people with commit rights in the list)
Thanks!
Fabio

qgp:
+1

PR #14098, opened 2020-04-20
PWGGA/GammaConv: added new IsoTree task & fixed bug in conversion skim

fjonasALICE:
- added task that creates a tree used to analyze isolated photons
- fixed a bug in the conversion skim task, that caused skimming for photons above phimax


nschmidtALICE:
+test

nschmidtALICE:
+1

PR #14112, opened 2020-04-21
Adding residuals for SoftDrop observables

mfasDa:
Residuals for zg, Rg, thetag and nsd monitored
vs part. level pt absolute and normalized by
the part level observable


mfasDa:
+1

PR #14118, opened 2020-04-22
OmegaToPiZeroGamma no secondary check

mhemmer-cern:
- The check for secondaries takes a good amount of time and considering
that there does not seem to be many secondaries (~0.1%)
this is commented out for now


dmuhlhei:
+1

PR #14124, opened 2020-04-22
Updating RUN3 converter to updated O2 data model

jgrosseo:
- All names made canonical (O2 var names correspond to f+branch name)
- Rename VZERO
- Trigger becomes BC table
- Primary key is BCID for most tables


jgrosseo:
+test

jgrosseo:
+1

PR #14133, opened 2020-04-23
Update to HF CF to allow online trkl mult equalization & event mult reweigting for pp@13 TeV samples

fcolamar:
Code is backward compatible to preserve standard usage.
Two new arguments are added at the end of the AddTask args list to enable new functionalities


cterrevo:
+1

cterrevo:
+test

PR #14169, opened 2020-04-27
PWGGA: Change of PHOS Tender

jluehder:
Triggered BC map of PHOS grp has different main directory name in .root file


dmuhlhei:
+1

peressounko:
Goal of commit not clear. No physics case to put trigger bad map into readout bad map. To be clarified

chiarazampolli:
Hello @peressounko ,

So we don't approve for now?

Chiara

peressounko:
Hi @chiarazampolli , 
we will chat with Jens then decide. I will approve it necessary. Dmitri

PR #14170, opened 2020-04-27
Adding QA histograms to SD and Spectrum tasks

mfasDa:
- Jet area (vs. pt, NEF, Nconst)
- (Max.) constituent Eta-phi
- (Max.) constituent distance to the main jet axis
In case of response tasks:
- Histograms for both part. and det. level jets
- Histograms for Matching deltaR between part.
  level jet and det. level jet


mfasDa:
+1

PR #14207, opened 2020-04-30
Adding Run2 SPD tracklets to Run3 converter

valzacc:
Adding eta and phi tracklets conversions to alpha, snp and tgl global track parameters to Run2 SPD tracklets to Run3 converter


jgrosseo:
+1

jgrosseo:
+1

PR #14223, opened 2020-05-01
New component to clone the cells branch on demand

gconesab:
Needed for studies with multiple cross-talk emulation settings, but it can be used for example to check the effect of different calibrations in one go


gconesab:
+test

gconesab:
I profitted to push a fix for another problem I found.  Please @raymondEhlers, have a quick look and approve so that this is available in today's tag

raymondEhlers:
+1 I only reviewed this quickly, but looks okay to me. Thanks! (There are a few typos in the docs, but they don't matter enough to delay approving it)

PR #14257, opened 2020-05-05
Upgrade Omega MBW to 2D correction R, pT; omega MBW is still 1D corre…

amaringarcia:
…ction but stored in 2D profile


dmuhlhei:
+1

PR #14262, opened 2020-05-06
Assign the cut on TPC tracks - VZERO correlation to the bit kTPCPileup

fprino:

PR #14264, opened 2020-05-06
Add case with training w/o signd0

andreaalici:


fgrosa:
+1

andreaalici:
Dear,
I cannot see any log explaining what is going wrong. It's the problem related to my commit only?
Thanks.
andrea

fgrosa:
Hi @andreaalici, I think that it is a general issue in AliPhysics not related to your commit, which looks good to me (indeed all the PR are failing without logs). Looking at other conversations it looks to me that Giulio is already aware of this https://github.com/alisw/AliPhysics/pull/14262#issuecomment-624515850

PR #14273, opened 2020-05-06
PWGGA/GammaConv: reenable rejection of events for which aod relabelin…

sstiefel19:
…g fails

but reject from within the AnalysisTask and not the V0Reader. This is
necessary since otherwise other AnalysisTasks, which do not have
conversion photons, would miss these events unnecessarily.


nschmidtALICE:
+1

PR #14303, opened 2020-05-10
Updated Run3 labels:

pzhristov:
- One label with the usual nomenclature:
positive - no mismatch, negative - mismatch;
- In addition a bit mask which encodes in the folllowing way which detector
has the mismatch:
a) ITS: 7 bits (one per layer)
b) TPC: 3 bits to count number of mismatch clusters encoding
the ranges: 0, 1, 2-3, 4-7, 8-15, 16-31, 31-63, >64
c) TRD: 1 bit (possibility to use 6 bits in the future)
d) TOF: 1 bit


pzhristov:
+test

pzhristov:
+1

PR #14305, opened 2020-05-10
add vertex histograms for c deuteron

jaimenorman:
- add vertex histograms and fill if running c-deuteron analysis
- add option, if running c-deuteron analysis, to only use tracks that come from signal (saves a lot of CPU time), and to remove injected particles


fgrosa:
+1

fgrosa:
+1

PR #14310, opened 2020-05-11
PWGGA/GammaConv: Added NLM variations for MergedClusterTask

jokonig:
Modified handling for energy splitting for clusters with NLM > 1


nschmidtALICE:
+1

PR #14318, opened 2020-05-12
PWG/Cocktail/parametrisations new PbPb PCMPHOS para added to cocktail…

mikesas:
… file


mikesas:
Additionally to the PCMPHOS spectra being parametrized, also the combined spectra that are preliminary are added. The PCMPHOS parametrization of the eta is using the combined spectrum.

PR #14319, opened 2020-05-12
PWGGA/GammaConv: GammaIsoTree added cell based iso & GJ weights

fjonasALICE:
- added cell experimental cell based isolation to GammaIsoTree
- added functionality to set cluster cuts separetly for signal clusters and background clusters (those use for isolation and taggin
- added functionality to calculate isolation for multiple radii on the fly
- added weighte for LHC17g5a1 GJ MC pp 8 TeV


dmuhlhei:
+1

PR #14320, opened 2020-05-12
Fix track extraction

raymondEhlers:
The JetConstituent classes do not store subtracted constituents. So
instead we have to access them via `Track(int)`.


raymondEhlers:
+test

raymondEhlers:
+1

raymondEhlers:
+1

raymondEhlers:
@ktf It appears that the root6 builds are stuck. Can you please have a look?

PR #14330, opened 2020-05-13
Store subtracted constituents in jet constituents

raymondEhlers:
When using constituent subtraction, the jet constituents should store
the subtracted constituents (when provided). This is the expected
behavior, and mirrors that of `AliEmcalJet::Track(int)`

@mfasDa Please take a look and let me know what you think. I stopped using these classes, but I hope this will others avoid the same issues that I had - it's easy to miss. I tested for charged constituents and it appears to work. However, I don't see a way to make an equivalent construction for charged constituents at the moment. I raise an exception rather than allow this to pass unnoticed. But suggestions are appreciated!

### DIFF
--- a/PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
+++ b/PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
@@ -566,9 +566,6 @@ void AliEmcalJetTask::FillJetConstituents(AliEmcalJet *jet, std::vector<fastjet:
         AliError(Form("Could not find track %d",tid));
         continue;
       }
-      if(fFillConstituents){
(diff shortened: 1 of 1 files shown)


raymondEhlers:
+test

mfasDa:
HI @raymondEhlers, with vscode it is rather trivial to find this out. Indeed there are a couple of users (AliAnalysisTaskNewJetSubstructure - I think this is @lhavener and @lcunquei, AliAnalysisTaskEmcalJetCorrection from Ruediger - perhaps used now by @hbossi, and task from you and me). As your changes affect mostly PbPb the corresponding users should be made aware of this, not to create confusion or double application.

raymondEhlers:
+1 After discussing privately with a few people, I'm merging this as I don't think it affects anyone else (nor myself at this point)

PR #14334, opened 2020-05-13
[PWGGA/CaloTrackCorr] Adaptations to embedding analysis with EMCal embedding framework

gconesab:


gconesab:
Hi @ktf 

can you have a look? I do not think it is normal that AliYAMLConfiguration.h is not found.

Thanks ... Gustavo

PR #14338, opened 2020-05-14
Code to reproduce Simone's Result

aga01935:
Analysis code to reproduce Simone's result on J/Psi Polarization, these macro only save muon tracks and the muon pair in terms of TLorenz Vector


mbroz84:
+1

PR #14345, opened 2020-05-14
Additional options for outlier cut

mfasDa:
- Cross check whether std::max_element gives the same
  result as trivial iteration over the accepted jets
  by hand (debug mode - energy scale task only)
- Add switch performing the outlier cut either on
  part. level jets or det. level jets (all tasks running
  on MC), default: part. level jets


mfasDa:
+1

PR #14379, opened 2020-05-19
fix compilation warnings in CF and yaml-cpp

dmuhlhei:


dmuhlhei:
Anyone?

phuhnGithub:
+1

dmuhlhei:
Anyone from EMCal?

gconesab:
+test

gconesab:
+1

dmuhlhei:
sorry @gconesab you again need to +1 - it seemed to be a little too early

raymondEhlers:
Hi, have you tested the changes to ensure it doesn't change yaml-cpp behavior? It's an external package, so I'm not super inclined to modify it unless we're taking changes from upstream

dmuhlhei:
I don't see how this can be the case - nothing of its functionality is touched

raymondEhlers:
I think I was too short - apologies. I don't mean to discourage you - I'm generally very much in favor of fixing warnings.

My concern is that if yaml-cpp relies on some undefined behavior that this fixes (which in general would be a good thing to fix), it might change the behavior for us. I have some yaml test code, but not the time to run it right now. Post HP, we could do this if you remind me.

Regarding the copy constructor and equality operator in the CorrFW, I think that would be a nice change. Where I'm a bit rusty: if they're not defined, does the compiler automatically define them? With the change, I'm fairly certain that it does (which be fine for this class, because we just want to copy the pointer address). If you can split out that change, we can add it now. Or we can wait until post HP for everything - up to you

PR #14385, opened 2020-05-20
[PWGGA/CaloTrackCorr] Add sum pT in cone vs photon pT at the generator level

gconesab:


gconesab:
Hi @ktf 
can you check why tests are failing?

PR #14388, opened 2020-05-20
PWGGA/GammaConv: IsoTree switched to TClonesArrays and fixed flush

fjonasALICE:
- switched from std::vector storage to TClonesArray in order to achieve proper splitting and compression in stored tree and proper viewing with tbrowser
- switched structs that store additional infos for particles to a little helper class that inherit from TObject and are easier to maintain and upgrade
- fixed a bug that caused the buffer of theTTree to not be properly flushed to disk and overflow when running over too many events


fjonasALICE:
don't approve yet. Building fails when checking the libs. Must have added the new classes in a wrong way ..

fjonasALICE:
fixed!

dmuhlhei:
+1

PR #14392, opened 2020-05-20
PWGGA/GammaConv/: introduce SDD+SSD Pileupcut

sstiefel19:
also move fHistoV0MultVsNumberTPCoutTracks from AnalysisTasks
to AliConvEventCuts


PWGGA/GammaConvBase/AliConvEventCuts.h
```diff
@@ -660,6 +670,8 @@ class AliConvEventCuts : public AliAnalysisCuts {
       Int_t                       fPastFutureRejectionHigh;               ///< sets bunch crossing event rejection in future. If both are 0, the cut is not applied
       Int_t                       fDoPileUpRejectV0MTPCout;               ///< reject event if # TPCout tracks does not follow expected V0M mult
       TF1 *                       fFPileUpRejectV0MTPCout;                ///< Pol1 function to compute the cut
+      Bool_t                      fRemovePileUpSDDSSDTPC;                 ///< reject event if too many TPC clusters with respect to SDD+SSD clusters 
+      TF1 *                       fFPileUpRejectSDDSSDTPC;                ///!<! Pol2 function to compute cut
```
sstiefel19, line 674: actually wanted to use //!<! here since according to my understanding, ///< would mean streaming the pointer and not the actual object and streaming the pointer would not be of use. Please correct me if Im wrong and ///< streams the object and that this is what is needed.

dmuhlhei: I think you need to use //< - when the AddTask is processed this TF1 is created, hence it has to be streamed

sstiefel19: if I want to stream the object a pointer points to, dont  I have to use "//->" ? At least thats what I understand from https://alice-doc.github.io/alice-analysis-tutorial/analysis/streamer.html

Another thing that confuses me: why is "///<" the right thing to do for fFPileUpRejectV0MTPCout? Again, since it is a pointer, I would have thought one needs to use "//->"
Or does "///<" also stream the object on which the pointer pointer points to?

dmuhlhei: You should use //< .

 "The string “->” in the comment field of the members *fH and *fTracks instruct the automatic Streamer to assume these will point to valid objects" which is not always the case here, isn't it?

sstiefel19: yes you are right, //-> is only for the case where the pointer for sure points to valid objects which is not the case here. I had overlooked the part in the root documentation where it sais that without further ado, the object behind a pointer data member is streamed, my bad. Thanks for explaining :)

PWGGA/GammaConvBase/AliV0ReaderV1.cxx
```diff
@@ -121,6 +121,7 @@ AliV0ReaderV1::AliV0ReaderV1(const char *name) : AliAnalysisTaskSE(name),
   fProduceImpactParamHistograms(kFALSE),
   fCurrentInvMassPair(0),
   fImprovedPsiPair(3),
+  fSDDSSDClusters(-1),
```
dmuhlhei, line 124: this will throw a warning (reorder), please fix

PWGGA/GammaConvBase/AliConvEventCuts.h
```diff
@@ -736,6 +747,8 @@ class AliConvEventCuts : public AliAnalysisCuts {
       Int_t                       fNSpecialSubTriggerOptions;
       TH2F*                       hSPDClusterTrackletBackgroundBefore;    ///< SPD tracklets vs SPD clusters for background-correction before cut
       TH2F*                       hSPDClusterTrackletBackground;          ///< SPD tracklets vs SPD clusters for background-correction
+      TH2F*                       hV0MultVsNumberTPCoutTracks;            //!<! correlation V=Mult vs number TPC out Tracks
+      TH2F*                       hTPCSDDSSDClusters;                     //!<! x: TPC clusters, y: SDD+SSD clusters
       // trigger information
       TString                     fV0ReaderName;                          ///< Name of V0Reader
       TString                     fCorrTaskSetting;                       ///< Name of Corr Task Setting
```
dmuhlhei, line 754: classdef needs update

PWGGA/GammaConvBase/AliV0ReaderV1.cxx
sstiefel19: done

PWGGA/GammaConvBase/AliConvEventCuts.h
sstiefel19: done

dmuhlhei:
@sstiefel19 sorry, for further processing of this request we need a rebase

dmuhlhei:
+1

PR #14405, opened 2020-05-24
Updated MC labels.

pzhristov:
From lbl back to label. Using UInt_t for fLabel and bit 15 of fLabelMask to keep the sign. Adding mccalolabel.


pzhristov:
+1

PR #14426, opened 2020-05-26
OmegaToPiZeroGamma: AODMCTrackArray member variable

mhemmer-cern:
- Added AODMCTrackArray as a member variable to the OmegaToPiZeroGamma task
- Changed MC True binning to match the normal same event binning
- AP like Plots now only get filled once per decay as they should


dmuhlhei:
+1

PR #14427, opened 2020-05-26
Downscale correction, Thresholds from CDB

mfasDa:
- Add histograms with downscale-corrected counts
- Take trigger threshods from OCDB instead of hard-coded thresholds


mfasDa:
+1

PR #14431, opened 2020-05-26
 PWGDQ EtaReconstruction Implemented methode to reconstruct etas from two gamma conversions

feisenhu:
In the new commit an additional method is implemented in order to measure and reconstruct two photo conversions from an decayed eta meson.


miweberSMI:
+1

PR #14438, opened 2020-05-27
PWGHF: Add unc. efficiency to total systematic

lvermunt:
Adding the relative statistical uncertainty of the prompt efficiency to the total systematic TGraph (gSigmaCorr). For most D meson analyses minor contribution (~1%), but can go up to ~5% for Lc in a D2H MC or in special HM MCs.

Now pp is also consistent with PbPb, where it is done here: https://github.com/alisw/AliPhysics/blob/66cc83e/PWGHF/vertexingHF/macros/ComputeDmesonYield.C#L996-L1007


fgrosa:
+1

PR #14508, opened 2020-06-03
Add option to cut on the number of non-0 FastORs

mfasDa:
Internally using a data grid with the fake FastOR
data filled for all cells with a cell time cut.
Number of contributing FastORs calculated from
looping over the data grid within the patch ranges.


mfasDa:
+1

PR #14514, opened 2020-06-04
Added AD timing and muon UPC triggers

ekryshen:
* AD (FDD) table with ADA and ADC timing
* Bypass of vertex cuts for muon UPC triggers with at least one muon track
* Selected trigger classes with index>50 added in bits 50-58 of fTriggerMask

### DIFF
--- a/RUN3/AliAnalysisTaskAO2Dconverter.cxx
+++ b/RUN3/AliAnalysisTaskAO2Dconverter.cxx
@@ -104,6 +104,7 @@ AliAnalysisTaskAO2Dconverter::AliAnalysisTaskAO2Dconverter(const char* name)
     , mucls()
     , zdc()
     , vzero()
+    , fdd()
     , v0s()
     , cascs()
 {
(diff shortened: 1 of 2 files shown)


pzhristov:
@ekryshen Do you foresee any truncation of the float mantissa? If yes, would you add the corresponding part as it was done for the tracks and calo clusters?

jgrosseo:
+test

jgrosseo:
@ekryshen Thanks! Please add what Peter has asked for, then we can merge this.

ekryshen:
@jgrosseo, @pzhristov Thanks for checking. Regarding float truncation, I have to discuss with AD experts

pzhristov:
@ekryshen As far as I see we store zeros for the amplitude, did you miss a call to fill the array? What concerns the truncation, we can probably use the number of bits in the ADCs and TDCs (we also may add one bit to be on the safe side).

ekryshen:
@pzhristov Indeed, I don't store the amplitude array for the moment. Actual AD array has 16 floats (there are 2 layers of 4 sectors per side) while we reserved 8 floats in O2 (may be having in mind to store merged amplitude from 2 layers?). I am not sure how to deal with it now. Since this amplitude info is not used in UPC studies anyway, I just left zeroes. 

Regarding truncation, I discussed it with Michal and Christoph. Typical time resolution of AD TDCs is about 0.1 ns (HPTDC unit was 25/256 ns). We should resolve ~60 ns (typical AD times) with 0.1 ns precision, so I would propose to use 0xFFFFF000 mask, e.g.:
60.1 = 1.111 0000 0110 0110 0110 0110 * 2^5 -> 1.111 0000 0110 * 2^5
60.2 = 1.111 0000 1100 1100 1100 1101 * 2^5 -> 1.111 0000 1100 * 2^5

Do you agree?

pzhristov:
@ekryshen Yes, 0xFFFF0000 is Ok. Please add it as separate flag in the begining of UserExec as it is done for the other truncated values.

ekryshen:
@jgrosseo, @pzhristov Done. I hope it is ready to be merged.

pzhristov:
+1

PR #14538, opened 2020-06-05
PWGGA: remove ! in //!

jluehder:
Change //! to //


kharlov:
Hi Jens, why do you change? Do you want to add these data members to the streamer to write to the root file for AnalysisManager? If these variables are transient and used only during analysis, their streamer is not needed.
Yuri

dmuhlhei:
Hi Yuri, yes these members should be streamed in general - we think we made a mistake here because the task is not working as expected on the grid

kharlov:
Usually I avoid writing unnecessary variables of analysis task to analysis manager. For this, all data members should be properly initialized in all constructors, including default ctor. Anyway, if you prefer writing them to a root file, I approve.
+1

kharlov:
+1

PR #14542, opened 2020-06-05
PWGGA/GammaConv: IsoTree track matching additions and fixes

fjonasALICE:
- for charged isolation, tracks originating from conversions are now properly detected. I noticed the same problem in the ClusterQA task and fixed it
- implemented a mode that allows to use the default track matching
- implemented track matching for secondaries


dmuhlhei:
+1

PR #14574, opened 2020-06-08
 adding a log meesage in case the fAODMCTrackArray is NULL

amaringarcia:


nschmidtALICE:
Thanks for adding it +1

nschmidtALICE:
+1

PR #14596, opened 2020-06-09
Added Task for He4 & triton analyses

sadkoeni:
Tasks to analyse He4 and triton in HM pp. Particularly triton to cross check absorption of He3 with isospin partner.


mkruegerGitHub:
+1

PR #14630, opened 2020-06-11
control histogram restricted to 3-body hypertritons

galocco:
In the previous version the pT and ct information of the generated hypertritons in 3-body were lost. The ITS/TPC request on the daughter tracks is removed.


lbariogl:
+1

PR #14643, opened 2020-06-12
First triplet approach try

lauraser:
Moved triplet creation to a function, made it general - now can create triplets for any particles.


a-mathis:
+1

PR #14645, opened 2020-06-12
deuterons - better parameters in sideband analysis

wrzesaCERN:
New parameters for the sideband analysis. Two possible sets of parameters for 2 and 3 sigmas in the TPC detector (+ now the pt range is wider than before).


altsybee:
+1

PR #14659, opened 2020-06-15
PWGHF: Possibility to run multiple wagons in same train

lvermunt:
Add suffix to TDirectory, so task can be run multiple times in same train


cterrevo:
+1

PR #14663, opened 2020-06-15
Add option for energy scale multiplicator

mfasDa:
Simulation of the energy-independent EMCAL
energy scale mismatch between data and MC
using constant mulitplicator applied to
the FEE amplitude when converting to FastOR
energy.


mfasDa:
+1

PR #14684, opened 2020-06-16
Fix ML model choice in prediction

fcatalan92:
This should fix the issue spotted by @lvermunt where there are crashes for candidates with pT above the last value defined in the config file. Moreover, the candidates falling in the last defined bin were rejected.

@pfecchio @mpuccio @fgrosa Please double check.


fgrosa:
Hi @fcatalan92, it looks good to me. Thanks!

mpuccio:
+1

PR #14701, opened 2020-06-17
Avoid double counting of subjet constituents for matching

raymondEhlers:
Without this change, a constituent could be counted twice during subjet
distance matching, which could cause the pt fraction to be greater than
1 (beyond just float rounding)


raymondEhlers:
+1

PR #14707, opened 2020-06-17
Update on the pi/K/p vs Nch

omvazque:
The momentum binning for the TOF analysis is now finer. This will allow a better parametrization of beta vs p.


bdoenigu:
+1

PR #14712, opened 2020-06-17
Match subjet constituent by label for hybrid-det matching

raymondEhlers:
It turns out that the labels (not the global index or other values) are
propagated to the subtracted constituents, so we can use the labels for
matching


raymondEhlers:
+1

PR #14718, opened 2020-06-18
PWGGA/GammaConv: OmegaToPiZeroGamma - Added AP-Like cut and changed base structure of the task

mhemmer-cern:
- Separated the calculation of Pio candidates from CalculateOmegaCandidates into own function called CalculatePi0Candidates
- CalculateRotationBackground is now an own function to be able to call it inside of CalculateOmegaCandidates to reduce redundant loops over the same Pi0s and Gammas. This should reduce CPU time.
- Added an AP-Like Cut which is overloading the SetToCloseV0sCut in the MesonCuts. For all other analyses, this change should not interfere!


nschmidtALICE:
+1

PR #14723, opened 2020-06-18
Update to Jet V2

SiyuTang:
Hi,

Here is the update of AliAnalysisTaskSEpPbCorrelationsJetV2.cxx, which can reduce the size of output.

Thanks!

Siyu


aohlson:
+1

PR #14729, opened 2020-06-18
Added MEMassQA histograms, which follow the logic of the MassQA (Inv …

dimihayl:
…mass vs k*), applied to the ME. N.B. There is one instance of MassQA in AliFemtoDreamControlSample.cxx, which I did not understood and did not implement for MEMassQA.

I think this "ControlSample" is irrelevant for most used cases, as it applies to the rotated phi, correct? Please confirm that no further work is needed in that file. Thanks! 

Successful local test


a-mathis:
+1

PR #14732, opened 2020-06-19
In NanoDimi Task: added S40 and S4 options to run the analysis. 40 an…

dimihayl:
…d 4 represent the IM window of the Lambda


a-mathis:
+1

PR #14737, opened 2020-06-19
PWGGA/GammaConv: OmegaToPiZeroTask - Added train configs for the AP-Like cut

mhemmer-cern:
- Added train configs for the AP-Like cut


FriederikeBock:
+1

PR #14741, opened 2020-06-19
PWGDQ EtaReconstruction implemented further secondary cuts

feisenhu:
, switch Dalitz or GammaGamma Methode

The commit consits of further cuts to identify secondaries from gamma conversions, as well a switch which is included to switch between the two implemented analysis methods (Dalitz / GammaGamma reconstruction)


miweberSMI:
+1

PR #14768, opened 2020-06-22
deuterons - new parameters to the sideband femto analysis, changes in…

wrzesaCERN:
deuterons - new parameters to the sideband femto analysis, changes in the previous method (the function ("IsDeuteron_sideband")  for sideband analysis was incorrect).


lgraczykCern:
+1

PR #14793, opened 2020-06-24
Support for EMCAL triggers in pp 8 TeV

mfasDa:
- Add trigger clusters used in older data sets
  (CENTNOPMD, ALL, ALLNOTRD)
- Add handling for EJE trigger (EMCAL jet trigger
  in 8 TeV) as extra case.


mfasDa:
+1

PR #14795, opened 2020-06-24
PWGHF: Add PreSelect function for pKpi

lvermunt:
To be used in productions where there are many candidates to be refilled (PbPb `18 or HF ITSUpgrade productions). So far I only added a pT, PID, and (when requested) a Lc invariant mass selection. More cuts, that don't need the secondary vertex, can still be added of course

PR includes also a minor fix for the Bayes PID strategies, to reject tracks that don't have TPC and TOF PID info. In rare cases these tracks could end up being selected by a loose PID strategy


fgrosa:
+1

PR #14800, opened 2020-06-25
add option to include bayesian PID in tree

jaimenorman:
add two additional PID options for the tree 
1. Include Bayesian PID variables
2. Include Bayesian and Nsigma variables

Currently the implementation includes default priors, and combines information from the TPC and TOF to calculate the Bayesian probabilities for each track to be a proton, kaon or pion.


cterrevo:
+1

PR #14802, opened 2020-06-25
PWGGA/GammaConv: Reduced memory, added a DCA cut and implemented a rotation method

Stev960:
Reduced memory, added a DCA cut and implemented a rotation method for the estimation of the background


dmuhlhei:
+1

PR #14809, opened 2020-06-25
fixed weights for invariant mass PairCleaner

magicep0t:
* fixed MC for Xi
	* removed checks for correct function values
	* fixed histogram ranges and bin sizes


sartozza:
+1

PR #14812, opened 2020-06-25
PWGHF: Typo fix in CombinedSoft PID strategy

lvermunt:
Spotted this typo/bug in previous PR where I added the PreSelect function, but forgot to actually fix it


cterrevo:
+1

PR #14838, opened 2020-06-30
Added extra variables to tree for c-deuteron analysis

mbucklan:
If the c-deuteron analysis flag is true then the TTree has some extra variables filled.


fgrosa:
+1

PR #14847, opened 2020-07-01
Combine SoftDrop code into helper base class for data and response

mfasDa:
- Common base class to have code calculating the SoftDrop
  parameters Zg, Rg, Nsd ... and include it via multiple
  inheritance into classes for data and response (safe is safe)
- Use struct instead of vector for softdrop results and
  softdrop params
- Outsource QA code that was inside MakeSoftdrop into
  separate functions so that MakeSoftdrop does not rely
  on dedicated class implementation and can be outsourced
- For casting from float to double in case of jet R use
  safe implementation truncating after the 3rd decimal
  digit (for jet R better precision is not needed)
- Force one overflow bin for Rg histograms instead of
  having it implicitly due to imprecise cast.


mfasDa:
+test

mfasDa:
+1

PR #14872, opened 2020-07-03
PWGHF: Add extra PreSelect for pK0s

lvermunt:
Option to reduce a bit further the CPU time when running pK0s in PbPb '18


fgrosa:
+1

PR #14894, opened 2020-07-07
Added centrality QA plot for AO2D conversion.

mgheata:
This PR adds a centrality histogram while converting ESD to AO2D files. The histogram is written in the AO2D.root file in the container QAlist. The centrality method can be selected in the conversion task by calling SetCentralityMethod(const char *method), the default value being "V0M". The multiplicity task becomes mandatory in the conversion train.


mgheata:
@jgrosseo can you check this PR?

jgrosseo:
+1

PR #14912, opened 2020-07-08
PWGDQ: LMee: added a pair variable for eta and phi for leg1 and leg2.

jjungIKF:
PWGDQ/LMee: added pair variable for eta and phi for both legs


rbailhac:
+1

PR #14927, opened 2020-07-10
Added rejection of pileup events in AO2D conversion via SetSkipPileup…

mgheata:
…, including PU rejection counter histogram.


jgrosseo:
+1

PR #14941, opened 2020-07-10
add pt-dependent Lb fragmentation

jaimenorman:
- pt dependent fragmentation fractions from LHCb measurement  - https://arxiv.org/pdf/1902.06794.pdf
- Uses fits defined in this paper to parameterise fragmentation fraction as a function of the beauty hadron pT, plus flat FF for pT<5
- minimum and maximum FF uncertainties are from the envelope of parameter variations
- includes option for 5 TeV for additional uncertainty taking into account possible difference in FF between 13 TeV and 5 TeV measurements

See D2H presentation for details: https://indico.cern.ch/event/922573/contributions/3925508/attachments/2066201/3467564/Fragmentation_Spectra_D2H.pdf


fgrosa:
+1

PR #14948, opened 2020-07-11
PWGGA/GammaConv: Modified handling of cluster splitting in calo merge…

jokonig:
…d task.
Before there was the possibility that neither the if case or the else if case was entered. Now all clusters with NLM = 1 is treated in the first case, and all clusters with NLM > 1 are treated in the second case.


nschmidtALICE:
+1

PR #14952, opened 2020-07-12
analyisis code to test polarization measurement by Simone

aga01935:
This is the analysis code to recreate the result based on original work by Simone


mbroz84:
+1

PR #14962, opened 2020-07-13
FemtoNanoLD: add (anti-)protons for pair cleaner

stheckel:
o Add protons and anti-protons, but not for CF
o These are used to clean pairs with lambdas and anti-lambdas
o Also add a switch for the proton-lambda pair cleaner


a-mathis:
+1

PR #14964, opened 2020-07-13
FONLL B hadron cross section in pp@13 TeV

zhujh126:


fgrosa:
Hi @zhujh126, could you please add one point keeping the same minimum and maximum values in order to have the bin width of 0.05 GeV/c? Thanks!

PR #14965, opened 2020-07-13
Added PU rejection by TPC, with setter and filling a bin in the PU QA…

mgheata:
… histogram.


mgheata:
@jgrosseo can you have a look?

jgrosseo:
+1

PR #14975, opened 2020-07-14
PWGGA:   PHOS Trigger

jluehder:
1) Added Histograms
2) Dont use triggered events without L0 Trigger
3) Demand Pi0 candidates to have cluster that triggered


kharlov:
+1

PR #14987, opened 2020-07-15
Add noise simulation to the trigger maker

mfasDa:
Noise is simulated FastOR-by-FastOR and event-by-event,
applied only to fastOR signal calculated from smeared FEE
singal. Two models are available:
- constant noise
- gaussian noise


mfasDa:
+1

PR #14996, opened 2020-07-16
PWGGA: Cluster Min E Cut on Meson

jluehder:
Add Cluster Cut analog to Omega Implementaion: Only if one Cluster has E > Threshold, Meson Candidate is used


dmuhlhei:
+1

PR #14997, opened 2020-07-16
PWGDQ EtaReconstruction Removing outdated parts of the code

feisenhu:
Updated and removed some old parts of the code.


miweberSMI:
+1

PR #15006, opened 2020-07-17
Reorganize Online patch QA task

mfasDa:
- Move to PWGJE::EMCALJetTasks
- Add data grids for FastOR energy, FEE energy, nmasked
  and map for fastors contributing to firing patches
- Change output histograms to THnSparses for patch
  energy, patch residuals, fastor energy and fastor
  residuals


mfasDa:
+1

PR #15016, opened 2020-07-17
Avoid calling the event cuts if no cuts requested

mgheata:
The logic is: if no cuts are requested, AcceptEvent is not called at all, otherwise it is called systematically. The result of AcceptEvent is directly used in case SetUseEventCuts or SetSkipPileup were set and the event does not pass cuts (or does not pass the pileup cut respectively). If SetSkipTPCPileup is set, the result of AcceptEvent is not used, but the QA histogram counting TPC pileup events is filled.

### DIFF
--- a/RUN3/AliAnalysisTaskAO2Dconverter.cxx
+++ b/RUN3/AliAnalysisTaskAO2Dconverter.cxx
@@ -611,7 +611,10 @@ void AliAnalysisTaskAO2Dconverter::UserExec(Option_t *)
   }
 
   // We can use event cuts to avoid cases where we have zero reconstructed tracks
-  bool skip_event = !fEventCuts.AcceptEvent(fESD) && fUseEventCuts;
+  bool skip_event = false;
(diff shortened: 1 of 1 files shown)


mgheata:
@jgrosseo can you have a look?

jgrosseo:
+1

ktf:
+1

mgheata:
The merge seems to be still pending, waiting for some approval, although 2 people commented +1. Any idea what went wrong?

ktf:
+1

ktf:
yeah, something is weird. I need to investigate.

ktf:
Ok... There was yet another bit which had troubles with python3. Fixed. Joys of dynamically typed languages... :-/

mpuccio:
Just out of curiosity: is there any reason not to run AcceptEvent? It is quite inexpensive...

mgheata:
We run AcceptEvent, but only if event cuts are requested for the conversion, or if we collect some statistics for pileup. This commit is only not to call it if we don't use it.

PR #15035, opened 2020-07-21
PWGDQ EtaReconstruction: Bool to analyse generated and reconstructed …

feisenhu:
…separatly

The new implemented bool is able to analyse the generated and reconstructed analysis method separately. This is improving the compiling and execution time.


miweberSMI:
+1

PR #15047, opened 2020-07-22
adapt to lambda

galocco:
Now it is possible to create of a tree of candidates lambda decaying in proton and pion with the same structure as for the hypertriton in 2body.


mpuccio:
+1

PR #15051, opened 2020-07-23
FemtoNanoLD: add flexiblity for pair cleaner

stheckel:
o Can now be completely deactivated (0), use default settings (1)
  or include proton-lambda cleaning (2)
o in addition, activate close pair rejection for subwagon 2


sartozza:
+1

PR #15054, opened 2020-07-23
PWGGA/GammaConv: fix streaming property of two histos

sstiefel19:
These histos need to be streamed since they are being
initialized within AliConvEventCuts::InitCutHistograms
which appears to be run before the distribution on
grid nodes.


dmuhlhei:
+1

PR #15056, opened 2020-07-24
AliFemto: significant changes to FsiWeightLednicky fortran code

majanik:
Exchanging old FORTRAN code with new one obtained from Lednicky, including new code for baryons.
The obtained FORTRAN code was not swapped directly but also corrected to meet proper programming practices (e.g. proper assignment of values in arrays, corrections to COMMON blocks, etc.). 
Several tests of the code were performed. The code works well for kaons and protons, but strong FSI for pions produces strange shapes (only QS and Coulomb options should be used for pion-pion pairs).

Also, AliFemtoModelWeightGeneratorLednicky.cxx file was modified, so that it allows to set by hand particle types in the pair.


majanik:
+1

PR #15061, opened 2020-07-24
PWGGA/GammaConv: GammaIsoTree added additional MC check

fjonasALICE:
introduced this check to catch photons that do not have a mother. This should normally not happen within Pythia, however it happens when running with DPMJet MC sometimes for some reason


dmuhlhei:
+1

PR #15081, opened 2020-07-28
Add option for run1 runrange in trending plots

mfasDa:
In addition
- Remove duplicated trigger cluster enum
- Fix assignment of max cluster in case of EDCAL triggers
  and only clusters in one of the two subdetectors


mfasDa:
+1

PR #15087, opened 2020-07-29
Move KF functions to AliVertexingHFUtils

zhujh126:
Move KF functions to AliVertexingHFUtils and add class AliAnalysisTaskSEXicZero2XiPifromKFP and AddTaskLc2pKs0FromKFParticle.C

### DIFF
--- a/PWGHF/correlationHF/CMakeLists.txt
+++ b/PWGHF/correlationHF/CMakeLists.txt
@@ -22,6 +22,7 @@ include_directories(${AliPhysics_SOURCE_DIR}/PWGHF/correlationHF)
 
 # Additional includes - alphabetical order except ROOT
 include_directories(${ROOT_INCLUDE_DIRS}
+                    $ENV{KFPARTICLE_ROOT}/include
                     ${AliPhysics_SOURCE_DIR}/CORRFW
                     ${AliPhysics_SOURCE_DIR}/OADB
                     ${AliPhysics_SOURCE_DIR}/PWG/Tools
@@ -85,7 +86,7 @@ add_library_tested(${MODULE} SHARED  ${SRCS} G__${MODULE}.cxx)
 
 # Generate the ROOT map
 # Dependecies
-set(LIBDEPS ANALYSISalice PWGHFhfe PWGHFvertexingHF)
+set(LIBDEPS ANALYSISalice PWGHFhfe PWGHFvertexingHF KFParticle)
 
 generate_rootmap("${MODULE}" "${LIBDEPS}" "${CMAKE_CURRENT_SOURCE_DIR}/${MODULE}LinkDef.h")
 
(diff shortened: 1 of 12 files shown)


fgrosa:
+test

zhujh126:
The configuration of KF should be added in CMakeLists For HFCJ, HFe and JE.
The following classes include "AliVertexingHFUtils" or use function of "AliVertexingHFUtils":
PWGHF/correlationHF/AliHFCorrelator.h
PWGHF/hfe/AliAnalysisTaskHFEMultiplicity.h
PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskSVtaskMCFilter.cxx

Since we add KF function in "AliVertexingHFUtils", if one wants to use "AliVertexingHFUtils", the corresponding KF configuration should be added in CMakeLists. Otherwise, when compiling, it will crash and return error "donot find KF class".

jezwilkinson:
+test

mfasDa:
Hi, 

Making specialized libraries dependent on each other creates the high risk to introduce circular dependencies. In case of EMCALJetTask the dependency is only introduced by 

```
  static TString  GetGenerator(Int_t label, AliAODMCHeader* header);
```

Also other functions are pretty common for many analysis. Have you thought about outsourcing the class or at least part of its functionality to PWG - that is usually the place for common functionality an is designed to have no dependency on any PWG library, the risk of circular dependencies is minimal there.

Cheers

Markus

fgrosa:
Hi Markus, 

I agree that we should limit as much as possible dependencies among specialized libraries. However, unfortunately it is not possible to move `AliVertexingHFUtils` to PWG, because it heavily depends on vertexingHF (since it was originally meant to be a class with utils methods for vertexingHF), but indeed it would be possible to move the `GetGenerator` method in another class in PWG if this is needed also for PWGJE.

Cheers,
Fabrizio

mfasDa:
Hi Fabrizio,

I think jets are not the only user of the utility functions but there are also users, i.e. DQ and HFE are using functionality from the class as well.

I will add the +test for EMCALJetTasks but please let's keep in discussion how we can disentangle the libraries again.

Cheers

Markus

mfasDa:
+test

fgrosa:
Hi Markus, 

yes indeed we should consider to move the methods that are used in other libraries in another util class. A possibility would be to move them in [AliAnalysisUtils.h](https://github.com/alisw/AliPhysics/blob/master/OADB/AliAnalysisUtils.h), if they do not require particular dependencies. Then we could remove the dependency on vertexinHF from the other libraries.
 
Cheers,
Fabrizio

fcolamar:
+test

jezwilkinson:
+1

jezwilkinson:
@mfasDa @fgrosa @fcolamar It seems that the tests went OK, are we happy to go ahead with the merge? (actually it seems somehow the build system only wants to have the +1 from HFCJ, but just to check with everyone given the discussion from before)
Cheers,
--Jeremy

fgrosa:
Hi @jeremy, for me the PR can be merged! It seems indeed that we need a +1 from @fcolamar. 
I will do the modifications discussed with @mfasDa in a second stage, to remove the dependencies on vertexingHF from the other libraries (as much as possible).

fcolamar:
+1

PR #15098, opened 2020-07-29
[PWG-CF] Add arrays for centrality bins

cimordas:
* Modify the task to not need external input of the centrality bins, using arrays of histograms and profiles.
* Add some configuration profiles.
* Add the AddTask to use the LEGO trains.
* Add option to compute some multiparticle correlators with 8 and 10 particles.


aohlson:
+1

PR #15100, opened 2020-07-29
PWGHF: Event ID bug fix TTreeCreatorApply

lvermunt:
Bug/typo fix. 

When the event ID was large, the ID was stored differently in the event properties TTree and the reconstructed HF candidate TTree due to Int_t vs UInt_t. When matching both TTrees at Pandas dataframe level, this mismatch rejected ~50% of the candidates

(not sure about classdef)


fgrosa:
+1

PR #15102, opened 2020-07-30
add the trigger selection for 2018 PbPb

XiaozhiBai:
HI
I shared the trigger selections for my 2018 PbPb analysis, others can use it.
Cheers
Xiaozhi


iarsene:
+1

PR #15104, opened 2020-07-30
Changes in the MC collision and MC particle:

pzhristov:
- Corrected TList navigation in case of cocktail generator;
- Arrays fMother[2] and fDaughter[2] replaced by individual columns
fMother0,1 and fDaughter0,1;
- Additional column for the impact parameter from MC. Implemented for
headers that inherit from AliCollisionGeometry.


pzhristov:
+1

PR #15114, opened 2020-07-30
Fix retrieving downscale factors for run1

mfasDa:
- Add -S- scheme instead of -B- scheme
- Reject TRD trigger classes containing INT7 string


mfasDa:
+1

PR #15116, opened 2020-07-31
PWGGA/GammaConv: GammIsoTree workaround for RhoTask problem in data

fjonasALICE:
This is a workaround to avoid crashes when the RhoTask is not found, which happens for some of LHC12 data periods. The crashes need to be further investigated, however rho is not really used at this stage so this workaround lets me at least get other output for now.


dmuhlhei:
+1

PR #15117, opened 2020-07-31
Changing to Lambda Lambda

gmantzar:
Changing the selection criteria to events with at least two Lambdas in order to study Lambda-Lambda correlations


otonvd:
+1

PR #15120, opened 2020-07-31
Disabeling MC and changing the selection for Lambda Lambda

gmantzar:
Write the nTuple only if there are at least two Lambdas. Also removing the MC info to run over data


sartozza:
+1

PR #15140, opened 2020-08-05
PWGGA: Use DDLs instead of modules for trigger rejection

jluehder:
PWGGA: Use DDLs instead of modules for trigger rejection


dmuhlhei:
+1

PR #15155, opened 2020-08-06
PWGLF/SPECTRA: alimultdepspec UE functionality

krschmit-cern:
This task adds the possibility to study the Underlying Event region in Events selected equivalently to the alimultdepspec base task.

- Find highest pT track (leading track) and define UE region relative to leading track in phi region
- Study the same observables defined in alimultdepspec base task
- Additionally look at leading track properties and study the difference between MC particle and track


mkruegerGitHub:
+1

PR #15177, opened 2020-08-10
Fix patch selection

mfasDa:
- Patch selection returned by mistake true also in case the patch was rejected
- In FastOR loop reject positions inside the PHOS region


mfasDa:
+1

mfasDa:
+1

PR #15190, opened 2020-08-12
Modifications to MCParticle class, to store IsPhysicalPrimary status.

smaff92:
As the title says, just a small modification to handle the primary status.


mkruegerGitHub:
+1

PR #15192, opened 2020-08-12
Rdev rsn timerangecut

preghenella:
This PR adds to the RSN analysis package the possibility to use `TimeRangeCut` event cuts.
The use of this extra cut is driven by a flag that can be set by the user via `AliRsnMiniAnalysisTask::SetUseTimeRangeCut`.
The default is `kFALSE` and the cut is not applied.

This development is useful to add those runs (8 runs in LHC18r) that are flagged by the DPG as being bad for TPC PID.

> 8 runs, 296749, 296750, 296849, 296890, 297029, 297194, 297219, 297481 have been excluded from lists including TPC, because they are bad for TPC PID due to a problem in the gain in one or two sectors. This problem happens at the end of the run: by excluding the problematic time range, the remaining events can be used for PID analyses.

The Lambda(1520) add-analysis-task macro is updated to make use of this addition.

### DIFF
--- a/PWGLF/RESONANCES/AliRsnMiniAnalysisTask.cxx
+++ b/PWGLF/RESONANCES/AliRsnMiniAnalysisTask.cxx
@@ -49,6 +49,8 @@
 #include "AliRsnMiniResonanceFinder.h"
 //#include "AliSpherocityUtils.h"
 
+#include "AliTimeRangeCut.h"
+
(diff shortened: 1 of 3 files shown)


preghenella:
@aknospe FYI
with @neelimaagrawal we have updated the RSN mini package to provide this extra feature.

PR #15205, opened 2020-08-13
Updates on the track selection

omvazque:
The track selection now includes the missing 'GetTPCsignalN' cut which was used for the preliminaries. This changes the shape of the efficiencies


mkruegerGitHub:
+1

PR #15207, opened 2020-08-13
PWGGA/GammaConv: Allow processing of multiple added particle at once

sstiefel19:
With this change, particles from all the injectors on the
AliConvEventCuts::fHeaderList will be processed.
This wont have any impact on existing analyses, since up to now
the GammaConvV1 task was such, that users would add only one item
to that list (either Pi0 or Eta).


dmuhlhei:
+1

PR #15208, opened 2020-08-13
FemtoNanoLD: cut settings for deuteron purities

stheckel:
o Without TPC nSigma cut or without TOF nSigma cut
o Again back to std pair cleaner and no CPR


a-mathis:
+1

PR #15211, opened 2020-08-14
Easy switch between MC and data and lowering disk space usage

gmantzar:
- switch between MC and data via ifdef (currently MC is activated)
- exclusion of the mass, eta and phi branches to reduce disk space usage


bhohlweger:
+1

PR #15220, opened 2020-08-14
Support for EMC7 trigger in pp 8 TeV

aschmier:
- Add handling for EMC7 trigger (EMCAL jet trigger
  in 8 TeV) as extra case.


mfasDa:
+1

PR #15221, opened 2020-08-16
Update TPCTPC correlation

SiyuTang:
Hi ,

Here are the updates for Jet v2 analysis, where the TPC-TPC correlation was added. Thanks a lot!

Siyu


altsybee:
+1

PR #15230, opened 2020-08-17
Missing CMakeList and LinkDef for PR 15228

a-mathis:
@miweberSMI @simons27 
this should do the trick


miweberSMI:
Hi @a-mathis,
but this is the same as we have in 15229 (since I resubmitted the ClassDef change)?
But if this works, then even better ;)

miweberSMI:
OK, now I understood the problem (.cxx in link def)

miweberSMI:
Thanks a lot @a-mathis !!!

PR #15246, opened 2020-08-18
nSigma of electrons

omvazque:
The nSigma of electrons was added to measure the contamination on thee kaons and protons at low pT


mkruegerGitHub:
+1

PR #15253, opened 2020-08-18
PWGGA/GammaConv: Changed configs for EDC (M02 upper cut) and added PCMEDC configs without smearing

jokonig:


FriederikeBock:
+1

PR #15263, opened 2020-08-19
Wider cuts to include the background for purity estimation.

gmantzar:
And: reduction of TTree Branches which are not needed for that to reduce disk space usage


a-mathis:
+1

PR #15266, opened 2020-08-19
Add trigger cluster information to the softdrop histograms

mfasDa:
For each jet 1 entry is created in the softdrop histograms.
2D histograms need to be projected to the trigger cluster
bin. Change is necessary for trigger matching in
reconstructions w TRD in combined tracking.


mfasDa:
+test

mfasDa:
+1

PR #15267, opened 2020-08-19
Modify the way to identify particles produced in the transport for AODs

fprino:
This is needed to properly tag pileup particles in productions with embedding+signal fitlering


fprino:
+1

fprino:
+1

PR #15270, opened 2020-08-20
Change task structure to allow to choose which particle to analyse and implement trigger selection based on centrality

alessandrobalbino:


bianchil:
+1

PR #15292, opened 2020-08-24
J/psi spectrum using EMCal triggers: MC studies

crisjahnke:
Changes to obtain J/psi spectrum uisng EMCal trigger.


miweberSMI:
+1

PR #15299, opened 2020-08-24
PWGGA/GammaConv: fix for c6fff645377685feb08c268ce4ca62ba6707d637

sstiefel19:
small fix for the case of signalRejection==4 and both
UseElecSharingCut() and UseToCloseV0sCut() disabled


dmuhlhei:
+1

PR #15327, opened 2020-08-26
PWGGA/GammaConv: refactor ProcessPhotonCandidates()

sstiefel19:
This change reduces the complexity of AliConversionPhotonCuts::
ProcessPhotonCandidates significantly. The semantics of the
sharedElectron and toCloseV0s cuts are slightly changed.
Before this change:
Say we have photons A B C. Then RejectToCloseV0s will cut photon A
if its too close to B. However, also C gets cut, if C is too close to A.
Even though A has already been thrown out.
For RejectSharedElectronV0s its the same: Say A and B have an electron track
in common -> A gets kicked out. Then C will still get kicked out,
if it is shares the positron track with A.
With this change, photon C does not get cut.

This affects typically 10% of photons for the shared electron
cut and sometimes one out of 200 photons for the tooclose cut.

Extensive automated tests were carried out to ensure that
the change does not break anything and does not have further
semantic consequences.

### DIFF
--- a/PWGGA/GammaConv/AliAnalysisTaskGammaConvV1.cxx
+++ b/PWGGA/GammaConv/AliAnalysisTaskGammaConvV1.cxx
@@ -23,7 +23,7 @@
 // Class used to do analysis on conversion pairs
 //---------------------------------------------
 ///////////////////////////////////////////////
-#include <unordered_set>
+#include <set>
 
 #include "TChain.h"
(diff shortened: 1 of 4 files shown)


nschmidtALICE:
+test

nschmidtALICE:
+test

nschmidtALICE:
+test

PR #15330, opened 2020-08-26
PWGGA/GammaConv: refactor ProcessPhotonCandidates()

sstiefel19:
This change reduces the complexity of AliConversionPhotonCuts::
ProcessPhotonCandidates significantly. The semantics of the
sharedElectron and toCloseV0s cuts are slightly changed.
Before this change:
Say we have photons A B C. Then RejectToCloseV0s will cut photon A
if its too close to B. However, also C gets cut, if C is too close to A.
Even though A has already been thrown out.
For RejectSharedElectronV0s its the same: Say A and B have an electron track
in common -> A gets kicked out. Then C will still get kicked out,
if it is shares the positron track with A.
With this change, photon C does not get cut.

This affects typically 10% of photons for the shared electron
cut and sometimes one out of 200 photons for the tooclose cut.

Extensive automated tests were carried out to ensure that
the change does not break anything and does not have further
semantic consequences.

### DIFF
--- a/PWGGA/GammaConvBase/AliConversionPhotonCuts.cxx
+++ b/PWGGA/GammaConvBase/AliConversionPhotonCuts.cxx
@@ -1641,7 +1641,7 @@ Bool_t AliConversionPhotonCuts::AcceptanceCuts(AliConversionPhotonBase *photon) 
 
   } else if (fDoShrinkTPCAcceptance == 4){   // accept only photons in eta-phi region from PHOS-PCM (pi0 and eta meson analysis)
     Double_t photonPhi = photon->GetPhotonPhi();
-      
+
     if(photon->GetPhotonEta() > fEtaForPhiCutMin && photon->GetPhotonEta() < fEtaForPhiCutMax ){
       //cout << "A and C side, eta=" << photon->GetPhotonEta() <<  endl;
       if(!(photonPhi>fMinPhiCut  && photonPhi<fMaxPhiCut )){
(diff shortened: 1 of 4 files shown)


dmuhlhei:
+test

PWGGA/GammaConvBase/AliConversionPhotonCuts.cxx
```diff
@@ -4779,6 +4779,63 @@ Bool_t AliConversionPhotonCuts::RejectToCloseV0s(AliAODConversionPhoton* photon,
...
+                    thePhoton.GetConversionZ());
+
+        TVector3 v2(iPhoton->GetConversionX(),
+                    iPhoton->GetConversionY(),
+                    iPhoton->GetConversionZ());
+
+        TVector3 d = v2 - v1;
+
+        if(d.Mag2() < fminV0Dist*fminV0Dist){
+          if(thePhoton.GetChi2perNDF() > iPhoton->GetChi2perNDF()) return kFALSE;
```
dmuhlhei, line 4825: there still is a bug, if the first photon to be included has a higher chi2 than a following candidate - then the first photon needs to be removed which is still not implemented.

please do not merge until fixed

PWGGA/GammaConvBase/AliConversionPhotonCuts.cxx
```diff
@@ -4779,6 +4779,63 @@ Bool_t AliConversionPhotonCuts::RejectToCloseV0s(AliAODConversionPhoton* photon,
...
+  // since it is quite rare that this cut cuts, we insert the positive track before checking if the negative one can be inserted as well
+  auto lItInsWorkedPositive = theLabels.insert(thePhoton.GetTrackLabelPositive());
+  if (!lItInsWorkedPositive.second){
+    return  kFALSE;
+  }
+
+  Bool_t lInsWorkedNegative = theLabels.insert(thePhoton.GetTrackLabelNegative()).second;
+  if (!lInsWorkedNegative){
+    theLabels.erase(lItInsWorkedPositive.first);
+    return kFALSE;
```
dmuhlhei, line 4795: we may want to check the chi2 values instead of simply accepting the first photon candidate in the list - to be discussed

dmuhlhei (changes requested):

PR #15341, opened 2020-08-27
Added trigger selection

simons27:
I added a private variable and according set-function to my class to allow for more flexibility with the trigger choice for analysis. Hard coded trigger selection in the AddTask was removed.


a-mathis:
+1

PR #15347, opened 2020-08-28
Splines and maps for LHC18q/r.pass3 without pileup correction 

wiechula:
As far as I saw since yesterdays tag we have the new AliRoot/v5-09-55-2 which includes the required modifications in AliTPCPIDResponse needed by the new splines. So it is fine to merge.


chiarazampolli:
+1

PR #15348, opened 2020-08-28
PWGGA/GammaConv: major speedup for ElectronStudies on JJ

fjonasALICE:
ElectronStudies
- in order to be able to run the ElectronStudies on JJ and GJ MC, I switched to default track matching for a major speedup, which is handled by a separate calo cut object to avoid rejection of matched clusters in the normal custer cuts
- following suggestions I added M20 to output

GammaIsoStudies
- added Sumw2 for all histos


nschmidtALICE:
+1

PR #15349, opened 2020-08-28
I want to add the new method to determine the scale rho function which trances form charge to full.

11kumaoka:
I want to add the new method to determine the scale rho function which trances form charge to full.
Now we need to read TF1 of the root file, but it could not be defined in each various region of the x-axis.
Therefore, I created the new option that the scale function was defined in this task.
@jdmuligan

### DIFF
--- a/PWGJE/EMCALJetTasks/AliAnalysisTaskRhoBase.cxx
+++ b/PWGJE/EMCALJetTasks/AliAnalysisTaskRhoBase.cxx
@@ -51,6 +51,7 @@ AliAnalysisTaskRhoBase::AliAnalysisTaskRhoBase() :
   fCompareRhoScaledName(),
   fRhoFunction(0),
   fScaleFunction(0),
+  fScaleFunctionKind(0),
   fInEventSigmaRho(35.83),
   fAttachToEvent(kTRUE),
   fIsPbPb(kTRUE),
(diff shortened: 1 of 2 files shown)


raymondEhlers:
Hi @11kumaoka ,

I don't mean to be difficult, but I'm sorry, but we can't accept this PR. I don't really understand you description above, but based on the code, there are a few problems here:

1. We already have this functionality. `SetScaleFunction()` already performs precisely this function. You pass it a TF1 which can depend on centrality of your own choosing (for example, the parametrization that you hard code). There are many options to provide it: you could define the TF1, or open it from a root file. There's a lot of jet framework documentation that covers these type of topics - I suggest that you take a look [here](http://alidoc.cern.ch/AliPhysics/master/READMEjetfw.html) if you haven't already.
2. You're modifying a base class - everyone performing jet analyses depends on this particular class operating correctly. So we have a very high bar for introducing changes to any of these classes - such changes need to be well tested and well understood. Beyond already having this functionality, this PR introduces at least a couple of bugs (not incrementing ClassDef, some members are now uninitialized, etc).

I know that you're still learning, and that's okay! We already have a lot of functionality, so I recommend that you look around at the code base, ask questions on the JE list, and continue speaking with others at your institution and elsewhere. For now, please close this PR and take another look at using the existing functionality.

Best regards,
Raymond

ps. If you want to tag James, you're missing an "l" in his name - it's @jdmulligan

11kumaoka:
Thank you for giving me a quick and kind response.
I will discuss it on the JE list.

Best regards,
Takuya

ps. Thank you for tell me my mistook

PR #15368, opened 2020-08-30
Fix type of EMCAL cells in trees

mfasDa:
The cell type branch (Char_t) got as type "C" which
is interpreted as char array, instead it should have gotten
"B".

### DIFF
--- a/RUN3/AliAnalysisTaskAO2Dconverter.cxx
+++ b/RUN3/AliAnalysisTaskAO2Dconverter.cxx
@@ -289,7 +289,7 @@ void AliAnalysisTaskAO2Dconverter::UserCreateOutputObjects()
     tCalo->Branch("fCellNumber", &calo.fCellNumber, "fCellNumber/S");
     tCalo->Branch("fAmplitude", &calo.fAmplitude, "fAmplitude/F");
     tCalo->Branch("fTime", &calo.fTime, "fTime/F");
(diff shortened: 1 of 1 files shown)


mfasDa:
@jgrosseo, @mpuccio In the development of a run3 full jet task we have seen problems reading the EMCAL cell information. 

```
[772139:internal-dpl-aod-reader]: Error in <TTreeReaderValueBase::GetBranchDataType()>: Must use TTreeReaderArray to read branch fCellType: it contains an array or a collection.
[772139:internal-dpl-aod-reader]: Error in <TTreeReaderValueBase::CreateProxy()>: The branch fCellType contains data of type {UNDETERMINED TYPE}, which does not have a dictionary.
``` 

I suspect the problem comes from the cell branch having type "C" which is interpreted as c-string, instead it should have gotten "B" as single char value. A second branch I adapted in the calo trigger struct. Could you please review and merge the pull request?

mfasDa:
Could someone please have a look at this pull request?

raymondEhlers:
Just to reiterate the need - it appears that this issue is preventing me from doing further development for full jets in O2

mpuccio:
+1 I approve because it is trivial, but please note that @jgrosseo is the right person to ping for the Run3 development (I have the permissions because I started the converter few years back)

jgrosseo:
@mfasDa Sorry, I get notifications for all AliPhysics PRs, so there is no way to filter out relevant ones. I will start a new conversion.

PR #15375, opened 2020-08-31
Enable debug code

feisenhu:
Enable commented code in order to check which kind of particles are not rejected by prefilter.


rbailhac:
+1

PR #15377, opened 2020-08-31
Make usage of negative noise optional

mfasDa:
Negative noise simulating baseline underfluctuations. In
case it is disabled negative noise values are set to 0.


mfasDa:
+1

PR #15407, opened 2020-09-03
Fix correlation of the interaction trigger with the EMCAL trigger dec…

mfasDa:
…ision for MC

Logics needs to be inversed: Instead of logical and
with the EMCAL trigger decision for each interaction
trigger the logical and has to be performed for each
interaction trigger in case the EMCAL trigger fired (if
not the trigger status has to be set to false). If not it
is possible that events pass the trigger selection which
are rejected by the interaction trigger.


mfasDa:
+1

mfasDa:
+1

PR #15428, opened 2020-09-04
PWGGA/GammaConv: OmegaToPiZero additional trainconfigs

mhemmer-cern:
- added trainconfigs for rotation background with rotation around Pi0 for triggered data


dmuhlhei:
+1

PR #15442, opened 2020-09-07
Modifications to RT/S0 Analysis Task, to re-introduce old feautures, …

smaff92:
…and completely remove TParticle dependence.

In addition, modifications have also been made to Track/Particle classes used in these analyses.


ddobrigk:
+1

PR #15443, opened 2020-09-07
PWG/EMCAL: Set the bool for the High Gain flag, to keep the information

jokonig:


FriederikeBock:
+1

FriederikeBock:
actual changes are around line 2200, the rest are trailing whitespaces & line feed back from windows to unix

PR #15445, opened 2020-09-08
[EMCAL-625] Fix AOD tables for EMCAL

mfasDa:
- Add track position (eta,phi) at the EMCAL surface to the
  track table. This is crucial for cluster-track matching
- Fix numbers for gain type based on CaloFlag definition
  in the ALTRO: HG - 1, LG - 0


jgrosseo:
+1

PR #15449, opened 2020-09-08
PWGGA/GammaConv:Added QA plots, optimized the rotationmethod and changed the labelling of the true information

Stev960:
Added some QA plots and changed the labelling of the true information. Furthermore optimized the rotationmethod by adding TGenPhaseSpace and increasing the number of swaps.


dmuhlhei:
+1

PR #15451, opened 2020-09-09
TriggerMaker: Enable noise by default in case of MC

mfasDa:
Change default configuration for MC so that noise is
always simulated with a sigma of 50 MeV


mfasDa:
+1

PR #15453, opened 2020-09-09
PWGCF/fix issues with running on grid

ShiQiu1:


aohlson:
+1

aohlson:
Hi @ShiQiu1, 
I noticed that you have CreateAlienHandler checked in here as a macro in AliPhysics, which has the disadvantage that you need to do a pull request any time you switch from "test" to "full" mode or change the AliPhysics tag.  Since you can submit and run jobs with just a local version of CreateAlienHandler, I suggest that you remove it from AliPhysics, unless there's some reason I'm not thinking of why it needs to be in this directory.  
Cheers, 
Alice

PR #15463, opened 2020-09-10
Femto dev

stheckel:
Two commits:
- First: changes in FemtoDream to obtain k* vs inv.mass of the pairs
- Second: needed changes in the lambda-deuteron AnalysisTask


bhohlweger:
+1

PR #15466, opened 2020-09-10
Trigger maker: Switch on loading FastOR mask by default

mfasDa:
- Load and enable FastOR mask by default
- Add messages for different configuration steps and
  display them via std::cout (so they always appear in the
  logs)
- Set license for trigger maker to BSD license


mfasDa:
+1

PR #15468, opened 2020-09-10
PWGGA/GammaConv: Modified configs for no NCell cut and changed defaul…

jokonig:
…t value for min. energy for the exotics cut
The value of the min. energy is necessary to not throw away all 1cell cluster which have an ecross of 0.


dmuhlhei:
+1

FriederikeBock:
+1

PR #15477, opened 2020-09-11
Added a cut for MC events jets with pt>4*pt_hard

osaarima:
Detector MC events with jets with pt>4*pt_hard are not processed. This
required a bit code reorganizing so that no histograms are filled before
this check is done.


aohlson:
+1

PR #15480, opened 2020-09-11
FemtoDream: use PDG masses for kstar vs inv.mass

stheckel:
o Put into new function to separate from 1D inclusive
  pair inv.mass with measured masses
o Change the scale and binning of kstar vs inv.mass hists


a-mathis:
+1

PR #15481, opened 2020-09-11
Quick bugfix for TPC variables

smaff92:
Just forgot to include the new class members in the Reset() function.
This has now been fixed!


lbariogl:
+1

PR #15485, opened 2020-09-11
Changing default eta cut of hadrons to 0.8

deepathoms:
Updating the default eta cut of hadrons from 0.9 to 0.8


deepathoms:
+test

PR #15495, opened 2020-09-14
kaon-deuteron femto analysis, small changes to additional variation o…

wrzesaCERN:
…f the selection. This is sth needed to the additional variation of the selection of deuterons. The new lines with "else if.."  help in the cuts that use the pt ranges instead of p range to avoid the contamination region that can be covered in the selection with tpc and tof. It is switched on with a specific set because it is an optional variation similar to the "if " above.

this code for deuterons will be further developed ...


lgraczykCern:
+1

PR #15504, opened 2020-09-15
Add function updating the collection name

mfasDa:
Function regenerating the collection name from the
settings in the jet container and the connected
constituent level containers (particles and clusters).
Intended to be called at runtime or after changes in the
wagon customization.

### DIFF
--- a/PWG/JETFW/AliJetContainer.cxx
+++ b/PWG/JETFW/AliJetContainer.cxx
@@ -12,8 +12,12 @@
  * about the suitability of this software for any purpose. It is          *
  * provided "as is" without express or implied warranty.                  *
  **************************************************************************/
+#include <iostream>
+#include <memory>
 
 #include <TClonesArray.h>
+#include <TObjArray.h>
+#include <TObjString.h>
 
 #include "AliVEvent.h"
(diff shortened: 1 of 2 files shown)


mfasDa:
Trivial function, but needed i.e. the users make changes in the wagon customization which have an impact on the collection name (i.e. when changing the constituent E cut). Needed because the container name is fixed when the jet container is created. Must be called explicitly. Otherwise doesn't affect other functions.

@raymondEhlers any objection?

mfasDa:
+test

raymondEhlers:
I don't really like this approach because it breaks the correspondence between array name and container name, which I think will make debugging more difficult. However, I know it's also annoying to remove and then recreate the jet container, so it's fine by me

mfasDa:
@raymondEhlers The other option would be to make all settings fields of the add macro which will force to change a huge amount of wagons every time we change something and trains become unmanagable. The problem is that the linking is done in a static way once the jet container is created, not when it is fully configured, for which we need a workaround. I understand your concern but this is the most straight forward solution. I will merge the PR.

mfasDa:
+1

PR #15514, opened 2020-09-15
1) Adds default bool cutit = false for  fTOFInvMass in TOF Mass PID, 2) adds systematic changes for sidebands pd Analysis

sbhawani:
1) Adds default bool cutit = false for  fTOFInvMass in TOF Mass PID, 2) adds systematic changes for sidebands pd Analysis


sartozza:
+1

PR #15515, opened 2020-09-16
Removing Stack and TParticle from the task 

crisjahnke:
Removing Stack and TParticle from the task (which were used on ESD case only)


iarsene:
+1

PR #15544, opened 2020-09-18
Correct set of minimum pt for the leading track

gluparel:
Correct small bug on the settings of the minimum pt for the leading track and filling of the control histogram


fgrosa:
+1

PR #15549, opened 2020-09-18
PWGGA/GammaConv: OmegaToPiZeroTask - Changed the way rotation background is calculated

mhemmer-cern:
- Changed the way rotation background is calculated for rotation around omega 
- Lowered the number of omega p_T bins to be more reasonable and use less memory
- Added OAC for omega mesons

### DIFF
--- a/PWGGA/GammaConvBase/AliConversionMesonCuts.h
+++ b/PWGGA/GammaConvBase/AliConversionMesonCuts.h
@@ -130,6 +130,7 @@ class AliConversionMesonCuts : public AliAnalysisCuts {
     Bool_t MesonIsSelectedMCChiCAODESD(AliDalitzAODESDMC* fMCMother,AliDalitzEventMC *mcEvent, Int_t &, Int_t &, Int_t &, Double_t fRapidityShift=0. ) const;
     Bool_t MesonIsSelectedPiZeroGammaAngle(AliAODConversionMother *omega, AliAODConversionMother *pi0, AliAODConversionPhoton *gamma,
                                            Bool_t DoPiZeroAngleCut, TF1 *maxfit, Double_t lowerFactor, Double_t upperFactor);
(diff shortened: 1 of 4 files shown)


PWGGA/GammaConvBase/AliConversionMesonCuts.h
```diff
@@ -345,7 +346,7 @@ class AliConversionMesonCuts : public AliAnalysisCuts {
   private:
 
     /// \cond CLASSIMP
-    ClassDef(AliConversionMesonCuts,45)
+    ClassDef(AliConversionMesonCuts,46)
```
dmuhlhei, line 349: not needed, keep 45

dmuhlhei (changes requested):

PWGGA/GammaConvBase/AliConversionMesonCuts.h
mhemmer-cern: Then I think I don't fully understand it. I added a new function, I thought that is a reason for an increment.

dmuhlhei: increment is only needed when adding/removing persistent member variables

dmuhlhei:
+1

PWGGA/GammaConvBase/AliConversionMesonCuts.h
mhemmer-cern: But what does persistent mean in that case? 
Also, the PR seems bugged for me, since there is no merging after the +1

dmuhlhei: Basically every member variable which does not have a //! (meaning it is transient)
Yes, there were some general issues with PRs - should work soon

dmuhlhei:
+1

PR #15554, opened 2020-09-18
PWGGA: Change Pt Range from ITS and TPC and TOF histograms to 50. GeV

jluehder:
and increase binning from 150 bins to 170


FriederikeBock:
+1

PR #15567, opened 2020-09-21
Fixes for XicTopKpi task

mfaggin:
- event selection in MC fixed: events having generated primary vertex with |vtx_z|<10cm rejected for both gen. and reco. candidates (main fix)
 - enlarge the pT range from 16 GeV/c to 24 GeV/c for reconstructed pKpi candidates
 - monitoring histograms for vtx_z added
 - missing delete command for a dynamic array in the UserExec added


fgrosa:
+1

PR #15574, opened 2020-09-21
Revert "Adding code to prefilter the Lb to Lc pi candidates (Anastasis)"

benedikt-voelkel:
This reverts commit 881503420f7f3cdefc730584880818389b345306.

Incomplete commit, needs revision --> revert

PR #15580, opened 2020-09-22
PWGHF: Update Lb ITS Upgrade TTreeCreator

lvermunt:
For possible extra selection on Lb variables and background shape study

Follow up of #15560 (8815034) and the reverting of this incomplete commit in #15574


cterrevo:
+1

PR #15589, opened 2020-09-23
Fill TTrees Only if Set

hdegen:
Really small changes applied. Now the TTrees will only be filled if someone wants to check something offline, otherwise it will not be filled. This will reduce the total output for quick and well known checks.


aortizve:
+test

aortizve:
+1

hdegen:
Dear all, thank you for verifying and adding this so fast! Unfortunately I tried to use the accepted addTask on the lego trains but that did not work. I checked and it seems that maybe the +1 was not understood by alibuild and the modifications were not merged. Can someone give the +1 again, please? Thanks!!

aortizve:
+1

PR #15590, opened 2020-09-23
Added a Pt selection cut for charged pions

pgamamed:
Added line 1337 "if (vCandidateCharged.Pt() <0.11)". Adding transverse momentum selection cut for charged pions.


mbroz84:
+1

PR #15594, opened 2020-09-23
Fix in MC event selection

mfaggin:
- requirement |vtx_z|<10cm left to store generated particles
 - ev. selections applied before looking at reconstructed particles


fgrosa:
+1

cterrevo:
+1

PR #15630, opened 2020-09-25
Split jet finding efficiency in efficiency and purity

mfasDa:
- Adding histogram at detector level (purity) with bins
  + all det jets
  + tagged jets
  + part jet inside det-level acceptance
- Adding histogram at particle level (efficiency) with bins
  + all part jets
  + tagged jets
  + det jet inside part-level acceptance


mfasDa:
+1

PR #15635, opened 2020-09-26
Fixes for missing include and forward declarations of TObjString

pzhristov:
Discovered when compiling with Root 6-22-02


pzhristov:
+test

PR #15650, opened 2020-09-28
Added a check for pythia header in the case that it is not available

osaarima:


osaarima:
Hello,
can this commit be checked? It is a very short fix.

dongjokim:
+1

PR #15655, opened 2020-09-28
PWGGA: Always use fHistoGoodMesonClusters, when phos trigger is used

jluehder:
Also add information, which ddls were bad and reduce amount of histograms in trigger helper


dmuhlhei:
+1

PR #15660, opened 2020-09-28
Charged Particle Spectra

phuhnGithub:
Added pile-up for MC

### DIFF
--- a/PWGLF/SPECTRA/ChargedHadrons/dNdPt/Framework/AliAnalysisTaskMKBase.cxx
+++ b/PWGLF/SPECTRA/ChargedHadrons/dNdPt/Framework/AliAnalysisTaskMKBase.cxx
@@ -8,6 +8,7 @@
 #include "AliESDtrack.h"
 #include "AliESDtrackCuts.h"
 #include "AliEventCuts.h"
+#include "AliAnalysisUtils.h"
 #include "AliGenDPMjetEventHeader.h"
 #include "AliGenEventHeader.h"
 #include "AliGenPythiaEventHeader.h"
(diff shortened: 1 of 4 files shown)


PWGLF/SPECTRA/ChargedHadrons/dNdPt/Framework/AliAnalysisTaskMKBase.cxx
```diff
@@ -1021,6 +1023,10 @@ Bool_t AliAnalysisTaskMKBase::InitMCParticle()
     fMCPrimSec = 2;
     fMCProdcutionType = AlidNdPtTools::kSecMaterial;
   }
+  if(fMCPileUpTrack){
+    fMCPrimSec = 3;
+    fMCProdcutionType = AlidNdPtTools::ProductionType::kPileUpTrack;
+  }
```
mkruegerGitHub, line 1029: maybe it is better to make lines 1014 - 1029 an if, else if, else if statement and add an else clause setting the two variables to -1 or something similar.
Otherwise fMCPrimSec and fMCProdcutionType could be uninitialized.

mkruegerGitHub: okay thinking about it again 'else if' might be a bad idea since fMCisPrim is probably also true for pileup (as we have seen in your efficiency plots) but maybe we could initialize the variables with something like this before the if clause:
fMCPrimSec = -1;
fMCProdcutionType = AlidNdPtTools::ProductionType::kUnknown;

mkruegerGitHub:
+1

PR #15671, opened 2020-09-29
Back to DecayAndDecay PC for LambdaXi and Paircleaning Fix always selecting Lambda by accident!

magicep0t:
Paircleaning fixes
	*stop checking trackID if first particle is unselected
	*corrected filling of mass histos of thrown out particles
	*now NOT absolute value of difference to PDG mass
	*better histo ranges
	*runtime optimization with better creation of objects and position of if's in loop


sartozza:
+1

PR #15681, opened 2020-09-30
PWGGA/GammaConv    ncell cut

AMechler:
changing phoslike ncell cut to be also taking into account in m02 cut
changing agound some cuts PCM-EDC pp 13 TeV


dmuhlhei:
+1

PR #15682, opened 2020-09-30
PWGGA/GammaConv: additional quantities added to ElectronStudies tree

fjonasALICE:
-  additional quantities added to ElectronStudies tree
- some fixes for running over ESD with GammaCalo task


nschmidtALICE:
+1

PR #15690, opened 2020-09-30
PWG/EMCAL: Added option to apply exotics cut on clusters directly

FriederikeBock:
- Added option to directly apply exotics as flagged by the ClusterExotics component on the clusters


FriederikeBock:
+1

PR #15693, opened 2020-10-01
PWGGA/GammaConv changing PCM-EDC pp 13 Tev cuts

AMechler:
changing PCM-EDC pp 13 Tev cuts MB, adding trigger cuts and changing lowb cuts


FriederikeBock:
+1

PR #15709, opened 2020-10-01
PWGGA/GammaConv: add MB header for LHC20g10

sstiefel19:
for LHC20g10: add posibility to find MB header
at other posistion than 0

also add a few todos


dmuhlhei:
+1

PR #15715, opened 2020-10-02
Update task with TPC tracks cuts and separate EMCal and DCal detector

pdhankhe:
The task is updated for the latest TPC track cut recommendations (eg. ncrossrows) and also for the separate analysis of EMCal and DCal. The updated event histogram shows the number of events as a function of different event selection cuts and different triggered sample.


adubla:
+1

PR #15722, opened 2020-10-02
Fix handling of passing the number of events

mfasDa:
The add macro for the generator is running on the
worker and must assume the environment is not set.
In this case the macro needs to get the number of
events from the analysis manager.

In addition added a little bit of documentation.


mfasDa:
The macro can only be used once https://github.com/alisw/AliRoot/pull/1248 is tagged.

PR #15734, opened 2020-10-05
PWGGA: Add Histograms for trigger checks in PHOS

jluehder:
ClusE vs Timing for each TRU
Distance for TRU patches


nschmidtALICE:
+1

PR #15750, opened 2020-10-06
PWGHF/hfe: adapted AliAnalysisTaskHFEBESpectraEMC.cxx for diff cluster lists

FriederikeBock:
Adapted AliAnalysisTaskHFEBESpectraEMC.cxx task in order to be able to load different cluster and track lists based on the Emcal correction framework. Default behaviour will stay as it is.


FriederikeBock:
+1

PR #15757, opened 2020-10-07
Xic0Semileptonic HM

jsbok:
adding an option for HM in Xic0Semileptonic


fgrosa:
+test

PWGHF/vertexingHF/AliAnalysisTaskSEXic0Semileptonic.cxx
```diff
@@ -56,57 +57,63 @@
 #include "AliESDtrackCuts.h"
 #include "AliESDtrack.h"
 #include "AliAODInputHandler.h"
-//#include "AliAnalysisTaskSEHFTreeCreator.h"
+#include "AliAnalysisTaskSEHFTreeCreator.h"
```
fgrosa, line 60: Please remove the inclusion of AliAnalysisTaskSEHFTreeCreatos since it is not needed and it would require the inclusion of the treeHF library in the dependencies of vertexingHF, which is not possible since treeHF already depends on vertexingHF.

PWGHF/vertexingHF/AliAnalysisTaskSEXic0Semileptonic.h
```diff
@@ -15,9 +9,9 @@
 #include "TVector2.h"
 #include "TSystem.h"
 #include "TProfile.h"
-#include <vector>
 #include "THistManager.h"
 #include "AliAnalysisTaskSE.h"
+#include "AliMultiplicitySelectionCPPWA.h"
```
fgrosa, line 14: It seems that this is not needed, is it correct? If so, please remove it because otherwise it is necessary to include the PWGUDdiffractive library in vertexingHF, which is not necessary (and in any case not optimal)

fgrosa (commented):
Hi @jsbok, thanks for the implementations! Could you please have a look at the comments regarding the includes? Cheers, Fabrizio

PR #15770, opened 2020-10-08
p-Pb AMPT

ysekiguc:


chiarazampolli:
+1

chiarazampolli:
I cannot really judge if this is fine or not, but it is a macro, in the worst case, fixes will be needed, nothing will break.

PR #15772, opened 2020-10-08
AliAnalysisTaskHFJetIPQA:: consider all final state V0 daughter parti…

kagarner:
…cles for max. IP estimation

### DIFF
--- a/PWGJE/FlavourJetTasks/AliAnalysisTaskHFJetIPQA.h
+++ b/PWGJE/FlavourJetTasks/AliAnalysisTaskHFJetIPQA.h
@@ -15,7 +15,6 @@ class AliHFEpidBayes;
 class TTree;
 class TH2D;
 class TCanvas;
-class TParticle;
 class TClonesArray;
 class AliAODMCParticle;
 class AliMCEvent;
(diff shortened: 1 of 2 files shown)


PWGJE/FlavourJetTasks/AliAnalysisTaskHFJetIPQA.h
```diff
@@ -298,7 +297,7 @@ class AliAnalysisTaskHFJetIPQA: public AliAnalysisTaskEmcalJet
     }
     Bool_t IsParton(int pdg);
     Bool_t IsParticleInCone(const AliVParticle* part, const AliEmcalJet* jet, Double_t dRMax);
-    Int_t NDaughterInCone(AliAODMCParticle* posdaugh, const AliAODMCParticle* negdaugh, const AliEmcalJet* jet, const AliAODEvent* event, Double_t dRMax, Double_t& ipsig);
+    Int_t NDaughterInCone(vector<Int_t>& vecDaughLabels, const AliEmcalJet* jet, const AliAODEvent* event, Double_t dRMax, Double_t& ipsig);
```
mfasDa, line 300: Please replace all remaining occurrences of vector with std::vector. The "using namespace std" comes in implicitly from AliJetShapes.h and will be removed in the near future - it is super dangerous.

mfasDa (changes requested):
Never rely on "using namespace std" in header files. This is super dangerous!

lcunquei:
+1

PR #15775, opened 2020-10-08
Add option for enable/disable time calib

mfasDa:
Default value is enabled, so the switch is there to
switch off time recalibration.

Load Phase calib also in case time calib is disabled


mfasDa:
@FriederikeBock @raymondEhlers Please review!

raymondEhlers:
LGTM

gconesab:
Hello, why do you need an extra switch? Is it not enough to diable the full component?

mfasDa:
Not sufficient. The component does L1 phase shift correction and time calib. I need this to produce data for run3 where we apply only the L1 phase shift correction but not the time calib for calibration and QC development, and for run3 for the moment we assume that the phase shift correction will be done in the firmware.

gconesab:
OK, thanks for the clarification.

FriederikeBock:
+1

PR #15785, opened 2020-10-09
Fixed a problem with folder specific includes.

osaarima:
This caused problem on the grid side when running. Local running was still ok.


dongjokim:
+test

dongjokim:
+1

PR #15786, opened 2020-10-09
Moved Finishing the TF in FinishTaskOutput.

mgheata:
Fixed final writing of the trees in the last TF_n folder created in AO2D.root during conversion.


pzhristov:
+1

PR #15790, opened 2020-10-10
Including New Case

hdegen:
New selection cut to study the effects w&w/o PU rejection to be able to compare to the Poisson correction applied in mu analyses


hdegen:
Dear all,

This is only a small change, only configuring a new set of cuts to be able to compare the results we obtain when applying pile up rejection and when we don't reject pile up but apply a Poisson correction. These 2 methods should present compatible results. Up to now, we have only histograms with all of the cuts applied together, but we do not have one histogram with all of these cuts except the pile up rejection. Can someone give a +test or +1, please? Feel free to contact me in private, if any doubts or problems with the modifications.

Thanks and cheers!
Hermann

aortizve:
+test

aortizve:
+1

PR #15839, opened 2020-10-15
Update hypertriton task using KFParticle

SHornung1:
Performance improvements
2-body and 3-body decay can be run within the same code now
Updated preselection


mkruegerGitHub:
+1

PR #15845, opened 2020-10-15
Fix cell time cut in filling the summed cell grid

mfasDa:
- use correct function (GetTime() instead of
  GetCellTime())
- Explicitly state that the time cut is implemented in
  nanoseconds and not in seconds by adding tag to
  variable and function names, in order to avoid
  confusion
- Add monitoring histograms for cell time before and
  after cell time cut
- Make use of fCaloCells provided by
  AliAnalysisTaskEmcal


mfasDa:
+1

PR #15852, opened 2020-10-15
added toward/away side, mult is now always UE mult

krschmit-cern:
- observables can now be studied in toward and away side (additionally to transverse side)
- multiplicity is now always counted in transverse side (UE)


mkruegerGitHub:
+1

PR #15866, opened 2020-10-16
PWG/EMCal: low gain time calib fix for Run1

fjonasALICE:
- added missing line for LowGainTimeCalib to properly deactivate it in case of Run1. Previously the time calibration completely failed if user has enable lowGainTimeCalib for Run1 data
- rest of commit are things for my cluster merging studies that are unrelated to timecalib


nschmidtALICE:
+1

mfasDa:
+1

PR #15868, opened 2020-10-18
Modification of AddTask to run with different set up in one LEGO train

SHornung1:
Moved trees inside a container with different name depending on the set up of the add task


bdoenigu:
+1

PR #15870, opened 2020-10-19
Add asso-pt information

SiyuTang:
Hi,

Here is an update for Jet v2 analysis code, where the asso pt information is added for trigger pair.

Thanks! 

Siyu


yzhouCERN:
+1

PR #15871, opened 2020-10-19
Add cluster energy type to setup of the cluster container in order to…

mfasDa:
… be able to run the soft drop response without hadronic correction


mfasDa:
+1

PR #15884, opened 2020-10-20
Fixes in histograms for SoftDrop response

mfasDa:
- Increase histogram range of the Residuals
  histogram and split residuals histograms
  for Rg and Thetag
- Move Jet finding efficiency histograms to
  parent group in order to have tehm always
  filled


mfasDa:
+1

PR #15886, opened 2020-10-20
Re-organization of failed part. level SoftDrop

mfasDa:
SoftDrop failure handling separated for det. level and
part. level:
- Failed handling at part. level treated as impurity in det.
  level loop
- Failed handling at det. level treated as inefficiency in
  part. level loop
Consequently increasing the number of tag bins from 3 to 4.
Only pairs of part. and det. level jets entering the
response matrix.


mfasDa:
+1

PR #15887, opened 2020-10-20
PWGGA/GammaConv: AliAnalysisTaskOmegaToPiZero - fixed bug in background calculation and added train configs

mhemmer-cern:
- In the background calculations the variable gamma0 was redeclared and overwritten inside the last loop. This is now fixed renaming the correct calls of gamma0 to gamma2.
- Added many new train configs which mostly are copies of existing ones to be able to run the same cut string, but with different AddTask call options.


dmuhlhei:
+1

PR #15894, opened 2020-10-20
Implement possibility to collect and embed signal MC files only for the data run block range used to anchor MC  

gconesab:
Make sure that given a grid job only the MC files anchored to a run  that corresponds to the block run range of the input data file are used for the embedding.
Implemented only for analysis in trains running analysis per run.

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
+++ b/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
@@ -634,7 +634,7 @@ void AliAnalysisTaskEmcalEmbeddingHelper::SetRunblockRange()
   
   if ( nBlocks == 0 ) return ;
   
-  AliDebug(1,Form("AliAnalysisTaskEmcalEmbeddingHelper::SetRunblockRange() - nRunBlocks %d\n", nBlocks));
+  AliDebug(1,Form("nRunBlocks %d\n", nBlocks));
   
   // Recover the run number from the input data file path.
   //
@@ -653,18 +653,17 @@ void AliAnalysisTaskEmcalEmbeddingHelper::SetRunblockRange()
   
   Int_t  runMin = -1,      runMax = -1;
   for (Int_t iblock = 0; iblock < nBlocks-1; iblock++) 
-  {
-    std::istringstream(fEmbeddedRunblock.at(iblock  )) >> runMin;
-    std::istringstream(fEmbeddedRunblock.at(iblock+1)) >> runMax;
-    //    runMin = std::stoi(fEmbeddedRunblock.at(iblock  ));
-    //    runMax = std::stoi(fEmbeddedRunblock.at(iblock+1));
-    AliDebug(1,Form("\t block %d, run min %d, run max %d\n",iblock,runMin,runMax));
+  {  
+    runMin = fEmbeddedRunblock.at(iblock  );
+    runMax = fEmbeddedRunblock.at(iblock+1);
+    
+    AliDebug(1,Form("\t block %d, run min %d, run max %d",iblock,runMin,runMax));
     
     if ( runMin <= fDataRunNumber && runMax > fDataRunNumber ) 
     {
       fEmbeddedRunblockMin = runMin;
       fEmbeddedRunblockMax = runMax;
-      AliDebug(1,Form("\t Selected range %d, run min %d, run max %d\n",
+      AliDebug(1,Form("\t Selected range %d, run min %d, run max %d",
                       iblock, fEmbeddedRunblockMin, fEmbeddedRunblockMax));
       
       return ;
@@ -684,14 +683,14 @@ bool AliAnalysisTaskEmcalEmbeddingHelper::IsRunInRunblock(const std::string & pa
 {
   if ( fEmbeddedRunblockMax < 0 && fEmbeddedRunblockMin < 0 ) return true;
   
-  AliDebug(1,Form("Run range [%d,%d]\n",
+  AliDebug(1,Form("Run range [%d,%d]",
          fEmbeddedRunblockMin,fEmbeddedRunblockMax));
-
+ 
   const char * cpath = path.c_str();
 
   Int_t mcrun = AliAnalysisManager::GetRunFromAlienPath(cpath);
   
-  AliDebug(1,Form("file: %s, run %d\n",cpath,mcrun));
+  AliDebug(1,Form("\t file: %s, run %d",cpath,mcrun));
 
   if ( mcrun >= fEmbeddedRunblockMin && 
        mcrun <  fEmbeddedRunblockMax ) 
@@ -898,6 +897,7 @@ void AliAnalysisTaskEmcalEmbeddingHelper::DetermineFirstFileToEmbed()
         fFilenameIndex = TMath::FloorNint(rand.Rndm()*fFilenames.size());
         const char* path = (fFilenames.at(fFilenameIndex)).c_str();
         ok = IsRunInRunblock(path);
+        //printf("\t Path %s, first? %d \n",path,ok);
       }
       
       if ( !ok ) 
@@ -1448,10 +1448,13 @@ Bool_t AliAnalysisTaskEmcalEmbeddingHelper::SetupInputFiles()
   // Keep track of the total number of files in the TChain to ensure that we don't start repeating within the chain
   fMaxNumberOfFiles = fChain->GetListOfFiles()->GetEntries();
 
-  if (fFilenames.size() > fMaxNumberOfFiles) {
+  if ( fFilenames.size() > fMaxNumberOfFiles && fEmbeddedRunblock.size() == 0 ) {
     AliErrorStream() << "Number of input files (" << fFilenames.size() << ") is larger than the number of available files (" << fMaxNumberOfFiles << "). Something went wrong when adding some of those files to the TChain!\n";
   }
-
+  else {
+    AliInfo(Form("Final number of input files to embed %d",fMaxNumberOfFiles));
+  }
+  
   // Setup input event
   Bool_t res = InitEvent();
   if (!res) return kFALSE;
--- a/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
+++ b/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
@@ -151,8 +151,6 @@ AliAnalysisTaskEmcalEmbeddingHelper::AliAnalysisTaskEmcalEmbeddingHelper() :
   fConfigurationPath(""),
   fEmbeddedRunlist(),
   fEmbeddedRunblock(),
-  fEmbeddedRunblockMin(-1),
-  fEmbeddedRunblockMax(-1),
   fDataRunNumber(-1),
   fPythiaCrossSectionFilenames(),
   fExternalFile(nullptr),
@@ -230,8 +228,6 @@ AliAnalysisTaskEmcalEmbeddingHelper::AliAnalysisTaskEmcalEmbeddingHelper(const c
   fConfigurationPath(""),
   fEmbeddedRunlist(),
   fEmbeddedRunblock(),
-  fEmbeddedRunblockMin(-1),
-  fEmbeddedRunblockMax(-1),
   fDataRunNumber(-1),
   fPythiaCrossSectionFilenames(),
   fExternalFile(nullptr),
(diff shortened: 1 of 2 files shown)


gconesab:
+test

gconesab:
Hi @raymondEhlers, whenever you have a moment, please have a look and let me know what you think. Not urgent, but I would like to test it soon.

raymondEhlers:
Hi @gconesab I'll take a look tomorrow morning. I'm a bit confused about your last comment - have you tested it locally?

Also, just from a very quick look, I'm confused why your runblock is a std::vector<std::string>> - why not a std::vector<std::pair<int, int>> (for a min, max for each range) or std::vector<int> (assuming they go from one to the next)? yaml can load this directly

gconesab:
Hi Raymond,
I am not used to vectors so I am not familiar with all functionalities, I just saw in the code how it was handled for the single runlist selection case and used the same. In the current approach I just assume that each entry of the list is a run number larger than the next entry and that a block range is defined from entry i (>= run number) to entry i+1 (< run number). I can try to implement the vector pair, not sure how to pass it via the yaml.
Thanks ... Gustavo

raymondEhlers:
Hi Gustavo,

Okay, I see. I don't recall why std::vector<std::string> was used for the run list, but I think let's try to do it right here. It will give us extra type validation, which should make it easier to use. Since you're just assuming that the block runs from one to the next (pairs aren't necessary in that case), you can do something like:

```yaml
runBlocks: [4, 6, 8, 19]
```

```c++
// Define embedding helper member as (pick the name as you like):
std::vector<int> fRunBlocks;
// Load from config with:
fYAMLConfig.GetProperty("runBlocks", fRunBlocks, false);
```

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
```diff
@@ -615,6 +623,83 @@ bool AliAnalysisTaskEmcalEmbeddingHelper::IsRunInRunlist(const std::string & pat
...
+ * Determine the MC run block range,  fEmbeddedRunblockMin and  fEmbeddedRunblockMax (used in IsRunInRunblock())
+ * Right now only done once, for the first data file
+ */
+void AliAnalysisTaskEmcalEmbeddingHelper::SetRunblockRange() 
+{
+  Int_t nBlocks = fEmbeddedRunblock.size();
+  
+  if ( nBlocks == 0 ) return ;
+  
+  AliDebug(1,Form("AliAnalysisTaskEmcalEmbeddingHelper::SetRunblockRange() - nRunBlocks %d\n", nBlocks));
```
raymondEhlers: Please use `AliDebugStream(1) << "nRunBlocks " << nBlocks << "\n";` No doubt it's not always used everywhere, but I'd like to try to maintain it when possible

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
```diff
@@ -1126,7 +1233,9 @@ Bool_t AliAnalysisTaskEmcalEmbeddingHelper::InitEvent()
  */
 void AliAnalysisTaskEmcalEmbeddingHelper::UserCreateOutputObjects()
 {
-  SetupEmbedding();
+  if ( fEmbeddedRunblock.size() == 0 ) 
+    SetupEmbedding();
```
raymondEhlers, line 1237: Please wrap in curly braces

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
```diff
@@ -798,7 +883,29 @@ void AliAnalysisTaskEmcalEmbeddingHelper::DetermineFirstFileToEmbed()
...
+        fFilenameIndex = TMath::FloorNint(rand.Rndm()*fFilenames.size());
+        const char* path = (fFilenames.at(fFilenameIndex)).c_str();
+        ok = IsRunInRunblock(path);
+      }
+      
+      if ( !ok ) 
+        AliFatal("No file found with the run range requeirements, check list of MC files and block run ranges STOP!");
+    }
+    else
+      fFilenameIndex = TMath::FloorNint(rand.Rndm()*fFilenames.size());
```
raymondEhlers, line 907: Please wrap in curly braces

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
```diff
@@ -798,7 +883,29 @@ void AliAnalysisTaskEmcalEmbeddingHelper::DetermineFirstFileToEmbed()
...
+      while ( !ok && iter <= nfiles*10 )
+      {
+        iter++;
+        fFilenameIndex = TMath::FloorNint(rand.Rndm()*fFilenames.size());
+        const char* path = (fFilenames.at(fFilenameIndex)).c_str();
+        ok = IsRunInRunblock(path);
+      }
+      
+      if ( !ok ) 
+        AliFatal("No file found with the run range requeirements, check list of MC files and block run ranges STOP!");
```
raymondEhlers, line 904: Please wrap in curly braces

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
```diff
@@ -615,6 +623,83 @@ bool AliAnalysisTaskEmcalEmbeddingHelper::IsRunInRunlist(const std::string & pat
...
+/**
+ * Get the data background run number from the data file path, fRunNumberBgk and 
+ * Determine the MC run block range,  fEmbeddedRunblockMin and  fEmbeddedRunblockMax (used in IsRunInRunblock())
+ * Right now only done once, for the first data file
+ */
+void AliAnalysisTaskEmcalEmbeddingHelper::SetRunblockRange() 
+{
+  Int_t nBlocks = fEmbeddedRunblock.size();
+  
+  if ( nBlocks == 0 ) return ;
```
raymondEhlers, line 635: Please wrap in curly braces

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
```diff
@@ -615,6 +623,83 @@ bool AliAnalysisTaskEmcalEmbeddingHelper::IsRunInRunlist(const std::string & pat
...
 }
 
+/**
+ * Get the data background run number from the data file path, fRunNumberBgk and 
+ * Determine the MC run block range,  fEmbeddedRunblockMin and  fEmbeddedRunblockMax (used in IsRunInRunblock())
+ * Right now only done once, for the first data file
+ */
+void AliAnalysisTaskEmcalEmbeddingHelper::SetRunblockRange() 
+{
+  Int_t nBlocks = fEmbeddedRunblock.size();
```
raymondEhlers, line 633: This should be `std::size_t`

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
```diff
@@ -615,6 +623,83 @@ bool AliAnalysisTaskEmcalEmbeddingHelper::IsRunInRunlist(const std::string & pat
...
+    if ( runMin <= fDataRunNumber && runMax > fDataRunNumber ) 
+    {
+      fEmbeddedRunblockMin = runMin;
+      fEmbeddedRunblockMax = runMax;
+      AliDebug(1,Form("\t Selected range %d, run min %d, run max %d\n",
+                      iblock, fEmbeddedRunblockMin, fEmbeddedRunblockMax));
+      
+      return ;
+    }
+  } // block loop
```
raymondEhlers: You may want to checkout `std::lower_bound` and `std::upper_bound` (in conjunction with converting to std::vector). It will do all of this for you. You just have to be a little careful with the index

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
```diff
@@ -615,6 +623,83 @@ bool AliAnalysisTaskEmcalEmbeddingHelper::IsRunInRunlist(const std::string & pat
...
+  AliAnalysisManager *mgr = AliAnalysisManager::GetAnalysisManager();
+   
+   if ( !mgr ) return ;
+  
+  Int_t runNumber = mgr->GetRunFromPath(); 
+  if ( fDataRunNumber != runNumber )
+  {
+    //printf("\t Run number, new %d, previous %d\n",runNumber,fDataRunNumber);
+    fDataRunNumber = runNumber;
+  }
```
raymondEhlers, line 652: To get the internal event (and the run number), please use AliAnalysisTaskSE::InputEvent()

raymondEhlers, line 652: I don't understand why you wouldn't want to set `fDataRunNumber`. This is the only place that you set it

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
raymondEhlers: Also, I'm not sure why you got through all of this work - I thought you were going to go in a different direction.

At this point, you're in a given job (ie only impacting a single job), so when you're in the runblock mode, why not just remove the filenames that you're not interested in from `fFilenames`? Something like `std::remove_if` could do this for you. And then I don't think you need many modifications further on in the code because all filenames are just ones that you want

gconesab:
Hi Raymond, I changed the vector of strings to ints and a couple minor fixes. My local limited tests seem to work, let me know what you think.

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
```diff
@@ -798,7 +883,29 @@ void AliAnalysisTaskEmcalEmbeddingHelper::DetermineFirstFileToEmbed()
...
+      {
+        iter++;
+        fFilenameIndex = TMath::FloorNint(rand.Rndm()*fFilenames.size());
+        const char* path = (fFilenames.at(fFilenameIndex)).c_str();
+        ok = IsRunInRunblock(path);
+      }
+      
+      if ( !ok ) 
+        AliFatal("No file found with the run range requeirements, check list of MC files and block run ranges STOP!");
+    }
```
raymondEhlers: If you remove the filenames that are not of interest, I think you don't need to modify this function at all

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
```diff
@@ -1605,6 +1721,15 @@ void AliAnalysisTaskEmcalEmbeddingHelper::UserExec(Option_t*)
...
   }
+  else if(fEmbeddedRunblock.size()!=0)
+  {
+    // I do not think this is enough, I am not sure how to know if the train 
+    // was setup for run by run or run mixed analysis
+    AliAnalysisManager *mgr = AliAnalysisManager::GetAnalysisManager();
+    if ( mgr && fDataRunNumber > -1 && fDataRunNumber!=mgr->GetRunFromPath() )
+      AliError(Form("CAREFUL! Check what you are doing, you are embedding a data run %d that belongs to another runblock [%d,%d]!, setup the train to do the analysis per Run",
+                 InputEvent()->GetRunNumber(),fEmbeddedRunblockMin,fEmbeddedRunblockMax));
+  }
```
raymondEhlers, line 1735: If you remove unwanted filenames, I think this also isn't an issue anymore. This is hard to get right (I think even my solution isn't ideal, but at least it's reasonably well tested - I could imagine modify https://github.com/alisw/AliPhysics/blob/719ea3248a354a47a2082a510d98600d2d735bf4/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx#L965-L968 to return early using the same mechanism as the internal event selection (here: https://github.com/alisw/AliPhysics/blob/719ea3248a354a47a2082a510d98600d2d735bf4/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx#L1792-L1795). But first let's focus on the general design)

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.h
```diff
@@ -358,6 +360,10 @@ class AliAnalysisTaskEmcalEmbeddingHelper : public AliAnalysisTaskSE {
   std::vector <std::string>                     fFilenames        ; ///<  Paths to the files to embed
   std::string                                   fConfigurationPath; ///<  Path to %YAML configuration
   std::vector <std::string>                     fEmbeddedRunlist  ; ///<  Good runlist for files to embed
+  std::vector <std::string>                     fEmbeddedRunblock ; ///<  Good runblock for files to embed, each run corresponds to the first run of a block
+  Int_t                                         fEmbeddedRunblockMin; ///< Minimum run in the run range valid for embedding
+  Int_t                                         fEmbeddedRunblockMax; ///< Maximum run in the run range valid for embedding, this one is first of next block  
+  Int_t                                         fDataRunNumber    ;   ///< Current run number in data file
```
raymondEhlers, line 366: If you remove files that you're not interested in, it's not clear to me if even this variables are necessary. But perhaps I'm being too simple for this

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
```diff
@@ -1310,6 +1422,10 @@ Bool_t AliAnalysisTaskEmcalEmbeddingHelper::SetupInputFiles()
       wrapped = true;
     }
 
+    // Accept if the MC file has a run number that belongs 
+    // to the same run block range as the data run used to anchor the MC
+    if ( !IsRunInRunblock(filename->c_str()) ) continue;
```
raymondEhlers, line 1427: If you remove unwanted filenames, I think this also isn't an issue anymore

raymondEhlers (changes requested):
Hi @gconesab ,

Assuming that I understand what you've done, I think you can simplify this a lot by just removing filenames from `fFilenames` that aren't of interest. As long as you're doing this in each job, it should be fine.

raymondEhlers:
> Hi Raymond, I changed the vector of strings to ints and a couple minor fixes. My local limited tests seem to work, let me know what you think.

Thanks, it looks like it's going in the right direction. I found time to fully review it, and I think we could go about it in a much simpler way (but let me know if I'm overlooking something)

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
gconesab: OK

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
gconesab: OK

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
gconesab: OK

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
gconesab: OK

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
gconesab: OK

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
gconesab: ok

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
gconesab: I am not sure how those work, I prefer to stick to what I know, but you can change it

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
gconesab: I tried, it does not work, I have the impression it takes the input event number from the external event and at this stage it is not yet set, I get 0, that is why I am using this. Maybe there is something wrong in my setup.

gconesab: Just in case we decide in the future the analysis of different data runs on the same analysis and then we have to decide when to change the list of files

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
gconesab: Note this is executed only once in the first event of the job. Here it is the place where the run range block is found based on the data run number. Here, I just to set once the members fEmbeddedRunblockMin and fEmbeddedRunblockMax and fDataRunNumber that are used later. I did like this in case in the future we decide to make this for the case other DataRuns can appear in the analysis.

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
gconesab: But this is what is done here, remove the filenames not of interest. I have not seen where to do this before here.

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.h
gconesab: This is just in case of future development when the run data changes and in case a mistake is done to know when the run number changed

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
gconesab: This is the place where the unwanted files are removed, I have not found the place it is done.
Let's say, I found it simpler to skip the files before adding them to the chain than modifying the vector with the list of files, I was not sure how to do it.

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
raymondEhlers: If this is called during `UserExec()` (which is should for your case), it has to be defined. Note that it must be `AliAnalysisTaskSE::InputEvent()`, not just `InputEvent()` (the latter **will** get the external event)

gconesab: Checked and it works, I am about to commit

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
```diff
@@ -615,6 +619,79 @@ bool AliAnalysisTaskEmcalEmbeddingHelper::IsRunInRunlist(const std::string & pat
...
+  std::cout << "    Selected run range block "<< iblock <<": ["<<runMin<<","<<runMax<<"]"<<std::endl;
+  
+  if ( runMin < 0 && runMax < 0 ) {
+    AliFatal("Runblock not found, stop!");
+    return ;
+  }
+
+  // Filter the list of files
+  //
+  std::cout << "    Size of filenames before filtering "<<  fFilenames.size() << std::endl;
```
raymondEhlers, line 680: When I suggested streams, I was thinking along the lines of `AliDebugStream`, `AliInfoStream`, etc. It would probably be best to cleanup these `std::cout`

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
```diff
@@ -1602,9 +1689,27 @@ bool AliAnalysisTaskEmcalEmbeddingHelper::PythiaInfoFromCrossSectionFile(std::st
...
+      runNumber = AliAnalysisTaskSE::InputEvent()->GetRunNumber();
+    }
+    
+    if ( fDataRunNumber > -1 && fDataRunNumber != runNumber ) {
+      AliError(Form("CAREFUL! Check what you are doing, you are embedding a data run %d"
+                    " but block range was anchored to %d"
+                    " setup the train to do the analysis per Run",
+                    AliAnalysisTaskSE::InputEvent()->GetRunNumber(), fDataRunNumber));
+    }
+  }
```
raymondEhlers, line 1712: I think this condition could be refined: it's fine if it changes runs, it just can't go outside the runblock. I didn't totally understand your conditions when we first discussed this. However, I think it's probably fine for now since you'll run it run-by-run

raymondEhlers (approved):
Hi @gconesab ,

Thanks for all of the improvements - it looks good to me. I added a few small comments, but I think it's basically ready to approve.

Best,
Raymond

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
gconesab: OK thanks

PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx
gconesab: At some point I was not seeing the prints and then used cout, now changed to AliInfoStream and AliDebugStream

gconesab:
+test

gconesab:
+1

PR #15902, opened 2020-10-21
For PID Reference Parameters

ailec:
To try this new parameters for the LHC18q pile up cut and see if better pion efficiency TRD performance is obtain.


mbroz84:
+1

mbroz84:
+1

PR #15904, opened 2020-10-21
Create new object with hard parton info from stack and add it as tran…

mfasDa:
…sient object to the event

Object providing info about the initial hard process(es)


mfasDa:
+1

mfasDa:
+1

PR #15906, opened 2020-10-21
Added a new way to set the total number of runs

hdegen:
Instead of a fixed set of nRuns, with this mod anyone will be available to run different run lists without the need to change the add task. The common number of runs used for stability is there, but if the option nRuns is set then the number of runs to fill the histograms will be modified accordingly.


hdegen:
Dear all,

That's just a really small change and should not affect even older configurations. Could someone please give a +test or +1, if available?

Thank you all in advance!
Hermann

valzacc:
+test

valzacc:
+1

PR #15913, opened 2020-10-22
PWGGA/GammaConv: ElectronStudies fixed types & added EMC CF default conf

fjonasALICE:
- fixed wrong types for momentum in ElectronStudies tree
- added support for eta meson in merging studies
- added yaml configs intended to be used as new default for CF of GA


FriederikeBock:
+1

PR #15914, opened 2020-10-22
Fixes for Herwig7 production

mfasDa:
- Add lightweight data object that is owned by the MC parton container
- Enable getting parton info object in the AliAnalysisTaskEmcal if the container name is set


mfasDa:
+1

mfasDa:
+1

mfasDa:
+1

PR #15920, opened 2020-10-22
changing time frame id to use event time

jgrosseo:


pzhristov:
May be you can increment the ClassDef even if we do not really store the converter.

pzhristov:
+1

PR #15924, opened 2020-10-23
update parametrizations from binshifted spectra and include all syst …

m-c-danisch:
…variations


nschmidtALICE:
There is at least one parametrization in this file that makes no sense and at least one that does not work at all. Please check.

nschmidtALICE:
+1

PR #15930, opened 2020-10-23
Provide function configuring MC type and pt-hard binning based on dat…

mfasDa:
…aset name

Function currently contains a list of known datasets.
More datasets added on request.


mfasDa:
+1

PR #15931, opened 2020-10-23
FemtoNanoLD: use FB 256 for deuteron purities

stheckel:
o For 3 subwagons, all with FB 256 for deuterons
o Std settings, no TPC PID, no TOF PID


aohlson:
+1

PR #15949, opened 2020-10-26
pre selection cut changes, updated vtxstatus content, and info for event loss estimation + restr…

tuvarichert:
…ucture and cleanup


bdoenigu:
+1

PR #15950, opened 2020-10-26
PWGGA: No Preselction for Omega in PCM when flag is set kFALSE

jluehder:


fjonasALICE:
Since this is for the omega analysis, let me just comment here that this looks good from my side!

nschmidtALICE:
+1

PR #15962, opened 2020-10-27
Switch order of acceptance and SoftDrop requirement

mfasDa:
Acceptance must be evaluated after the SoftDrop
at other (part or det and vice versa) level, as having
the same acceptance type is not a strict requirement
for the analysis, while having the SoftDrop for both
part. and det. level jet is. Evaluating the SoftDrop after
the acceptance cut masks the acceptance status in
the efficiency and purity histogram in almost all cases
leading to efficiency and purity ~1


mfasDa:
+1

PR #15975, opened 2020-10-29
Fixing bug related to passing container names

deepathoms:
Changing the way cluster and track container names from the EMCal correction framework are passed to the task


deepathoms:
+1

PR #15977, opened 2020-10-29
Add possibility to perform application for multi-class models

fgrosa:
Add several methods for multiclass model application: 

- `PredictMultiClass`
- `IseSelectedMultiClass` 

which work for both binary and multiclass models.
The old methods for binary models 

- `Predict`
- `IseSelected` 

are kept for retrocompatibility.

The yaml config file can be as the old ones, or can take a list of threshold cuts (`cuts` entry) instead of a single value. An optional entry can also be added (`cuts_opt`) to specify if they are "lower" or "upper" cuts, which can be also either a single value or a list in case of multiclass classification.

### DIFF
--- a/ML/AliExternalBDT.cxx
+++ b/ML/AliExternalBDT.cxx
@@ -140,11 +140,15 @@ bool AliExternalBDT::Predict(double *features, int size, std::vector<double> &ou
   }
 
   std::vector<float> output(fOutSize);
-  TreelitePredictorPredictInst(fPredictor, entries.data(),
+  int predict = TreelitePredictorPredictInst(fPredictor, entries.data(),
       static_cast<int>(useRawScore), &output[0],
       &fOutSize);
+  if(predict<0)
+    return false;
+
   for (std::size_t iEntry = 0; iEntry < fOutSize; ++iEntry) {
     outputScores.push_back(static_cast<double>(output[iEntry]));
   }
+
   return true;
 }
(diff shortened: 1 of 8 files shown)


fcatalan92:
@pfecchio @fmazzasc @mpuccio FYI

ML/AliExternalBDT.cxx
```diff
@@ -120,24 +122,33 @@ bool AliExternalBDT::LoadLightGBMModel(std::string path) {
...
+
   if (status != 0) {
     std::cerr << "Library loading failed" << std::endl;
     return false;
   }
   return true;
 }
 
-double AliExternalBDT::Predict(double *features, int size, bool useRawScore) {
+bool AliExternalBDT::Predict(double *features, int size, std::vector<double> &outputScores, bool useRawScore) {
```
fcatalan92, line 136: I would not use a vector but an array since its dimension is known a priory when you write the task I think. However, I don't have a strong opinion here.

Instead the casting of the features from double to float and the results from float to double do not make much sense. Probably we should move to float since this is the intrinsic precision of the model application. We can address this in another PR eventually.

fgrosa: For the first comment, it is a bit more complex to use an array because the size of the output is not necessarily known a priori, and there might be cases in which this is not even considered in a task (i.e. if you only call IsSelctedMulticlass without the output scores). One should define an array with a not defined size that depends on the size of the output score taken from treelite, so personally I would keep a vector at least for the time being.

Regarding the second comment, I completely agree, I did not change it not to modify all the tasks that use this method but we can consider to do it in the future.

mpuccio:
+1

PR #15978, opened 2020-10-29
PWGLF/NUCLEX/Exotica/Dibaryons: Enable background estimation using sideband technique

nashimiz:
The task are modified in order to estimate the backgrounds of dibaryon's invariant mass distributions, using sideband technique


bdoenigu:
+1

PR #15986, opened 2020-10-29
Add Pb-Pb 2018 event selections for ITS-TPC matching studies

mfaggin:
1. time-range cut for LHC18r period
   2. out-of-bunch pile-up rejection according to ITS cluster vs. TPC cluster correlation


fgrosa:
+1

fprino:
+1

PR #16015, opened 2020-11-02
Add switchs to forced track extrapolation to the EMCal surface on AODs

gconesab:
Avoid unnecessary processing of tracks already discarded during AOD creation
For discussion


gconesab:
+1

PR #16029, opened 2020-11-02
Add selection of D+ meson candidates

fgrosa:
- Added selection of D+ candidates in a way that it will be easy to set different HF particles
- Included vertexingHFML library to CMakeList (works only with ROOT6 because of ML application)


a-mathis:
+1

PR #16039, opened 2020-11-03
PWGHF: Improvements for running PbPb

lvermunt:
* Reduce the number of branches when running PbPb
* Fix for downsampling seed != 0
* Added PreSelect for ITS upgrade and general for Ds
* Fix for eventID in MC


fgrosa:
+1

PR #16043, opened 2020-11-03
PWGHF: Fix EvID + reduce PbPb output size

lvermunt:
* Fix evID for MC another time (bring GetEvID function back to working QM19 state)
* Reduce PbPb output size further by removing more unnecessary branches


fgrosa:
+1

PR #16051, opened 2020-11-04
PWGGA/GammaConv: add CheckRootFilesForEquality.C macro

sstiefel19:


FriederikeBock:
please add this to the gitlab of the Afterburner software instead this shouldn't be in here.

sstiefel19:
> please add this to the gitlab of the Afterburner software instead this shouldn't be in here.

I put here it here in preparation of the automated testing framework I wanna set up. 
Namely to deploy a testing framework that will, before a change gets merged into the master, run a couple of tasks (depending on which tasks are affected by the change), to make sure the change has no unexpected sideeffects.
I thought that the code for such a framework would have to be within AliPhysics in order to be accessible for the machines involved,  thats why I put it here. Could it also get accessed if its within the Afterburner repo?

FriederikeBock:
+1

PR #16055, opened 2020-11-04
PWGGA/GammaConv: OmegaToPiZero - fixed non photon selection compatibility 

mhemmer-cern:
- Added non-photon selection support for the background to reduce CPU time
- Fixed 6000er train configs to be able to run EDC-EDC-EDC
- Changed pT binning for some QA plots for omega candidates


nschmidtALICE:
+1

PR #16062, opened 2020-11-04
[EMCal] New method to calculate shower shape reducing or enlarging the cluster size like NxN clusterizer

gconesab:


gconesab:
+1

PR #16066, opened 2020-11-05
Protection for jets with 0 jet mass

mfasDa:
- Add cut on non-0 jet mass
- More elaborate error handling
  and monitoring


mfasDa:
Fastjet softdrop crashes when it finds a jet with 0 jet mass, due to calculation of mu.

mfasDa:
+1

PR #16069, opened 2020-11-05
[EMCal] Implement methods to calculate number of local maxima 

gconesab:
(adapted from AliCalorimeterUtils) and use in method to recalculate shower shape with NxN selection


gconesab:
+1

PR #16077, opened 2020-11-05
update of previously existing files to allow a cut in the mass spectrum

rarnaldi:
Files have been modified in order to allow the possibility to set a mass cut from an external macro (previously it was harcoded in the .cxx). In this way it will be possible to create trees with a given cut in the dimuon mass spectrum


rarnaldi:
+1

miweberSMI:
+1

PR #16081, opened 2020-11-05
PWGHF: Adding TOF mismatch selection (TreeCreator)

lvermunt:
Fix for extracted nsigma TOF stored in TTrees. There are two methods:
 * `AliAODPidHF::GetnSigmaTOF()`
 * `AliPIDResponse::NumberOfSigmasTOF()`
First one is adding a TOF mismatch cut (default 0.01) as extra condition for flagging bad TOF-PID tracks (standard procedure in HF). Second option isn't, leading to random nsigma TOF values in a huge range, which were given as input to the ML training. These candidates should have been flagged with nsigmaTOF = -999 instead. 

Switched now from 2nd to 1st method, fixing the issue


fgrosa:
+1

PR #16083, opened 2020-11-06
PWGHF: Fix in case of PID post calibration PbPb

lvermunt:
PID post calibration is performed already in the newly used AliAODPidHF class, so we were double correcting now. Fix = only correct in task in case when the PID response class is used


fgrosa:
+1

PR #16108, opened 2020-11-09
Add D+ variable histrograms to task for O2 comparisons

fcatalan92:
@fprino In addition to adding the D+ variable histograms, I uncommented also the lines to set the primary and secondary vertices for 3-prongs


fprino:
+1

PR #16114, opened 2020-11-09
PWGHF: Downsampling methods to reduce output size PbPb

lvermunt:
Two additional methods:
 * Only fill reconstructed TTrees with selected events
 * Pseudorandom candidate rejection factor (as defined for other HF studies), to store less low pT candidates when creating sample for training only


fgrosa:
+1

PR #16120, opened 2020-11-10
Add option to select events in a certain pt-hard range

mfasDa:
Option is in particular of relevance when combining min. bias and pt-hard events, in order to remove events with pt-hard values in ranges where the jet-jet MC is used, in order to avoid double counting.


mfasDa:
+1

PR #16122, opened 2020-11-10
PWGGA/GammaConvBase: Fixed bug in PID-selection

Stev960:


PWGGA/GammaConvBase/AliCaloSigmaCuts.cxx
```diff
@@ -682,7 +682,7 @@ Bool_t AliCaloSigmaCuts::TrackIsSelected(AliAODTrack* track, AliPIDResponse* fPI
   
   if(!(fPIDResponse->NumberOfSigmasTPC(track, AliPID::kProton))) return kFALSE;
   if(track->P() <= 0.1) return kFALSE;
-  if(track->P() <= 0.8 || track->P() > 0.1){
+  if((track->P() <= 0.8) && (track->P() > 0.1)){
```
nschmidtALICE, line 685: You could actually remove the "&& (track->P() > 0.1)" part as you catch this case already in the line above.

PR #16127, opened 2020-11-10
PWGGA: Turn off Preselection in same train

jluehder:


PWGGA/GammaConv/macros/AddTask_GammaConvNeutralMesonPiPlPiMiNeutralMeson_MixedMode_pp.C
```diff
@@ -183,6 +184,7 @@ AliVEventHandler *inputHandler=mgr->GetInputEventHandler();
 
   AliAnalysisTaskNeutralMesonToPiPlPiMiNeutralMeson *task=NULL;
   task= new AliAnalysisTaskNeutralMesonToPiPlPiMiNeutralMeson(Form("GammaConvNeutralMesonPiPlPiMiNeutralMeson_%i_%i",neutralPionMode, trainConfig));
```
fjonasALICE, line 186: running multiple pion selectors in same trains does not work in this case, since the output files for two different pion preselections will have the same name, causing an error

nschmidtALICE:
+1

fjonasALICE (changes requested):
running multiple pion selectors in same trains does not work in this case, since the output files for two different pion preselections will have the same name, causing an error. Please use implementation of EMCal correction framework as guidance on how to implement this

PR #16128, opened 2020-11-10
PWGGA: Changes to run same config with preselection on and off

jluehder:


fjonasALICE (changes requested):
please add also for other AddTasks

nschmidtALICE:
+1

PR #16131, opened 2020-11-10
Exception handling for ZDC in LHC16qrst

ddobrigk:
At the request of @coppedis, @valzacc (in the context of having ZNA, ZNC fully available in LHC16qrst)


ddobrigk:
+1

PR #16132, opened 2020-11-10
Using time period if the event time stamp is not available, for example in MC

pzhristov:


pzhristov:
Hi @jgrosseo, this fixes the issue with the MC productions without time stamps. However, the time stamp we use is the host system time taken the moment the event is created (in seconds). It is OK for the purposes of the converter, but we may look also in the orbit/BC time of the event, which is what will be used in O2.

pzhristov:
+1

PR #16140, opened 2020-11-11
Add product of prong impact pararameters in Dplus tree

fcatalan92:
Add an optional flag to store the d0K x d0pi1 and d0K x d0pi2 variables in the tree


fgrosa:
+1

PR #16154, opened 2020-11-12
PWGLF/NUCLEX: Add out-of-bunch pileup cuts for daughters of V0s and Cascades

nashimiz:
- Add track cuts to reject V0 and Cascade candidates with daughter tracks stemming from pile-up events
- Fix mass window selection of anti-Lambda


bdoenigu:
+1

PR #16161, opened 2020-11-13
PWGGA/GammaConv PHOS trigger mimicking and train config changes (pp 13 TeV)

AMechler:
PHOS trigger mimicking:
- PHOS trigger mimic helper only created if mimicing enabled in addtask
- enableTriggerMimicking==3 one leg of meson must be triggered
- enableTriggerMimicking==4 all meson candidates in triggered event

train config changes (pp 13 TeV):
- add configs to test PHOS standard with different mimicing settings in same train
- change pcm-edc lowB NL 01 -> 11
- adding additional RBin settings for pcm-edc lowB


jluehder:
From my side, this is ok

FriederikeBock:
+1

PR #16168, opened 2020-11-13
Updated resolution of LHC15o/pass2 to 66 ps (P. Antonioli, N. Jacazio)

chiarazampolli:
From Pietro Antonioli, and @njacazio .


chiarazampolli:
+1

PR #16170, opened 2020-11-13
PWGGA/GammaConv: GammaIsoTree photon truth definition

fjonasALICE:
Changed definition for true photons, now including conversions


nschmidtALICE:
+1

PR #16171, opened 2020-11-13
Add option for requiring a cluster with a min. energy for run3 trigge…

mfasDa:
…r studies

Run3 trigger in hardware expected to be similar to the L0 trigger
which can be mimiced with adding a requirement of a cluster with min.
energy (non-lin corr) in event. Existing cluster containers are used
(if available), however the cluster selection is implemented by hand as
the cluster container will use had. corr energy in case of full jets.


mfasDa:
+1

PR #16172, opened 2020-11-13
Added possibility of an offline trigger for data

jokonig:


FriederikeBock:
Hi @jokonig ,
did you test this carefully on data and MC, so that the logic for the MC didn't change. Your changes at first glance seem ok, but I wanna make sure we are not introducing funny things for MC running now.
Regards,
Fredi

jokonig:
Hi @FriederikeBock ,
I ran train test with this modification and without and compared the output. It was consistent for min. bias triggers as well as the EMCal triggers in Data and MC.
Kind regards,
Joshua

FriederikeBock:
+1

PR #16177, opened 2020-11-14
PWG/EMCal cell-level trackmatcher implementation

mikesas:


PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellTrackMatcher.h
```diff
@@ -0,0 +1,65 @@
...
+
+ private:
+  AliEmcalCorrectionCellTrackMatcher(const AliEmcalCorrectionCellTrackMatcher &);               // Not implemented
+  AliEmcalCorrectionCellTrackMatcher &operator=(const AliEmcalCorrectionCellTrackMatcher &);    // Not implemented
+
+  // Allows the registration of the class so that it is available to be used by the correction task.
+  static RegisterCorrectionComponent<AliEmcalCorrectionCellTrackMatcher> reg;
+  
+  /// \cond CLASSIMP
+  ClassDef(AliEmcalCorrectionCellTrackMatcher, 1); // EMCal cell energy variation component
```
raymondEhlers, line 61: Typo in the comment here

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellTrackMatcher.cxx
```diff
@@ -0,0 +1,252 @@
...
+
+
+}
+
+/**
+ * Called before the first event to initialize the correction.
+ */
+void AliEmcalCorrectionCellTrackMatcher::ExecOnce()
+{
+  fParticleContainerIndexMap.CopyMappingFrom(AliParticleContainer::GetEmcalContainerIndexMap(), fParticleCollArray);
```
raymondEhlers, line 105: As far as I can tell, you don't use this, so better to remove it

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellTrackMatcher.h
```diff
@@ -0,0 +1,65 @@
...
+protected:
+  
+  Bool_t        IsTrackInEmcalAcceptance(AliVParticle* part, Double_t edges=0.9) const;
+
+  Double_t               fMipE;                           ///< Energy of MIP used to subtract from cell
+
+  #if !(defined(__CINT__) || defined(__MAKECINT__))
+  // Handle mapping between index and containers
+  AliEmcalContainerIndexMap <AliParticleContainer, AliVParticle> fParticleContainerIndexMap; //!<! Mapping between index and particle containers
+  #endif
```
raymondEhlers, line 45: As far as I can tell, you don't use this, so better to remove it

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellTrackMatcher.h
raymondEhlers, line 61: Typo in the comment here

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellTrackMatcher.cxx
```diff
@@ -0,0 +1,252 @@
...
+  AliVTrack* track = 0;
+
+  AliParticleContainer * partCont = 0;
+  TIter nextPartCont(&fParticleCollArray);
+  while ((partCont = static_cast<AliParticleContainer*>(nextPartCont()))) {
+    auto partItCont = partCont->accepted_momentum();
+    for (AliParticleIterableMomentumContainer::iterator partIterator = partItCont.begin(); partIterator != partItCont.end(); ++partIterator) {
+      track = static_cast<AliVTrack *>(partIterator->second);
+      // Propagate the track
+      AliEMCALRecoUtils::ExtrapolateTrackToEMCalSurface(track, 440, 0.1396, 20, 0.35, kFALSE, kTRUE, kFALSE);
```
raymondEhlers, line 125: You should check the cluster-track matcher for the cases where we actually have to propagate the tracks. For AODs, this is stored for Run2, which can save a lot of computation time. We recently introduced some improvements there. (This may have actually been in the RecoUtils, so maybe this is already accounted for. Take this as a suggestion to crosscheck this if you haven't already)

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellTrackMatcher.h
```diff
@@ -0,0 +1,65 @@
+#ifndef ALIEMCALCORRECTIONCELLTRACKMATCHER_H
+#define ALIEMCALCORRECTIONCELLTRACKMATCHER_H
+
+#include "AliEmcalCorrectionComponent.h"
+
+#if !(defined(__CINT__) || defined(__MAKECINT__))
+#include "AliEmcalContainerIndexMap.h"
+#endif
```
raymondEhlers, line 8: If you drop the indexmap, you can also drop this

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellTrackMatcher.h
```diff
@@ -0,0 +1,65 @@
...
+  Bool_t        IsTrackInEmcalAcceptance(AliVParticle* part, Double_t edges=0.9) const;
+
+  Double_t               fMipE;                           ///< Energy of MIP used to subtract from cell
+
+  #if !(defined(__CINT__) || defined(__MAKECINT__))
+  // Handle mapping between index and containers
+  AliEmcalContainerIndexMap <AliParticleContainer, AliVParticle> fParticleContainerIndexMap; //!<! Mapping between index and particle containers
+  #endif
+
+  TClonesArray *fEmcalTracks;                  //!<!emcal tracks
```
raymondEhlers, line 47: Perhaps I'm being dense, but it also appears that you don't use this?

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellTrackMatcher.cxx
```diff
@@ -0,0 +1,252 @@
...
+      particleContainerNames += "_";
+    }
+    else {
+      firstLoop = false;
+    }
+    particleContainerNames += partCont->GetArrayName();
+  }
+
+  fEmcalTracks = new TClonesArray("AliEmcalParticle");
+  fEmcalTracks->SetName(Form("EmcalTracks_%s", particleContainerNames.c_str()));
```
raymondEhlers, line 84: In fact, I think all of this stuff can be removed.

Now I think I see what's going on - all of this info is only necessary if you're storing the matching information. But as far as I can tell, you're not, so it's not necessary

raymondEhlers (commented):
Hi Mike,

This looks a really nice addition - thanks!

As a general comment, we may want to rename this task because it handles both the matching and subtraction of the MIP energy. For clusters, this is handled by two separate tasks: the cluster-track matcher and the hadronic correction. The cluster-track matcher task _only_ does the matching, while the hadronic correction does the actual subtraction (it can do track E or MIP energy, so it generalizes the apporach here). I assume that you're not storing matching information here because I'm not aware of a cell field for such info - fair enough. But then I think it's important to be clear on precisely what this does.

Best,
Raymond

raymondEhlers:
+1 looks good to me!

nschmidtALICE:
+1

PR #16181, opened 2020-11-15
Implement replay mode of EJE triggers in data

mfasDa:
- Require physics selection to be kINT7
- Require trigger selection from AliEmcalTriggerSelectionTask


mfasDa:
+1

PR #16211, opened 2020-11-18
Two changes to improve tree creation

rarnaldi:
1) save in the tree only the events which contain at least a dimuon with mass higher than a threshold
2) save also a TH1D containing the infos on the trigger statistics


miweberSMI:
+1

PR #16212, opened 2020-11-18
PWGGA/GammaConv: OmegaToPiZeroGammaTask - changed background scheme for two cut settings

mhemmer-cern:
- Changed background scheme for two cut settings to use TGPS
- Added check for gamma -> e -> gamma from Pi0/omega for true omegas


FriederikeBock:
+1

PR #16230, opened 2020-11-19
Increase ClassDef of the AliAnalysisTaskEmcal

mfasDa:
Recently data fields have been added, consequently
the ClassDef of the task should be increased.


mfasDa:
+1

Trivial change increasing ClassDef of the AliAnalysisTaskEmcal due to new data members

PR #16232, opened 2020-11-19
Add separate check for L1 phase shift enabled in RecalibrateCells

mfasDa:


mfasDa:
@gconesab @raymondEhlers Any problem with this? Up to now the check for L1 phase shift enabled was part of the check for the time calibration, so if the time calibration is switched off the phase shift correction cannot be applied separately.

raymondEhlers:
Seems fine to me

gconesab:
I am also fine with it
+1

mfasDa:
+1

mfasDa:
@gconesab @raymondEhlers Thanks for your review!
@gconesab I think the `+1` must be in the first line, otherwise it is not recognised as approval. Adding comments in lines below, in addition to the `+1` seems to work

PR #16238, opened 2020-11-20
Set next splitting to harder branch in the iteration

mfasDa:


mfasDa:
Following the harder branch was missing in the SoftDrop implementation, therefore it was resulting in an infinity loop

mfasDa:
+1

PR #16239, opened 2020-11-20
PWG/EMCAL: fix for nonlinearity application

nschmidtALICE:
The order of SetNonLinCorrEnergy() and fSetForceClusterE must be inverted to work correctly. Otherwise the correction is applied twice!

Also fixed that finetuning correction is also applied on DCal clusters.


gconesab:
+1

PR #16251, opened 2020-11-22
Register 2013 p-Pb datasets properly

ddobrigk:


ddobrigk:
+1

ddobrigk:
I am afraid this too will require a manual merge... There is no real issue and it is again the system bug :-(

chiarazampolli:
Ciao David,

I can merge, but why do we have also again the 15o pass2 calibration here?

Chiara

ddobrigk:
Ciao Chiara, 
I actually don't know - I think the system flagged it as not done before (it was a manual merge before, too) and now it appears again (which is wrong). Something similar happened in O2 and then I solved it by closing and a full reset of my fork - I don't know why but that did the trick. Let me do the same thing here... closing. 
Thanks! 
Cheers,
David

PR #16255, opened 2020-11-23
Register 2013 p-Pb data properly

ddobrigk:


ddobrigk:
...and now without the LHC15o OADB update (which anyway was merged but the system was still listing it for some reason)

ddobrigk:
+1

chiarazampolli:
Merging, assuming you checked the ranges (I didn't :-) ).

Chiara

PR #16258, opened 2020-11-23
Macro to produce PbPb trees for dimuon analysis

lucamicheletti93:
Macro to produce PbPb trees for dimuon analysis. The information of polarization angles in the helicity, Collins-Soper and Event-Plane reference frames is also included.


miweberSMI:
+1

PR #16261, opened 2020-11-23
Reduction of the number of output histograms, 

fkrizek:
Reduction of the number of output histograms plus
incorporation of histograms for 2D unfolding


raymondEhlers:
+1

PR #16274, opened 2020-11-24
PWGGA/GammaConv/ fix seg faults

AMechler:
fixing seg fault for many mc particles in cluster while using mclabelsorting
fix memory seg faul in minimal output


PWGGA/GammaConvBase/AliAODConversionPhoton.h
```diff
@@ -132,6 +148,3 @@ class AliAODConversionPhoton : public AliAODConversionParticle, public AliConver
```
nschmidtALICE, line 149: Please increase the classdef here. Otherwise the PR looks fine.

AMechler: Thanks for catching, classdef is incremented now

nschmidtALICE:
+1

PR #16287, opened 2020-11-25
Add Unlike-Sign Test

SiyuTang:
Hi all,

Here is an update for Jet v2 analysis, about the unlike-sign test.

Thanks a lot!

Siyu Tang


yzhouCERN:
+1

SiyuTang:
There is no any log, so it seems not related to the code. Could someone help me to deal with it? Thanks!

PR #16289, opened 2020-11-25
Put the TList to DataContainer

zhaomr13:
Change way of reading weight. 
In the past I put the TFile to DataContainer. 
Now, it is changed to put the TList.


zumoravc:
+test

PR #16304, opened 2020-11-26
PWGGA/GammaConv: Fix TooClose photon cut for Linux architectures

sstiefel19:
This fixes a runtime crash on linux architectures (see train 1625 GA_pp_AOD)
that does not occur on Mac.


nschmidtALICE:
+1

PR #16316, opened 2020-11-27
PWGGA/GammaConv: Updated OmegaToPiZeroGamma Pi0Rotation Background Calculation

mhemmer-cern:
- Updated the calculation for the Pi0 rotation background method
- Fixed bug which would flag clusters that are bad after rotation as good
- Fixed bug which would lead to skipping good clusters after rotation if
another cluster was flagged as bad before
- Added pT dependent MesonIsSelectedByMassCut variations which use
fMassParamFunction value of 0 instead of 5
- Updated all the cut strings accordingly to change above
- Added Dalitz cut as an option in the AddTask
- Added more Dalitz QA plots to monitor this cut and the background
- Updated the fDoMesonQA flag to work as a bitmask to freely activate
different kind of QA plots
- Set light output flag for OmegaToPiZeroGammaTask now correctly in
AddTask implemented


nschmidtALICE:
+1

PR #16329, opened 2020-11-30
fix(PWGCF): changing counter type to double

pkarczma:
This fix allows to sum entries with non-unitary weights, which may happen depending on what exactly a user asks for.


pkarczma:
@jgrosseo please have a look if this is fine

jgrosseo:
That's look fine to me

jgrosseo:
+1

jgrosseo:
@TimoWilken Is this a known issue or related to the fact that this was a draft PR first (which triggered wrong comments from alibuild above)?

TimoWilken:
@jgrosseo the failing builds are due to GNOME's GitLab being down (so we can't fetch libxml2). I can look into the wrong comments.

PR #16333, opened 2020-11-30
Add Primary FMD Information

SiyuTang:
Hi,

Here is an update for Jet v2 analysis, for estimating the secondary FMD effect.

Thanks a lot!

Siyu


yzhouCERN:
+1

PR #16360, opened 2020-12-02
Generalize hardest kt grooming methods

raymondEhlers:
Makes dynamical grooming methods available (ie. usable) and add method
for selecting jet core


raymondEhlers:
+1

ktf:
@TimoWilken looks like in this case the ROOT5 builds are lagging behind. Could you check the builders?

PR #16370, opened 2020-12-03
PWGGA/GammaConv: E/p studies nonlin variations

fjonasALICE:


nschmidtALICE:
+1

ktf:
@TimoWilken again I think there is something wrong going on with this PR.

TimoWilken:
@ktf the CI builders are busy (mesosci/slc7/9 in particular), hence the delay. See alisw/ali-bot#899 for a fix to the build order of untested PRs.

ktf:
Ok, but this has been the case for the last 4 hours. Maybe we need to give higher priority to O2 / AliPhysics PRs vs alidist ones (which can be indeed very long).

TimoWilken:
@ktf for no particular reason, PRs are currently built in alphabetical order of .env file name (and within that, newest first). If you want to deprioritise alidist PRs, a quick hack would be to rename e.g. `o2-alidist.env` to `zz-o2-alidist.env` (and `aliroot-alidist.env` similarly). Or I can work out a better way of doing it tomorrow.

ktf:
Will that change the name of the check? If not, I think it's a very good solution.

TimoWilken:
@ktf no, the name of the check (e.g. as shown on GitHub) is set by the `CHECK_NAME` variable inside the .env file, not by the file name itself. I'll send a PR with the change in a minute.

ktf:
However it will not requery now right? Maybe we should undo the other change so that:

* It builds aliphysics.
* If none, it builds O2
* If none, it builds alidist.

TimoWilken:
@ktf that's true, it won't requery with the other change. Undoing it will mean that the queue numbers don't really make sense though (they'll change on each iteration as new PRs come in). Let's leave it to clear its backlog overnight and work out what order things should be done in tomorrow morning, then I can make the thing make sense as a whole?

PR #16372, opened 2020-12-03
Fixing some issues in the task for LEGO trains

lucamicheletti93:
Fixing some issues in the task for LEGO trains


miweberSMI:
+1

PR #16415, opened 2020-12-08
PWGLF/NUCLEX: Add filter-bit selection for AOD tracks

nashimiz:
- Add filter-bit selection for AOD in order to select TPC-only tracks
- Remove background estimation using sideband


bdoenigu:
+1

PR #16432, opened 2020-12-09
Added flag to accept V0M events beyond 90% to MeanPtV2Corr task 

vvislavi:
Decoupled dN/deta calculation from mpt. Added a quick-fix for covariance production to be calculated as a function of new Nch, but using old Nch


vvislavi:
+1

vvislavi:
Merge crashed for whatever reason -- logs say no error. Will push a dummy commit and reopen the merge request, so that test with /release can be redone

PR #16445, opened 2020-12-10
Correcting the definitions for Helicity and Collins-Soper in Pb-Pb

lucamicheletti93:
Correcting the definitions for Helicity and Collins-Soper in Pb-Pb


iarsene:
+1

PR #16451, opened 2020-12-11
Add HMPID to AO2D conversion

njacazio:
Hi @gvolpe79 this is the changes that we would port to the O2 data format, do we need anything else?

### DIFF
--- a/RUN3/AliAnalysisTaskAO2Dconverter.cxx
+++ b/RUN3/AliAnalysisTaskAO2Dconverter.cxx
@@ -593,6 +593,7 @@ void AliAnalysisTaskAO2Dconverter::InitTF(ULong64_t tfId)
     tTracks->Branch("fTPCSignal", &tracks.fTPCSignal, "fTPCSignal/F");
     tTracks->Branch("fTRDSignal", &tracks.fTRDSignal, "fTRDSignal/F");
     tTracks->Branch("fTOFSignal", &tracks.fTOFSignal, "fTOFSignal/F");
+    tTracks->Branch("fHMPIDSignal", &tracks.fHMPIDSignal, "fHMPIDSignal/F");
     tTracks->Branch("fLength", &tracks.fLength, "fLength/F");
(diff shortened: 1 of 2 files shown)


pzhristov:
+1

ktf:
Please remove the draft if you are ready.

ktf:
Changing to non draft, since there was no reply.

njacazio:
Hi @ktf, I did not reply because I would prefer not to be merged yet as I would like Giacomo to validates it first

ktf:
ok, i closed it to avoid interference with the builds while we understand the permissions race condition.

njacazio:
@ktf shall we reopen this one or is it better if we open a new one?

RUN3/AliAnalysisTaskAO2Dconverter.cxx
```diff
@@ -593,6 +593,7 @@ void AliAnalysisTaskAO2Dconverter::InitTF(ULong64_t tfId)
     tTracks->Branch("fTPCSignal", &tracks.fTPCSignal, "fTPCSignal/F");
     tTracks->Branch("fTRDSignal", &tracks.fTRDSignal, "fTRDSignal/F");
     tTracks->Branch("fTOFSignal", &tracks.fTOFSignal, "fTOFSignal/F");
+    tTracks->Branch("fHMPIDSignal", &tracks.fHMPIDSignal, "fHMPIDSignal/F");
```
njacazio, line 596: @gvolpe79 do you think we miss anything else apart from the HMPID signal?

gvolpe79:
The HMPID signal (Cherenkov angle) is the only parameter needed for the HMPID PID! What it is not clear to me it is if the track quality cuts to select good tracks are applied at the AOD creation or at analysis level! if the latter case, we should add other information.

njacazio:
So we need to add the mip charge, the distance between the extrapolated track and the matched cluster. 
The mip charge is maximum at 4000 (a short would be enough), this is used to cut, tracks with q_mip > 120 are accepted.
The distance would be a float to be used to cut tracks with distance > 5 cm.
I will investigate now if this has to go in a separate table

PR #16491, opened 2020-12-16
PWGLF/NUCLEX: Fix the size of track array

nashimiz:
- The max size of track array is set larger (2000->2500) for HM event
  - The size of array is determined by new member variable
- p-Omega invariant mass calculation is fixed


lramona:
+1

PR #16495, opened 2020-12-16
PWGGA/GammaConv: Added new NCell cut option for isolated 1cell cluste…

jokonig:
…rs and added respective configs. Additionally modified NonLin for 1 cell clusters and added new exotic cluster cut variations


nschmidtALICE:
+1

PR #16497, opened 2020-12-16
Fixing minor issues and removing some useless prints

lucamicheletti93:
Fixing minor issues and removing some useless prints


miweberSMI:
+1

PR #16506, opened 2020-12-17
Better QA plot for HM events

mpuccio:


mpuccio:
Ciao @fprino , can you please have a look?

akalweit:
+1

PR #16515, opened 2020-12-18
PWGGA/GammaConv: OmegaToPiZeroGamma - Added train config for rotation method with weighting

mhemmer-cern:
- Added train config for rotation method with weighting


FriederikeBock:
+1

PR #16542, opened 2020-12-23
PWGLF/NUCLEX: Fix rejection of pairs with shared tracks

nashimiz:
- Fix pair rejection with shared tracks for AOD
  - Track ID is reset in case of TPC-only tracks
- Add selection cuts for omega (tighter cuts)
- Candidates are stored in TClonesArray


bdoenigu:
+1

PR #16543, opened 2020-12-23
Opening of the Lambda and Xi Cuts for systematic variations

gmantzar:
Also reincluding Branches in the TTree to make that possible


a-mathis:
+1

PR #16556, opened 2020-12-28
AliAnalysisTaskHFJetIPQA:: fix inconsistencies for determination of g…

kagarner:
…enerator-level V0 jets

- consider also V0 daughters which aren't generator-level final state particles but are reconstructed tracks
- apply V0 acceptance cuts also for MC Veto for V0-daughter tagging


lcunquei:
+1

PR #16557, opened 2020-12-28
adding trigger flag to the tree

lcunquei:


lcunquei:
+1

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskLundPlane.cxx
```diff
@@ -344,6 +345,16 @@ Bool_t AliAnalysisTaskLundPlane::FillHistograms() {
       if ((fCentSelectOn == kFALSE) && (jet1->GetNumberOfTracks() <= 1))
         continue;
 
+
+      Int_t mytrig=0;
+      if(fStoreTrig==kTRUE){
+        
+        if(fInputEvent->GetFiredTriggerClasses().Contains("INT7")) mytrig=0;
+        if(fInputEvent->GetFiredTriggerClasses().Contains("EJ1")) mytrig=1;
+	if(fInputEvent->GetFiredTriggerClasses().Contains("EJ2")) mytrig=2;
```
mfasDa, line 354: @lcunquei this is unfortunately not enough: If you request AliVEvent::kINT7 then you will only get the EMCAL triggers that are a subset of INT7. Thus you always have to request AliVEvent::kINT7 when requesting INT7 triggers, and AliVEvent::kEMCEJE when requesting EJ1/EJ2 triggers, in order to require for the event to be a proper collision candidate excluding out-of-bunch pileup. The condition has to be

``` c++
if(fInputHandler->IsEventSelected() & AliVEvent::kINT7 && fInputEvent->GetFiredTriggerClasses().Contains("INT7")) mytrig=0;
if(fInputHandler->IsEventSelected() & AliVEvent::kEMCEJE && fInputEvent->GetFiredTriggerClasses().Contains("EJ1")) mytrig=1;
if(fInputHandler->IsEventSelected() & AliVEvent::kEMCEJE && fInputEvent->GetFiredTriggerClasses().Contains("EJ2")) mytrig=2;
```
The order (trigger selection bit before trigger string) comes from the fact that when chaining multiple conditions the later conditions are only evaluated if the previous conditions all are true. As your task inspects all three triggers at the same time you must not apply a global trigger selection in the task (via __R_ADDTASK__->SelectCollisionCandiate) or set the global trigger selection to AliVEvent::kAny as you perform the trigger selection for each trigger separately inside the task.

mfasDa (commented):
Hi @lcunquei, you need to slightly modify the trigger selection, because depending on how you would configure the trigger selection you either select only trigger that are a subset of Min. Bias or allow for out-of-bunch pileup.

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskLundPlane.cxx
lcunquei: hey Markus, thanks for taking a look! ah i see, thanks, will change that.

PR #16558, opened 2020-12-29
PWGLF/NUCLEX: Reject other particle from proton candidates using PID info

nashimiz:
- The nσ is computed assuming different particle hypotheses (kaons, electrons and pions).
  -  if the nσ value found to be smaller,  the track is rejected from proton candidates.
- New histograms (multiplicity, TOF β, etc...) are added.


bdoenigu:
+1

PR #16559, opened 2020-12-29
implementation was incorrect

lcunquei:


lcunquei:
+1

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskLundPlane.cxx
```diff
@@ -349,9 +351,9 @@ Bool_t AliAnalysisTaskLundPlane::FillHistograms() {
       Int_t mytrig=0;
       if(fStoreTrig==kTRUE){
         
-        if(fInputEvent->GetFiredTriggerClasses().Contains("INT7")) mytrig=0;
-        if(fInputEvent->GetFiredTriggerClasses().Contains("EJ1")) mytrig=1;
-	if(fInputEvent->GetFiredTriggerClasses().Contains("EJ2")) mytrig=2;
+        if(fInputHandler->IsEventSelected() & AliVEvent::kINT7 && fInputEvent->GetFiredTriggerClasses().Contains("INT7")) mytrig=0;
+        if(fInputHandler->IsEventSelected() & AliVEvent::kEMCEJE && fInputEvent->GetFiredTriggerClasses().Contains("EJ1")) mytrig=1;
+	if(fInputHandler->IsEventSelected() & AliVEvent::kEMCEJE && fInputEvent->GetFiredTriggerClasses().Contains("EJ2")) mytrig=2;
```
mfasDa, line 356: Sorry @lcunquei, but the implementation still has a couple of problem***s***:
- The default value (mytrg == 0) should never be part of the value range. In this case if non of the conditions is selected you will accept the event as INT7 event, even though you shouldn't. Rather use -1.
- After the last condition you need a check:
  ``` c++
  if(mytrg < 0) continue;
  ```
  In order to reject the triggers you don't want. mytrg is the same for the whole event so you need to evaluate this only once per event so I suggest you to put the block at the beginning of the function. Normally I would overwrite the function `IsTriggerSelected` and store the result in a class member. This is for the last problem.
- Multiple triggers can fire at the same time, because triggers can be substets of other triggers. Your code assigns one trigger class to the event, the last to be tested in case multiple fire. You either need to store all trigger classes for the event, reflecting all possible permutations of the three cases, or make different trees for different triggers. The later option is probably the simpler one - this is the one I usually use(d) - now with histograms.

mfasDa (commented):
Sorry, but this still doesn't work as you expect, see my comments. Modifications are however rather mild.

PR #16575, opened 2021-01-04
Optimization for TPCFMD correlation

SiyuTang:
Hi,

Here is the update for TPC-FMD correlations in jet v2 analysis. Thanks!

Siyu


yzhouCERN:
+1

PR #16595, opened 2021-01-08
Check to prevent dividing by zero when calculating ct. Also added pi0/eta weights to closure test histos.

efgauger:


sashingo:
+1

PR #16596, opened 2021-01-08
PWGGA/GammaConv: iso analysis effi additions & ABCD method

fjonasALICE:
- fixed to efficiency (inclusion of MC iso requirement also required for true rec)
- implemented histo to test ABCD method for purity as cross check
- implemented external jet finder for outlier rejection
- changed debug output in AliConvEventCuts


nschmidtALICE:
+1

PR #16600, opened 2021-01-08
Adding pi0 and eta weight for p-Pb

deepathoms:
Adding the weight for pi0 and eta pT spectra for tagging efficiency calculation


deepathoms:
+1

PR #16616, opened 2021-01-12
PWGHF: Jpsi to e+e- task and candidate selection for HF

zhangbiao-phy:
Add Jpsi to ee task with selection cut in aliPhysics

### DIFF
--- a/PWGHF/vertexingHF/AliAnalysisTaskHFSimpleVertices.cxx
+++ b/PWGHF/vertexingHF/AliAnalysisTaskHFSimpleVertices.cxx
@@ -89,6 +89,22 @@ AliAnalysisTaskHFSimpleVertices::AliAnalysisTaskHFSimpleVertices() :
   fHistD0SignalVertZ{nullptr},
   fHistInvMassD0Signal{nullptr},
   fHistInvMassD0Refl{nullptr},
+  fHistInvMassJpsi{nullptr},
+  fHistPtJpsi{nullptr},
+  fHistPtJpsiDau0{nullptr},
+  fHistPtJpsiDau1{nullptr},
+  fHistImpParJpsiDau0{nullptr},
+  fHistImpParJpsiDau1{nullptr},
+  fHistd0Timesd0Jpsi{nullptr},
+  fHistCosPointJpsi{nullptr},
+  fHistDecLenJpsi{nullptr},
+  fHistDecLenXYJpsi{nullptr},
+  fHistDecLenErrJpsi{nullptr},
+  fHistDecLenXYErrJpsi{nullptr},
+  fHistJpsiSignalVertX{nullptr},
+  fHistJpsiSignalVertY{nullptr},
+  fHistJpsiSignalVertZ{nullptr},
+  fHistInvMassJpsiSignal{nullptr},
   fHistInvMassDplus{nullptr},
   fHistPtDPlus{nullptr},
   fHistPtDplusDau0{nullptr},
@@ -133,6 +149,7 @@ AliAnalysisTaskHFSimpleVertices::AliAnalysisTaskHFSimpleVertices() :
   fDo3Prong(kFALSE),
   fMaxDecVertRadius2(8),
   fMassDzero(0.),
+  fMassJpsi(0.),
   fMassDplus(0.),
   fMassDs(0.),
   fMassLambdaC(0.),
@@ -146,11 +163,15 @@ AliAnalysisTaskHFSimpleVertices::AliAnalysisTaskHFSimpleVertices() :
   fNPtBins(25),
   fMinPtDzero(0.),
   fMaxPtDzero(9999.),
+  fMinPtJpsi(0.),
+  fMaxPtJpsi(9999.),
   fCandidateCutLevel(1),
   fSelectD0(1),
   fSelectD0bar(1),
   fMinPt3Prong(0.),
+  fNPtBinsJpsi(9),
   fNPtBinsLc(10),
+  fSelectJpsi(1),
   fSelectLcpKpi(1)
 {
   //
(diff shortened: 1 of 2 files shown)


vkucera:
**Please do not approve yet! These changes should not be automatically merged!**

vkucera:
@fprino Could you please have a look at these changes?

PWGHF/vertexingHF/AliAnalysisTaskHFSimpleVertices.cxx
```diff
@@ -1267,10 +1479,15 @@ Int_t AliAnalysisTaskHFSimpleVertices::SelectInvMassAndPtDzero(TObjArray* trkArr
   Double_t hilim=fDzeroSkimCuts[2];
   pdg2[0]=211; pdg2[1]=321;   // pi+ K- --> D0
   minv2 = rd4massCalc2->InvMass2(nprongs,pdg2);
-  if(minv2>lolim*lolim && minv2<hilim*hilim ) retval+=1;
+  if(minv2>lolim*lolim && minv2<hilim*hilim ) retval += (1 << kbitDzero);
   pdg2[0]=321; pdg2[1]=211;   // K+ pi- --> D0bar
   minv2 = rd4massCalc2->InvMass2(nprongs,pdg2);
-  if(minv2>lolim*lolim && minv2<hilim*hilim ) retval+=2;
+  if(minv2>lolim*lolim && minv2<hilim*hilim && (retval & (1 << kbitDzero)) == 0) retval+=(1 << kbitDzero);
```
fprino, line 1485: This line will create problems for the D0 because it breaks the logic used to tag separately D0 and D0bar candidates. A candidate (+- pair) can be OK as D0 (which is what is tested at L1482), D0bar (tested here) or both. In case it is OK only as D0 we should return 1, and this is done correctly in the new code. In case it is OK for both we should return 3, and this is also OK in the new code. What does not work is the case in which it is OK only as D0ber, in which we should return 2, and instead the new code returns 1.
The simplest way would be to redefine the selection bits as kbitDzero, kbitDzerobar and kbitJpsi.

PWGHF/vertexingHF/AliAnalysisTaskHFSimpleVertices.cxx
```diff
@@ -1139,6 +1308,49 @@ Int_t AliAnalysisTaskHFSimpleVertices::DzeroSkimCuts(AliAODRecoDecayHF2Prong* ca
...
+  Double_t dca[2];
+  Double_t dca1[2];
+  Double_t covtrack[2]; Double_t covtrack1[2];
+  Double_t ptCand = cand->Pt();
+  if (ptCand < fMinPtJpsi || ptCand > fMaxPtJpsi) return 0;
+  Int_t jPtBin = GetPtBin(ptCand);
+  if (jPtBin==-1) return 0;
+  
+  if (TMath::Abs(cand->InvMassJPSIee()-fMassJpsi) > fJpsiCuts[jPtBin][0] ) isJpsi=false;
+  if (!isJpsi ) return 0;
```
fprino, line 1340: This is an extra if condition that you test and can be avoided by doing:
if (TMath::Abs(cand->InvMassJPSIee()-fMassJpsi) > fJpsiCuts[jPtBin][0] ) return 0;
It seems to me that isJpsi is not needed if you modify in the same way at lines 1343 and 1346.
If you want to preserve isJpsi, you can do:
if (TMath::Abs(cand->InvMassJPSIee()-fMassJpsi) > fJpsiCuts[jPtBin][0] ){
  isJpsi=false;
  return 0;
}

PWGHF/vertexingHF/AliAnalysisTaskHFSimpleVertices.cxx
```diff
@@ -1139,6 +1308,49 @@ Int_t AliAnalysisTaskHFSimpleVertices::DzeroSkimCuts(AliAODRecoDecayHF2Prong* ca
...
  
+}
+//______________________________________________________________________________
+Int_t AliAnalysisTaskHFSimpleVertices::JpsiSkimCuts(AliAODRecoDecayHF2Prong* cand)
+{
+  bool isJpsi = true;
+  Double_t ptCand = cand->Pt();
+  if (ptCand < fJpsiSkimCuts[0]) return 0;
+  Double_t m0=cand->InvMassJPSIee();
+  if(m0 < fJpsiSkimCuts[1] || m0 > fJpsiSkimCuts[2]) isJpsi=false;
```
fprino, line 1319: I think you can avoid using isJpsi and just do 
if(m0 < fJpsiSkimCuts[1] || m0 > fJpsiSkimCuts[2]) return 0;

Below you can then replace:
  Int_t returnValue=0;
  if(isJpsi) returnValue+=1;
  return returnValue;
with a simple: 
  return 1;
This would allow you to spare the extra if conditions at line 1320 and 1324

PWGHF/vertexingHF/AliAnalysisTaskHFSimpleVertices.cxx
zhangbiao-phy: Thanks a lot for comments, I will make a commit to redefine the selection bits (add kbitDzerobar).

PWGHF/vertexingHF/AliAnalysisTaskHFSimpleVertices.cxx
zhangbiao-phy: Thanks for comments, Yes, It will make the code more clear.

PWGHF/vertexingHF/AliAnalysisTaskHFSimpleVertices.cxx
```diff
@@ -1482,7 +1470,7 @@ Int_t AliAnalysisTaskHFSimpleVertices::SelectInvMassAndPt2prong(TObjArray* trkAr
   if(minv2>lolim*lolim && minv2<hilim*hilim ) retval += (1 << kbitDzero);
   pdg2[0]=321; pdg2[1]=211;   // K+ pi- --> D0bar
   minv2 = rd4massCalc2->InvMass2(nprongs,pdg2);
-  if(minv2>lolim*lolim && minv2<hilim*hilim && (retval & (1 << kbitDzero)) == 0) retval+=(1 << kbitDzero);
+  if(minv2>lolim*lolim && minv2<hilim*hilim && (retval & (1 << kbitDzerobar)) == 0) retval+=(1 << kbitDzerobar);
```
fprino, line 1473: we do not need to check ((retval (1 << kbitDzerobar)==0) here. It is sufficient to have:
if(minv2>lolim*lolim && minv2<hilim*hilim) retval+=(1 << kbitDzerobar);

fprino:
Hi zhangbiao-phy,

thanks for implementing my suggestions.
I have still one minor remark on the new commit (see comment above).
If you can fix it, I would then approve the PR.

Best regards, Francesco

PWGHF/vertexingHF/AliAnalysisTaskHFSimpleVertices.cxx
zhangbiao-phy: Ok, Thank you so much for your reminding.

fprino:
+test. Thank you for implementing my last suggestion!

zhangbiao-phy:
> Hi zhangbiao-phy,
> 
> thanks for implementing my suggestions.
> I have still one minor remark on the new commit (see comment above).
> If you can fix it, I would then approve the PR.
> 
> Best regards, Francesco

Hi Francesco,

Many thanks for your comments! I have fixed now. I will squash the 3 commit into 1.

Best regards,
Biao

fprino:
+test

fprino:
+1

PR #16623, opened 2021-01-13
Fixing PU rejection and timing cuts

hdegen:
Fixing the PU rejection and timing cuts, also adding some more configurations to the stability study


valzacc:
+test

valzacc:
+1

PR #16632, opened 2021-01-13
remove MC gen pileup particles from track tree

jdmulligan:
Implement [rejection of pileup particles at generator level](https://twiki.cern.ch/twiki/bin/view/ALICE/AliDPGtoolsPileup#Pileup_in_Monte_Carlo_simulation) in track tree 
(not actually needed for jet-jet productions, but may be useful for other productions)

FYI @nzardosh (shouldn't affect you, just FYI)


pzhristov:
+test

PR #16638, opened 2021-01-14
Add settable threshold for EMCAL cell level compression

mfasDa:
Mimic run3 cell level compression type 1 and 2 (only amplitude
above threshold, no pre-clusterization)


mfasDa:
@jgrosseo This is for the compression of the cell table storing only cells above cluster cell threshold (which have a chance to end up in a cluster). I think PHOS also has a compression intended but the design and implementation is up to them.

jgrosseo:
+1

PR #16639, opened 2021-01-14
Use mass value of Omegac0 2.6952 instead of 2.6975 (default in PYTHIA6)

zhujh126:
Use mass value of Omegac0 2.6952 instead of 2.6975 (default in PYTHIA6)


fgrosa:
+1

PR #16647, opened 2021-01-15
Protection against duplicate entries in the AliVCaloTriggers

mfasDa:
For trigger bits and ampitudes: If the entry is 0 or negative the
value for the current trigger should be ignored, as the grids are
in any case initialized with 0.


mfasDa:
+test

mfasDa:
+1

PR #16648, opened 2021-01-15
Update AliAnalysisTaskNuclei.cxx

sadkoeni:
Added Chi2TPCConstrainedVsGloabal and number of shared TPC Clusters to ntuple variables in order to investigate discrepancy between different filterbits.


lbariogl:
+1

PR #16663, opened 2021-01-17
Event counting fix and multiplicity branch add for MC efficiency

Nikolajal:


bdoenigu:
+1

Nikolajal:
@bdoenigu Is there an issue with the root6 check? The log says that 100% tests were succesful and the commit makes minor changes that shouldn't bother the test

bdoenigu:
I also don‘t see anything problematic in the logs

PR #16669, opened 2021-01-18
Passed the name of the particle level jet finder to access the truth …

JColburn-CERN:
…level particles without needing to adopt the container in the Steering macro


raymondEhlers:
+1

PR #16675, opened 2021-01-19
PWGGA/GammaConv: OmegaToPiZeroGamma - Adjusted Rotation Method 

mhemmer-cern:
- Changed fixed rotation angle from 90° to (90+-5)° to reduce a possible bias.
- Added two new Trainconfigs


FriederikeBock:
+1

PR #16714, opened 2021-01-22
 Including TPC calibrations on 2017 and 2018 data for high pT analysis

crisjahnke:
Added a data-driven calibration on TPCnsigma, due to the missing splines at high pT for 2017 data and part of the 2018 periods.


dsekihat:
+1

PR #16715, opened 2021-01-22
Add histograms of generated V0 (Xi) pt vs eta to compute eta-dependent efficiency + compute fraction of events with trigger particle in mult classes

ChiaraDeMartin95:


lramona:
+1

PR #16739, opened 2021-01-25
Possibility to make THnSparses lighter when exploring PID possibilities

mfaggin:
Possibility to avoid the THnSparses to be filled with standard PID selections when the case with Bayes PID is checked only for proton.


fgrosa:
+1

PR #16748, opened 2021-01-26
PWGGA/GammaConv: Added probability of in-Jet analysis to merged cluster task

jokonig:
- Also modified mult dependent train configs for GammaCalo task
- Corrected centrality binning for high mult trigger setting


FriederikeBock:
+1

PR #16760, opened 2021-01-27
PWGGA/GammaConvBase: Changed AliConversionMesonCuts - Background Setting 'z'

mhemmer-cern:
- Changed the 'z' background setting to only use TGPS and _not_ forbid similar decays since it yields better background description when looking at TGPS omega decays and not forbid similar decays compared to TGPS omega decays and _do_ forbid similar decays


FriederikeBock:
+1

PR #16773, opened 2021-01-28
New MC pT-weights for D in PbPb(2018)

strogolo:
I add the new pT-weights for topological analysis of non-strange D mesons in Pb-Pb (2018 data sample).
The weights are obtained with:
- corrected pT spectra in 0-10 and 30-50 (data)
- MC(GenLimAcc) spectra from LHC20g2a and LHC20g2b


fgrosa:
+1

PR #16775, opened 2021-01-29
Allowing the time calibration to run on Run1 data

hahassan7:
The time calibration code has been modified such that it can analyze Run1 data. I also implemented a new method for the extraction of the time calibration coefficients using a Gaussian fit to the time distribution for every cell.


FriederikeBock:
+1

PR #16776, opened 2021-01-29
Nowa wiolata rz

wrzesaCERN:
Changes in AliFemtoESDTrackCut: cleaning up a little messy code for deuterons implemented in the past by me. Removing temporary options of method and some specific cases used for tests. Functions and options for deuterons that I have left in this class are sufficient for performing basic analysis with deuterons. I added/moved more specific features to AliFemtoWRzTrackCut.

Changes in  AliFemtoWRzTrackCut: this class I have created in the past but over time, become unnecessary. I decided to adapt it mainly for deuterons that I use in femto analysis. This class in comparison to  AliFemtoESDTrackCut contains some additional features: a base for the sideband analysis, a dEdx cut more apt to my future checks, and an exclusive n sigma selection. This class is under development and testing.


lgraczykCern:
+1

PR #16833, opened 2021-02-04
PWG/EMCAL: Modified cell track matcher to use correct eta/phi on EMCa…

jokonig:
…l surface. Added posibility to switch propagation on/off


FriederikeBock:
+1

PR #16841, opened 2021-02-05
PWGGA/GammaConv: iso tree SM performance improvements & fixes

fjonasALICE:
* mainly improvements for performance, removed old track matchers not needed, streamlined elements stored during runtime etc. * * fixed bug that causes neutral particles to be counted for MCIso perp cone


nschmidtALICE:
+1

PR #16842, opened 2021-02-05
Add Task to analysis Jetv2 in PbPb

SiyuTang:
Hi all,

Here is the new task for the analysis of Jet v2 analysis in PbPb. Please check it. Thanks a lot!

Siyu


yzhouCERN:
+1

PR #16853, opened 2021-02-07
AliFlowEventSimpleMakerOnTheFly.cxx memory leak fix and optimization

florishermsen:
**Memory Leaks**
When pTracks are rejected from the sample, the pointer was not deleted. Especially for a low mean acceptance, this causes OOM issues when scaling to many collisions. This has been addressed.

**Optimization**
When pTracks are checked for acceptance, it is more efficient to do this straight after the relevant properties are set. In simulations performed with a low mean acceptance rate for some ranges of properties, this reduced execution time by orders of magnitude (e.g. ~1e5 seconds to ~2e3 seconds). Re-ordering some of the operations addresses this.


florishermsen:
Would anyone be so kind to review this? Thanks

yzhouCERN:
+test

yzhouCERN:
+1

yzhouCERN:
+1

PR #16872, opened 2021-02-10
Updated Function to included a "Unmatched" catagory for Jets that PDG…

JColburn-CERN:
… Codes were not Identified for. This allows a comparision of the full data sample inclusive of the failed jets.

Additionally increased the limited on PDG agreement from 0.66 to 0.8.

### DIFF
--- a/PWGJE/EMCALJetTasks/AliAnalysisTaskJetChargeFlavourTemplates.cxx
+++ b/PWGJE/EMCALJetTasks/AliAnalysisTaskJetChargeFlavourTemplates.cxx
@@ -53,7 +53,7 @@ AliAnalysisTaskJetChargeFlavourTemplates::AliAnalysisTaskJetChargeFlavourTemplat
   JCDown(0x0),
   JCGluon(0x0),
   JCOther(0x0),
-
+  JCUnmatched(0x0),
 
   JCLow(0x0),
 
@@ -61,6 +61,7 @@ AliAnalysisTaskJetChargeFlavourTemplates::AliAnalysisTaskJetChargeFlavourTemplat
   JCDownLow(0x0),
   JCGluonLow(0x0),
   JCOtherLow(0x0),
+  JCUnmatchedLow(0x0),
 
 
   JCMid(0x0),
@@ -69,6 +70,7 @@ AliAnalysisTaskJetChargeFlavourTemplates::AliAnalysisTaskJetChargeFlavourTemplat
   JCDownMid(0x0),
   JCGluonMid(0x0),
   JCOtherMid(0x0),
+  JCUnmatchedMid(0x0),
 
   JCHigh(0x0),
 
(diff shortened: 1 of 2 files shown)


PWGJE/EMCALJetTasks/AliAnalysisTaskJetChargeFlavourTemplates.cxx
```diff
@@ -650,6 +672,36 @@ Bool_t AliAnalysisTaskJetChargeFlavourTemplates::FillHistograms()
             }
 
 
+        }
+
+        //Add Unmatched JetCharge Catagory
+        else if(IndexOfMaximum == -1)
+        {
+
+          fTreeBranch[23] = jetCharge;
```
raymondEhlers, line 681: This is less critical than my other comment, but if I understand your tree correctly, you're storing the jet pt in the tree. So your entries at 24, 25, and 26 are redundant with 23. You can make this cut in post processing, saving space with your tree output. However, I didn't look through your entire task, so perhaps you have some other reason

raymondEhlers (changes requested):
The comment on array indices is critical, but the other one on output branches may or may not be important

PWGJE/EMCALJetTasks/AliAnalysisTaskJetChargeFlavourTemplates.cxx
JColburn-CERN: Hi Raymond,
Your right this I am binning based on momentum within the analysis function. This was just the simplest way I could think of to allow easier comparison when running on the grid.  Thus how I've done it so far. Is it worth changing to save on the output size? How would I approach doing this in post-processing instead?

raymondEhlers:
+1

PWGJE/EMCALJetTasks/AliAnalysisTaskJetChargeFlavourTemplates.cxx
raymondEhlers: Hi @JColburn-CERN, the purpose of a tree is precisely so that you can select on variables when processing your tree, after running on the grid (as opposed to histograms, where you of course need this binning at the time that you fill it). During the post-processing of the tree, you apply whatever logic you want based on the branches you have available. Since it looks like you store your jet pt in your tree, you can always make selections on the jet pt later. There are a ton of ways to make these selections - there are a ton of examples in the ROOT docs and tutorials (a non-comprehensive list: retrieving values directly from branches, TTree::Draw, RDataFrame (my current favorite)). You can of course do your analysis any way you like and I haven't read your analysis closely, so maybe this is really, truly necessary. But making a selection on jet pt to decide which branches to fill while storing the jet pt in your tree suggests to me that you could be doing things more efficiently. Again, there are some decent ROOT tutorials.

As for whether it's worth changing to save on output size, it depends on how big your output is currently, how nicely that works with the LEGO train, your local post processing resources, etc. I always aim to minimize my output size (although it's not always feasible, or not worth a ton of effort for a minor savings).

PR #16873, opened 2021-02-10
Update hypertriton task employing the KFParticle package

SHornung1:
- activated storing of secondary vertex coordinates also for data
- correct y axis of pT-difference plot
- moved EventCuts Class QA histograms inside the area which can be turned off via doQA switch
- added and moved some documentation/comments


bdoenigu:
+1

PR #16885, opened 2021-02-11
PWGGA/GammaConv: Added new modes for merged pi0 in Jets (out of jet, perpendicular to Jet, around jet)

jokonig:


FriederikeBock:
+1

PR #16890, opened 2021-02-11
Small updates in Lc->pKpi systematics in pp at 13 TeV

mfaggin:
- Raw yield extraction: no more sigma+-20% cases
 - PID: valuies from coarser bin analysis confirmed


fgrosa (commented):
+1

fgrosa:
+1

PR #16899, opened 2021-02-12
Fixed adding handler in AliMCWeightTask

phuhnGithub:
Also included debug cout for performance checks; disabled via preprocessor macros.


mario-krueger:
+1

PR #16911, opened 2021-02-13
PWGGA/GammaConv: Prevent deletion of object in TList

jokonig:
this caused a seg. fault in some cases because the object in the TList was deleted because it got out of scope.
Now the object in the list is only deleted when list->Clear() is called


FriederikeBock:
+1

PR #16917, opened 2021-02-14
Remove offline trigger selection call in default function

mfasDa:
Even though no longer used better remove the hard coded
settings for pPb 2013 in order to avoid unforced errors.


mfasDa:
+1

PR #16941, opened 2021-02-16
Add possibility to apply multi-class models for Ds

stefanopolitano:


PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.cxx
```diff
@@ -1670,7 +1701,7 @@ void AliAnalysisTaskSEDs::CreateCutVarsAndEffSparses()
 
   Int_t nSparseAxes = knVarForSparse;
   if (!fApplyML)
-    nSparseAxes--;
+    nSparseAxes -= 3;
```
fgrosa, line 1704: Here you should differentiate between binary and multi-class classifications, because otherwise in case of binary classification you get 2 extra axes all filled with -1

PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.cxx
```diff
@@ -1837,19 +1869,20 @@ void AliAnalysisTaskSEDs::CreatePIDMLSparses()
...
-                                    nPIDbins, nPIDbins};
-  Double_t xminPID[knVarPID] = {0., fMLOutputMin, PIDmin, PIDmin, PIDmin, PIDmin, PIDmin, PIDmin, PIDmin, PIDmin, PIDmin, PIDmin, PIDmin, PIDmin};
-  Double_t xmaxPID[knVarPID] = {fPtLimits[fNPtBins], fMLOutputMax, PIDmax, PIDmax, PIDmax, PIDmax, PIDmax, PIDmax, PIDmax, PIDmax, PIDmax, PIDmax, PIDmax, PIDmax};
-
-  TString PIDvarnamesComb[knVarPIDcomb] = {"#it{p}_{T}", "ML model output", "n#sigmaCombPi_0", "n#sigmaCombK_0", "n#sigmaCombPi_1", "n#sigmaCombK_1",
-                                                 "n#sigmaCombPi_2", "n#sigmaCombK_2"};
-  Int_t nBinsPIDcomb[knVarPIDcomb] = {nPtBins, fNMLBins, nPIDbins, nPIDbins, nPIDbins, nPIDbins, nPIDbins, nPIDbins};
-  Double_t xminPIDcomb[knVarPIDcomb] = {0., fMLOutputMin, PIDcombMin, PIDcombMin, PIDcombMin, PIDcombMin, PIDcombMin, PIDcombMin};
-  Double_t xmaxPIDcomb[knVarPIDcomb] = {fPtLimits[fNPtBins], fMLOutputMax, PIDcombMax, PIDcombMax, PIDcombMax, PIDcombMax, PIDcombMax, PIDcombMax};
+  TString PIDvarnames[knVarPID] = {"#it{p}_{T}", "ML model output 0", "ML model output 1", "ML model output 2", "n#sigmaTPCPi_0", "n#sigmaTPCK_0", "n#sigmaTOFPi_0", "n#sigmaTOFK_0",
```
fgrosa, line 1872: also for this it would be convenient to differentiate between binary and multi-class classification. In particular you could move the axes of the ML outputs at the end of the array so that if it is binary classification you just reduce the number of the axes by 2

PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.cxx
```diff
@@ -1013,36 +1013,52 @@ void AliAnalysisTaskSEDs::UserExec(Option_t * /*option*/)
...
-            Double_t var4nSparsePID[knVarPID] = {ptCand, modelPred,
-                                                fMLResponse->GetVariable("nsigTPC_Pi_0"), fMLResponse->GetVariable("nsigTPC_K_0"),
-                                                fMLResponse->GetVariable("nsigTOF_Pi_0"), fMLResponse->GetVariable("nsigTOF_K_0"),
-                                                fMLResponse->GetVariable("nsigTPC_Pi_1"), fMLResponse->GetVariable("nsigTPC_K_1"),
-                                                fMLResponse->GetVariable("nsigTOF_Pi_1"), fMLResponse->GetVariable("nsigTOF_K_1"),
-                                                fMLResponse->GetVariable("nsigTPC_Pi_2"), fMLResponse->GetVariable("nsigTPC_K_2"),
-                                                fMLResponse->GetVariable("nsigTOF_Pi_2"), fMLResponse->GetVariable("nsigTOF_K_2")};
-
-            Double_t var4nSparsePIDcomb[knVarPIDcomb] = {ptCand, modelPred,
+            std::vector<Double_t> var4nSparsePID = {ptCand, modelPred[0], modelPred[1], modelPred[2],
```
fgrosa, line 1035: See comment in the `CreateCutVarsAndEffSparses` method, but it would be convenient to move the ML outputs at the end of this vector, so that if it is binary classification you can simply reduce by 2 the number of axes

PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.cxx
```diff
@@ -1080,35 +1096,50 @@ void AliAnalysisTaskSEDs::UserExec(Option_t * /*option*/)
...
```
fgrosa, line 1116: same as above

fgrosa (commented):
Hi @stefanopolitano thanks a lot! For me the PR is basically ok, except for a couple of points where it would be better to slightly modify the code to better handle the case of binary classification

PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.cxx
```diff
@@ -1013,36 +1013,52 @@ void AliAnalysisTaskSEDs::UserExec(Option_t * /*option*/)
 
         if (fApplyML)
         {
-          isMLsel = fMLResponse->IsSelected(modelPred, d, aod->GetMagneticField(), Pid_HF, 0);
+            isMLsel = fMLResponse->IsSelectedMultiClass(modelPred, d, aod->GetMagneticField(), Pid_HF);
```
fgrosa, line 1016: here you missed as last argument `0`

PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.cxx
```diff
@@ -1080,35 +1096,50 @@ void AliAnalysisTaskSEDs::UserExec(Option_t * /*option*/)
 
         if (fApplyML)
         {
-          isMLsel = fMLResponse->IsSelected(modelPred, d, aod->GetMagneticField(), Pid_HF, 1);
+          isMLsel = fMLResponse->IsSelectedMultiClass(modelPred, d, aod->GetMagneticField(), Pid_HF);
```
fgrosa, line 1099: here you missed as last argument `1`

PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.cxx
```diff
@@ -1013,36 +1013,52 @@ void AliAnalysisTaskSEDs::UserExec(Option_t * /*option*/)
...
+            isMLsel = fMLResponse->IsSelectedMultiClass(modelPred, d, aod->GetMagneticField(), Pid_HF);
+
+          if (isMLsel)
+          {
+            if (modelPred.size() == 0)
+            {
+              modelPred.push_back(-1.);
+              modelPred.push_back(-1.);
+              modelPred.push_back(-1.);
+            }
```
fcatalan92, line 1025: Why do you should have this case? If `isMLsel` is true `modelPred` should be filled.

PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.cxx
fcatalan92, line 1099: Here you should add `1` as the last argument to match the mass hypothesis

PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.cxx
```diff
@@ -1080,35 +1096,50 @@ void AliAnalysisTaskSEDs::UserExec(Option_t * /*option*/)
...
-          isMLsel = fMLResponse->IsSelected(modelPred, d, aod->GetMagneticField(), Pid_HF, 1);
+          isMLsel = fMLResponse->IsSelectedMultiClass(modelPred, d, aod->GetMagneticField(), Pid_HF);
+          if (isMLsel)
+          {
+            if (modelPred.size() == 0)
+            {
+              modelPred.push_back(-1.);
+              modelPred.push_back(-1.);
+              modelPred.push_back(-1.);
+            }
```
fcatalan92, line 1107: See above, it should not be necessary

fcatalan92 (changes requested):

PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.cxx
fgrosa: `isMLsel` is true also if no ML is applied and in this case if ve don't fill `modelPred` with three dummy values we get an error when we fill `var4nSparse`, so I would say that it is ok

PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.h
```diff
@@ -212,12 +222,13 @@ class AliAnalysisTaskSEDs : public AliAnalysisTaskSE
...
   /// variables for ML application
   Bool_t fApplyML = kFALSE;                     /// flag to enable ML application
+  Bool_t fMultiClassML = kFALSE;                /// flag to enable multi-class models (Bkg, Prompt, FD)
   Bool_t fUseMinimalVarForSparse = kFALSE;      /// flag to keep only inv. mass, pt and prob. in the sparse
   TString fConfigPath = "";                     /// path to ML config file
   AliHFMLResponseDstoKKpi* fMLResponse = nullptr;  //!<! object to handle ML response
-  Int_t fNMLBins = 300;                         /// number of bins for ML output axis in THnSparse
-  Double_t fMLOutputMin = 0.85;                 /// min for ML output axis in THnSparse
-  Double_t fMLOutputMax = 1.0;                  /// max for ML output axis in THnSparse
+  Int_t fNMLBins[3] = {100,100, 100};           /// number of bins for ML output axis in THnSparse
```
fgrosa, line 229: Since you modified these data members you should also update the `ClassDef` from 41 to 42

PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.cxx
fcatalan92: Effectively, `isMLsel` is used only if we are applying the model. See https://github.com/alisw/AliPhysics/blob/05b88d37eb04282ac4af04305501f7e9d8096ff5/PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.cxx#L1048
probably we could also initialize it to false.
Since now `var4nSparse` is a std::vector we could just insert the content of `modelPred` (without the need to add dummy values) in it. Then, the length of the vector is different depending on where we are using ML or not and multiclass or not but it will match the THnSparse structure by default.
So, we could remove L1018-1031 and initialize the vector like this:
```
std::vector<Double_t> var4nSparsePID = {..., ..., ...};
var4nSparsePID.insert(var4nSparsePID.begin(), modelPred.begin(), modelPred.end())
```

fgrosa: Indeed this solution would also work, but you mean `var4nSparse` and not `var4nSparsePID`, since the latter is only filled in case of `fApplyML` (and it's not a `std::vector`), is it correct?

fcatalan92: In this PR also `var4nSparsePID` is a `std::vector` https://github.com/alisw/AliPhysics/blob/820a05353136d1c245223975b6ce59e3d1712f4a/PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.cxx#L1035

fgrosa: Right, so to recap for @stefanopolitano:
- initialise `isMLsel` to `false`
- remove the all the `push_back(-1.)` both in case of null size and size equal unity
- initialise the `var4nSparse`, `var4nSparsePID`, and `var4nSparsePIDcomb` as @fcatalan92 suggested

stefanopolitano: I agree with Fabio, this could be a good way to fill var4nSparse according to the size of modelPred without introducing dummy values and we could use it when we have to insert the ML output in a vector (var4nSparse, var4nSparsePIDcomb). 
However, instead of:
   var4nSparsePID.insert(var4nSparsePID.begin(), modelPred.begin(), modelPred.end())
I would use:
   var4nSparsePID.insert(var4nSparsePID.end(), modelPred.begin(), modelPred.end())
In this way, we could have ML outputs at the end of the vector(s): if it is binary classification, we can simply reduce by 2 the number of axes as Fabrizio was suggesting. What do you think?

fgrosa: Yes I agree too! I would also do as you suggest in order to have the ML outputs at the end of the final vector and reduce the number of axes in case of binary classification. Thanks!

fcatalan92: Yes, indeed it was a typo! You should add them at the end in order to to use the trick.

PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.cxx
```diff
@@ -1669,8 +1672,10 @@ void AliAnalysisTaskSEDs::CreateCutVarsAndEffSparses()
...
 
   Int_t nSparseAxes = knVarForSparse;
-  if (!fApplyML)
-    nSparseAxes--;
+  if (!fApplyML && !fMultiClass)
+    nSparseAxes -= 3;
+  if (fApplyML && !fMultiClass)
+    nSparseAxes -= 2;
   if (fApplyML && fUseMinimalVarForSparse)
     nSparseAxes = knVarForSparseMLMinimal;
```
fcatalan92, line 1680: Now, the "minimal" configuration of the THnSparse has always 5 axes, right? We should have 3 for binary classification.

fcatalan92 (changes requested):

PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEDs.cxx
fcatalan92, line 1680: Hi @stefanopolitano, now the logic starts to become a bit cumbersome. You can rewrite everything as:
```
Int_t nSparseAxes = knVarForSparse;
if (!fApplyML) {
    nSparseAxes -= 3;
} else {
    if(fUseMinimalVarForSparse)
        nSparseAxes = knVarForSparseMLMinimal;
    if (!fMultiClass)
        nSparseAxes -= 2;
}

fgrosa:
+1

fgrosa:
+1

fgrosa:
+1

PR #16954, opened 2021-02-18
Fix a bug on centrality

SiyuTang:
Hi ,

Here is an update for Jet v2 analysis, please check it.

Thanks a lot!

Siyu


yzhouCERN:
+1

PR #16957, opened 2021-02-18

lhusova:
change the cascade cuts to by configurable from outside in the train vagon


zumoravc:
+1

PR #16968, opened 2021-02-19
 PWGGA/GammaConv/ fix bug for phos trigger mimic 

AMechler:
when running multiple wagons with the same calotriggermimichelper


nschmidtALICE:
+1

PR #16970, opened 2021-02-19
PWGGA/GammaConv: AliCaloTrackmatcher implemented new mode 7

fjonasALICE:
- implemented fRunningMode=7 for AliCaloTrackMatcher which considers all hybrid tracks for track matching (which should be used after testing as default imo)
- current default mode 0 will only match hybrid tracks with SPD hits that have not been constrained to primary vertex (good global tracks) and will ignore other tracks of hybrid sample that can be selected in analysis
- fixed a minor bug that only affects cases where a TM cut is used which is not pt dependent. Some matches were not matched by accident
- fixed an issue that should only occur when choosing a very wide TM cut. If, for some reason, one track is matched to multiple clusters, there was an issue that the second matched cluster was thrown away and never considered as matched
- added warnings for fRunningMode = 7. Negative IDs are not unique, even though hybrid tracks contain no doubles and no double IDs. However, if the user runs with fRunningMode = 7, but in his analysis has tracks that are not hybrid, this can lead to unwanted behavior. Asking e.g. in analysis for dEta dPhi values of a TPCOnly track when the trackmatcher ran over hybrid tracks would return the dEta and dPhi of the hybrid track instead of the TPCOnly track. This is because the TM can't distinguish the two types at the point where the user asks in the analysis. Also, if the user for some reason considers multiple copies of the same track, the trackmatcher will faithfully return multiple matches.Therefore it is important that the user is not looking for something that was never checked by fRunningMode = 7

### DIFF
--- a/PWGGA/GammaConv/AliAnalysisTaskGammaIsoTree.cxx
+++ b/PWGGA/GammaConv/AliAnalysisTaskGammaIsoTree.cxx
@@ -5174,12 +5174,29 @@ void AliAnalysisTaskGammaIsoTree::FillCaloTree(AliAODCaloCluster* clus,AliAODCon
   Float_t trackdPhi = 0;
   Float_t trackP = -1;
   Float_t trackPt = -1;
+
+  Bool_t fishyCase = kFALSE;
+  Double_t iso = isoCharged.isolationCone.at(2);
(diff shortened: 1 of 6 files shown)


nschmidtALICE:
+1

PR #16973, opened 2021-02-19
[PWGJE-23] Add singleton handling access to PYTHIA cross section file

mfasDa:
In order to minimize the amount of calls to
the pyxsec(_hists).root files a singleton is
used to cache the cross section value and
just pass it instead of load it again, if the file
is the same.

The reading of the pythia cross section via
singleton class is implemented in
AliAanlysisTaskEmcal as second fast option,
default is kept with the standard option. The
fast option can be selected on request and
will replace the default method after a
transition period.

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcal.cxx
+++ b/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcal.cxx
@@ -49,6 +49,7 @@
 #include "AliEmcalDownscaleFactorsOCDB.h"
 #include "AliEMCALGeometry.h"
 #include "AliEmcalMCPartonInfo.h"
+#include "AliEmcalPythiaFileHandler.h"
 #include "AliEmcalPythiaInfo.h"
 #include "AliEMCALTriggerPatchInfo.h"
 #include "AliESDEvent.h"
@@ -136,6 +137,7 @@ AliAnalysisTaskEmcal::AliAnalysisTaskEmcal() :
   fMCRejectFilter(kFALSE),
   fCountDownscaleCorrectedEvents(kFALSE),
   fUseBuiltinEventSelection(kFALSE),
+  fReadPyxsecFast(false),
   fPtHardAndJetPtFactor(0.),
   fPtHardAndClusterPtFactor(0.),
   fPtHardAndTrackPtFactor(0.),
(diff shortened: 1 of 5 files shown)


mfasDa:
@raymondEhlers Please review!

mfasDa:
+test

mfasDa:
+test

raymondEhlers (approved):
Hi @mfasDa , nice work!

I was going to suggest a modification of the implementation, but having written the explanation, I think it's overkill for our purposes. However, I would suggest a modification to be more explicit: I would change `UpdateFile(...)` to `GetInfo(...)` (or GetPythiaInfo or GetPythiaScaling or whatever), and I would have it return the scaling info directly. So practically, it's something like:

```c++
// This needs a more thoughtful name
class PythiaInfo {
  double fXsec;
  double fTrials;
}

// In AliAnalysisTaskEmcal
auto pythiahandler = PYTHIAFileHander::Instance();
try {
  PythiaInfo info = pythiahandler->GetInfo(currFile);
  fXsec = info.GetCrossSection();
  fTrials = info.GetNumberOfTrials();
} catch(PYTHIAFileHander::FileNotFoundException &e) {
  ...
}
```

There are a few benefits:

1) It's a lot more explicit and I think better encapsulates the implementation. I don't foresee any generalization, etc, but if so, this approach will be much easier (ie. change the cache values to a map, and now you can support many files in the cache).
2) Before reading all of the code, the functionality of `UpdateFile(...)` wasn't clear to me based on the name. After reading it, I now understand why it's called that, but I think it could be clearer. And if we directly return the info, it will be.

I don't want to take more of your time, so it's up to you if you actually want to change it. This is just something to consider

mfasDa:
+test

mfasDa:
@raymondEhlers Thanks for your suggestion! They sound fine and were straight forward to implement, so I just adapted the code. I think for the purpose as container for 2 values a POD struct is sufficient - what do you think?

As this method is still handled as optional while being tested and the default is for the moment the old method I think it can be merged right away.

raymondEhlers:
@mfasDa Thanks for implementing them! Yeah, POD struct sounds good here. Also agreed that you can merge when you're ready since it's off by default for now

mfasDa:
+1

@raymondEhlers Thanks for your review! Merging

PR #16976, opened 2021-02-20
included SPD tracklet vs. Zvertex

moeshita:
included SPD tracklet vs. Zvertex


fcolamar:
Ciao Momo, 
just a couple of comments about your commit:
- you could just use the function AliVertexingHFUtils::GetNumberOfTrackletsInEtaRange(aod,-1.,1.) to immediately retrieve the number of SPD tracklets in a given eta range (in this case, [-1,1]), without having to evaluate manually the eta for each tracklet
- I see that you define a method 'GetCorrectedNtrackletsD' in the header file, but never use such method
- The same for 'Nref' variable: it's defined in the .h, and initialised in the .cxx constructors, but then is used only in some commented lines of the .cxx, maybe is a leftover)?
Could you please update your commit accordingly? Thanks!

fcolamar:
+1

PR #16978, opened 2021-02-21
Update the jet v2 analysis in Pb-Pb

SiyuTang:
Hi,

Here is an update for Jet v2 analysis in Pb-Pb. Please check it.

Thanks!

Siyu


yzhouCERN:
+1

PR #16989, opened 2021-02-23
[PWGJE-23] Fix determination of file type in file handler

mfasDa:
File type (ESD/AOD) determination must be done before
path handling to cross section file. In addion adding a
few AliDebugStream in both AliAnalysisTaskEmcal and
AliEmcalPythiaFileHandler.


mfasDa:
+1

Trivial bug fix, spotted in local tests. Otherwise component works fine locally and refreshes the cache as expected once for all files.

PR #16951, opened 2021-02-17
Including the k0p task

lramona:
Commit for M. Urioni

PR #17003, opened 2021-02-24
Fix object names in Event mixing class

lramona:


vkucera:
Does this fix the bug in https://github.com/alisw/AliPhysics/pull/16951 ?

lramona:
yes, it should fix the bug

vvislavi:
+1

PR #17004, opened 2021-02-24
Include possibility for CVMFS paths to Treelite utility

lvermunt:
* Possibility to use storage on CVMFS to reduce pressure on network from HF application jobs
* Possibility to directly use precompiled models to avoid copying+compilation step as well

### DIFF
--- a/ML/AliMLModelHandler.cxx
+++ b/ML/AliMLModelHandler.cxx
@@ -18,6 +18,7 @@
 #include <TFile.h>
 #include <TGrid.h>
 #include <TSystem.h>
+#include <AliDataFile.h>
 #include "AliLog.h"
 #include "AliExternalBDT.h"
 
@@ -152,6 +153,24 @@ std::string AliMLModelHandler::ImportFile(std::string path) {
       AliFatalClass(Form("Error file %s not found! Exit", path.data()));
     }
     return path;
+  }
+
+  // check if file is on cvmfs
+  if (path.find("/cvmfs") != std::string::npos) {
+    bool checkFile = gSystem->AccessPathName(path.c_str());
+    if (checkFile) {
+      AliFatalClass(Form("Error file %s not found on CVMFS! Exit", path.data()));
+    }
+    return path;
+  }
+
+  // check if file is on cvmfs, but still requires to set the full path (so starts with PWG or AODB)
+  if (path.rfind("PWG", 0) != std::string::npos || path.rfind("AODB", 0) != std::string::npos) {
+    std::string path_cvmfs = AliDataFile::GetFileName(path);
+    if (path_cvmfs == "") {
+      AliFatalClass(Form("Error file %s not found on CVMFS! Exit", path.data()));
+    }
+    return path_cvmfs;
   }
     
   // check if file is on alien
--- a/ML/AliMLModelHandler.cxx
+++ b/ML/AliMLModelHandler.cxx
@@ -161,7 +161,11 @@ std::string AliMLModelHandler::ImportFile(std::string path) {
     if (checkFile) {
       AliFatalClass(Form("Error file %s not found on CVMFS! Exit", path.data()));
     }
-    return path;
+    // only skip copying for certain file types
+    std::string fileform = path.substr(path.find_last_of(".")+1,path.size());
+    if(fileform == "so" || fileform == "yml" || fileform ==  "yaml"){
+      return path;
+    }
   }
 
   // check if file is on cvmfs, but still requires to set the full path (so starts with PWG or AODB)
@@ -170,7 +174,12 @@ std::string AliMLModelHandler::ImportFile(std::string path) {
     if (path_cvmfs == "") {
       AliFatalClass(Form("Error file %s not found on CVMFS! Exit", path.data()));
     }
-    return path_cvmfs;
+    // only skip copying for certain file types
+    std::string fileform = path.substr(path.find_last_of(".")+1,path.size());
+    if(fileform == "so" || fileform == "yml" || fileform ==  "yaml"){
+      return path_cvmfs;
+    }
+    path = path_cvmfs;
   }
     
   // check if file is on alien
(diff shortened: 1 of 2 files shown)


mpuccio:
+test

ML/AliMLModelHandler.cxx
```diff
@@ -153,6 +154,33 @@ std::string AliMLModelHandler::ImportFile(std::string path) {
...
+    bool checkFile = gSystem->AccessPathName(path.c_str());
+    if (checkFile) {
+      AliFatalClass(Form("Error file %s not found on CVMFS! Exit", path.data()));
+    }
+    // only skip copying for certain file types
+    std::string fileform = path.substr(path.find_last_of(".")+1,path.size());
+    if(fileform == "so" || fileform == "yml" || fileform ==  "yaml"){
+      return path;
+    }
+  }
```
fcatalan92, line 169: I discussed a bit with @fgrosa which explained to me that this is intended to allow users to access the data on cvmfs locally using their installation of AliPhyscis. 
However, from my understanding, this won't work because locally you can access only data exporting `ALICE_DATA=root://eospublic.cern.ch//eos/experiment/alice/analysis-data' and reading it using ROOT methods.

ML/AliMLModelHandler.cxx
```diff
@@ -153,6 +154,33 @@ std::string AliMLModelHandler::ImportFile(std::string path) {
...
+    if (path_cvmfs == "") {
+      AliFatalClass(Form("Error file %s not found on CVMFS! Exit", path.data()));
+    }
+    // only skip copying for certain file types
+    std::string fileform = path.substr(path.find_last_of(".")+1,path.size());
+    if(fileform == "so" || fileform == "yml" || fileform ==  "yaml"){
+      return path_cvmfs;
+    }
+    path = path_cvmfs;
+  }
```
fcatalan92, line 183: The previous use case can be added here, checking if `path_cvmfs` contains `root://` and defaulting to always copy the files in that case. I would add also a disclaimer like in https://github.com/alisw/AliPhysics/blob/1e729f477822aaf60ba6f6c7f74617566df97f55/PWGHF/vertexingHF/AliAnalysisTaskSEImproveITSCVMFS.cxx#L1519-L1522
Finally, looking for `"PWG"` or `"AODB"` in the path could lead to conflicts for certain paths when you try to use the code reading models stored on your machine. So some sort of protection could be added, like checking for `cvmfs:` at the beginning of the path (and then removing it) like it is done for files on alien. Maybe there are smarter solutions BTW

fcatalan92:
Hi @lvermunt, thank you very much for the effort! In addition to the specific comments on the code, I have some additional considerations: 

- In principle, it is not necessary to read the configuration `.yaml` file for each job as in the current implementation of the code. We could read it one time in the `Init()` of the task (the relevant classes can be streamed) and then only do the model compilation/initialization in the `UserCreateOutputObjects()`. This is minor I would say and can be done in a future PR.
- I'm not sure that `.so` files compiled on a specific machine can be executed without problems on all the GRID sites. Maybe some investigation with experts and testing is needed, so you can add a warning when this option is used for the moment.

ML/AliMLModelHandler.cxx
lvermunt: Hi @fcatalan92, indeed, from a local installation this is not possible, because as you write it should use ROOT methods. However this method is intended for lxplus (or machines like aliceml) where one can access directly the files via a `/cvmfs` path

fgrosa: Hi Luuk, what Fabio and I realised is that if you run it loading AliPhysics from `cvmfs`, it works from scratch because it is covered by the next if statement (since `AliDataFile::GetFileName()` returns the `cvmfs` path if you do so), so indeed we can avoid to have this additional if.

lvermunt: Hi Fabrizio, sorry I should have clarified better. This part should be seen as a safety if the user provides the full cvmfs path (for some reason). Indeed the part with `AliDataFile` is the "main" part. So this can be removed if you think it is not worth to keep this as safety

lvermunt:
> * I'm not sure that `.so` files compiled on a specific machine can be executed without problems on all the GRID sites. Maybe some investigation with experts and testing is needed, so you can add a warning when this option is used for the moment.

Good point, I didn't think of this. I will check if locally compiled models work on my installation on aliceml. But to see if it works for all GRID sites we should test it after merging I guess

ML/AliMLModelHandler.cxx
fgrosa: Ok I got it now, so I have no strong opinion on this, indeed if you think that someone could set directly the `cvmfs` path we can also keep it, even if maybe in general it would be better to let `AliDataFile::GetFileName()` getting the correct path.

ML/AliMLModelHandler.cxx
fcatalan92: Hi @lvermunt, you can have this structure by accident if you are working locally and reading the models from your machine. In any case, it is an unlikely situation.

ML/AliMLModelHandler.cxx
fcatalan92: I'm for supporting only the case with `AliDataFile::GetFileName()`, but it is not a strong opinion.

mpuccio: I prefer having the possibility of specifying the path manually even from cvmfs

mpuccio:
On this last point: if you compile the models on a machine with SLC6 it should work without any issue (also on sites with CentOS7). Still, not many people use SLC6 and we should avoid uploading directly .so IMO, so maybe putting a warning like @fcatalan92 says would be good.

lvermunt:
Hi all, thanks for the feedback! I added some extra warning/disclaimers + a fix in case one runs locally using files from cvmfs (so it still copies). 

I kept the  `/cvmfs` part since Max wrote people in LF provide sometimes the full path. I also think it can be useful for tests to decouple the tag used for AliPhysics and the one for the cvmfs directory in case of old/overwritten/removed files.

mpuccio:
+1

PR #17005, opened 2021-02-24
PWGGA/GammaConv: Inclusion of a merged Cluster Analysis with Tree's for my Machine Learning Project

jhonerma:
Based on the existing GammaCaloMerged AnalysisTask I implemented an analysis to include trees. These trees will contain data about merged clusters that will be used by me to train neural networks. The end goal is to establish whether neural networks can distinguish between merged clusters from neutral pion decay's and background. For the moment i only included tree's that will contain data from MC, since these will be used as training data.


nschmidtALICE:
+1

PR #17019, opened 2021-02-25
feature-l0pp8TeV

mfasDa:
- Adding L0 trigger (2 GeV) for MC pp 8 TeV
- Setting trigger aliases for L0 triggers with names
  "EMCL0;EMC7;EMC8" and similar for DCAL
- Adding MC sample lhc17g6b anchored to pPb
  8 TeV


mfasDa:
+1

PR #17020, opened 2021-02-25
OADB for LHC15f with ZDC-based estimators

ercolessi:
@ddobrigk @noferini


ddobrigk:
@ercolessi , thanks a lot! But should we perhaps not overwrite the existing OADB-LHC15f.root file instead? It seems to me that this OADB makes the previous one completely obsolete, right? (it would be good to also cross-check that this is the case, even browsing through the calibration histograms that you did not touch to see if there's a good match)

ercolessi:
@ddobrigk, yes sure! Let me have a look to cross check that the two OADBs are matching and then I will change the name.

ercolessi:
Hi @ddobrigk, I noticed my OADB does not include all of the estimators contained in the current OADB-LHC15f.root, in particular these are missing:

- SPDTracklets08
- SPDTracklets08to15
- SPDTracklets08Corr
- SPDTracklets08to15Corr
- CL0
- CL1

I guess the current one is extracted using a different macro than the one I modified and used:
https://github.com/alisw/AliPhysics/blob/master/OADB/COMMON/MULTIPLICITY/macros/calibration/CalibratePeriodPP.C

I am not sure which one should I refer to then, can you help me? Then I can run the calibration again if necessary!

For what concerns the rest of estimators, most of the time I get a perfect match among the two OADBs, some runs have little differences, I attach an example for run 225716
immagine
immagine

Thanks!

ddobrigk:
Hi @ercolessi, thanks for the cross-checks... I think this is because @ffionda calibrated this dataset and I am not sure we committed the calibration macro in her case - I will write her an e-mail to be sure (with you in cc!). As for these tiny differences that you see in the existing calibrator, this, I think, is nothing very worrying.

Let's see if @ffionda can provide us with the code to include the custom objects - if she can do it, I would prefer making the old OADB obsolete.

ddobrigk:
+1

PR #17026, opened 2021-02-25
N_ch is measured in respective area

krschmit-cern:
- N_ch is again measured in respective area
- added sum pT of whole event and of respective area
- added mult of whole event vs mult of respective area


mario-krueger:
+1

PR #17037, opened 2021-02-26
PWGDQ: Add tasks to create tree from charmonium MC productions

rarnaldi:
PWGDQ: add tasks to create tree from charmonium MC productions. It works for both JPsi or Psi2S.


miweberSMI:
+1

miweberSMI:
+1

miweberSMI:
+1

PR #17039, opened 2021-02-27
fixed calculation of delta_phi

alcaliva:
Fixed calculation of DeltaPhi: 

    Double_t delta_phi = (180.0/TMath::Pi())*TVector2::Phi_0_2pi (phi_trk-phi_ref);


mario-krueger:
+1

PR #17040, opened 2021-02-27
Handle dependency on KFParticle as a CMake module

mfasDa:
Provinding FindKFParticle.cmake copied and
adpated from FindRooUnfold.cmake which finds
the KFParticle package based on KFParticle_ROOT
and export include path and library as cmake
properties. CMakeLists.txt in libraries which
make use of KFParticles are adapted, and
find_package(KFParticle) is added to the global
CMakeLists.txs

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -236,6 +236,9 @@ else()
   # ZEROMQ
   find_package(ZeroMQ)
 
+  # KFParticke
+  find_package(KFParticle)
+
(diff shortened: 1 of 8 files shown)


mpuccio:
+1

mfasDa:
+1

for PWGJE/EMCALJetTasks

pzhristov:
+1

pzhristov (approved):

PR #17042, opened 2021-02-28
Allowing ctests to run also without being inside an ALICE environment

mfasDa:


mfasDa:
+1 

For PWG/EMCAL and PWG/Tools

phuhnGithub:
+test

mfasDa:
Excluding fixes for EMCAL ctests in order to get them merged first as it is not sure when this pull request is going to be processed.

mfasDa:
At least launching the test would not harm ...

mfasDa:
Let's see whether someone from the few people who have approval rights will ever look at this ...

In 2 months I don't want to see a comment by @ktf "There was no activity on this PR since 2 months, can this be closed?". No - I want to have this merged! There is only no activity because it is ignored by the reviewers.

ktf:
+test

I am not sure this is a good idea. This will save in the cache the value of the PATH at the time you run cmake.

ktf:
+1

ktf:
we already have LD_LIBRARY_PATH...

mfasDa:
> we already have LD_LIBRARY_PATH...

Exactly this was the idea behind: You export the LD_LIBRARY_PATH but not the path, so this worked only half way. There was also a commit that added env, I assume with the same purpose to export the full path, but this doesn't work, because cmake expands this to /usr/bin/env instead of evaluating this.

@ktf Can you please also look at the corresponding commit in AliPhysics?

In O2 this worked all out-of-the box ...

ktf:
This is aliphysics?

mfasDa:
@ktf yes, the same is here for AliRoot: https://github.com/alisw/AliRoot/pull/1322

PR #17043, opened 2021-02-28
Fixing EMCAL track selection unit tests

mfasDa:
Condition for category 2 hybrid track selection
had to be negated as the if condition counts
the errors and not the successes.


mfasDa:
+1

PR #17049, opened 2021-03-01
PWGGA/GammaConv: Added two Cutstrings to the AddTask CaloMergedML

jhonerma:
I updated the AddTask for my own Analyses with two cutstrings that will become my default cut for trains.


nschmidtALICE:
+1

PR #17050, opened 2021-03-01
Adding PID hypothesis used during tracking

jgrosseo:


jgrosseo:
@akalweit

RUN3/AliAnalysisTaskAO2Dconverter.cxx
```diff
@@ -1239,6 +1239,10 @@ void AliAnalysisTaskAO2Dconverter::FillEventInTF()
     if (goldenChi2Status)
       tracks.fFlags |= TrackFlagsRun2Enum::GoldenChi2;
 
+    // Uppermost 4 bits contain PID hypothesis used during tracking
+    if (track->GetPIDForTracking() >= 0 && track->GetPIDForTracking() <= 15)
```
shahor02, line 1243: in principle, the track always have a PID assigned (default pion), no need for `if`.

jgrosseo: I agree. The return type is int, therefore I wanted the protection to make sure the other bits are not touched.

Btw the 3 first bits are also free for run3 tracks : we have a separate enum for run2 and run3 (distinguishable by the track type)

jgrosseo:
+1

jgrosseo:
+1

jgrosseo:
@pzhristov This fixes the MC association problem you mentioned today

pzhristov:
@jgrosseo Thanks, I confirm the tutorial MC example works. We can merge also https://github.com/AliceO2Group/AliceO2/pull/5566

jgrosseo:
Thanks. The O2 PR still has to wait. I want to validate with the converted data (and then also convert pp).

ktf:
+1

PR #17062, opened 2021-03-02
Fixing type of MC index columns

jgrosseo:


jgrosseo:
Interesting that this did not cause a warning in O2. Seems we do not check the type...

jgrosseo:
+1

PR #17073, opened 2021-03-02
Add pt weights in the generation tree and correct pt weights in the reconstruction tree

zhujh126:
Add pt weights in the generation tree and correct pt weights in the reconstruction tree


fgrosa:
+1

PR #17075, opened 2021-03-03
Splitting BC and Collision writing

jgrosseo:
- BC for all triggers
- Collisions if there was a vertex
- Store also vertexer Z and 3D
Updating extra run 2 tables

### DIFF
--- a/RUN3/AliAnalysisTaskAO2Dconverter.cxx
+++ b/RUN3/AliAnalysisTaskAO2Dconverter.cxx
@@ -68,9 +68,9 @@
 
 ClassImp(AliAnalysisTaskAO2Dconverter);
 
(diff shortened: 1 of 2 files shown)


RUN3/AliAnalysisTaskAO2Dconverter.cxx
```diff
@@ -900,65 +885,121 @@ void AliAnalysisTaskAO2Dconverter::FillEventInTF()
...
+      SETBIT (run2collinfo.fEventCuts, kINELgtZERO);
+    
+    if( multSelection->GetThisEventIsNotPileupInMultBins() )
+      SETBIT (run2collinfo.fEventCuts, kPileupInMultBins);
+    
+    if( multSelection->GetThisEventHasNoInconsistentVertices() )
+      SETBIT (run2collinfo.fEventCuts, kConsistencySPDandTrackVertices);
+    
+    if( multSelection->GetThisEventPassesTrackletVsCluster() )
+      SETBIT (run2collinfo.fEventCuts, kINELgtZERO);
```
mpuccio, line 993: Shouldn't this be a different bit?

ddobrigk: Yes, I noticed it about 15 minutes before you wrote here :-) 
Thanks!

jgrosseo:
+1

PR #17079, opened 2021-03-03
PWGGA/GammaConv: GammaIsoTree fixed bug in cluster processing & more

fjonasALICE:
- fixed loading of clusters which previously caused memory problems when running over multiple train configs in one train
- added trackingQA histos to check pPb with partial TPC coverage
- added primary track multiplicity to tree to study efficiency
- changed perp cone to include left _and_ right side in phi (previously only left), area is now 2xcone


nschmidtALICE:
+1

PR #17082, opened 2021-03-03
PWG/EMCAL: Adding correction component for low energetic clusters

jokonig:
This new correction component corrects the mismatch between the fraction of 1 cell clusters in MC and data.
A larger fraction of 1 cell clusters are observed in MC. The new correction flags a fraction of the 1 cell clusters via the Chi2() variable, those cells are then treated as if they would have 2 cells and thus passing the exotics and M02/M20 cut.
- The correction can be added via the yaml file with: ClusterLowEnergyEfficiency
- The functions for the corrections are still under preparation. For now the functions: kGammaAndElec or kAllClusters should be considered


gconesab:
+test

FriederikeBock:
+test

FriederikeBock:
+1

PR #17084, opened 2021-03-03
[PWGJE-23] Fix ESD type determination

mfasDa:
ESD type determination needs to be done at the
beginning of the function, similar as done for AODs
as the path becomes modified later and the type
no longer is available.


mfasDa:
+1

PR #17095, opened 2021-03-04
PWGDQ: Files modified in order to read AliAOD.root instead of AliAOD.Muons.root

rarnaldi:
code modified to run on AliAOD.root. Furthermore the run number is no more an argument in the AddTask


miweberSMI:
+1

PR #17109, opened 2021-03-05
Updated MC event header behaviour + new histos

osaarima:
The MC event header is now able to try to use several different mc headers.


osaarima:
Hello,
can this be checked?

Best,
Oskari

dongjokim:
+1

PR #17116, opened 2021-03-05
Added syst. unc. for Lc->pK0S pPb 5TeV BDT, low pT analysis

AnnalenaKa:
Added the systematic uncertainties for the Lc->pK0S analysis in pPb at 5.02TeV using BDTs including low pT interval below 1GeV/c


fcatalan92:
+1

PR #17119, opened 2021-03-05
Include iostream

pzhristov:


pzhristov:
+1

PWG/EMCAL/EMCALbase/AliEmcalPythiaFileHandler.cxx
```diff
@@ -37,6 +37,7 @@
 #include <TSystem.h>
 #include <TTree.h>
 #include <memory>
+#include <iostream>
```
mfasDa, line 40: Should in principle not be needed - iostream should come via AliLog.h

pzhristov: It is needed with the master of Root, the compilation breaks in

AliDebugGeneralStream("AliEmcalPythiaFileHandler::GetCrossSectionAndNTrials", 2) << "Returning cached value for cross section and number of trials" << std::endl;

since std::endl is not declared.

PR #17131, opened 2021-03-07
[PWGJE-23] Make new method for cross section loading default

mfasDa:
Switch to new method to load cross section and
number of trials from pythia file to default. Old method
will be kept as fallback for a couple of weeks in case
of problems observed in transition period.


mfasDa:
Tests started already now, due to the huge amount of time pull request testing needs by now, but merging should be held back until the ESD test is fully evalutated, both on the file server side and on the merging side.

mfasDa:
+test

mfasDa:
+1

mfasDa:
Commit is merged after it was verified by Costin that the number of file accesses is also reduced in the ESD test train.

PR #17147, opened 2021-03-08
Add luminosity histogram and trigger cluster counters

mfasDa:
- Add luminosity histogram filled for min. bias with
  number of events and for EMCAL triggers with
  downscale-corrected min. bias events * emcal
  trigger downscale factor
- Check run-by-run whether EMCAL triggers were
  active based on CTP configuration in the OCDB
  and fill histograms for EMCAL triggers only if the
  trigger was active
- Add histograms for trigger cluster counter based
  on triggered events


mfasDa:
+1

mfasDa:
Stupid ROOT5!

mfasDa:
+1 

Incredible

PR #17152, opened 2021-03-08
add a pT cut for the summed pTs of the pair

jungmichael:
Hi there,
I have implemented a cut for the sum of the transverse momenta of the pair (pT1 + pT2). Only pairs within specified limits (limit1 < summed pT < limit2) will be used for the same-event distribution. The default values of the limits are: limit1 = 0.0 and limit2 = 999.0. These limits can be changed using "config->SetSummedPtCut(limit1,limit2)" in the AddTask.
regards
Michael


otonvd:
+1

PR #17158, opened 2021-03-09
Automatically configure trigger normalization task based on run numbers

mfasDa:
Define default configuration for various datasets
(pp 13 TeV, p-Pb at various energies and years,
...), and use run number to select proper
configuration.


mfasDa:
+1

PR #17183, opened 2021-03-11
minor fix to denominator of efficiency is calculated + add rejection of h from K0s when analysing hybrid MC

ChiaraDeMartin95:


mario-krueger:
+1

PR #17186, opened 2021-03-12
Modify MC mother tracking

JinjooSeo:


xinyepeng:
+1

PWGHF/vertexingHF/AliAnalysisTaskSEXic0Semileptonic.cxx
```diff
@@ -1896,28 +1841,47 @@ void AliAnalysisTaskSEXic0Semileptonic::DefineEventTree()
...
+	 		if ((abspdgGranma > 500 && abspdgGranma < 600) || (abspdgGranma > 5000 && abspdgGranma < 6000)) isFromB=kTRUE;
+	    if(abspdgGranma==4 || abspdgGranma==5) isQuarkFound=kTRUE;
+			mother = mcGranma->GetMother();
+    	}else{
+	      printf("CheckOrigin: Failed casting the mother particle!");
+	      break;
+    }
+  }
+
+	if(!isQuarkFound){
```
fgrosa, line 1873: I would suggest you not to consider particles that do not have a charm or beauty quark in the decay chain as we do for D and Lc, also because if you will run this task in p-Pb or Pb-Pb in the future, you will include HIJING signals. Hence I would suggest you either to reject those signal or add a flag to choose if keep them or not (analogous of `searchUpToQuark` in https://github.com/alisw/AliPhysics/blob/16e39139f8a8b04b2b3b2060f330e17f63502ab4/PWGHF/vertexingHF/AliVertexingHFUtils.h#L166).

fgrosa (commented):
Hi @JinjooSeo, even if @xinyepeng already gave you +1 I would suggest you a minor modification. Could you implement this? Thanks!

PWGHF/vertexingHF/AliAnalysisTaskSEXic0Semileptonic.cxx
JinjooSeo: Hi @fgrosa , Now I understanding your motion. I modified it and will open PR. Thanks a lot!

PR #17198, opened 2021-03-13
Using AliVertexingHFUtils method to tag MC efficiency

chengtt0406:
correct tag for prompt and feed-down efficiency

### DIFF
--- a/PWGHF/vertexingHF/AliAnalysisTaskSEXic2eleXifromAODtracks.cxx
+++ b/PWGHF/vertexingHF/AliAnalysisTaskSEXic2eleXifromAODtracks.cxx
@@ -3139,7 +3139,7 @@ void AliAnalysisTaskSEXic2eleXifromAODtracks::DefineCascTreeVariables()
   fCandidateVariableNames[25]="EvNumber";
 
   for (Int_t ivar=0; ivar<nVar; ivar++) {
-    fCascVariablesTree->Branch(fCandidateVariableNames[ivar].Data(),&fCandidateCascVariables[ivar],Form("%s/f",fCandidateVariableNames[ivar].Data()));
+    fCascVariablesTree->Branch(fCandidateVariableNames[ivar].Data(),&fCandidateCascVariables[ivar],Form("%s/F",fCandidateVariableNames[ivar].Data()));
   }
 
   return;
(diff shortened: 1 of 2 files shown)


PWGHF/vertexingHF/AliAnalysisTaskSEXic2eleXifromAODtracks.cxx
```diff
@@ -3142,7 +3139,7 @@ void AliAnalysisTaskSEXic2eleXifromAODtracks::DefineCascTreeVariables()
   fCandidateVariableNames[25]="EvNumber";
 
   for (Int_t ivar=0; ivar<nVar; ivar++) {
-    fCascVariablesTree->Branch(fCandidateVariableNames[ivar].Data(),&fCandidateCascVariables[ivar],Form("%s/F",fCandidateVariableNames[ivar].Data()));
+    fCascVariablesTree->Branch(fCandidateVariableNames[ivar].Data(),&fCandidateCascVariables[ivar],Form("%s/f",fCandidateVariableNames[ivar].Data()));
```
fgrosa, line 3142: With this you fill the tree with `Float16_t` instead of `Float_t`, is is intended?

fgrosa (commented):
Ciao @chengtt0406 thanks for the implementation! Before giving you +1, I just have a minor question to be sure that a modification that you implemented is intended.

PWGHF/vertexingHF/AliAnalysisTaskSEXic2eleXifromAODtracks.cxx
chengtt0406: Hi, Fabrizio. Yes, I remember it is  because of the ROOT version.

fgrosa: Hi Tiantian, yes in the old ROOT versions `f` corresponded to `Float_t`, now `F` is the argument for `Float_t`. However now you are changing it to `f` that in recent ROOT versions corresponds to `Float16_t` (hence less precision, since `Float16_t` are 24 bit floating point with truncated mantissa instead of the usual 32 bits, see https://root.cern.ch/doc/master/classTTree.html). Did you want to reduce the precision of your variables to reduce the dimension of your tree?

chengtt0406: yes, you are right. Sorry, just now I did not read it carefully. It should be "F".  I remember I did the commit (changed 'f' to 'F')few days ago, it is strange that it is back again. It could be due to my git has a problem.  I will push again. Thanks for your check.

chengtt0406: The new version was committed just now. I was using the code version that the one before I made 'f' to 'F'. That's why the 'f' to 'F' was back again... sorry for the mistake

fgrosa:
+1

PR #17203, opened 2021-03-14
Add function ot obtain cross section weight from event header to AliAnalysisTaskEmcal

mfasDa:
- Add function to get the cross section weight in
  AliAnalysisTaskEmcal for the current event (PYTHIA
  or HepMC only)
- Add option in jet spectrum task to use the cross
  section weight to reweight the spectra on-the-fly


mfasDa:
+1

PR #17211, opened 2021-03-15
Fill cluster histograms with cross section weight from event header

mfasDa:
Only done for MC events. Needed for weighted
productions.


mfasDa:
+1

PR #17214, opened 2021-03-15
New 3D PsiPair Cut

eddoming:
New implementation in the Psipair, Optimization at 5 and 8 %, plus histograms to study the invariant mass of the electron-positron coming from the primary vertex


jokonig:
+1

PR #17219, opened 2021-03-15
AOD version of Femto task for Lambda-deuteron

stheckel:
o Up to now only NanoAOD version available
o Now this new task is for AODs, but along the lines of the Nano task
o Using same AddTask, which has been adopted


sartozza:
+1

PR #17228, opened 2021-03-16
PWGGA/GammaConv: OmegaToPiZeroGamma Rotation Method Hotfix

mhemmer-cern:
- Added previously missing check in rotation method when combining a pi0 from an original cluster and a changed cluster with another original cluster, that the two original clusters are not the same.


fjonasALICE:
+1

PR #17234, opened 2021-03-16
Update systematic uncertainties for LcTopK0S in pPb at 5TeV BDT lowpT

AnnalenaKa:
Update systematic uncertainties for LcTopK0S in pPb at 5TeV for BDT analysis at low pT


jezwilkinson:
+1

PR #17235, opened 2021-03-17
Fixed the implmentation of the SubWagon Suffix parameter in order to …

JColburn-CERN:
…properly add it to the End of the wagonName in Both Add Tasks.


raymondEhlers:
+1

PR #17245, opened 2021-03-17
 PWGGA/GammaConv: OmegaToPiZeroGamma - Fixed Pi0 pT Binning differences

mhemmer-cern:
- All Pi0 Pt bins should now have 200 bins and go from 0 to 20 GeV/c
- Also changed the isSignal flag to kFALSE when calling MesonIsSelected() for background Pi0s


fjonasALICE:
+1

PR #17249, opened 2021-03-18
PWGGA/GammaConv: Fixing problem for a case 1 in PsiPair

eddoming:
Correction to Background histograms in terms of the invariant mass of the electron-positron pair and correction to lineal dependence in PsiPair.


jokonig:
Hi Edgar, please increase your classdef since the persistent members have changed

jokonig:
+1

PR #17264, opened 2021-03-18
solved problem on missing generated particles + merged trees of Omega and Xi

iravasen:
Not all the generated particles were filled into the tree (check performed locally on single AliAOD.root files). In addition the trees of Omega and Xi have been merged to reduce the output size especially on data (expected a 20% reduction) even if the present solution is not yet optimal (in my opinion)


mario-krueger:
+1

PR #17267, opened 2021-03-18
Update for jet v2

SiyuTang:
Hi all,

Here is an update for the code of Jet v2 in Pb-Pb.
Please approve it.

Thanks a lot!

Siyu


yzhouCERN:
+1

PR #17275, opened 2021-03-19
Fixed bug on nsigma TPC cut + added cut on competing mass settable fo…

iravasen:
…r xi and omega (default: 0 for both)


bdoenigu:
+1

PR #17279, opened 2021-03-19
Fix loading of MC headers from cocktail headers (ESD case)

mfasDa:
In addition add static function creating the MCEventHandler


mfasDa:
+1

PR #17284, opened 2021-03-20
Added automatic conversion to MeV for mean and sigma

jackal1-66:
Added automatic conversion to MeV for mean and sigma labels shown in the fitter canvas output (blue labels)


fgrosa:
+1

PR #17292, opened 2021-03-22
Stupid ROOT5

mfasDa:
Workaround which would be handled in a safer way
with std::optional from c++17, but since we are forced
to be compatible with ancient ROOT5 and cannot
profit from the various improvement coming from
c++17 this workaround is needed.


mfasDa:
+1

PR #17294, opened 2021-03-22
Femto Lambda-deuteron: fix ITS PID, change pt

stheckel:
o Fix cut on ITS PID
o Change the pT range of (anti)deuterons to high purity only


a-mathis:
+1

PR #17297, opened 2021-03-22
Added cut variations for DCA_{xy} and PID

alcaliva:
Added extra dimension for PID systematics (TPC & TOF).
Added DCA_{xy} variations into track cuts
Successfully tested on the GRID.


mario-krueger:
+1

PR #17317, opened 2021-03-23
Possibility to cut on V0 radius

chiarazampolli:


chiarazampolli:
Ciao @fgrosa , @xinyepeng ,

This is the implementation to be able to cut on the V0 radius. I have not tested it yet, which is why I used the [WIP] label. If you want, you can already check if you spot mistakes. Otherwise I will try to check on 1 file locally tomorrow.

Chiara

chiarazampolli:
Mmmh... despite the [WIP], it seems that the PR will be merged if the tests are fine, because I can merge PRs in AliPhysics. Let's see tomorrow. Of course, I will test it.

fgrosa:
> Ciao @fgrosa , @xinyepeng ,
> 
> This is the implementation to be able to cut on the V0 radius. I have not tested it yet, which is why I used the [WIP] label. If you want, you can already check if you spot mistakes. Otherwise I will try to check on 1 file locally tomorrow.
> 
> Chiara

Ciao @chiarazampolli, the implementation looks ok for me, thanks!

chiarazampolli:
Ciao @fgrosa ,

I added one more change, since the class was expecting the V0 type as last variable and printing some messages if it was not. 
Now it will of course not work with old files, because variable 20 is not the V0 type anymore. 
If I had not changed this, I am also not sure it would have worked, because the variable is not initialized. 
So I think that people should regenerate the cut file, adding these two new variables.
Let me know what you think.

Chiara

fgrosa:
> Ciao @fgrosa ,
> 
> I added one more change, since the class was expecting the V0 type as last variable and printing some messages if it was not.
> Now it will of course not work with old files, because variable 20 is not the V0 type anymore.
> If I had not changed this, I am also not sure it would have worked, because the variable is not initialized.
> So I think that people should regenerate the cut file, adding these two new variables.
> Let me know what you think.
> 
> Chiara

Ciao @chiarazampolli , 
ok perfect! As soon as the PR gets merged @xinyepeng and I will write to the pK0s analysers asking them to reproduce their cut object.

chiarazampolli:
Ok, so I remove [WIP], and the PR will get merged. 

Note that like for every cut there is no default value. I think that if one does not want to apply any cut, the min Radius should be set to 0, and the max could be set to 100. 

I also now committed a further slight change, so that the value of the cut used is then used at the 2nd power, because AliAODv0::RadiusV0() returns (x*x + y*y). 

If you could check just once more, then I can remove the [WIP].

Chiara

PS: I tested the change this morning, which is how I found the issue with the order of the cuts, and things look fine to me.

chiarazampolli:
For the records: from GRP object:

**PbPb**
fVertexerV0Rmin               0.9                 min radius of the fiducial volume
fVertexerV0Rmax               100                 max radius of the fiducial volume

**pp**
fVertexerV0Rmin               0.2                 min radius of the fiducial volume
fVertexerV0Rmax               200                 max radius of the fiducial volume

One can check like this:
```
AliCDBManager* man = AliCDBManager::Instance()
man->SetDefaultStorage("raw://")
man->SetRun(296553) //PbPb 2018
AliCDBEntry* e = man->Get("GRP/Calib/RecoParam")
TObjArray* a = (TObjArray*)e->GetObject()
a->Print()
```
and then `fEventSpecie = 4` --> PbPb, `fEventSpecie = 3` --> pp

fgrosa:
> Ok, so I remove [WIP], and the PR will get merged.
> 
> Note that like for every cut there is no default value. I think that if one does not want to apply any cut, the min Radius should be set to 0, and the max could be set to 100.
> 
> I also now committed a further slight change, so that the value of the cut used is then used at the 2nd power, because AliAODv0::RadiusV0() returns (x_x + y_y).
> 
> If you could check just once more, then I can remove the [WIP].
> 
> Chiara
> 
> PS: I tested the change this morning, which is how I found the issue with the order of the cuts, and things look fine to me.

Great, thanks a lot! We will specify that they have to add the default values, which is indeed an important point.

chiarazampolli:
Ciao Fabrizio,

Note that I updated in the comment the value to not cut on the radius (I forgot 1 zero):
- min --> 0
- max --> 1000
As you can see from the reco + WDF values, these should have no effect.

Chiara

PR #17321, opened 2021-03-24
Add handling to load trigger offline masked FastORs from the OADB

mfasDa:
Masked FastOR location must be specified with
"oadb"


mfasDa:
+1

PR #17329, opened 2021-03-25
Add limit on constituents pT of a particle level jet

KotliarovAr:
Code author Filip Krizek has approved changes to the task


raymondEhlers:
+1

PR #17331, opened 2021-03-25
PWGGA/GamaConv/: Size of the cast arrays (ROOT Streamer)

eddoming:
Size of the cast arrays fixed with member persistence


jokonig:
+1

PR #17341, opened 2021-03-26
Add min radius and max radius cuts for V0 candidates

fgrosa:
@chiarazampolli @lvermunt


lvermunt:
@fgrosa, thanks! You missed the few config files in AliPhysics/PWGHF/vertexingHF/upgrade/ though

fgrosa:
> @fgrosa, thanks! You missed the few config files in AliPhysics/PWGHF/vertexingHF/upgrade/ though

Thanks! Indeed I forgot them so I updated them now. However there are few very old, that I did not modify, with only 10 and 17 variables (to be seen if we want to update them, but I fear that they will have to be revisited for all the channels)

chiarazampolli:
Ciao @fgrosa , 
I don't think we need to touch old configs, as you say, they will anyway not work if they have the wrong number of variables defined.

PR #17348, opened 2021-03-28
Update for Jet v2 analysis in Pb-Pb

SiyuTang:
Hi all,

Here is  an important update of Jet v2 analysis in Pb-Pb collisions, where we use the new V0 calibration from Alex. Please check them and approve it. Thanks!!

Siyu


yzhouCERN:
+1

PR #17359, opened 2021-03-29
PWGDQ: task to create a tree containing muon/dimuon infos from MC embedding

rarnaldi:
PWGDQ: task to create a tree containing muon/dimuon infos from MC embedding
Works with JPsi and Psi2S


miweberSMI:
+1

PR #17372, opened 2021-03-30
Add missing trigger cluster

mfasDa:
Cluster CALOFAST was missing, resulting in a segfault
when initializing the trigger cluster counters.


mfasDa:
+1

PR #17375, opened 2021-03-30
Update for jet v2 in PbPb

SiyuTang:
Dear all,

Here is an important update for Jet v2 analysis, which is related today's pag meeting. Please approve it .

Thanks!

Siyu


yzhouCERN:
+1

PR #17382, opened 2021-03-31
Correct shift for trigger cluster index

mfasDa:
Trigger clusters were shifted by 1 (value vs ID),
therefore entries were appearing for shifted
labels. Corrected with this commit.


mfasDa:
+1

PR #17390, opened 2021-03-31
Fixing naming of AliHFSystErr objects

strogolo:
1. In the first commit as discussed and agreed with @fgrosa, the default name of AliHFSystErr object in the constructor is moved to "AliHFSystErr" that is the default used in all macros.
2. In the second commit the naming of D0 syst err has been fixed


fgrosa:
+1

PR #17395, opened 2021-04-01
Update for jet v2

SiyuTang:
Hi all,

Here is an update for jet v2 analysis, where the more narrow binning is used, 

Please approve it! Thanks!

Siyu


yzhouCERN:
+1

PR #17407, opened 2021-04-01
Not using squared value for cut on RadiusV0

chiarazampolli:
...and a histogram 3D to study the signd0 effect, with optimized binning.


chiarazampolli:
@lvermunt , @andreaalici : when you set the cut for the V0radius, note that from now on it will not be squared in the RDHFcuts, so the value in the cut object is directly the one used to cut (as in the RecoParams).

PR #17413, opened 2021-04-02
Bugfix for DMC7 and DG1 trigger + support for mixed trigger

mfasDa:
- Bugfix in filling norm hist and cluster counter for
  DMC7 and DG1 (missing "," in trigger class vector)
- Add mixed triggers to trigger downscale factor
  cache if at least one trigger (EMCAL or DCAL) is
  present. If both are found the downscale factor
  is taken from the EMCAL side.


mfasDa:
+1

PR #17421, opened 2021-04-05
Update the bins of trigger pT

SiyuTang:
Hi all,

An update about binning changes in Jet analysis of p-Pb. Please approve it.

Thanks!

Siyu


yzhouCERN:
+1

PR #17423, opened 2021-04-06
Task to create a tree containing only single muons kinematics

rarnaldi:
Task to create a tree containing only single muons kinematics (to be used for event mixing)


iarsene:
+1

PR #17438, opened 2021-04-07
PWGGA/GammaConvBase: Changed MesonIsSelectedPiZeroGammaOAC

mhemmer-cern:
- Firstly, now only checks the angle between the gamma cluster coming
from the omega and the two gamma clusters coming from the Pi0. The
OAC for the two Pi0 gamma clusters should be handled by the Pi0MesonCuts
when calling MesonIsSelected!
- Secondly, now only one angle needs to be smaller or bigger instead of
all angles to reject the omega candidate

 Some blank changes are made automatically by my editor. Sorry for that.


jokonig:
+1

PR #17441, opened 2021-04-07
PWGHF: Add safety check for V0

lvermunt:
In recent MCs I noticed crashes due to zero pointer V0. Added safety check


jezwilkinson:
+1

PR #17449, opened 2021-04-08
PWGDQ: add TLorentzVectors with muon kinematics as a branch of the tree

rarnaldi:
PWGDQ: add TLorentzVectors with muon kinematics as a branch of the tree. To be used for event mixing code


iarsene:
+1

PR #17473, opened 2021-04-11
fixing HF selection

hdegen:
After several tests between tasks I found a better way to select HF electrons. This commit is to fix this and also remove (and understand with histograms) one unsafe cut that have been used in this task.


rbailhac:
+1

rbailhac:
+1

PR #17479, opened 2021-04-12
PWGHF: Update Lc PbPb syst

lvermunt:
Update systematics from preliminary values to ones of paper proposal. Another iteration still intended


fgrosa:
+1

PR #17481, opened 2021-04-12
[PWGJE] Add new task: AliAnalysisTaskJetExtractorPP

sbysiak:
I would like to add my new analysis task PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskJetExtractorPP.
This is almost a copy of older version of AliAnalysisTaskJetExtractor (newer is not compatible with pp), but with fixed memory leak.

### DIFF
--- a/PWGJE/EMCALJetTasks/CMakeLists.txt
+++ b/PWGJE/EMCALJetTasks/CMakeLists.txt
@@ -292,6 +292,7 @@ if(FASTJET_FOUND)
         UserTasks/AliAnalysisTaskHJetDphi.cxx
         UserTasks/AliAnalysisTaskHJetSpectra.cxx
         UserTasks/AliAnalysisTaskJetExtractor.cxx
+        UserTasks/AliAnalysisTaskJetExtractorPP.cxx
         UserTasks/AliAnalysisTaskJetPP.cxx
         UserTasks/AliAnalysisTaskSVtaskMCFilter.cxx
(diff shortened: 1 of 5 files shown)


hjbossi:
Hi @sbysiak, 

Looking over this quickly, I think there are a few additional adjustments that might be useful when adapting this task for pp. For example, this task assumes rho*A background subtraction because it is designed for PbPb and additionally there are some hard coded cuts (such as the min perc area cut, see lines 2022-2026) that you might not want in pp. This task is designed to also provide information on MC jets embedded into a PbPb background through variables such as the true pT fraction. As you won't be using this task for embedding, to avoid confusion it would be better to remove this functionality. Anyway, this was just a few thoughts after taking a quick look, I am happy to clarify my thoughts over email or mattermost!

Cheers, 

Hannah

sbysiak:
After private discussion with Hannah, I would like to ask to merge this PR @matplo @vkucera .
Background subtraction is to be used also in small systems and hard coded cuts appear only in AddTask, which is the default and should be anyway tuned by the user.
We will consider developing current AliAnalysisTaskJetExtractor to work also in pp, but for the time being I would like to have ability to run the LEGO train on this task, which is already tested, except for the memory leak, which was also fixed in this PR.

sbysiak:
After further discussion with Hannah (thanks!) it appeared that it is also possible to obtain desired output with custom settings in the updated AliAnalysisTaskJetExtractor, so I am closing the PR.

PR #17488, opened 2021-04-12
Adding option for DeltPhi bins, isMC

deepathoms:
EH Correlation:
- Add switch to change DeltaPhi bins in THnsparse
- Add switch for MC to apply exotic cluster removal, EMCal cluster TOF cut only in data

HFEBEtask:
- Add switch for MC to apply exotic cluster removal, EMCal cluster TOF cut only in data


deepathoms:
+1

PR #17500, opened 2021-04-14
 added a few QA histos

crisjahnke:
QA histos added to the task
Additional check on the trigger efficiency at J/psi pT pair also added


dsekihat:
+1

PR #17518, opened 2021-04-16
Make sure online bad channel map is applied to smeared signal also wh…

mfasDa:
…en the smear model is configured manually


mfasDa:
+1

PR #17521, opened 2021-04-16
Add pT weights (FONLLxMCatsHQ) for syst Pb-Pb 2018

strogolo:
Update of pT weights for D meson Pb-Pb 2018 analysis.
1. I extend the pT weights up to 50 GeV/c also in the 30-50 centrality class
2. I introduce the pT weights (FONLL x MCatsHQ) for systematic unc. evaluation


fgrosa:
+1

PR #17540, opened 2021-04-19
fix enabling of PID variables

jaimenorman:


jaimenorman:
@fgrosa I noticed after running a TPC variable was missing, which looks like it was due to the variable being enabled only if the ITS setting is enabled. I swapped an ITS and TPC branch definition which should fix this.

fgrosa:
+1

fgrosa:
> @fgrosa I noticed after running a TPC variable was missing, which looks like it was due to the variable being enabled only if the ITS setting is enabled. I swapped an ITS and TPC branch definition which should fix this.

thanks @jaimenorman!

PR #17555, opened 2021-04-21
AliAnalysisTaskHFJetIPQA:: change from considering TrackPt dependence…

kagarner:
… instead of Chi2 dependence for JP estimation


mfasDa:
+1

PR #17565, opened 2021-04-22
Add option to do trigger selection based on online patches

mfasDa:
In case of online patches the STU decision for a given
trigger patch is required in addition to energy / ADC
above threshold


mfasDa:
+1

PR #17574, opened 2021-04-23
Restructuring Lc --> K0Sp task to be able to run multiple models at t…

chiarazampolli:
…he same time.

Configuration changes correspondingly in the AddTask.
It was tested comparing the output with what one could get with the old implementation,
with 2 models.
Improved also the binning of the 2D histograms, needed to run multiple models.
Removed use of external library, to simplify the task (in any case, we don't use it).


chiarazampolli:
@andreaalici , for your information.

PR #17581, opened 2021-04-25
Small Fix on MUON tracks AOD converter to RUN3 forward coordinate system

vfeuilla:
This change is to avoid that the parameter "Sign/pT" takes the value zero, which would cause division by zeros in the in the evaluation of pT.


vfeuilla:
@iarsene, @rpezzi, @mguilbau.
Sorry I missed it, but there is a change needed here. With the way the protection was made, it was possible for fSigned1Pt to be equal to zero, which causes obvious divisions by zero when computing pT and Eta in O2.

iarsene:
@vfeuilla  Thanks Victor. It is fine from my side.
@jgrosseo  This can be approved from my side

vfeuilla:
@jgrosseo, @mpuccio, Sorry to ping you but can one of you approve this? Thanks!

jgrosseo:
+1

PR #17587, opened 2021-04-26
Update AliAnalysisTaskEmcalJetEnergyFlow.cxx

ChrisPliatskas:
Enabled the SetMakeGeneralHistograms option in order to get the Nevents histogram but also to facilitate scaling with pt hard binning when running om Pythia productions


raymondEhlers:
+1

PR #17588, opened 2021-04-26
PWGHF: Fix for multitrial fitter

lvermunt:
Multitrial fitter crashed in case background only fit failed (when `Background()` in `Significance()` was called). Added safety check using return of fitter


fprino:
+1

PR #17594, opened 2021-04-27
Registering output container connected to AO2D.root file.

mgheata:
Due to the non-standard way to handle the output file in the AO2D conversion task, we need to create an un-connected output container having as output file AO2D.root. This helps the system in the lego environment understand that this file needs to be registered, which is not the case when running the wagon standalone in a train.


mgheata:
@pzhristov or @jgrosseo can you approve this?

pzhristov:
+1

PR #17599, opened 2021-04-27
Increase of multiplicity reach for Pb-Pb

sustripathy:
This commit increases the multiplicity reach in the generation macro for Pb-Pb collisions.


aortizve:
+1

PR #17622, opened 2021-04-29
Updated the Jet Charge Template function to include generator level calculation

JColburn-CERN:
Updated the Template function to also report particle and detector level jets and a connection between the two, Additionally now the Values of K used and the Mother fraction used can be set in the Steering macro.


raymondEhlers:
+1 I don't have time to review the code, but I know you're under time pressure, so I'm approving it now

PR #17638, opened 2021-05-03
Add missing "," separating datasets

mfasDa:
Typo leading to misconfiguration when running on
J/Psi MC sample anchored to 2018 pp 13 TeV. Bug
introduced in AliPhysics/#17611


mfasDa:
+1

PR #17643, opened 2021-05-03
Correction on MUON tracks AOD converter to RUN3 forward coordinate system

vfeuilla:
Modification to skip muon tracks which parameters are left to thir default values. These tracks don't correspond to real muon tracks and can cause crashes if converted.


vfeuilla:
@iarsene, @jgrosseo, @rpezzi 
Hi all, here is a small modification to skip some problematic muon tracks in the converter to AODs. Thanks!

jgrosseo:
+1

jgrosseo:
Usually equal comparisons are discouraged for floating point. Does the == work for FLT_MAX under all cases?

PR #17663, opened 2021-05-06
 PWGGA/GammaConv/    AliAnalysisTaskGamma.. Calo & ConvCalo

AMechler:
Include additional cases to the additional True meson search

### DIFF
--- a/PWGGA/GammaConv/AliAnalysisTaskGammaCalo.cxx
+++ b/PWGGA/GammaConv/AliAnalysisTaskGammaCalo.cxx
@@ -6100,6 +6100,7 @@ void AliAnalysisTaskGammaCalo::ProcessTrueMesonCandidates(AliAODConversionMother
   }
   Int_t gamma1MCLabel = TrueGammaCandidate1->GetCaloPhotonMCLabel(0);   // get most probable MC label
   Int_t gamma1MotherLabel = -1;
+  tmpGammaMotherlabel = -1;
   // check if
 
   TParticle * gammaMC1 = 0x0;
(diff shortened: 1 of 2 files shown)


PWGGA/GammaConv/AliAnalysisTaskGammaCalo.cxx
```diff
@@ -6084,8 +6086,21 @@ void AliAnalysisTaskGammaCalo::ProcessTrueMesonCandidates(AliAODConversionMother
   }
   if (TrueGammaCandidate1->GetIsCaloPhoton() == 0) AliFatal("CaloPhotonFlag has not been set. Aborting");
 
+  Bool_t previouslyNotFoundTrueMesons = kFALSE;
```
fjonasALICE, line 6089: probably will not/rarely happen but maybe a good idea to put a depth counter to avoid infinite loops. If I recall correctly it can sometimes happen for some MCs that there are loops in the mother daughter chain and then this while loop goes on forever so maybe depth = 20 and increase a counter during each while loop

PWGGA/GammaConv/AliAnalysisTaskGammaConvCalo.cxx
```diff
@@ -5831,6 +5831,20 @@ void AliAnalysisTaskGammaConvCalo::ProcessTrueMesonCandidates(AliAODConversionMo
     }
     if (TrueGammaCandidate1->GetIsCaloPhoton() == 0) AliFatal("CaloPhotonFlag has not been set. Aborting");
 
+    Bool_t previouslyNotFoundTrueMesons = kFALSE;
+    Int_t tmpGammaMotherlabel = gamma0MotherLabel;
+    while (tmpGammaMotherlabel > 0) {
```
fjonasALICE, line 5836: see comment from above

PWGGA/GammaConv/AliAnalysisTaskGammaCalo.cxx
AMechler: ok I can do that

PR #17670, opened 2021-05-07
PWGGA/GammaConv(Base): OmegaToPiZeroGamma - changed OAC Pi0 Gamma Function

mhemmer-cern:
- Replaced the function which was used to determine the lower and upper bound for the angle between PiZero and Gamma with two separate functions
- Changed the corresponding function in AliConversionMesonCuts


jokonig:
+1

PR #17671, opened 2021-05-09
PWGDQ: add new muon and dimuon variables to the MC embedding tree

rarnaldi:
PWGDQ: add new muon and dimuon variables to the MC embedding tree


iarsene:
+1

PR #17678, opened 2021-05-10
TRDPID_parameters

ailec:
Committing the TRDPID parameters for LHC17 and LHC18 pp data sets. Also the TRDPID parameters for LHC18r PbPb data taking period


fprino:
+1

PR #17691, opened 2021-05-11
Allows for extraction of the pT spectra for particles used in kt and …

makor:
…mult analyses.

Note: Only records pairs of particles if k*<200 MeV/c.


sartozza:
+1

PR #17733, opened 2021-05-19
Add include unistd.h for compilation on MacOS

strogolo:
The compilation on MacOS was failing because of the missing "#include <unistd.h>". 
This compilation issue is fixed after including unistd.h in the .cxx of the task.


altsybee:
+1

PR #17742, opened 2021-05-20
Add flag and method for R/S study

strogolo:
Adding flag to activate the splitting of candidates for R/S studies. The default of the flag is kFALSE.
It will fill the output for three cases separately:
- only D0
- only D0bar
- D0D0bar


fgrosa:
+1

PR #17749, opened 2021-05-21
Attempt to fix memory leak in event-wise CS

raymondEhlers:
It seems to only occur with jemalloc (ie. on the grid)

@lhavener


raymondEhlers:
+1

PR #17757, opened 2021-05-24
Add simple analysis task monitoring the patch position of online patches from STU data

mfasDa:
Mainly in order to check the acceptance of the
trigger at L1
- Purely rely on information in AliVCaloTrigger
- Monitor trigger classes and
  combined/mixed/exclusive classes between
  EMCAL and DCAL (no distinction between high
  and low threshold)


mfasDa:
+1

PR #17764, opened 2021-05-25
Adding efficiency weight in order to propagte leg PID efficiency to the pair level

aharlend:
Adding efficiency weight in PWGDQ/reducedTree/AliReducedVarManagerin in order to enable propagating of leg PID efficiency to the pair level via standard J/psi MC


iarsene:
+1

PR #17773, opened 2021-05-26
Updates toward finalization including up-to-date ODC/BBD correction

Chongk:
Updates toward final vdM analysis: major changes are as follows.
 - Steering script (Execute.sh) updated tor better syst. err handling
 - ODC/BBD and Optical rate correction part updated/integrated
 - QA codes are added/updated


Chongk:
It looks all conveners are very busy. I'll open the pull request again. The purpose of these vdM code update is general maintenance, after all.

PR #17786, opened 2021-05-27
Fix masking of DCAL TRUs

mfasDa:
In case a DCAL TRU is masked in the STU the global
TRU index need to be used. The global TRU index  is
including TRUs in EMCAL STU region and virtual TRUs
in PHOS region. As luckly the only the EMCAL TRUs
were masked by mistake, and the DCAL 1/3rd
supermodules were dead in readout as well the bug
did not harm so far.


mfasDa:
+1

PR #17789, opened 2021-05-27
RUN3 conversion: If no TOF is available set negative time

njacazio:
- If no TOF is available set negative time
- Set name/title arrays for easier diff


njacazio:
@jgrosseo this fixes an issue spotted by @pbuehler indeed I gave it for granted that the TOF times were written negatives when TOF was not available but this was not the case.
I also took the initiative to flatten the array of names/titles as conflicts are easier to solve.
This is now compatible with O2, before we were not

njacazio:
This should also fix the empty TOF Chi2

jgrosseo:
+1

PR #17795, opened 2021-05-27
Mask hander providing EMCAL trigger mask in various formats from OCDB

mfasDa:
Providing trigger mask from OCDB at L0 and L1
- as list of FastOR abs. ID
- as list of position in the EMCAL space


mfasDa:
+1

mfasDa:
Stupid ROOT5!

zhaomr13:
> Stupid ROOT5!

seems that I have the seem problem. I agree, stupid ROOT5!

mfasDa:
+1

PR #17820, opened 2021-05-31
Check that the expected time is positive

njacazio:


njacazio:
This should protect against cases when no expected time is available for the track, I will align the O2 code for this possibility

jgrosseo:
+1

PR #17826, opened 2021-06-01
Added AddTask macros for the inclusive j/psi cross section analysis at 13TeV to attic

ilofnes:
Added AddTask macros for the inclusive j/psi cross section analysis at 13TeV to attic.
Added the possibility of setting the background fit function options using a TString in AliRessonanceFits. The default is set to be "MEI0", which was previously hardcoded.


iarsene:
+1

PR #17835, opened 2021-06-01
Add container class for offline trigger mask

mfasDa:
- Mask can have status dead or bad
- FastORs handled via FastOr abs ID
- Container browsable, in this case a
  2D col-row plot is created
- Corresponding unit tests added for
  - In-memory set/get
  - ROOT File-I/O
  - Modify
  - Clean


mfasDa:
+1

All unit tests passing, commit can safely be merged

mfasDa:
Adaption in trigger maker will follow in second dedicated commit

mfasDa:
Stupid ROOT5!

mfasDa:
+1

PR #17842, opened 2021-06-02
Add handling for offline trigger mask via trigger mask container

mfasDa:
Support for both types of trigger mask containers
(new trigger mask container and old TObjArray with
list of FastORs) implemented, however on in case of
the new trigger mask container it is possible to
distinguish bad from dead channels and therefore
allow to enable/disable them separately.


mfasDa:
+1

PR #17845, opened 2021-06-02
Fix browse function

mfasDa:
Based on the implementation of
AliEMCALTriggerMapping access to the EMCAL
geometry is mandatory. The geometry is initialized
based on the name of the geometry, selecting for
run2 the standard geometry and for run1 the
geometry from 2013, as the TRU layout for the
existing supermodules has no effects on earlier
setups in run1 with partial geometry.


mfasDa:
+1

PR #17849, opened 2021-06-03
Dividing the Charge Particle level tracks by 3 

JColburn-CERN:
since the charge on a particle level track are in units of 1/3 e instead of units of e. Thus correcting the mistake causing Particle level templates to be much broader.


raymondEhlers:
+1

PR #17853, opened 2021-06-03
Fix several default arguments of AliAnalysisTaskEmcalLight

mfasDa:
Were overlooked when adding handling for
Herwig and HepMC MC input. Without default
handling the parameters might artificially be set
to true, leading in MC information from headers
being overwritten with 0.


mfasDa:
+1

PR #17855, opened 2021-06-03
Add getter for event weight (weighted production) in AliAnalysisTaskEmcal

mfasDa:
Getter only implemented for pythia and HepMC-based
productions. Use getter also in the spectrum and
part. level task. In case of the jet analysis task the
user has to define whether apply cross section or
event weighting.


mfasDa:
+1

PR #17856, opened 2021-06-03
Update AliAnalysisTaskUpcFourPi.cxx

pgamamed:
changed the number of bins from 300 to 500 in fAllPtFourPions histogram


mbroz84:
+1

PR #17864, opened 2021-06-04
Add plotting function for easy plotting of the L0 trigger mask for a given run

mfasDa:
- Plotting masked channels as histogram
- For easy reading a grid separating FastORs,
  TRUs and Supermodules is added as well. The
  grid is shown based on run1 or run2 geometry
  depending on the run number


mfasDa:
+1

mfasDa:
+1

PR #17866, opened 2021-06-04
Fix handling of the 1/3rd supermodules in run2

mfasDa:
The mask for the 1/3rd supermodules is the same as
in run1, while the masking handler tried to apply the
run2 mapping. Fixed in the way that the mapping
handlers need an argument for the SM type, and in
case it is the 1/3 type the old mapping is applied.


mfasDa:
+1

PR #17879, opened 2021-06-08
Use mask handler in trigger maker to read OCDB trigger mask

mfasDa:
Use mask handler for trigger mask reading from the
OCDB and remove builting trigger mask reading. With
this the same code can be used for visualization of the
trigger mask and utilization in the trigger maker, so
the trigger maker can make use of a well tested and
unit-testable code.


mfasDa:
The new mask handler has been tested to work against the mask obtained directly from DCS. Nevertheless before merging one test run should verify that the mask is ported properly to the trigger maker.

mfasDa:
+test

mfasDa:
+test

mfasDa:
+1

All parts checked - PR is ready for merging

PR #17884, opened 2021-06-08
PWG/EMCAL: Added possibility to modify/scale the cell energy

jokonig:
This can be configured in the yaml file via "enableAdditionalScale" which is set to 1 (no modification) as default. This functionality is introduced to make checks for future cell calibrations, it is not supposed to be used for physics analyses!


raymondEhlers:
+1

PR #17886, opened 2021-06-08
Adjust scaling histograms for AliAnalysisTaskEmcalLight

mfasDa:
Exclude historgrams with tag "PtHard" from list of
scaling histograms (affects only tasks based on
AliAnalysisTaskEmcalLight. Also add getters for
the scaling histogram names for user access.


mfasDa:
+1

PR #17888, opened 2021-06-08
Update of jet particle v2 in p-Pb of AMPT

SiyuTang:
Dear all,

Here is an update of jet particle v2 in AMPT p-Pb, please approve it.

Thanks!

Siyu


yzhouCERN:
+1

PR #17897, opened 2021-06-09
Draw L1 mask and fix double delete

mfasDa:
- Add draw funtion for mask at L1 similar to the draw
  function for the mask at L0
- Don't delete the cache object, just set the pointer to
  nullptr, because the handler does not have
  ownership of the OCDB entry of the DCS config


mfasDa:
+1

PR #17901, opened 2021-06-10
Preparing the move to KFParticle package, but with the flag as it is nothing changes

amaringarcia:
with guidance from Victor Gonzalez

### DIFF
--- a/PWGGA/GammaConv/AliAnalysisTaskConvCaloCalibration.cxx
+++ b/PWGGA/GammaConv/AliAnalysisTaskConvCaloCalibration.cxx
@@ -43,7 +43,7 @@
 #include "AliVParticle.h"
 #include "AliESDtrack.h"
 #include "AliESDtrackCuts.h"
-#include "AliKFVertex.h"
+#include "AliGAKFVertex.h"
 #include "AliGenCocktailEventHeader.h"
 #include "AliConversionAODBGHandlerRP.h"
 #include "AliAODMCParticle.h"
--- a/PWGGA/GammaConv/AliAnalysisTaskEtaToPiPlPiMiGamma.cxx
+++ b/PWGGA/GammaConv/AliAnalysisTaskEtaToPiPlPiMiGamma.cxx
@@ -42,9 +42,9 @@
 #include "AliESDv0.h"
 #include "AliESDEvent.h"
 #include "AliESDpid.h"
-#include "AliKFParticle.h"
+#include "AliGAKFParticle.h"
 #include "AliMCEventHandler.h"
-#include "AliKFVertex.h"
+#include "AliGAKFVertex.h"
 #include "AliTriggerAnalysis.h"
 #include "AliCentrality.h"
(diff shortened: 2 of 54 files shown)


amaringarcia:
Dear @ktf , @pzhristov and @davidrohr, we did not expect this PR will reach you because it is just in the scope of the GammaConv and GammaConvBase modules. That's why we were not adding to the PR any explanations about what we did. This comment tries to cover that lack
The main goal of the step we took is to move from AliKFParticle to KFParticle support within Gamma Conversions.
To do that we have redirected all references to AliKFParticle to a new 'complex', AliGAKFParticle within GammaConvBase, which behavior is decided at production time with the flag set(USEKFLIBRARY FALSE) in CMakeLists.txt of GammaConvBase. The fact that AliKFParticle was referenced from quite a number of source files within both GammaConv and GammaConvBase made quite large the impact of this PR.
As you can easily check, with the current value of the USEKFLIBRARY flag (FALSE), the behavior of the AliGAKFParticle complex is just to redefine all references to AliGAKFParticle to AliKFParticle which, basically, leaves the scene exactly the same as before the PR.
We have tested the whole AliPhysics production with this value and all modules which reference any of the GammaConv and GammaConvBase modules are produced correctly and the whole AliPhysics is correctly built. The same success was obtained with the setting set(USEKFLIBRARY TRUE).
In this case, with the USEKFLIBRARY flag set to TRUE, the AliGAKFParticle complex introduces a wrapper class which transforms the AliKFParticle interface to the KFParticle interface in the points that they differ. 
The interface transformation is being completed right now but we wanted to have in advance the 'transparent' version of what we did to have an analysis results baseline in order to compare them with the ones introducing the final KFParticle support.
Hope we managed to explain our intentions and, please, do not hesitate to contact us back should you need any additional clarifications.

ktf:
+test

pzhristov:
Hi @amaringarcia,

Would it be possible to avoid creating GA-specific KF particle and to patch AliKFParticle so that it uses the new KFParticle package? I do not see any reason to keep the obsolete and buggy KF implementation in AliRoot.

Cheers,
Peter

amaringarcia:
Dear @pzhristov ,
the idea we have is to use this transition ONLY in Run2 software , and for O2 move to the newest KFParticle package.
We also need to do some large tests to find/quantify  the differences, as we have used until now the AliKFParticle package.
So we need a tag version where we will switch on the flag to true.
The AliKFParticle is used in many other tasks (not related to photons), so I do not think we should touch it. We are also NOT the authors. My understanding is that in general ALL users should move to the external KFParticle package, and in the future the AliKF should be declare obsolete and not used anymore.
Regards,
Ana

amaringarcia:
Dear  @ktf (Giulio Eulisse), @pzhristov (Peter Hristov), @davidrohr (David Rohr), sorry for this error. I could only test it for the master... Hopefully it is ok now.
Best regards,
Ana

pzhristov:
+test

PR #17910, opened 2021-06-11
PWGGA/GammaConv: Fixed output for JJ MC-Events in my Analysis

jhonerma:
The trees of my analysis will now also be connected to the output for JJ MC-events, which was previously not the case.


fjonasALICE:
+1

PR #17922, opened 2021-06-13
PWGLF/SPECTRA: dont compile deprecated tasks that produce tons of warnings

mario-krueger:


mario-krueger:
FYI:
- @phuhnGithub I had to add an include in your code (frankly I dont really understand why removing the other tasks had that side-effect...)
- @yelmard I had to remove some unneded includes in your INEL0 task

PR #17932, opened 2021-06-15
AliAnalysisTaskHFJetIPQA:: enable splitting of data sample in even/un…

kagarner:
…even events for unfolding and tree output


mfasDa:
+1

PR #17935, opened 2021-06-15
Add the TPC dE/dx_tuned_on_data to check the TPC PID

lupengzhong:
Hi Ionut, 

Could you please have a look at my pull request?  I implemented a new variable TPC dE/dx_tuned_on_data on reducedTree, in order to check the TPC PID on MC, the goal is trying to improve the agreement between Data and MC. This suggestion is from the TPC expert Jens and the TPC PID calibration group. 
Thanks!

Pengzhong & Xiaozhi


iarsene:
Hi,
This should be fine, however, i do not see from where the tuned on data TPC signal is computed. I see that you fill it from the dielectron VarManager, however, this variable does not seem to be defined there. This code in principle would not compile so i can't approve it. 
Cheers,
Ionut

PWGDQ/reducedTree/AliAnalysisTaskReducedTreeMaker.h
```diff
@@ -92,6 +92,7 @@ class AliAnalysisTaskReducedTreeMaker : public AliAnalysisTaskSE {
 
   // TStrings with active or inactive branches
   void SetTreeActiveBranch(TString b)   {fActiveBranches+=b+";";}
+
```
iarsene, line 95: Please remove this space. There is no reason to touch this file.

PWGDQ/dielectron/core/AliDielectronVarManager.h
```diff
@@ -1257,6 +1258,7 @@ inline void AliDielectronVarManager::FillVarESDtrack(const AliESDtrack *particle
 
 
   values[AliDielectronVarManager::kTPCsignal]=particle->GetTPCsignal();
+  values[AliDielectronVarManager::kTPCsignalTunedOnData]=particle->GetTPCsignalTunedOnData();
```
iarsene, line 1261: This should be fine. What about the AOD tracks? Is there a tune on Data tpc signal also for AOD tracks? Please implement it also there.

PR #17940, opened 2021-06-16
AliPhysics Run3 converter saves a bitmap with fired muon chambers

nicolovalle:
A bitmap for the 10 muon chabers (each bit is equal to 1 if the chamber is fired) is saved into the AO2D. 
It propagates the information from AliESDMuonTrack::fMuonClusterMap


jgrosseo:
+1

PR #17942, opened 2021-06-16
Small update and cleanup of AliAnalysisTaskEmcalJetEnergyFlow task

ChrisPliatskas:
The following changes are included:
a) The IsMCprod data member is properly initialized and assigned the value from the add task method so that it can handle the analysis of MC as well as data.
b) Small fixes in the if conditions of the Response matrix creation to avoid errors due to array indices
c) Added 1 debug/printout option.

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.cxx
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.cxx
@@ -822,11 +822,12 @@ void AliAnalysisTaskEmcalJetEnergyFlow::FillEFHistograms(){
           DetjetCont1 = dynamic_cast<AliJetContainer*>(fJetCollArray[i]);
           DetjetCont1->SetJetEtaLimits(-max_eta,max_eta);
           for(auto jet:DetjetCont1->accepted()) DetLowRJetsList.Add(jet);
-  
+          std::cout<<"If you see this when you run over MC then you're in trouble- point A"<<std::endl;
           // Casting the higher R half of the pair to a container and getting the accepted jets to a list
           DetjetCont2 = dynamic_cast<AliJetContainer*>(fJetCollArray[i+1]);
           DetjetCont2->SetJetEtaLimits(-max_eta,max_eta);
           for(auto jet:DetjetCont2->accepted()) DetHighRJetsList.Add(jet);
+          std::cout<<"If you see this when you run over MC then you're in trouble- point B"<<std::endl;
         // For the case of data, match and calculate the energy flow only on the detector level jets for this R pair iteration
           if(DetLowRJetsList.GetEntries()==0||DetHighRJetsList.GetEntries()==0) continue;
           iLowRIndex_det.Set(DetHighRJetsList.GetEntries()); iHighRIndex_det.Set(DetLowRJetsList.GetEntries());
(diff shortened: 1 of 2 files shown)


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.cxx
```diff
@@ -398,7 +400,7 @@ void AliAnalysisTaskEmcalJetEnergyFlow::AllocateEnergyflowHistograms(){
 
 
   Int_t Pair_number = 3;
- if (isMCprod)Int_t Pair_number = fJetCollArray.GetEntries()/2-1;
+ if (IsMCprod)Int_t Pair_number = (fJetCollArray.GetEntries()/2)-1;
  else Int_t Pair_number = fJetCollArray.GetEntries()-1;
```
raymondEhlers, line 404: This won't do what you want. Your redefinition shadows the initial definition. As a result, `Pair_number` will always be 3.

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.h
```diff
@@ -67,7 +67,7 @@ class AliAnalysisTaskEmcalJetEnergyFlow: public AliAnalysisTaskEmcalJet {
 	void                    DoClusterLoop()                         ; ///<May remove later	
 	void                    DoCellLoop()                            ; ///<May remove later
 
-        Bool_t                  isMCprod                                ;///<Flag for MC productions
+        Bool_t                  IsMCprod                                ;///<Flag for MC productions
```
raymondEhlers, line 70: If you change a streaming member name, you have to update the LinkDef

raymondEhlers (changes requested):

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.h
ChrisPliatskas: Even if it's only a name change?Okay I didn't know that, I will update it.

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.cxx
ChrisPliatskas: Okay, I will fix that. Thanks for the comment.

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.h
raymondEhlers: I might matter - I'm not 100% sure (since it's the same type, it could very well be fine). But incrementing is cheap, and will for sure avoid any issue :-)

ChrisPliatskas: Yes that's true, thank you for the feedback. I will keep it in mind for future updates :)

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.cxx
raymondEhlers: You didn't fix the issue. See this example:

```c++
int i = 3;
if (true) int i = 10;
// This will always print 3 because you redefined `i`
std::cout << "i: " << i << "\n";
```

You can fix it by removing the `int` from the second line. So it would be: `if (true) i = 10;`

ChrisPliatskas: I see, that was a very silly mistake. Now it should be okay, sorry for the email/notification spam

raymondEhlers:
+1

PR #17954, opened 2021-06-18
PWGPP/EMCAL: add trigger selection Pi0Calib

EwaGlimos:


gconesab:
+test

mfasDa:
Not sure this is of relevance but as far as I recall we also took data with 0EMC/0DMC, those have only EMCAL in readout as EMCAL was in PHYSICS2 and no tracking/trigger detectors. The physics selection should identify them as kEMC7. Not sure you want to add them.

gconesab:
> Not sure this is of relevance but as far as I recall we also took data with 0EMC/0DMC, those have only EMCAL in readout as EMCAL was in PHYSICS2 and no tracking/trigger detectors. The physics selection should identify them as kEMC7. Not sure you want to add them.

Well, the calibration data was always taken with kEMC7, so I would think it should be considered

EwaGlimos:
> Not sure this is of relevance but as far as I recall we also took data with 0EMC/0DMC, those have only EMCAL in readout as EMCAL was in PHYSICS2 and no tracking/trigger detectors. The physics selection should identify them as kEMC7. Not sure you want to add them.

I had some problems with getting a proper response from the Physics Selection. I believe the workaround I implemented should include also those events for the EMC7

FriederikeBock:
+1

PR #17965, opened 2021-06-21
PWGGA/GammaConv: Fixed the AddTask of GammaCaloMergedML to work with JJ-MC-Events

jhonerma:
Updated the Addtask of my analysis to get output for non-generic MC-events.


fjonasALICE:
+1

PR #17987, opened 2021-06-23
Update of the streamed data members

ChrisPliatskas:
This commit includes: a) Initialization of the MC flag in the class constructors, b) a clean up of the streamed / transient data members c) a reset of the jet container pointes at each jet pair iteration and d) a couple of debug print statements


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.cxx
```diff
@@ -109,7 +109,7 @@
   Bool_t AliAnalysisTaskEmcalJetEnergyFlow::Run()
   {
           if(fOutput->IsUseScaling()) Printf("Scaling is enabled! \n");
-  
+          if(IsMCprod) Printf("MC production");
```
raymondEhlers, line 112: Since this can't be turned off (as in the case of AliDebugStream, for example) and it runs every event, this will inflate the logs on the LEGO trains. Please remove this line or use something like AliDebugStream

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.h
```diff
@@ -50,32 +50,32 @@ class AliAnalysisTaskEmcalJetEnergyFlow: public AliAnalysisTaskEmcalJet {
...
-	void                    DoClusterLoop()                         ; ///<May remove later	
-	void                    DoCellLoop()                            ; ///<May remove later
+	void                    DoClusterLoop()                         ; // /<May remove later	
+	void                    DoCellLoop()                            ; // /<May remove later
 
-        Bool_t                  IsMCprod                                ;///<Flag for MC productions
-	THistManager            fHistManager                            ;///< Histogram manager
-//	TList*			fOutput					;///!<! Output list
+        Bool_t                  IsMCprod                                ;// /<Flag for MC productions
+	THistManager            fHistManager                            ;// /!<! Histogram manager
```
raymondEhlers, line 71: I don't know if the extra space will work with the parser. I think it needs to be `///<` for streaming and `//!<!`. Same for below. If you've already check this and confirmed that the streamers work as expected, then you can disregard it

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.cxx
ChrisPliatskas: I will move this to ExecOnce(), so this will avoid filling the log file and it makes sense since I don't expect the flag's value to change per event. Similarly I will also move the preceding line to ExecOnce(). Thank you for checking!

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.h
```diff
@@ -56,20 +56,20 @@ class AliAnalysisTaskEmcalJetEnergyFlow: public AliAnalysisTaskEmcalJet {
...
-	void                    DoClusterLoop()                         ; // /<May remove later	
-	void                    DoCellLoop()                            ; // /<May remove later
+	void                    DoClusterLoop()                         ; ///<May remove later	
+	void                    DoCellLoop()                            ; ///<May remove later
 
-        Bool_t                  IsMCprod                                ;// /<Flag for MC productions
-	THistManager            fHistManager                            ;// /!<! Histogram manager
-//	TList*			fOutput					;// /!<! Output list
+        Bool_t                  IsMCprod                                ;///<Flag for MC productions
+	THistManager            fHistManager                            ;///!<! Histogram manager
```
raymondEhlers, line 71: This is close, but still won't quite do what you want. You have `///!<!`, but it should be `//!<!` (only two `/`) (Currently, I suspect this will interpret fHistManager as streaming due to the extra `/`, but don't recall precisely). I know I seem perhaps annoyingly specific on this, but streamers can cause a lot of problems, so it can save you a lot of trouble to have them correct from the start

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.cxx
raymondEhlers: Very good, thanks!

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.h
ChrisPliatskas: Oh thanks, so my situation at the moment is that the train test fails and I was trying to see what causes the problem. I "managed" to replicate the error with my local AliPhysics version so I should probably close this PR until I fix that error. Then you won't be spammed by notifcations right? Thank you for your help so far Raymond! :)

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.cxx
mfasDa: AliAnalysisTaskEmcal provides UserExecOnce as framework function which user tasks can use to implement custom initialisations, please use that instead and never overwrite framework functions like ExecOnce directly.

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetEnergyFlow.h
raymondEhlers: Sure, that sounds like a good plan. I'll look for your PR when it's ready. You're welcome!

PR #17999, opened 2021-06-24
Changes

KotliarovAr:
Particle level jets have constituents pT > 1 MeV/c
Add histograms for QA: if detector level jet is withing 0.9 - R and particle level one is outside 0.9 - R -> fill impurity histograms

Filip approved the code


raymondEhlers:
+1

PR #18000, opened 2021-06-24
 PWGGA/GammaConv pcm gamma E smearing

AMechler:
changes for MC pcm gamma smearing
enable bremsstrahlung without gaus smearing


jokonig:
+1

PR #18011, opened 2021-06-25
PWGGA/GammaConv: OmegaToPiZeroGamma - Added Cluster Cut Variations

mhemmer-cern:
- Added cluster cut variations (train configs X9X1 till X9X6) for:
1. NonLin
2. timing
3. track matching
4. exotic cluster
5. Ncell
6. M02 max

- changed the ordering of some train configs, so they are grouped together in a more logical sense
- Increased Pi0 pT binning from 20 to 30 GeV/c as the max value


jokonig:
+1

PR #18020, opened 2021-06-29
Fix for overwriting nSigmaTPC_bach_pi. Set to nSigmaTOF_bach_pi.

AnnalenaKa:
Fix name: nSigmaTPC_bach_pi was overwritten. It is meant to be nSigmaTOF_bach_pi.


cterrevo:
+1

PR #18023, opened 2021-06-29
Change condition in conversion task for avoid division by 0.

mgheata:
This should fix FPEs observed in Grid conversion jobs.


jgrosseo:
+1

PR #18037, opened 2021-06-30
PWGJE/EMCALJetTasks/UserTasks multiplicity bug fix for AliAnalysisTaskEmcalJetEnergyFlow

ChrisPliatskas:
This update contains three changes:
1. Fixed the bug where the multiplicity histogram at gen level was filled with det level jets (this caused the LEGO train test run to crash).  
2. Removed the class inheritance check from the track analysis functions since it produced errors for the mcparticles container.
3.  Changed the order that particle containers are added in the task. Now the particle container defined by the AddTask args is added first and if IsMC is true an additional particle container is added. This information is useful when adding the jet containers to the macro in deciding what particle container each jet finder should use


raymondEhlers:
+1

PR #18040, opened 2021-06-30
correct CF bin limits for the RT case and add the setter

gluparel:
correct CF bin limits for the RT case and add the setter for the aver… age multiplicity in the trnasverse region


adubla:
+test

adubla:
+1

adubla:
+1

PR #18044, opened 2021-06-30
optimising for unfolding

JColburn-CERN:
I have altered the binning of the jet charge histograms and added a response matrix for each flavour to optimise for unfolding.


raymondEhlers:
+1

PR #18064, opened 2021-07-03
PWGGA/GammaConv/macros: OmegaToPiZeroGamma AddTask - Fixed Train Configs

mhemmer-cern:
- There can be no train config with both mixed event and rotation method in one. Fixed this.


jokonig:
+1

PR #18069, opened 2021-07-05
Updated particle-composition correction task

phuhnGithub:
included Lambda in first train iteration for MC simulation
updated default input paths for pp and PbPb


mario-krueger:
+1

PR #18087, opened 2021-07-06
LMeeCocktail update

hmurakam:
1. Cocktail input files ROOT5 and ROOT6 
- Following particles are parametrized:  neutral pion (111), charged pion (211), eta (221), rho (113), omega (223), phi (333), J/psi (443), charged kaon (321), and proton (2212). 
- Note1: Neutral pion is charged pion divided by pi+-/pi0
- Note2:J/psi is taken from preliminary results
- etaprime is parametrized via mT scaling (scaling factor = 0.50 )
- Spectra are shifted +-10% with respect to 13TeV_default case and fitted -> 13TeV_[low,high]

2. Added J/psi to the cocktail
- one can include J/psi in addition to LFs.
- in case motherSelect need to be set "127" (=1+2+4+8+16+32+64)  pi0, eta, etaprime, rho, omega, phi, Jpsi.


rbailhac:
+1

PR #18098, opened 2021-07-07
Adding in task that uses Embedding to fill Response with matched Jets and Constituents

chughes90:
Added Files:

AddTaskEmcalJetHPerformanceConstits.C
AliAnalysisTaskEmcalJetHPerformanceConstits.cxx
AliAnalysisTaskEmcalJetHPerformanceConstits.h

Description:

Builds on the embedding done in AliAnalysisTaskEmcalJetHPerformance by adding in lines to loop to through matched jets and fill a response with their matched constituent pT. Intended to use for unfolding.


raymondEhlers:
Hi @chughes90 , nice to see your developments here. Two comments:

- You have two copies of your new classes:
  1. in ` PWGJE/EMCALJetTasks/UserTasks/*`, where they should be located.
  2. in ` PWGJE/EMCALJetTasks/*`, where they should not be.
  Please remove the files in 2
- You can update the author info :-)

chughes90:
> Hi @chughes90 , nice to see your developments here. Two comments:
> 
>     * You have two copies of your new classes:
>       
>       1. in ` PWGJE/EMCALJetTasks/UserTasks/*`, where they should be located.
>       2. in ` PWGJE/EMCALJetTasks/*`, where they should not be.
>          Please remove the files in 2
> 
>     * You can update the author info :-)

Hi Raymond,

Thanks for the reply and directions - I still am not sure what I am doing. I got rid of the .cxx and .h in PWGJE/EMCALJetTasks/ as you suggested.

I may have made a mistake which is I did a git pull in AliPhysics before applying the above corrections. I also don't know if I should have opened a separate pull request. Please let me know if I need to correct this.

raymondEhlers:
+1

Ideally, you should squash the commits, but it's fine for now

PR #18115, opened 2021-07-12
PWGDQ:LMee: changed variable definition in VarMangager.h

jjungIKF:
included the charge sign in the definition to exclude the charge assymmetry between ULS and LS


rbailhac:
+1

PR #18121, opened 2021-07-12
Option for pT cut on generated MC

sustripathy:
The commit includes option for pT cut on generated MC. This is necessary for f0 analysis in pp collisions at 5.02 TeV.


mario-krueger:
+1

PR #18124, opened 2021-07-13
PWGGA/GammaConv/macros OmegaToPiZeroGamma AddTask changed timing cut …

mhemmer-cern:
…variation

- The -12.5 to 13 ns timing cut for the timing cut variation now uses the timing cut efficiency


fjonasALICE:
+1

PR #18132, opened 2021-07-14
Increasing statistics: More open cuts for systematics + timing condition relaxed

gmantzar:
Setting new open Cuts for systematic variations and releasing timing information for the Pion Track of the Xi candidates


sartozza:
+1

PR #18143, opened 2021-07-16
Multi-charm added to MC predictions task

ddobrigk:


ddobrigk:
@ginnocen, as discussed yesterday, this task should be 100% fine for doing yield predictions using LEGO trains. Note there is also some 2pc functionality but untested. I will ask for tests using the LEGO trains soon. Note that we could test regular Monash, mode 2 pp and even angantyr, in principle (this task was originally designed for PbPb and does multiplicity dependence in various ways)

Heads up also for @arossi81 @marcovanleeuwen @qgp @urasantonio, as these tests may be of interest (@arossi81 has many done already :-D)

arossi81:
Hi @ddobrigk ,
I had a quick look at the code (very quick). I just have a doubt: it seems to me that you do not remove the feed-down from beauty for charm. 
Cheers,
Andrea

ddobrigk:
Hi @arossi81, 

That is correct: there is no explicit removal of feeddown for now, as this would break a bit the symmetry of the code, but we could still implement this. Do you have something maybe already in AliPhysics that I could take a peek at, out of curiosity? 

Thank you! 
Cheers,
David

arossi81:
Hi @ddobrigk,
you can use the CheckOrigin methods in [PWGHF/vertexingHF/AliVertexingHFUtils.cxx](https://github.com/alisw/AliPhysics/blob/master/PWGHF/vertexingHF/AliVertexingHFUtils.cxx)

In my pythia simulations (outside AliRoot, but what is done is exactly the same) I use the piece of code below.
Cheers,
Andrea

int indmum=pythia.event[i].mother1();
while(indmum){
	  if(TMath::Abs(pythia.event[indmum].id())==4)break;// May not be needed
	  if((TMath::Abs(pythia.event[indmum].id())%1000)/100==5||(TMath::Abs(pythia.event[indmum].id())%10000)/1000==5){
	    isfromb=true;	    
	    break;
	  }
	  	  indmum=pythia.event[indmum].mother1();
	}

PR #18147, opened 2021-07-16
Add correlation histograms for trigger classes

mfasDa:
Correlation histograms are implemented for each
trigger cluster separately. Diagonal elements are
absolute event counters, off-diagonal elements
show the correlation with the trigger.

Mainly used to cross check downscale factors.


mfasDa:
+1

PR #18150, opened 2021-07-16
Add prompt HF

ddobrigk:
@arossi81 this implementation includes another 13 particles with a promptness check using 

`AliVertexingHFUtils::CheckOrigin(lMCevent, lMCPart, kTRUE)==0`

Not sure if I understood the usage of the method so well, so please jump up in case I did it wrong :-D Thanks a lot!!!

### DIFF
--- a/PWGLF/STRANGENESS/CMakeLists.txt
+++ b/PWGLF/STRANGENESS/CMakeLists.txt
@@ -28,6 +28,7 @@ include_directories(${ROOT_INCLUDE_DIRS}
                     ${AliPhysics_SOURCE_DIR}/OADB/COMMON/MULTIPLICITY
                     ${AliPhysics_SOURCE_DIR}/PWG/Tools
                     ${AliPhysics_SOURCE_DIR}/PWGCF/Correlations/Base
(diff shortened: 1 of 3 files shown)


ddobrigk:
@arossi81 my first mistake I already caught: `AliVertexingHFUtils::CheckOrigin(lMCevent, lMCPart, kTRUE)==0`  should have been `AliVertexingHFUtils::CheckOrigin(lMCevent, lMCPart, kTRUE)!=0` because I _skip_ particles that are _not prompt_. Please shout at me if I still have more mistakes :-D

arossi81:
Hi @ddobrigk,
you need to check whether you get 4 or 5, not !=0. The method returns an integer: 5 if the particle comes from beauty, 0 if you pass "true" to the argument "searchUpToQuark" and a quark charm or beauty is not found, 4 otherwise. So, I suggest to set searchUpToQuark==kTRUE and check whether you get 4 (-> prompt) or 5 (->from B). However, this may not be ok for all generators, namely when the decay history is not stored down to the quarks (e.g. for HIJING if you want to use the charm and beauty particles produced by HIJING, if HIJING just gives you the background event that's ok).  

Cheers,
Andrea

PR #18152, opened 2021-07-18
improvement on sum pT and multiplicity correlation

krschmit-cern:
- Added direct calculation of <sump_T> and <N_acc> in regions
- Added MC Truth information for correlation observables


bdoenigu:
+1

PR #18171, opened 2021-07-20
feature-ncellcut

mfasDa:
- Cut by default 0 (= no cut)
- In case the 1-cell clusters are included the
  cut is bypassed in case a 1-cell cluster is found
  in the same way as for the exoticity cut and the
  shower shape cut
- Be more explicit with parameter names in the
  interface for better readability


mfasDa:
+test

mfasDa:
@jokonig Please review!

jokonig:
Hi, I think it all looks good!

mfasDa:
+1

@jokonig Thanks for the review! Merging ...

PR #18190, opened 2021-07-23
Updated 3Body Trigger

lauraser:
Additional QA histograms included, third limit for sampling added, track cleaning included in trigger


sartozza:
+1

PR #18194, opened 2021-07-24
Fix rotation on resonances

romainschotter:
Fix rotation to compute combinatorial background on resonances 
- rotation was done but not save


bdoenigu:
+1

PR #18196, opened 2021-07-25
Missing ; in LambdaPhi AddTask but compilation not fail

sartozza:
I apologize for the silly mistake but the compilation of AliPhysics went fine and I spotted it today.


sartozza:
+1

PR #18200, opened 2021-07-26
Add flag to remove unused variables, add Combined Bayesian PID with TPC only

AnnalenaKa:
A flagg is added to the task to switch off the storing of variables in the tree that are not used in the current analyses. Furthermore a new variable is added which is the combined bayesian PID using only the TPC and no TOF information.


fgrosa (commented):
Hi @AnnalenaKa thanks for the implementation! There is a small detail, when fixed I will approve it. Cheers, Fabrizio

PWGHF/vertexingHF/AliAnalysisTaskSELc2pKs0fromKFP.h
```diff
@@ -147,6 +148,7 @@ class AliAnalysisTaskSELc2pKs0fromKFP : public AliAnalysisTaskSE
         Bool_t                  fIsMC; ///< Flag of MC analysis
         Bool_t                  fUseWeights; ///< Flag to use pT weight functions
         Bool_t                  fKeepOnlyMCSignal; ///< flag to keep only signal candidates
+        Bool_t                  fKeepAllVariables; ///<flag to keep all possible variables that were removed to reduce the tree size
```
fgrosa, line 151: Every time you add a data member you should update the `ClassDef` (in this case from `7` to `8`)

AnnalenaKa: Thanks a lot for spotting this. I did not see this.

fgrosa:
+1

PR #18215, opened 2021-07-28
Configure macro for generate events in p-Pb 8.16 TeV

dmallick2021:
Hi, 
I have created two configure macro for event generation using on the fly for p-Pb collisions at 8.16 TeV


dmallick2021:
Need to update the macro

PR #18236, opened 2021-08-02
add histogrammed response, rho shift variations and tree rejection factor

jaimenorman:
- add THnSparse to allow for more entries in response
- 3d histogram of d phi vs pt vs rho shift for systematic studies
- allow to reject a fraction of entries for the response tree (will be useful to fill for QA)


raymondEhlers:
+1

jaimenorman:
Hi @raymondEhlers I see you have added a +1 but the build fails on an unrelated class (AliMUONPainterPlotSelector), is this something that should be fixed? I saw that other pull requests are also failing

raymondEhlers:
There's nothing that we can do - we just have to wait for a fix in AliRoot from offline

pzhristov:
The problem with the build will be fixed by a new AliRoot tag that is under test.

PR #18241, opened 2021-08-03
option to scale DCA for background studies

jaimenorman:
- add option to scale DCA by a constant factor (the track properties can in addition be smeared by the usual method - if wanting to just scale the DCA, the same 'improver' file should be given as both arguments to the add task)
- add option to only smear a fraction of tracks in case the DCA is being scaled

this functionality will be used for background studies in cases of low background for the ITS3


fgrosa:
+1

PR #18242, opened 2021-08-03
Fix double counting of mixed triggers

mfasDa:
Extra loop over the mixed triggers (OR of EMCLA or
DCAL trigger) became obsolate once the downscale
factors were added to the standard cache and
therefore treated as normal trigger class, however the
extra loop was not removed, leading to a double
counting of the trigger class when filling the
luminosity counting. Extra loop is removed in order to
prevent the double counting - mixed triggers will
always be treated as standard trigger class.

### DIFF
--- a/PWG/EMCAL/EMCALtasks/AliAnalysisTaskEmcalTriggerNormalization.cxx
+++ b/PWG/EMCAL/EMCALtasks/AliAnalysisTaskEmcalTriggerNormalization.cxx
@@ -174,8 +174,6 @@ Bool_t AliAnalysisTaskEmcalTriggerNormalization::Run(){
   auto normhist = static_cast<TH2 *>(fHistos->FindObject("hTriggerNorm")),
        luminosityHist = static_cast<TH2 *>(fHistos->FindObject("hTriggerLuminosity")); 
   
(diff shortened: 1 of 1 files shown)


PWG/EMCAL/EMCALtasks/AliAnalysisTaskEmcalTriggerNormalization.cxx
```diff
@@ -206,29 +205,6 @@ Bool_t AliAnalysisTaskEmcalTriggerNormalization::Run(){
...
-            } else {
-              // Only DCAL triggers would enabled
-              // this should never happen
-              downscaleTrigger = fCacheDownscaleFactors[dcaltriggerclass];
-            }
-            auto triggerbin = luminosityHist->GetXaxis()->FindBin(edtrigger.data());
-            auto triggerweight = downscaleTrigger * mbweight;
-            luminosityHist->Fill(luminosityHist->GetXaxis()->GetBinCenter(triggerbin), centralitypercentile, triggerweight);
-          }
-        }
```
mfasDa, line 231: Loop no longer needed because the luminosity histogram is alley filled for the mixed triggers in the previous loop over all cached triggers.

mfasDa:
+1

PR #18260, opened 2021-08-04
Remove the pointer declarations for the TFormula (remove memory leak)

romainschotter:


bdoenigu:
+1

romainschotter:
Hello @bdoenigu 
Do you have an idea of what happened ? It seems to me that the error is not coming from my changes...

bdoenigu:
It is very likely not your code! I have no idea, this happens since a few days - not sure who feels responsible. Last time we had this some servers needed to be restarted...

romainschotter:
In this case, what do you suggest ? Should I send a mail to the analysis task force to see if someone can arrange that ? Or should I just wait ?

PR #18275, opened 2021-08-09
Add ldl_XicPlus, ct_XicPlus and PA_XiToXicPlus, make a sequence of pion_0 and pion_1 from XicPlus decays according to pt ( pt(pion_0)<pt(pion_1) ); add ldl and ct for Xic0 and Omegac0

zhujh126:
Add ldl_XicPlus, ct_XicPlus and PA_XiToXicPlus, make a sequence of pion_0 and pion_1 from XicPlus decays according to pt ( pt(pion_0)<pt(pion_1) )


xinyepeng:
+1

PR #18296, opened 2021-08-12
Special Tag for KFParticle test

amaringarcia:


amaringarcia:
During June we did a commit to move from  from AliKFParticle to KFParticle support within Gamma Conversions.
To do that we have redirected all references to AliKFParticle to a new 'complex', AliGAKFParticle within GammaConvBase, which behavior is decided at production time with the flag set(USEKFLIBRARY FALSE) in CMakeLists.txt of GammaConvBase. The fact that AliKFParticle was referenced from quite a number of source files within both GammaConv and GammaConvBase made quite large the impact of that PR.
 The  value of the USEKFLIBRARY flag  was FALSE, then the behavior of the AliGAKFParticle complex is just to redefine all references to AliGAKFParticle to AliKFParticle which, basically, leaves the scene exactly the same as before the PR.
Today , I am setting the (USEKFLIBRARY TRUE) to be able to test the behavior with KFParticle in the grid.
In this case, with the USEKFLIBRARY flag set to TRUE, the AliGAKFParticle complex introduces a wrapper class which transforms the AliKFParticle interface to the KFParticle interface in the points that they differ.

PR #18341, opened 2021-08-20
Add PID function kTPCpidTOFveto in AliRsnCutSetDaughterParticle

syano0822:
Add the new PID scheme function, kTPCpidTOFveto into AliRsnCutSetDaughterParticle class.
The PID function requires acceptable TPC sigma and veto TOF sigma.


mario-krueger:
+1

PR #18344, opened 2021-08-20
Add histograms assessing the charged and neutral components of the jet energy scale

mfasDa:
Part. - and det. level jet pt are weighted with its
corresponding NEF (1-NEF) in order to approximate
the neutral (charged) part of the  jet pt.


mfasDa:
+1

PR #18375, opened 2021-08-25
PWGGA/GammaConv+EMCal: EDC cuts for iso analysis & experimental track propagation

fjonasALICE:
- implemented AliEMCALRecoUtils::ExtrapolateTrackToEMCalSurfaceExperimental which is an experimental function to propagate tracks to the EMCal surface and not used anywhere so far. The function is not validated, however, I think the name of the function containing "Experimental" should make clear enough that this can not be used without validation. The function uses PropagateTrackToBxByBzExperimental for track propagation, which properly takes into account different materials and does not assume Silicon as default for the material passed. Furthermore, the critical energy for electrons is calculated, after which Bremsstrahlung is the dominant source of energy loss. Furthermore, the calculation of density fluctuation corrections has been imroved
- added new configs for iso analysis to include EDC
- fixed crash in clusterizer when using single particle gun ESDs


loizides:
+1

FriederikeBock:
+1

PR #18377, opened 2021-08-25
removing TPC signal from AliFemtoNanoAODReader

lgraczykCern:
remocing TPC signal from NanoAOD reader (no information in NanoAOD)


lgraczykCern:
+1

PR #18384, opened 2021-08-26
cmake:: fix doxygen finding on f34

adriansev:
on cmake 3.20.5 there are problems if doxygen is already found. Just invalidate the already detected doxygen executable.
solution from https://gitlab.kitware.com/cmake/cmake/-/issues/18708


adriansev:
@ktf Hi! can you please take a quick look and review? Thanks a lot!

ktf:
+1

PR #18401, opened 2021-08-30
Bug fix + new QA plots

VasyaDi:


jezwilkinson:
Hi @VasyaDi , 
One minor comment - As the commit is adding new data members to the task, the "ClassDef" number at the bottom of AliAnalysisTaskSEDvsRT.h should be incremented (3 --> 4). Apart from that the commit looks fine.

VasyaDi:
Hi @jezwilkinson ,
I've updated the number. Can you please elaborate on the reason for changing the number? Is it needed for keeping track of the task iterations?

jezwilkinson:
> Hi @jezwilkinson ,
> I've updated the number. Can you please elaborate on the reason for changing the number? Is it needed for keeping track of the task iterations?

The ClassDef is for ROOT's data streamer, basically to avoid compatibility issues when different AliPhysics tags encounter a version of the task with different data members present. If a commit only changes the CXX or adds new functions then we don't need to increment it, but generally if there are new members added into the .h then we normally do increment.

jezwilkinson:
+1

PR #18416, opened 2021-08-31
[WIP] Adapt EMCAL trigger table to STU format in run3

mfasDa:
- Only 2 columns (FastOR ID and ADC)
- All FastOR ADCs only for fraction of events (settable)
- FastOR ADCs from the 3 leading patches for the other events

### DIFF
--- a/RUN3/AliAnalysisTaskAO2Dconverter.cxx
+++ b/RUN3/AliAnalysisTaskAO2Dconverter.cxx
@@ -18,6 +18,7 @@
  * Convert Run 2 ESDs to Run 3 AODs (AO2D.root).
  */
 
+#include <algorithm>
 #include <TFile.h>
 #include <TDirectory.h>
 #include <TChain.h>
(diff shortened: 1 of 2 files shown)


mfasDa:
Work in progress - don't merge yet!

mfasDa:
Test could be launched.

jgrosseo (changes requested):
(copied from mail exchange) Your PR makes a new table for the emcal trigger, but we agreed to have both in the same table, didn't we? Wouldn't it be better you adjust the calotrigger table as suggested by Dimitri and keep the filling there?

jgrosseo:
Needs merging with phos changes first

mfasDa:
There was no reason to close this pull request!

Indeed as I mentioned in the private discussion the developments from PHOS and EMCAL are parallel and therefore conflicting, however I would have rebased the branch and update the PR once the PHOS changes are merged. Also therefore the PR was marked as [WIP].

Please for the future keep the pull requests open!

PR #18428, opened 2021-09-02
Add PHOS trigger filtering

peressounko:


RUN3/AliAnalysisTaskAO2Dconverter.cxx
```diff
@@ -253,6 +260,24 @@ void AliAnalysisTaskAO2Dconverter::NotifyRun(){
   if (!triggerContainer) AliFatal("Cannot fetch OADB container for trigger analysis");
   AliOADBTriggerAnalysis* oadbTriggerAnalysis = (AliOADBTriggerAnalysis*) triggerContainer->GetObject(fCurrentRunNumber, "Default");
   fTriggerAnalysis.SetParameters(oadbTriggerAnalysis);
+  //read PHOS trigger bad map
+  AliOADBContainer phosBadmapContainer(Form("phosTriggerBadMap"));
+  phosBadmapContainer.InitFromFile(Form("%s/PHOS/PHOSTrigBadMaps.root", AliAnalysisManager::GetOADBPath()),              
+                                   "phosTriggerBadMap");
+  TObjArray *maps = (TObjArray*)phosBadmapContainer.GetObject(fCurrentRunNumber,"phosTriggerBadMap");
+  if(!maps){
+    AliError(Form("Can not read PHOS Trigger Bad map for run %d. \n",fCurrentRunNumber)) ;
```
jgrosseo, line 269: fPHOSBadMap will be used below, so this has to be AliFatal, otherwise it will crash anyway later.

However, I wonder will we have such a map also for all MCs (most Run 2 MCs will be converted)? 
If not, the usage of the PHOS bad trigger map should be enabled by a flag in the task.

RUN3/AliAnalysisTaskAO2Dconverter.cxx
```diff
@@ -374,12 +399,41 @@ void AliAnalysisTaskAO2Dconverter::UserExec(Option_t *)
     ::Fatal("AliAnalysisTaskAO2Dconverter::UserExec", "Something is wrong with the event handler");
   }
 
+  // populate meta data only once
+  if (fMetaData.GetEntries() == 0) {
```
jgrosseo, line 403: I think you need to simply rebase for your PR because these metadata changes are not from you but from something I did long ago.

jgrosseo: This explains the conflict...

peressounko:
+1

jgrosseo (changes requested):
Thanks for updating the PR!
Before we merge, please:
- Increase the ClassDef
- In `eventextra.fNentries[kCaloTrigger] = nphostriggers_filled;` you need to change it to `+=`. There should not be a second branch because branch here corresponds to one table.
Thanks!

peressounko:
> * Increase the ClassDef
done
> * In `eventextra.fNentries[kCaloTrigger] = nphostriggers_filled;` you need to change it to `+=`. There should not be a second branch because branch here corresponds to one table.
I already increased ncalotriggers_filled in l 2088. Do I miss something here?
      Best regards, Dmitri

jgrosseo:
Yes, it looks OK now.

I think you can remove `Int_t nphostriggers_filled = 0;`. It is not used anymore...

peressounko:
Removed Int_t nphostriggers_filled = 0;

jgrosseo (approved):

PR #18432, opened 2021-09-02
Some methods to work without top. selections added

mfaggin:
- method to switch off the topological selections added
 - method to switch off als othe PID selection after filtering added
 - method to reject events without at least one identified proton, one identified kaon and one identified pion added, to spare CPU time


mfaggin:
@arossi81 here the updates we were thinking about. Please have a look at them.

arossi81:
Hi @mfaggin ,
for the rejection of the event in case no tracks is identified as a proton (or pi or K), you should do the OR rather than the AND.
Cheers,
Andrea

mfaggin:
Hi @arossi81,
thanks for spotting it. Now it should be ok.
Best regards.
Mattia

fgrosa:
+1

PR #18437, opened 2021-09-03
Fix AliAnalysisTaskAR

ariedel-cern:
Fix logic and setters for correlation cuts on multiplicity and centrality to be more flexible.


dongjokim:
+1

PR #18447, opened 2021-09-06
Correct track selections for leading track

gluparel:
Use same track selections for the leading track used in the analysis task and correct the eta acceptance for the leading track


cterrevo:
+1

PR #18450, opened 2021-09-07
Add glueball hunting task in PWGLF/RESONANCE

syano0822:
Add AddTaskGlueball.C and ConfigGlueball.C macros in PWGLF/RESONANCE/macros/mini. The macros look for glueball signal in KK pair.


JihyeSSong:
+1

PR #18469, opened 2021-09-10
MID words: bitmap and local boards

nicolovalle:
RUN3 converter modified. MIDBitMap and MIDBoards added to the tFwdTrack tree

### DIFF
--- a/RUN3/AliAnalysisTaskAO2Dconverter.cxx
+++ b/RUN3/AliAnalysisTaskAO2Dconverter.cxx
@@ -95,7 +95,6 @@ const TString AliAnalysisTaskAO2Dconverter::TreeName[kTrees] = {
   "O2fwdtrackcov",
   "O2calo",
   "O2calotrigger",
-  "O2muoncluster",
   "O2zdc",
   "O2fv0a",
   "O2fv0c",
@@ -124,7 +123,6 @@ const TString AliAnalysisTaskAO2Dconverter::TreeTitle[kTrees] = {
   "Forward tracks Covariances",
   "Calorimeter cells",
   "Calorimeter triggers",
-  "MUON clusters",
   "ZDC",
   "FV0A",
   "FV0C",
@@ -225,7 +223,6 @@ AliAnalysisTaskAO2Dconverter::AliAnalysisTaskAO2Dconverter(const char* name)
     calo(),
     calotrigger(),
     fwdtracks(),
-    mucls(),
     zdc(),
     fv0a(),
     fv0c(),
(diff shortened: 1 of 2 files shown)


RUN3/AliAnalysisTaskAO2Dconverter.cxx
```diff
@@ -2551,8 +2553,20 @@ AliAnalysisTaskAO2Dconverter::FwdTrackPars AliAnalysisTaskAO2Dconverter::MUONtoF
...
+  UInt_t midpattern = MUONTrack.GetHitsPatternInTrigChTrk();
+  UShort_t midbitmap = 0;
+  for (int icath=0; icath < 2; icath++){
+    for (int ich=0; ich < 4; ich++){
+      if (AliESDMuonTrack::IsChamberHit(midpattern,icath,ich)) midbitmap |= 1<<(ich+icath*4);
+    }
+  }
+  convertedTrack.fMIDBitMap = midbitmap;
+  UChar_t midboard = (AliESDMuonTrack::GetCrossedBoard(midpattern) & 0xFF) ;
+  convertedTrack.fMIDBoards = ((midboard << 24) | (midboard << 16) | (midboard  << 8) | midboard) & 0xFFFFFFFF;
```
dstocco, line 2568: Hi @nicolovalle ,
I'm not 100% sure about this line. You defined the midboard as a UChar_t, meaning that the compiler knows it has 8 bytes. Now, if you do midboard<<24, I think that the compiler still thinks that the result of the operation is a UChar_t...and therefore it cuts whatever it finds after bit 8. This means that you will always get 0.
I think it would be better to do:
`UInt_t middboard = static_cast<UInt_t>(AliESDMuonTrack::GetCrossedBoard(midpattern));`
(notice that 0xFF is done already by GetCrossedBoard: no need to do it again).
And then:
`convertedTrack.fMIDBoards = ((midboard << 24) | (midboard << 16) | (midboard << 8) | midboard);`
(notice again that 0xFFFFFFFF should not be necessary)

RUN3/AliAnalysisTaskAO2Dconverter.cxx
```diff
@@ -2713,7 +2727,18 @@ AliAnalysisTaskAO2Dconverter::FwdTrackPars AliAnalysisTaskAO2Dconverter::MUONtoF
...
+  //MCH and MID words
   convertedTrack.fMCHBitMap = MUONTrack.GetMUONClusterMap();
+  UInt_t midpattern = MUONTrack.GetMUONTrigHitsMapTrk();
+  UShort_t midbitmap = 0;
+  for (int icath=0; icath < 2; icath++){
+    for (int ich=0; ich < 4; ich++){
+      if (AliESDMuonTrack::IsChamberHit(midpattern,icath,ich)) midbitmap |= 1<<(ich+icath*4);
+    }
+  }
+  convertedTrack.fMIDBitMap = midbitmap;
```
dstocco, line 2739: Same comment as above

RUN3/AliAnalysisTaskAO2Dconverter.h
```diff
@@ -463,6 +463,12 @@ class AliAnalysisTaskAO2Dconverter : public AliAnalysisTaskSE
     Int_t fMatchMFTTrackID = -1;
     Int_t fMatchMCHTrackID = -1;
     UShort_t fMCHBitMap = 0u;
+    // MID bit map
+    // | non-bending plane (4bit) | bending plane (4bit) |
+    // i-th chamber can be tested with: fMIDBitMap & (1<<i)
+    UShort_t fMIDBitMap = 0u;
```
dstocco, line 469: Why do you need to add `u` after 0? It does not hurt, but it is redundant.

RUN3/AliAnalysisTaskAO2Dconverter.cxx
nicolovalle: Hi @dstocco , 
thanks for having checked this. Actually the code works (`MIDBoards` is filled properly).. maybe an automatic cast is done during the bit shift? 
In any case what you say is correct, so I'm going to make the change.

jgrosseo:
Waiting for removal of muon clusters in this PR (as discussed by mail)

jgrosseo:
+1

PR #18482, opened 2021-09-13
PWGGA/GammaConv: Updated Analysistask to be more memory efficient

jhonerma:
I changed branches in the trees to be more efficient with regard to memory by using variable length arrays. The clusters can be reconstructed from the information in the trees later, instead of writing a 2d-array containing the cluster into a tree. Also added additional branches to extract more information about clusters.

### DIFF
--- a/PWGGA/GammaConv/AliAnalysisTaskGammaCaloMergedML.cxx
+++ b/PWGGA/GammaConv/AliAnalysisTaskGammaCaloMergedML.cxx
@@ -933,8 +933,8 @@ void AliAnalysisTaskGammaCaloMergedML::UserCreateOutputObjects(){
     UChar_t MaxClusterCellN = 100;
     fMergedClusterTreeCluster                 = new Float_t[MaxClusterCellN];
     fMergedClusterTreeClusterModNum           = new UChar_t[MaxClusterCellN];
-    fMergedClusterTreeClusterCol             = new UChar_t[MaxClusterCellN];
-    fMergedClusterTreeClusterRow             = new UChar_t[MaxClusterCellN];
+    fMergedClusterTreeClusterCol              = new UChar_t[MaxClusterCellN];
+    fMergedClusterTreeClusterRow              = new UChar_t[MaxClusterCellN];
 
 
     fHistoMCHeaders                               = new TH1I*[fnCuts];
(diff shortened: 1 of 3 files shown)


PWGGA/GammaConv/AliAnalysisTaskGammaCaloMergedML.cxx
```diff
@@ -3993,26 +4017,3 @@ void AliAnalysisTaskGammaCaloMergedML::GetClusterReadout(AliVCluster* cluster, A
   //std::cout<<"\n";
   //}
 }
-
-void AliAnalysisTaskGammaCaloMergedML::ResetBuffer(){
```
fjonasALICE, line 3997: I can only see a fraction of the code, but since you removed the ResetBuffer function: do you ensure that the tree buffer is reset properly elsewhere now?

PWGGA/GammaConv/AliAnalysisTaskGammaCaloMergedML.h
```diff
@@ -327,26 +328,29 @@ class AliAnalysisTaskGammaCaloMergedML : public AliAnalysisTaskSE {
...
-    Float_t                 fMergedClusterTreeClusterM02;                                 //! cluster m02
-    Float_t                 fMergedClusterTreeClusterM20;                                 //! cluster m20
+    UChar_t                 fMergedClusterCellN;                  //! number of cells in cluster
+    Float_t*                fMergedClusterTreeCluster;		        //! energy of cluster cells
+    Float_t*                fMergedClusterTreeClusterTiming;      //! timing of cluster cells
+    UChar_t                 fMergedClusterTreeClusterType;			  //! cluster classification
+    Float_t                 fMergedClusterTreeClusterEnergy;			//! cluster energy
+    Float_t                 fMergedClusterTreeClusterPt; 				  //! cluster pt
+    UChar_t*                fMergedClusterTreeClusterModNum;			//! cells supermodule numbers
+    UChar_t*                fMergedClusterTreeClusterCol;				  //! coloumn numbers of cells
```
fjonasALICE, line 338: just a suggestion for readability: I think the naming scheme of your branches is rather confusing, i.e. it is not clear which branches store cell infos and which branches store cluster infos. E.g. fMergedClusterTreeClusterCol suggests you are talking about a cluster property. fMergedClusterTreeCluster does not make clear at all what property is stored (apparently cell energies)

fjonasALICE (approved):
since the code is not used by anyone else i approve as it is. You may consider for the future to implement my changes for clarity. 

Also you may consider switching from arrays to std:vector to properly deal with variable length in the future

fjonasALICE:
+1

PR #18492, opened 2021-09-14
Update AliAnalysisTaskAR

ariedel-cern:
Updates to AliAnalysisTask.{h,cxx} in PWG/FLOW.
- add booleans to turn cuts on and off (so far they were always on)
- rework setters for correlation cuts
- add histograms holding values of all cuts


dongjokim:
+1

PR #18499, opened 2021-09-14
Femto L-d: change nSigma ranges for deuteron PID

stheckel:
o 2 subwagons: nSigma = 2.5 (0) and 2.0 (1)
o Both with deuteron pT range: 0.4 - 2.0


otonvd:
+1

PR #18506, opened 2021-09-15
Fix type of the field fLnAmplitude

mfasDa:
Type of fLnAmplitude was forgotten to be changed
to short in tree->Branch, only changed in the struct.


jgrosseo:
+1

PR #18508, opened 2021-09-15
PWG/EMCAL: optional LG/HG determination via ADC values in CF

fjonasALICE:
When using the option enableLGDetermination: true   in the cell energy component of the CF, the information if a cell is low gain or high gain will be determined via the ADC value of a cell. This option should not be used for data where the information is already stored in the cells itself. 
It is meant to be used with run1 data where the LG/HG info is not stored yet in the cell information. Accuracy of this method was evaluated in local tests using run2 data and will be further tested with full statistics on the grid


gconesab:
It looks good to me the changes in RecoUtils. I let others final approval

raymondEhlers:
LGTM in the CorrFW

raymondEhlers:
+1

PR #18510, opened 2021-09-15
Corrections related with memory leak

PaolaVT:
Two lines were added to the program in order to avoid memory leak.


aortizve:
+test

aortizve:
+1

PR #18520, opened 2021-09-17
PWGGA/GammaConv/macros/ V0M and SPD multiplicity

eddoming:
New cuts for V0M and SPD slices in multiplicity for 5 TeV and 13TeV (Nominal and low magnetic field)


jokonig:
+1

PR #18538, opened 2021-09-20
mc weight initialised to 1

rosarioturrisi:


rosarioturrisi:
@fprino no need to rerun the train, just fixed a parameter for future ease of use

rosarioturrisi:
needed for the track matching analysis

mfaggin:
@fprino we realized that a further detail was missing, so we need your +1 and new trains

fprino:
+1

PR #18540, opened 2021-09-20
Deuteron coalescence simulation in pp collisions

alcaliva:
Simulation of deuteron formation by coalescence in pp collisions.


lramona:
+1

PR #18567, opened 2021-09-24
AO2D converter fix for using event cuts for both ESD/AOD.

mgheata:
There was a recent change for accepting events based on cuts on ESD event, not working for AOD. This fixes for the general ESD/AOD case.


mgheata:
@pzhristov and @jgrosseo can you have a look?

pzhristov:
+1

PR #18572, opened 2021-09-24
DeltaP and DeltaX calculation in deuteron rest frame, implemented mT-…

alcaliva:
…dependence of source radius, finer p0 bins


lramona:
+1

PR #18582, opened 2021-09-27
all cuts applied to MC tracks before filling DCAxy templates + symmet…

maciacco:
…ric nSigmaTPC cut for DCAxy distribution in data


bdoenigu:
+1

PR #18598, opened 2021-09-29
Adding luminsity calculation tool for normatask output

mfasDa:
Automatically calculating the integrated luminosty
for certain known collision systems from the standard
output of the normalization task.


mfasDa:
+1

mfasDa:
+1

PR #18602, opened 2021-09-30
Update AliAnalysisTaskAR

ariedel-cern:
- Fix mistakes in the compuation of multiparticle correlators (mostly the implementation using nested loops)
- Rework UserExec such that the use of real and/or monte carlo data is more seamless
- Expand GetPointers method to initalize more histograms and add more convenient Getters to extract them from a reiniatlized task


dongjokim:
+1

PR #18616, opened 2021-10-04
Fix for ROOT5

MartinSchmidtTuebingen:
Unfortunately my laptop does not compile ROOT5, but it should work


mfasDa:
+1

mfasDa:
Only the array with the array dimensions needs to be constexpr, because the array dimension is needed at compile time to initialise the size of the static arrays.

MartinSchmidtTuebingen:
I actually have no idea how it changed from TMath::Pi() to 3.14159265358979323846f, was it your master merge? Anyway, this is fine I guess. Thanks.

PR #18622, opened 2021-10-06
differential proton re-weighting

alcaliva:
Proton spectra from Pythia 8 are re-weighted to match the data in toward, transverse and away separately.


mario-krueger:
+1

PR #18634, opened 2021-10-08
updated code structure for highest trigger flag

gyronaut:
Added functionality to select events with different number of triggers, as well as additional flags to select only the highest trigger from an event for the correlation for cross checks.


gyronaut:
Hello,
Can someone please approve this pull request for testing?  This is an
update to only the two analysis files for my h-phi angular correlation
analysis.
Best,
Justin

PR #18636, opened 2021-10-08
Update AliAnalysisTaskAR

ariedel-cern:
- add setter method to set default values for configuration, binning and cuts
- fix smaller mistakes in the naming scheme


dongjokim:
+1

PR #18642, opened 2021-10-08
Add flexibility for cocktail histograms

hscheid:
The changes should not affect any other analysis. However, if one wants it should now be possible to change the number of bins and the min/max of the past pt and mass axis.


rbailhac:
+1

PR #18661, opened 2021-10-12
for HM event selection

Tao-Fang:


AIS8:
event selectiong from 0 to 0.1 the centrality

PR #18673, opened 2021-10-14
Add histograms for checking Omega (not from Omega_c^0 decay) efficiency

zhujh126:
Add histograms for checking Omega (not from Omega_c^0 decay) efficiency


fgrosa:
+1

PR #18691, opened 2021-10-17
fix "pass by value" issue, add "per event" scaling flag

gyronaut:
fixed issue with the MakeCorrelation function not passing parameters by reference, and added new flag to do per-event scaling


rihanphys:
+1

PR #18697, opened 2021-10-18
Protection for missing AD in ESD data.

mgheata:
This commit protects the code filling the O2fdd tree against `nullptr` returned by `fVEvent->GetADData()`


mgheata:
@jgrosseo could you have a look?

PR #18708, opened 2021-10-19
Add the flag function for multiplicity var 

zhangbiao-phy:
- Add function to set the flag for multiplicity var in ML tree
- fix a minor error for the parameter


fgrosa:
+1

PR #18717, opened 2021-10-20
Added p0 values for systematics

alcaliva:


mpuccio:
Be careful, the CODEX modifications sneaked in

PR #18721, opened 2021-10-20
Changing the jet analysis task so it does background subtraction

hahassan7:
The jet analysis has been modified so it can do background subtract of the jets. Also added 2 histograms for the background fluctuation Delta Pt for two methods, Random Cone and embedding.


mfasDa:
+1

PR #18723, opened 2021-10-20
Added features to PWGDQ/dielectronJet for PbPb analyses

FabioCanedo:
I added two features to the analysis in dielectronJet to allow PbPb analysis:

- UE background subtraction from jets;
- Centrality determination;

I also did some cleanup in the code making several functions shorter.


iarsene (approved):

iarsene:
+1

PR #18762, opened 2021-10-26
Create gen_eposlhc_ppb_5TeV_noqgp.sh

ahkhanm:
no QGP in EPOS LHC


maireiphc:
Hello Ahsan,

can you add a comment l.17 in your bash script to document the intent of the substitution (disable the 'fusion' option in EPOS-LHC parameter file).
Better to do be complete.

Cheers

PR #18776, opened 2021-10-26
Reduce verbosity from geometry initialization

mfasDa:
Check with GetInstance whether a geometry instance
has already been created for the current run, and only
if not call GetInstanceFromRunNumber to build the
geometry object.


mfasDa:
@jgrosseo Could you please review?

pzhristov:
+test

PR #18821, opened 2021-11-01
Minor fix of the histogram name of DeltaPt to include centrality

ChrisPliatskas:
Fix the missing centrality definition (without typos this time) for the hJetDeltaPt_R histogram


raymondEhlers:
+1

PR #18834, opened 2021-11-02
PWGGA: Change Omega Systematic Configs, Ch.Pi Mass Cut,  TPC Sigma Cut

jluehder:


PWGGA/GammaConv/AliPrimaryPionCuts.cxx
```diff
@@ -1511,6 +1551,11 @@ Bool_t AliPrimaryPionCuts::SetMassCut(Int_t massCut){
             fDoMassCut = kTRUE;
             fMassCut = 0.520;
             break;
+        case 22: //m overload mass cut for chi2 of vParticle
```
jokonig, line 1554: shouldnt this be case 23 as case 22 already exists? This way the old case 22 will never be reached

jokonig (changes requested):

jokonig:
+1

PR #18855, opened 2021-11-03
To debug memory leak

hanseopark:
These have problems of memory leak using clear function of vector array in c++.
It maybe solve change to use swap function that is origin to empty included memory.


hanseopark:
Could you check it?

jokonig:
Hi, is there a reason swap should be used instead of just clear()? 
Because clear() should not produce any memory leak. The elements are removed from the vector, however the vector does not loose its capacity as it does when using swap(). But since the vector is constantly filled again I dont see a reason to always free up the memory completely. But maybe I am missing something

PR #18858, opened 2021-11-04
Fix bug that braks compilation on macOS

lbariogl:
This bug is present only for macOS users. On Ubuntu it is fine. Maybe it would be better to include tests also for macOS, as for O2.


sartozza:
+1

PR #18869, opened 2021-11-04
Improved output handelling

JColburn-CERN:
Removed the Majority of Histograms and simplified the Trees. The idea behind this is that TTrees will be used primarily for analysis in future in order to help with performing offline Analysis. Additionally, This should reduce the size of the output and memory usage to help with train outputs.

	# On branch master
	# Your branch is up to date with 'upstream/master'.

### DIFF
--- a/PWGJE/EMCALJetTasks/AliAnalysisTaskJetCharge.cxx
+++ b/PWGJE/EMCALJetTasks/AliAnalysisTaskJetCharge.cxx
@@ -155,7 +155,9 @@ Bool_t AliAnalysisTaskJetCharge::FillHistograms()
     // Technical detail; fix possibly corrupted jet container ID
     JetCont->ResetCurrentID();
     // Jet is acceptable?
-    while((Jet1=JetCont->GetNextAcceptJet())) {
+    for (auto Jet1 : JetCont->accepted())
+    {
+
       if(!Jet1)
       {
 
--- a/PWGJE/EMCALJetTasks/AliAnalysisTaskJetCharge.cxx
+++ b/PWGJE/EMCALJetTasks/AliAnalysisTaskJetCharge.cxx
@@ -153,7 +153,6 @@ Bool_t AliAnalysisTaskJetCharge::FillHistograms()
 
   if(JetCont) {
     // Technical detail; fix possibly corrupted jet container ID
-    JetCont->ResetCurrentID();
     // Jet is acceptable?
     for (auto Jet1 : JetCont->accepted())
     {
--- a/PWGJE/EMCALJetTasks/AliAnalysisTaskJetCharge.cxx
+++ b/PWGJE/EMCALJetTasks/AliAnalysisTaskJetCharge.cxx
@@ -4,7 +4,10 @@
 // Basic analysis task template for analysis jets storing information in both tree
 // branches and histograms
 
+
 #include <TH1F.h>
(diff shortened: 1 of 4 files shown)


raymondEhlers (changes requested):
Hi Sammy, you need to go back and check this commit carefully. You've renamed the class in `AliAnalysisTaskJetCharge.{cxx,h}` to `AliAnalysisTaskJetChargeFlavourTemplate`, but you already define that class in `AliAnalysisTaskJetChargeFlavourTemplate.{cxx,h}`. At best, this is going to lead to unexpected behavior, if it even complies. Please check what you want to do and revise as appropriate

JColburn-CERN:
Hi Raymond, Thanks for pointing this out to me that would have led to some very annoying behaviour had it complied. The issue should be fixed now. Cheers

PWGJE/EMCALJetTasks/AliAnalysisTaskJetCharge.cxx
```diff
@@ -186,112 +134,111 @@ Bool_t AliAnalysisTaskJetCharge::FillHistograms()
...
+
+  //TClonesArray *trackArr = dynamic_cast<TClonesArray*>(InputEvent()->FindListObject("HybridTracks"));
+  Float_t JetPhi=0;
+  Float_t JetPt_ForThreshold=0;
+
   if(JetCont) {
     // Technical detail; fix possibly corrupted jet container ID
     JetCont->ResetCurrentID();
     // Jet is acceptable?
     while((Jet1=JetCont->GetNextAcceptJet())) {
```
raymondEhlers, line 158: I guess this was already here, but be aware this is basically a deprecated way of accessing the jets. Preferred is:

```c++
for (auto Jet1 : JetCont->accepted()) {
  // ...
}
```

PWGJE/EMCALJetTasks/AliAnalysisTaskJetChargeFlavourTemplates.h
```diff
@@ -48,141 +46,29 @@ class AliAnalysisTaskJetChargeFlavourTemplates : public AliAnalysisTaskEmcalJet
...
+  Float_t ParticlePt;             //
+  Float_t ParticlePhi;            //
+  Float_t ParticleEta;            //
+  Float_t ParticleJetCharge;      //
+  Float_t LeadingTrackPt;         //
+  Int_t PdgCode;                  //
+
+
+
   TChain                              *pChain;                                                     //
```
raymondEhlers, line 72: Need to bump the ClassDef here

raymondEhlers (changes requested):
You don't need to change the jet iterator, but you need to bump the classdef in your flavour task

JColburn-CERN:
Thanks for the help. That classdef dump always gets me.
Your correct that the way of looping through the jets was always there, it's not surprising it's deprecated since the original code I learnt it from was back in root 5. Updated it now to the new preferred method

PWGJE/EMCALJetTasks/AliAnalysisTaskJetCharge.cxx
raymondEhlers: You need to remove the `ResetCurrentID` part too. This can some issues in some causes, which is part of why it was deprecated

raymondEhlers:
+1

PR #18891, opened 2021-11-07
Fixing a problem with objects not found

hahassan7:
The analysis code was crushing because some objects were not found.


mfasDa:
+1

PR #18920, opened 2021-11-10
ChangeMultiV0toProfile

wenyaCern:
This version changes the filled multiplicity in each V0 channel from TH1D to TPrifile, calculating the averaged (previous is the total multiplicity for the run) multiplicity for the run of the channel to do the Gain Equalization. 
In fact, the gain equalization process was misunderstood before.


rihanphys:
+1

PR #18924, opened 2021-11-10
Update in spherocity calculation for AOD datasets

sustripathy:
This commit updates the spherocity calculation for AOD datasets. The track cuts for spherocity calculation are now similar for both AOD and ESD datasets.


bdoenigu:
+1

PR #18925, opened 2021-11-10
dowang pdtHe3 cf(k*) vs proton pt

dowangFDU:
in order to study proton pT dependent, create a new class to record it.


sartozza:
+1

PR #18926, opened 2021-11-10
Follow-up fix of the Energy flow task for the analysis of MC

ChrisPliatskas:
Fixed the title of hJetPtDeltaRDeltaPt to include centrality during the filling of the histogram for the generator level in the MC case


raymondEhlers:
+1

PR #18928, opened 2021-11-10
Spherocity analysis of charged kstar using AOD datasets

sdeb0101:
Added Features:
(1). Root6 compatible
(2). Compatible with AOD datasets


bdoenigu:
+1

PR #18948, opened 2021-11-12
Add v24 slope and diff v22 vs pt to CMWESE code

wenyaCern:
This code adds the process for extracting the slope between v24 Vs. Ach, as well as the differential v22 spectrum as QA.
Also adds information of LHC11h for the same work.


rihanphys:
+1

PR #18962, opened 2021-11-15
DQ:LMee: Updating RandomRejection Task: trying to determine random re…

jjungIKF:
…jection on pair level


jjungIKF:
Why was this not approved after the +1?

hmurakam:
+1

PR #18965, opened 2021-11-15
dowang add control function and delete cout

dowangFDU:
this function is to control where should be TPC+TOF & delete cout in AliFemtoCorrFctnpdtHe3.cxx


rihanphys:
+1

PR #18966, opened 2021-11-15
Change AliContainer to fOutput

wenyaCern:
Histograms in AliContainer cannot be merged even if a single run has filled them well.
Now I change the storage strategy from AliContainer to a common output list.


rihanphys:
+1

PR #18978, opened 2021-11-16
Apple to apple for two Omegac0 classes

chengtt0406:
1) Apple to apple for the two Omegac0 classes - - to check the compatibility of 2 results  
2) Add flag for fTree_Electron
3) Add variables in fTree_Omegac0 - - to check the KFParticle performance on the cascade


xinyepeng:
+1

ktf:
@TimoWilken could you check what is wrong with the aliphysics tests? Also the nightly cache seems to be still broken.

PR #18988, opened 2021-11-16
PWGHF: Fixes new Dstar tasks

lvermunt:
- Fixed a typo AliAnalysisTaskSECharmHadronMLSelector task (D0 -> Dstar)
- Fixed an error in case the tasks would run over PbPb with the PreSelect function enabled
- Fixed an issue that entered for the SetOwnPrimaryVtx(), RecalcOwnPrimaryVtx(), UnsetOwnPrimaryVtx(), and CleanOwnPrimaryVtx() calls. The same strategy as in the AliRDHFCutsDStartoKpipi.cxx class should be adopted, where these functions are called for the D0-daughter of the Dstar candidate

@fgrosa, your tasks so please have a look if you are ok with it :)


fgrosa:
+1

fgrosa:
Thanks @lvermunt!

PR #18989, opened 2021-11-16
General improvements for analysisTask and subsequent analysis

MartinSchmidtTuebingen:
MTF: Update add macros

MTF: Use correct indentation

MTF: Adding some helper tasks

MTF: Fix PID macros for ROOT6

MTF: Fix call of incrementEventCounter after pile-up check


raymondEhlers:
+1

PR #18995, opened 2021-11-17
Add unique event ID for both Data and MC

zhujh126:
Add unique event ID for both Data and MC

### DIFF
--- a/PWGHF/vertexingHF/AliAnalysisTaskSEXicPlusToXi2PifromKFP.cxx
+++ b/PWGHF/vertexingHF/AliAnalysisTaskSEXicPlusToXi2PifromKFP.cxx
@@ -77,6 +77,7 @@ AliAnalysisTaskSEXicPlusToXi2PifromKFP::AliAnalysisTaskSEXicPlusToXi2PifromKFP()
   fVar_XicPlus(0),
   fTree_XicPlus_QA(0),
   fVar_XicPlus_QA(0),
+  fVar_XicPlus_EvtID(0),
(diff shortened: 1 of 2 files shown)


PWGHF/vertexingHF/AliAnalysisTaskSEXicPlusToXi2PifromKFP.cxx
```diff
@@ -2398,10 +2436,18 @@ void AliAnalysisTaskSEXicPlusToXi2PifromKFP::FillTreeRecXicPlusFromCasc(AliAODEv
   fVar_XicPlus_QA[74] = fVar_XicPlus_QA[72] - fVar_XicPlus_QA[46]; // SV: pt diff of Xic+ (w/ topo. constraint recalPV - PV)
 
   fVar_XicPlus_QA[75] = fpVtx->CountRealContributors(); // PV: count daughter primary tracks
+  fVar_XicPlus_QA[76] = GetMCEventID(); // Event ID for MC
+  }
+  if (!fIsMC) {
+    fVar_XicPlus_QA[76] = GetEventIdAsLong(AODEvent->GetHeader()); // Event ID for Data
```
fgrosa, line 2442: `fVar_XicPlus_QA` is a `Float_t*` so you are rather casting the `ULong64_t` (which has 64 bits) to a `Float_t` (that has 32 bits), so you lose some information and likely your EventID will not be unique

fgrosa (changes requested):

PR #19000, opened 2021-11-17
PWGHF: Fix for Dstar structure

lvermunt:
Fixed bug for future runs on PbPb with Preselect enabled (overlooked this in previous PR)


fgrosa:
+1

PR #19002, opened 2021-11-17
Background subtraction using the perp cone method

hahassan7:
We implemented a new code for subtraction the background from jets using the perpendicular cone method.


mfasDa:
+1

PR #19029, opened 2021-11-20
include the event cuts for pp Run 2 to AliRsnEvent cuts+AddTask and config file for sigma star in pp at 5 TeV

pganoti:
include the event cuts for pp Run 2 to AliRsnEvent cuts+AddTask and config file for sigma star in pp at 5 TeV


bdoenigu:
+1

PR #19030, opened 2021-11-20
 Fix the GetUniqueID(), after the TParticle changes. Now the True pho…

amaringarcia:
…tons are found


jokonig:
Please start you PW with PWGGA/GammaConv in the future

jokonig:
+1

PR #19036, opened 2021-11-22
PWGHF: Fix issue Dstar MLSelector

lvermunt:
- Fixed a bug that I introduced when cleaning before opening PR #18988 (used before defined) 
- Added @fgrosa PbPb fix also for this task


fgrosa:
+1

PR #19037, opened 2021-11-22
Add histograms for closure test

mfasDa:
- Add histograms for jet finding efficiency and purity
  as well as particle level spectrum for closure and non-
  closure sample
- Apply spliting of closure and non-closure sample at
  event and not at jet level


mfasDa:
+1

PR #19042, opened 2021-11-23
Split MC background sources (-2: Xic+ -> Xi*0pi+ -> Xi-pi+pi+) into rompt (-4) and feed-down (-5); Add pointing angle in x-y plane

zhujh126:


fgrosa:
+1

PR #19071, opened 2021-11-26
PWG/EMCal: Added correction component for cluster position correction

jokonig:
- New correction component to correct the cluster position
- Shifts are evaluated by comparing the position of matched tracks to
the cluster
- Shifts are very small and will not affect most analyses, however for
pi0 reconstruction with PCM-EMCal, this correction is necessary as
otherwise the pi0 peak width differs in data and MC
- Correction can be applied to data or to MC


loizides:
+1

PR #19072, opened 2021-11-26
Add some switches, prompt code

wenyaCern:
This version aims to prompt the code to solve the following problems:
1, Add more switches to separate code into different parts containing different functions like : Int Cov with ESE, V24 Vs. Ach, Diff V22, etc.
2, Reduce output files size for V24 Vs. Ach
3, Prompt some input parameters


rihanphys:
+1

PR #19089, opened 2021-11-27
correct path of the config file

pganoti:
correct path of the config file for sigma star pp at 5 TeV analysis


mario-krueger:
+1

PR #19114, opened 2021-11-30
resolving problem with ghost particles, Gleb Romanenko

glromane:
Added checking for non-zero particle pdg code before filling histograms (line 215) to prevent presence of unidentified ("ghost") particles in the resulting histograms. Changing 'fPtPDG' monitoring histogram range (line 46) to add pT-distribution of unidentified particles (line 253).


erogocha:
+1

PR #19116, opened 2021-12-01
Code only for CMWESE

wenyaCern:
Since adding v24 into CMW ESE code has met many errors, even cannot pass lego train test. We reduce the code only to calculate the CMWESE effect for 11h and 10h. 
Systematic loop produced in code file labeled by "syst".


yzhouCERN:
+1

PR #19117, opened 2021-12-01
A blank line is added at the end of Addtask.

ChunzhengLab:
1. Adjusted the bins of lambda mass;
2. Set the default parameter for bCheckPIDFlow;
3. **Add a blank line to the end of Addtask. If there is no blank line as EOF in the end of AddTask, a cling errors occurs when running LEGO train.**


rihanphys:
+1

PR #19120, opened 2021-12-01
PWG/EMCal: fix in new cluster position correction task

jokonig:
- Remove condition that task can only run on MC events


mfasDa:
+1

PR #19129, opened 2021-12-02
PWG/EMCal: changed member presistence to keep values of variables

jokonig:
- changed member variables to presistent as they need to be streamed to
keep the values assigned by the yaml file


raymondEhlers:
+1

PR #19177, opened 2021-12-08
[ALIROOT-8678] Add support for weighted pythia productions

mfasDa:
- Add field, setter and tester for weighted prod
- Include type weighed prod in ConfigureMCDataset
- Fill general histograms with weight (1 default, event
   weight in case the production is configured as weighted
   pythia


mfasDa:
+1

PR #19189, opened 2021-12-09
Fix AddTaskTPCPIDEtaQA and AliTPCPIDEtaQA class

MartinSchmidtTuebingen:
Fix type of AddTaskTPCPIDEtaQA variables and comment not working and not needed phicut functions


MartinSchmidtTuebingen:
Will be continued in PR together with more adjustments of PID framework for ROOT6

PR #19201, opened 2021-12-10
fix little bug & add default calculation

wenyaCern:
This code added to :
1. fix little bug
2. add default measurement because the previous run for default 10h show wrong results(maybe from the input files)


rihanphys:
+1

PR #19220, opened 2021-12-13
Change qn_cent Yaxis Bins

wenyaCern:
Subdivide the qn distribution bins ---- we need thinner qn distribution slices, to divide qn bins more precisely.


yzhouCERN:
+1

PR #19229, opened 2021-12-13
Switcher to accept fits with ROOT::Fit::FitResult::IsValid() true also for multi-trial

mfaggin:


mfaggin:
@fprino this modification is in line with that of this morning, but for the multi-trial class that bases on AliHFMassFitterVAR.cxx

xinyepeng:
+1

PR #19233, opened 2021-12-14
Modification on the PWGJE/EMCALJetTasks/CMakeLists.txt so that the PW…

rhosokawa:
…GJE/EMCALJetTasks/FlowVectorCorrections/macros directory is copied to an install directory


raymondEhlers:
+1

PR #19236, opened 2021-12-14
Remove chi2 cut when constructing primary vertex with KF

zhujh126:
Remove chi2 cut when constructing primary vertex with KF


xinyepeng:
+1

PR #19237, opened 2021-12-14
Add topological variables Lc and Ks0 to reco tree

AnnalenaKa:
Adding the variables (Decay legth Lc 3D, l/dl Lc in xy plane, cos PA Lc in xy plane, d0 Lc in xy plane, DCA Lc, DCA Ks0, signed d0) to the reconstruction tree.


jezwilkinson:
+1

PR #19240, opened 2021-12-14
AliAnalysisTaskJetExtractor: include calibrated event-plane angle

caitiebeattie:
Updated the JetExtractor task to include an option for using a fully calibrated event-plane angle.

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskJetQnVectors.h
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskJetQnVectors.h
@@ -8,16 +8,10 @@
 
 //*************************************************************************************************
 // \class AliAnalysisTaskJetQnVectors
-// \brief task used to load the Qn calibrations and get the calibrated Qn vectors for HF analyses
+// \brief task used to load the Qn calibrations and get the calibrated Qn vectors for JE analyses
 // \authors:
-// F. Grosa, fabrizio.grosa@cern.ch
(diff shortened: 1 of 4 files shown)


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskJetQnVectors.h
```diff
@@ -93,6 +88,7 @@ class AliAnalysisTaskJetQnVectors : public AliAnalysisTaskSE {
     unsigned long long fTriggerMask;                 /// trigger mask
 
     double fq2V0;                                    /// q2 vector from the V0     
+    double fEPangleCF;                               /// EP Angle with callibrations from V0
```
raymondEhlers, line 91: You should initialize this in the constructor. And since you've implicitly set this to be a streaming (I think - it's a bit ambiguous), you should bump the ClassDef

raymondEhlers (changes requested):
Hi Caitie, generally looks good, but you should make a few small changes

caitiebeattie:
Hi @raymondEhlers, thanks for the feedback! I've made the corrections.

raymondEhlers:
Hi @caitiebeattie , I actually meant for my comments to apply to the class in `AliAnalysisTaskJetQnVectors.h`, not in the EventExtractor. It's kind of late, so I'm not sure if I'm following it correctly, but at first glance, I think you don't need to add the EP angles as members of the EventExtractor. Sorry for the confusion

raymondEhlers:
+1

PR #19241, opened 2021-12-14
PWGHF: Additional possibilities PID Dstar

lvermunt:
* Added the possibility to require TPC **or** TOF for the PID nsigma selection, instead of TPC **and** TOF as was the default (and still is the default)
* Added the possibility to include the soft pion candidate track in the PID selection when PID info is available (turned off by default)


fgrosa:
+1

PR #19280, opened 2021-12-19
[PWG-CF]Add Input Files Path

wenyaCern:
This pull adds input files paths which are needed for 10h track wise systematic estimation for CMWESE


rihanphys:
+1

PR #19331, opened 2021-12-28
Event Wise Syst Estimation with Subwagons

wenyaCern:
This version of code for CMWESE aims to:
1, Tune AddTask.C to utilize subwagon function in train
2, Aim to do the calculation for event-wise systematic sources : vz cut and pile up


yzhouCERN:
+1

PR #19335, opened 2021-12-29
AliAnalysisTaskJetExtractor, AliAnalysisTaskJetQnVectors: allow user to select detector for measuring q2

caitiebeattie:
Update to allow user to choose between V0M, V0A, and V0C for calculating q2, EPangle. Involves using SetQ2Detector flag. Default is V0M. Also includes relative EPangle for matched jets at detector level.


raymondEhlers:
+1

PR #19343, opened 2022-01-01
[PWG-CF] CMWESE Fill for NUAweight

wenyaCern:
This version of the code is edited for NUA weight with different track cuts. Prepared for track-wise systematic in CMWESE analysis.


rihanphys:
+1

PR #19366, opened 2022-01-05
Adjust Jet ID Analysis to use AliVEvent if possible

MartinSchmidtTuebingen:
Using InputEvent() when calling other classes where AliVEvent is sufficient

Make task run with ESDs and AODs, small code improvements

Use AliAnalysisTask centrality estimator

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskIDFragmentationFunction.cxx
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskIDFragmentationFunction.cxx
@@ -68,6 +68,7 @@
 #include "AliHelperClassFastSimulation.h"
 
 #include "AliAnalysisTaskIDFragmentationFunction.h"
+
 using std::cout;
 using std::endl;
 using std::cerr;
@@ -82,7 +83,6 @@ AliAnalysisTaskIDFragmentationFunction::AliAnalysisTaskIDFragmentationFunction()
    ,fAODJets(0)  
    ,fAODExtension(0)
    ,fNonStdFile("")
-   ,fCentralityEstimator("V0M")
    ,fMinMultiplicity(-1)
    ,fMaxMultiplicity(-1)
    ,fNameTrackContainer("Tracks")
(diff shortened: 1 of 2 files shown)


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskIDFragmentationFunction.cxx
```diff
@@ -944,31 +931,38 @@ Bool_t AliAnalysisTaskIDFragmentationFunction::FillHistograms()
...
+  if (fDebug > 1)
+      Printf("%s:%d event ACCEPTED ...",(char*)__FILE__,__LINE__); 
+  
   fh1EvtSelection->Fill(0.);
-  fh1VtxSelection->Fill(primVtx->GetType());
+  
+  const AliAODVertex* aodVertex = dynamic_cast<const AliAODVertex*>(primVtx);
+  if (aodVertex) {
+      fh1VtxSelection->Fill(aodVertex->GetType());
+  }
```
mfasDa, line 964: Not AliVVertex?

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskIDFragmentationFunction.cxx
```diff
@@ -2039,9 +2033,7 @@ AliEmcalJet* AliAnalysisTaskIDFragmentationFunction::GetPerpendicularCone(AliEmc
   if (!processedJet)
     return 0x0;
   
-  Double_t dDistance = 2.0 * TMath::Abs(GetFFRadius());
-  
-  AliEmcalJet* perpJet = new AliEmcalJet(processedJet->Pt(), processedJet->Eta(), processedJet->Phi()+perpAngle, processedJet->M());
+  AliEmcalJet* perpJet = new AliEmcalJet(processedJet->Pt(), processedJet->Eta(), processedJet->Phi() + perpAngle, processedJet->M());
```
mfasDa, line 2036: In case this is just used locally inside the function it should be better put on the stack (AliEmcalJet perpJet(...);) in order to avoid memory leaks.

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskIDFragmentationFunction.cxx
MartinSchmidtTuebingen: In this specific case, no. GetType only exists in AliAODVertex.

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskIDFragmentationFunction.cxx
MartinSchmidtTuebingen: No, they are returned and saved in a list to be processed later. The list is the owner of the stored objects and is properly deleted, so there should be no memory loss.

mfasDa:
+1

PR #19386, opened 2022-01-10
PWGHF: Fix PID option MLVarHandler

lvermunt:
The Bayes probability was always present in the TTree because of a "typo"


fcatalan92:
+1

PR #19389, opened 2022-01-10
PWGHF: Update Dstar invmass range THnSparse

lvermunt:
Increased upper limit of Dstar invariant mass in THnSparse to have more room for the multitrial fitter (+ now the bin width is 0.1 MeV/c^2 instead of 0.044)


fgrosa:
Hi @lvermunt, is it really necessary to go up to 0.188 for the invariant-mass axis? If I recall well most of the time the fits go up to ~0.15, and anyway to not exceed 0.16 (but I might be wrong)

fgrosa:
+1

PR #19390, opened 2022-01-10
New task for delta-phi correlations wrt event plan for MCKine train

marcovanleeuwen:
This pull request is on behalf of Darius Keijdener. He has prepared this task for studies with generator-only events, in particular with AMPT, using the MC kine train.


yzhouCERN:
+1

PR #19393, opened 2022-01-10
PWGHF: Fix Dstar impact parameter prongs

lvermunt:
Fix for the impact parameter of the prongs of the Dstar. It was only filled correctly for the soft pion (D0 prong 1 was the impact parameter of the D0 candidate itself, the value for the D0 prong 2 was random)


fgrosa:
+1

PR #19396, opened 2022-01-10
[ALIROOT-8678] Add histogram for sum of weights before event selection

mfasDa:
Needed in case of EMCAL trigger selection being part of the
event selection

In addition exclude histograms with sum of weights from the
automatic scaling.


mfasDa:
+1

PR #19414, opened 2022-01-12
[PWG-CF] CMW ESE with Trk Wise Syst

wenyaCern:
Add the path for NUA files, for track-wise syst estimation: chi2cut, Nhits, FB, etc.


rihanphys:
+1

PR #19418, opened 2022-01-12
AliAnalysisTaskJetQnVectors, AliJEQnVectorHandler: allow usage of calibration files not stored on EOS

caitiebeattie:
Enables use of calibration files stored in working directories.


raymondEhlers:
+1

PR #19421, opened 2022-01-12
Make AddTaskESDFilterPWGJETrain ROOT6 ready

MartinSchmidtTuebingen:
+ Small log improvement in AliAnalysisTAskIDFragmentationFunction

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskIDFragmentationFunction.cxx
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskIDFragmentationFunction.cxx
@@ -765,7 +765,8 @@ Bool_t AliAnalysisTaskIDFragmentationFunction::FillHistograms()
   
   fESD = dynamic_cast<AliESDEvent*>(InputEvent());
   if(!fESD){
-    if(fDebug>3) Printf("%s:%d ESDEvent not found in the input", (char*)__FILE__,__LINE__);
+    if(fDebug>3) 
+      Printf("%s:%d ESDEvent not found in the input", (char*)__FILE__,__LINE__);
   }
   
   // get AOD event from input/ouput
@@ -806,12 +807,12 @@ Bool_t AliAnalysisTaskIDFragmentationFunction::FillHistograms()
   }
   
   if(!fAOD){
-    Printf("%s:%d AODEvent not found", (char*)__FILE__,__LINE__);
-//     return kFALSE;
+    if(fDebug>3) 
+      Printf("%s:%d AODEvent not found", (char*)__FILE__,__LINE__);
   }
   if(!fAODJets){
-    Printf("%s:%d AODEvent with jet branch not found", (char*)__FILE__,__LINE__);
-//     return kFALSE;
+    if(fDebug>3) 
+      Printf("%s:%d AODEvent with jet branch not found", (char*)__FILE__,__LINE__);
   }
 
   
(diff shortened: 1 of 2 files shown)


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskIDFragmentationFunction.cxx
```diff
@@ -765,7 +765,8 @@ Bool_t AliAnalysisTaskIDFragmentationFunction::FillHistograms()
   
   fESD = dynamic_cast<AliESDEvent*>(InputEvent());
   if(!fESD){
-    if(fDebug>3) Printf("%s:%d ESDEvent not found in the input", (char*)__FILE__,__LINE__);
+    if(fDebug>3) 
+      Printf("%s:%d ESDEvent not found in the input", (char*)__FILE__,__LINE__);
```
mfasDa, line 769: In general I recommend you to use the AliLog facility. In this case you would do
``` c++
AliDebug(3, "ESD Event not found in input");
```
and you can steer this in the macro / ROOT prompt via
``` c++
AliLog::SetClassDebugLevel("AliAnalysisTaskIDFragmentationFunction");
```
Similar loggers exist for error messages as warning (AliWarning), error (AliError) and fatal (AliFatal), the later will terminate execution.

In many cases it now makes more sense to use the c++ streaming facility, particularly if one wants to stream objects where the operator<< is implemented. With AliLog this looks like
``` c++
AliDebugStream(3) << "ESD Event not found in input" << std::endl;
```
(of course using the same debug levels as AliDebug). O2 will use a similar mechanism, called FairLogger.

PWGJE/macros/AddTaskESDFilterPWGJETrain.C
```diff
@@ -1,4 +1,10 @@
+#ifdef __CLING__
+// ROOT6, bindet andere macros wie Header files ein, include Pfad kann mit R__ADD_INCLUDE_PATH hinzugefügt werden
+R__ADD_INCLUDE_PATH($ALICE_PHYSICS)
+#include <PWGJE/macros/CreateTrackCutsPWGJE.C>
 
+#include "AliAnalysisTaskESDfilter.h"
```
mfasDa, line 6: Please do not include headers of classes in ROOT/AliRoot/AliPhysics. Those are handled automatically by cling via so-called pre-compiled headers. Includes should be used only in 2 cases:
- Code using functions from other macros (this is the case for CreateTrackCutsPWGJE.C)
- Code using functions from libraries for which no pre-compiled headers exists (examples: direct used of fastjet, pythia, other c++ generators in macros)
`AliAnalysisTaskESDfilter` is in libANALYSISalice, for which pre-compiled headers exist, ROOT should find them without including the header file.

mfasDa (changes requested):
Please remove the unnecessary include of `AliAnalysisTaskESDfilter.h`, those can lead to troubles if paths are screwed up or you have local versions of the file in the local directory. Otherwise fine for merging. The logging facility is just a suggestion to make life easier.

PWGJE/macros/AddTaskESDFilterPWGJETrain.C
MartinSchmidtTuebingen: Done

mfasDa:
+1

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskIDFragmentationFunction.cxx
MartinSchmidtTuebingen: Yes, I know about AliDebug. The class is sprinkled with this old logging facility, I will remove it in another PullRequest all together.

PR #19438, opened 2022-01-14
AddTaskMacro for EP dependent dihadron task (kine train)

marcovanleeuwen:
We forgot to add an AddTask macro for the EP dependent dihadron task which was committed last week (Darius' analysis).


rihanphys:
+1

PR #19473, opened 2022-01-18
Add sparse for reflections.

mfaggin:


arossi81:
Ciao Mattia,
- I would add a flag for filling the new sparse because it will double the size of the output and it is not always needed.
- increase the class number (though it is not formally needed)
Cheers,
Andrea

mfaggin:
@arossi81 thanks for the suggestions, I followed them

arossi81:
+1

PR #19480, opened 2022-01-19
Fix streaming of angularity handler

mfasDa:
- Extract AngularityHandler to separate class
- Set ClassDebugLevel for AngularityHandler and
  AngularityBin
- Add streaming operators for debugging
- Increase debug levels in AliAnalysisTaskEmcalJetEnergyScale


mfasDa:
+1

PR #19483, opened 2022-01-19
Adding the possibility to build candidates only with signal tracks an…

arossi81:
…d expanding the NTuple for studying fast selections.


arossi81:
+1

PR #19486, opened 2022-01-19
Add system to cycle between masses and PDGs for Ps MC Analysis

JakeMartinez:
4 different mass states with different PDG numbers need to be analyzed, I am including a variable that, when it is on during MC, will cycle through all necessary mass states and PDG numbers


mario-krueger:
+1

PR #19491, opened 2022-01-20
Fix AliAnalysisTaskAR

ariedel-cern:
Protect against floating point exception (division by zero) in AliAnalysisTaskAR when there are not enough tracks for calculations after applying all cuts.


dongjokim:
+1

PR #19495, opened 2022-01-20
track label, number of casc daughters

chengtt0406:
Using the absolute track label;
Adding one condition about the numbers of daughters of cascade


jezwilkinson:
+1

jezwilkinson:
@xinyepeng I think that when you comment on the commit directly instead of here on the main thread, the build machine doesn't see the +1 properly

xinyepeng:
> @xinyepeng I think that when you comment on the commit directly instead of here on the main thread, the build machine doesn't see the +1 properly

@jezwilkinson yes, I realize it, thanks!

PR #19502, opened 2022-01-21
Adding overflow bin to angularity handler using the same value as the…

mfasDa:
… highest bin

In addition adding catch for BinNotFoundException and monitor number of errors


mfasDa:
+1

PR #19569, opened 2022-01-27
Fix: fix GetPointers Method

ariedel-cern:
Fix error prone method for initalizing vectors of symmetric cumulants.


dongjokim:
+1

PR #19582, opened 2022-01-28
PWGGA/GammaConv: Added config for pPb 5TeV omega PCM

nstrangm:
-Added argument doWeightingPart as it was not declared and task was crashing


jokonig:
+1

PR #19587, opened 2022-01-28
PWG/EMCAL: Added control histogram to cell track matcher

jokonig:
- Control histogram to check how many cells are good matches and how
many are fake matches
- Added option to only run cell track matching for primaries (R < 5cm)
such that conversions can still be fully rejected. Electrons should
leave their full energy in the cluster instead of just the MIP


FriederikeBock:
+1

PR #19594, opened 2022-01-30
Add pT of beauty mother to reconstructed and generated tree

AnnalenaKa:
Add the pT of the beauty mother in the case of non-prompt Lc to the reconstructed tree and generated tree.


fgrosa:
+1

PR #19601, opened 2022-01-31
Add rotational background for Lc w/o vertexing - kaon rotation

mfaggin:


mfaggin:
@arossi81 here the code to insert the Lc rotations. In the end I opted for the rotation of the kaon track (namely the one in the middle). The logic is that everything is done in one function before the sparse filling, when for each selected "real" candidate we build 19 rotated candidates (rotation in phi, with step pi/10) and after calculating invariant mass and pt we fill the analysis sparse. The rotated candidates are flagged with infoMC==-1 (underflow of axis 6 of the fhSparseAnalysis)
As discussed, this properly works ok only w/o topological cuts.

arossi81:
+1

PR #19605, opened 2022-02-01
PWGHF: Adding First version of Lc-h codes (by Samrangy Sadhu)

fcolamar:


fcolamar:
Hi, Samrangy tried to compile the code and to run the ctest before the pull request, and everything worked fine (I also cross-checked and get no issues as well). Anyway here we got errors in the checks, does anybody know why this could happen? Thanks

fgrosa:
Hi @fcolamar for sure there is one mistake in your `CMakeLists.txt`: the `vertexingHF/vHML` library is present only in case of `ROOT6`, see

https://github.com/alisw/AliPhysics/blob/0b6ced918c38a6c11455ee06748ce403f470b398/PWGHF/vertexingHF/vHFML/CMakeLists.txt#L16

because the library needed for the BDT inference (treelite) is not included in the tags with `ROOT5`. Hope it helps!

fcolamar:
Hi @fgrosa, thanks! Indeed this info was helpful and solved the issues on ROOT5.
Though ctest on ROOT6 still crashes when trying to load PWGHFcorrelationHF, and we cannot reproduce the issue locally: does anybody know if it's possible to retrieve further information (the log is not particularly exhaustive)?

fcolamar:
Problem fixed, it was a missing library dependence in the CMakeLists.txt

PR #19637, opened 2022-02-04
Mark PHOS cells as calo type 0

mfasDa:
PHOS cells were not marked. As EMCAL cells are processed
before PHOS the PHOS cells were assigned calo type 1 which
is the index for EMCAL cells, consequently PHOS cells were
appearing as EMCAL cells.


mfasDa:
The previous commit message mentioned a change of the "cell type". Obviously the cell type was not changed but rather the "calo type", as the calo type distinguishes between the calorimeters. The commit message was fixed accordingly.

jgrosseo:
+1

PR #19647, opened 2022-02-06
Switch on pileup cuts in AliTriggerAnalysis

ekryshen:
kIsPileupFromSPD and kIsV0PFPileup bits were always set to 0 in run2bcinfo.fEventCuts. Need to switch on pileup cuts in AliTriggerAnalysis


jgrosseo:
+1

PR #19698, opened 2022-02-11
Fix missing infos

romainschotter:
Fix missing infos (event counting, vertex quality info, missing kink info in AOD) and add minor fixes


bdoenigu:
+1

PR #19714, opened 2022-02-11
Add syst. unc. non-prompt Lc(->pKpi) pp 13 TeV

strogolo:
The following PR is to commit the systematic uncertainties presented and approved during the D2H meeting (11th Feb. 2022).


xinyepeng:
+1

PR #19732, opened 2022-02-14
PWGGA/GammaConv: added option to run without pion preselection for heavy meson task

fjonasALICE:


FriederikeBock:
+1

pzhristov:
+test

fjonasALICE:
@pzhristov Info for debugging purposes: I have approval rights in this part of the code, so normally the PR would also have been auto approved. I guess since it never reaches the test stage it gets stuck even before people are assigned to the PR. Also aliBuild seems to be stuck in some sort of loop now ....

PR #19737, opened 2022-02-14
Class for Sigma->nbar+pi+ added

peressounko:


peressounko:
+1

pzhristov:
I will merge the PR, but again, there is no information about the checks. At least we have to be sure the PR compiles on Linux and Mac. @TimoWilken Would you check what caused the change of behaviour?

PR #19738, opened 2022-02-15
Minor fix: Return task pointer from AddTaskEPDependentDiHadronOnTheFly.C

marcovanleeuwen:
Minor fix in the AddTask macro, needed for train runs (committed on behalf of Darius Keijdener)


yzhouCERN:
+1

PR #19762, opened 2022-02-16
Add implementation of the uncertainty on the integrated luminosity

mfasDa:
The uncertainty is calculated based on
the difference using the observed
downscaling instead of the expected
downscaling.

In additinon users must specify the
unit in which the integrated luminosity
is expressed. Internally the integrated
luminosity is handled in pb-1.


mfasDa:
+1

PR #19770, opened 2022-02-17
Add low limit mass cut for eXi pairs

chengtt0406:
To have a look the effect from low limit pair mass cut, which was observed in D2H talk recently


fgrosa:
+1

PR #19772, opened 2022-02-17
Added primary vtx parameters output

rosarioturrisi:
Added to the THnSparse output four parameters of primary vertex: transverse and longitudinal resolution, chi-square, longitudinal position.


fprino:
+1

PR #19780, opened 2022-02-18
Updated analysis task for Rho+J/psi analysis

abylinkin:
Updated track cuts and added PIDTPC branches for proton and kaon identification


mbroz84:
+1

PR #19804, opened 2022-02-20
Update the name of container

SiyuTang:
Hi,

This is an update of the task of jet v2. The name of container is updated to avoid the error in lego-train.

Please approve it. Thanks!

Siyu


yzhouCERN:
+1

PR #19820, opened 2022-02-22
updates to jet-Event-Shape-Engineering tasks

caitiebeattie:
JetExtractor, JetQnVectors Updates, AliJEQnVectorHandler: Allow for LHC18q+r to be run simultaneously. Addition of branch for Q2 cross-check with 2nd detector.

### DIFF
--- a/PWGCF/FEMTOSCOPY/FemtoDream/AliAnalysisTaskOtonkd.cxx
+++ b/PWGCF/FEMTOSCOPY/FemtoDream/AliAnalysisTaskOtonkd.cxx
@@ -27,6 +27,7 @@ AliAnalysisTaskOtonkd::AliAnalysisTaskOtonkd()
     fIsMCtruth(false),
     fdoFDpairing(false),
     fisIncludeSomeProtons(false),
+    fisPions(false),
     fdoSideband(false),
     fSigmaUp(0.0),
     fSigmaLow(0.0),
@@ -72,7 +73,7 @@ AliAnalysisTaskOtonkd::AliAnalysisTaskOtonkd()
 }
 
 AliAnalysisTaskOtonkd::AliAnalysisTaskOtonkd(
-  const char *name, bool isMC, bool isMCtruth, bool isIncludeSomeProtons, bool doFDpairing)
+  const char *name, bool isMC, bool isMCtruth, bool isIncludeSomeProtons, bool isPions, bool doFDpairing)
   : AliAnalysisTaskSE(name),
     fisLightWeight(false),
     fTrackBufferSize(2000),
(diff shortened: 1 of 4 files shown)


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskJetQnVectors.h
```diff
@@ -75,12 +74,15 @@ class AliAnalysisTaskJetQnVectors : public AliAnalysisTaskSE {
...
     bool fEnableQvecTPCVsCentrDistr;                 /// flag to enable histos of TPC Q-vector vs. centr
 
-    AliJEQnVectorHandler* fJEQnVecHandler;           /// Qn-vector handler
+    AliJEQnVectorHandler* fJEQnVecHandler1;           /// Qn-vector handler
+    AliJEQnVectorHandler* fJEQnVecHandler2;           /// Qn-vector handler
     int fHarmonic;                                   /// Qn-vector harmonic
     int fCalibType;                                  /// type of calibrations used by handler
     int fNormMethod;                                 /// normalisation of Q vector
 
     TString fOADBFileName;                           /// OADB input file name
```
raymondEhlers, line 83: Based on the changes in the constructor, I think you meant to remove this variable since it's replaced by the other two? You should also update the variable set in `SetUseAODBCalibrations1(...)` as appropriate

raymondEhlers (commented):
LGTM - there's just one minor piece you should check

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskJetQnVectors.h
caitiebeattie: Thanks for catching that typo with the SetUseAODBCalibrations - I've just fixed that. I'm not sure I'm looking in the right spot for the other comment. Which variable are you referring to?

raymondEhlers: You replaced `fOADBFileName` in the constructor with `fOADBFileName1` and `fOADBFileName2` (ie. `fOADBFileName` is not even initialized anymore). It's only settable via `SetUseAODBCalibrations(...)`, but neither `fOADBFileName1` or `fOADBFileName2` are settable. So I assumed that you wanted to replace `fOADBFileName` with those two everywhere. Right now, you seem to have kind of a half and half split. If you want to keep `fOADBFileName`, you should include initialize it in the constructors. If you don't, then I would remove it entirely, and make `fOADBFileName1` and `fOADBFileName2` settable.

PR #19853, opened 2022-02-23
Fix copy/paste error for effective luminosity

mfasDa:
EJ2 trigger must use EJ1 trigger for
the calculation of the effective
(observed downscaling)


mfasDa:
+1

PR #19854, opened 2022-02-23
Modified output

rosarioturrisi:
Added four histos to keep the distribution of primary vtx parameters included in the thnsparse. Modified the range of few thnsparse axes to take into account the range of different centralities.


chiarazampolli:
+1

PR #19867, opened 2022-02-24
Fix mass hypothesis for jet substructure

raymondEhlers:
Was supposed to be assuming the charged pion mass, but instead was using
the PID mass hypothesis. Note that this isn't a problem in the main jet
finding - it's only for the reclustering


raymondEhlers:
+1

PR #19873, opened 2022-02-24
Output upgrade

rosarioturrisi:
Added 1d histos for vtx quality checks and the same variables to the TpcItsMatch list.


chiarazampolli:
+1

PR #19876, opened 2022-02-25
[PWG-CF] Fix a little bug and add wCent

wenyaCern:
This version aims to :
1, fix a little bug on fill Qx/y mean for LHC18r
2, calculate 15o CMWESE with the weight of centrality


yzhouCERN:
+1

PR #19879, opened 2022-02-26
Fix substructure mass hypothesis for constituent subtraction

raymondEhlers:
We don't want to use the pion mass hypothesis for CS since that's an
choice of the algorithm. All cases are now enumerated in the comments


raymondEhlers:
+1

PR #19913, opened 2022-03-01
[PWG-CF] Pile-up Syst with new approch

wenyaCern:
This version of code added a new method to estimate pile-up systematic variation


yzhouCERN:
+1

PR #19938, opened 2022-03-03
Further fixes for mass hypothesis for reclustering

raymondEhlers:
Need to skip the pion mass hypothesis for both jet-wise and event-wise
CS, but was missing the check for the event-wise case


raymondEhlers:
+1

PR #19944, opened 2022-03-04
Added AliEventCut as a possible event selection.

osaarima:
This can be turned on or off by a flag in the addtask macro


dongjokim:
+1

PR #19961, opened 2022-03-06
TPC: Add splines and eta maps for LHC16d.pass2 offline and HLT clusters

wiechula:
* small fix to drawing macro


chiarazampolli:
+1

PR #19989, opened 2022-03-09
AliAnalysisTaskJetQnVectors: TH3Fs for resolution calculation

caitiebeattie:
addition of 3D histograms for detector resolution calculation


raymondEhlers:
+1

PR #19991, opened 2022-03-10
LHC18qr update: V0MNew

ddobrigk:
@rolavick @strogolo @ekryshen @chiarazampolli @noferini: This is the LHC18qr update that provides the new estimators that are needed for cross-checks. It implements "V0MNew" as well as minus05/10 and plus05/10 variations simultaneously in data and MC for both q and r periods.


ddobrigk:
+1

PR #19992, opened 2022-03-10
Fix constructor and update addtask macro (event-plane dependent dphi)

marcovanleeuwen:
Further fix and update for the on-the-fly task from Darius.


yzhouCERN:
+1

PR #19996, opened 2022-03-10
Fix bug for grid analysis for MTFPID

MartinSchmidtTuebingen:
Fix bug happening when copying analysis class like on grid to convolutedGaussFunction (does not get copied correctly). Initialise parameters in UserCreateOutputObjects


mfasDa:
+1

PR #19997, opened 2022-03-10
More fixes for EPDependentDphi task

marcovanleeuwen:


marcovanleeuwen:
Just realised that removing the default constructor is in principle not a good idea; will do some more tests before committing this.

PR #20013, opened 2022-03-14
PWGHF: Introduce Lc->pK0s background from mixed events

lvermunt:
Based on code from 15o PbPb analysis. Required for Referee comment


adubla:
+1

PR #20021, opened 2022-03-15
PWGHF: Fix to run mixed event stand-alone

lvermunt:
PID object was not initialised when running mixed-event stand alone. Fixed + added warnings


fgrosa:
+1

xinyepeng:
+1

PR #20022, opened 2022-03-15
fix tfile-related bug

alindner14:


mario-krueger:
+1

PWGLF/SPECTRA/MultEvShape/AliMEStenderV2.h
```diff
@@ -129,6 +129,7 @@ class AliMEStenderV2 : public AliAnalysisTaskSE
   AliMESeventInfo *fMCevInfo;  //!
   TClonesArray *fMCGenTracksIO;   //!
   TClonesArray *fMCtracksMissIO; //!
+  TFile *fTreeFile;
```
mario-krueger, line 132: Probably this makes no difference at all since you override the pointer anyhow, but I suppose you do not want to stream this TFile* member variable. I would add there a //!

mario-krueger: Also, thinking about it a bit more, it is not clear to me where this file would be created.
Suppose it will work on the grid (which I doubt) every instance of your task would create another root-file but then what will be done with it? If I read the code correctly, it is not used anywhere, right?

alindner14: Thanks for the streamer heads-up. I changed that. 
The reason for creating this TFile was to solve an error realted to the tree crated as a memory-resident Tree, which is caused due to the fact that the tree is created before the output file, thus, relates the tree to the input file instead to the output file. (e.g. https://root-forum.cern.ch/t/this-error-is-symptomatic-of-a-tree-created-as-a-memory-resident-tree/38891)

mario-krueger: I see. Interesting root issue. But why does it appear specifically for this one MC dataset, while in the past the code worked fine?
We can try it, however I am not entirely sure if the system will just let you actually create files on the grid computers.
In any case, even if it works they should be deleted / closed / unlinked somewhere in the destructor.

alindner14: It appears because I have reconfigured my output objects as a workaround for the runtime crashes. 
I have also deleted the tfile in the destructor.

PR #20034, opened 2022-03-16
Fix a bug of TPC Plane for 10data and add ZDC Plane for 18data of CVE

ChunzhengLab:
Fix a bug of TPC Psi2 and Format the code.
Add ZDC


yzhouCERN:
+1

PR #20051, opened 2022-03-17
First Version Omegac0 analysis 2XiPi

ZFederica:
AnalysisTask and AddTask for the analysis of omegac0 decaying into XiPi (new decay channel)


cterrevo:
+1

PR #20069, opened 2022-03-21
Add option to cut separately on charged and neutral particle pt

mfasDa:
Cut implemented in
AliMCParticleContainer as it is relevant
only in case of simulation. The title string is only modified in case a cut
is set, in order to keep backward comp
for existing tasks.


mfasDa:
+test

mfasDa:
@ktf @TimoWilken Does anyone of you understand why this PR was merged even though I just commented with +test?

TimoWilken:
Hi @mfasDa, thanks for the report -- I'll look into why this was wrongly approved. Sorry for the inconvenience!

PR #20075, opened 2022-03-22
Add virtual function for user code to be filled before event selection

mfasDa:
Provide funtion "UserRunBeforeEventSelection" for
user code before event selection. Notably
this is used to fill part. level histograms
without bias from det. level event selection


mfasDa:
+1

PR #20109, opened 2022-03-30
Allows JEQnVectorHandler to work for any harmonic, not just n =2

moliver813:
Allows JEQnVectorHandler to work for any harmonic, not just n = 2
Adds TPC Full, negative, positive to extractable event planes
Should not change anything for n=2 harmonic


raymondEhlers:
+1

PR #20130, opened 2022-04-05
Revert "Fix variable-size array initialisation"

alindner14:
Reverts alisw/AliPhysics#20119

### DIFF
--- a/PWGLF/SPECTRA/MultEvShape/AliMEStenderV2.cxx
+++ b/PWGLF/SPECTRA/MultEvShape/AliMEStenderV2.cxx
@@ -37,13 +37,11 @@
 #include "AliMEStrackInfo.h"
 #include "TTree.h"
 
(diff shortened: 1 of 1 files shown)


bdoenigu:
+1

pzhristov:
@alindner14  @bdoenigu I am sorry, but the code does not compile on Mac if you try to initialise variable size arrays or declare them with square brackets. The fix was quite OK, what was the reason to revert it? Of course, you can new the arrays and then delete them using delete [], however std::vector does the same automatically.

alindner14:
Hi @pzhristov , the changes were not ok, I had problems with memory allocation with that version. To fix this problem asap, I created a PR (https://github.com/alisw/AliPhysics/pull/20138) reverting the commit that I made in the first place, that gave problems when compiled on macos.

PR #20145, opened 2022-04-09
Change to allow multiple intervals (Run 3 MC tests)

ddobrigk:


ddobrigk:
+1

ddobrigk:
+1

ddobrigk:
+test

ddobrigk:
+1

ddobrigk:
+1

ddobrigk:
+test

ddobrigk:
+1

ddobrigk:
+1

ktf:
@ihrivnac could you check what is going on? Do I remember correctly you bumped vmc?

ihrivnac:
> @ihrivnac could you check what is going on? Do I remember correctly you bumped vmc?

New Geant4 version (11.x) requires build with c++17 standard.

ddobrigk:
Thanks a lot! Does this mean we need to update alidist with that compilation option? 
(should this not have been caught by the CI while testing alidist, maybe?)

ihrivnac:
> Thanks a lot! Does this mean we need to update alidist with that compilation option? (should this not have been caught by the CI while testing alidist, maybe?)

The C++ standard  is set to '17' in o2-defaults.sh; where this value is set for building AliPhysics?

ktf:
I do not remember if we actually allow C++17 in AliPhysics. @pzhristov do you remember?

pzhristov:
> I do not remember if we actually allow C++17 in AliPhysics. @pzhristov do you remember?

Yes, when we use --defaults o2, we use CXXSTD: '17' and it works well also for AliPhysics.

pzhristov:
I suppose adding   CXXSTD: '17' to the environment in defaults-next-root6.sh would solve the issue. Probably we can also move to defaults-o2.sh for the Root6 analysis builds.

pzhristov:
I crated https://github.com/alisw/alidist/pull/4058

PR #20160, opened 2022-04-12
JEQnVector divide-by-zero fix

moliver813:
Prevents a divide-by-zero when multiplicity in a subdetector is zero.


raymondEhlers:
+1

PR #20166, opened 2022-04-13
Add files via upload

scostanz:
Hi,
I would like to add the method SetPtWeightsFromFONLL13overLHC20f4abc() in the AliCFTaskVertexingHF class.
The function contains the values of the histogram ratio of the LHC20f4a/b/c MC with PYTHIA8 and FONLL calculations for pp data at 13 TeV using D0 simulated pT distribution.

Thanks,
Susanna


cterrevo:
+test

PR #20191, opened 2022-04-19
PWGGA/GammaConv: Added option to enable material budget weights in pPb omega AddTask...

nstrangm:
Added the 13 TeV standard PCM Trainconfig to pPb AddTask


jokonig:
+1

PR #20203, opened 2022-04-20
A small fix about the file path of alien_cp

lupengzhong:
Add "file:" before the local file path when using the alien_cp.


iarsene (approved):

iarsene:
+1

PR #20209, opened 2022-04-21
Update AliAnalysisTaskR2P2multClass.h

dbauri:


rihanphys:
+1; @dbauri you should create one PR with all changes, otherwise it might cause error (as your .cxx depends on .h file)

PR #20225, opened 2022-04-22
Add MC info at the TPC entrance

romainschotter:
Add MC info at the TPC entrance by saving the corresponding AliTrackReference object (+add a boolean to turn on/off this feature)


bdoenigu:
+1

PR #20246, opened 2022-04-26
PWG/EMCAL: Added Cell energy and time compression component

jokonig:
- Component to try settings for a compression algorithm used in
O2/DataFormats/EMCAL/...Cell.cxx
- Cell energy/time gets multiplied with a certain resolution factor and
then gets transformed in an integer
- All values for resolution etc. can be set in the yaml file

### DIFF
--- a/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellEnergyCompression.h
+++ b/PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellEnergyCompression.h
@@ -1,0 +1,63 @@
+#ifndef ALIEMCALCORRECTIONCELLENERGYCOMPRESSION_H
+#define ALIEMCALCORRECTIONCELLENERGYCOMPRESSION_H
+
+#include "AliEmcalCorrectionComponent.h"
+
+#include "TF1.h"
+
+/**
+ * @class AliEmcalCorrectionCellEnergyCompression
+ * @ingroup EMCALCORRECTIONFW
+ * @brief Cell energy Compression component in the EMCal correction framework.
+ *
+ * This component is to test different compression algorithms as planned to be used in O2.
+ * The cell energy is multiplied with a certain Compression factor (typially 1/0.0153) and is then stored as int16_t (Short_t)
+ * To retrieve the energy the int16_t (Short_t) is devided by the Compression factor
+ * All
+ * @author Joshua Koenig, joshua.konig@cern.ch
+ * @date Apr 25 2022
+ */
+
+class AliEmcalCorrectionCellEnergyCompression : public AliEmcalCorrectionComponent {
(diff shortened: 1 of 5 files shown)


PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellEnergyCompression.h
```diff
@@ -0,0 +1,63 @@
...
+  void SetEnergy(Double_t e);
+  Double_t GetEnergy() const;
+  void SetTime(Double_t t);
+  Double_t GetTime() const;
+
+  TH1F                  *fCellEnergyCompressionLoss;            //!<! difference between energy with no compression and with compression
+  TH1F                  *fCellTimeCompressionLoss;              //!<! difference between cell time with no compression and with compression
+
+  Bool_t                 fDoEnergyCompression;                  ///< switch to turn on energy Compression
+  int16_t                fEnergy;                               ///< current energy
```
mfasDa, line 42: This needs the header <cstdint> to be included, but not sure CINT can use it, better use short here and convert to int16_t in the cxx file, or exclude the class from the useless ROOT5 builds. There might be a different trick using typedefs, like

```c++
#if ROOT_VERSION_CODE < ROOT_VERSION(5,99,0)
typedef eint16_t short;
#else
#include <cstdint>
typedef eint16_t int16_t;
#endif
...
eint16_t fEnergy;
...
```

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellEnergyCompression.h
```diff
@@ -0,0 +1,63 @@
...
+  Double_t               fTruncationMaxE;                       ///< Maximum cell energy
+  Double_t               fEnergyCompression;                    ///< Energy Compression
+  Double_t               fCompressionBias;                      ///< bias added to energy before converting it to int16_t
+  Bool_t                 fDoTimeCompression;                    ///< switch to turn on time Compression
+  int16_t                fTime;                                 ///< current time
```
mfasDa, line 47: Same as above.

mfasDa (changes requested):
Some change is needed for the useless compatibility to ROOT5 only there thanks to nonsense decisions of the ALICE management.

mfasDa:
+test

PWG/EMCAL/EMCALtasks/AliEmcalCorrectionCellEnergyCompression.h
jokonig: I now just went with Short_t.
Is there any disadvantages the shorts could have in our use case? (I could not think if one)

mfasDa:
+1

PR #20254, opened 2022-04-27
Comments for Systematics

rfacen:
config->SetPtQA(true);
config->SetMassQA(true); config->SetExtendedQAPairs(pairQA); 

Commented


sartozza:
+1

PR #20259, opened 2022-04-28
Removes unneeded generator index check for MC-matched tracks

moliver813:
Removes generator index check that was erroneously applied to only MC-matched tracks. This prevented the task from being usable for MC productions with more than one generator (e.g Pileup + HIJING).

### DIFF
--- a/PWGJE/EMCALJetTasks/AliAnalysisTaskPWGJEQA.cxx
+++ b/PWGJE/EMCALJetTasks/AliAnalysisTaskPWGJEQA.cxx
@@ -76,6 +76,7 @@ AliAnalysisTaskPWGJEQA::AliAnalysisTaskPWGJEQA() :
   fDoEventQA(kTRUE),
   fGeneratorLevelName(),
   fDetectorLevelName(),
+  fMCGeneratorIndex(-1),
   fRejectOutlierEvents(kFALSE),
   fIsPtHard(kFALSE),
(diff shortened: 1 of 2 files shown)


moliver813:
The check makes sure the MC particles used in the matched-tracks THnSparse (tracks matched to an MC particle) are from the MC generator of index 0 in the event. For some reason, this check was not applied to the THnSparse of MC tracks, which one would need to do to calculate the tracking efficiency properly by dividing the matched tracks by all findable MC particles. It would make sense if it was applied to both. One related issue that came up while using this task is that for the LHC20e3 production, almost no particles had the MC generator index of 0, instead having the index of 1, which I infer means that the index 0 referred to the pileup that was included in the production, while most particles from HIJING had index 1, and were thus not included in the matched track THnSparse.
An alternative to removing this would be to add an option to the task to set which generator to use particles for determining tracking efficiency and to apply it to both the MC track and MC matched track THnSparse. This would require the user to determine which index the generator of interest is. This would indeed be useful if one wanted to use just the HIJING (or similar MB generator) particles from a production. In that case, I think the default behavior should still be to accept all generators unless the user specifies otherwise.

moliver813:
Yes, I'll add the switch option.

moliver813:
Added the switch option with the fMCGeneratorIndex variable

raymondEhlers:
+1

PR #20284, opened 2022-05-03
D0 pT weights @ 13 TeV with PYTHIA 8 and FONLL

scostanz:
Hi,
I would like to add the method SetPtWeightsFromFONLL13overLHC20f4abc() in the AliCFTaskVertexingHF class.
The function contains the values of the histogram ratio of the LHC20f4a/b/c MC with PYTHIA8 and FONLL calculations for pp data at 13 TeV using D0 simulated pT distribution.

Thanks,
Susanna

### DIFF
--- a/PWGHF/vertexingHF/AliCFTaskVertexingHF.cxx
+++ b/PWGHF/vertexingHF/AliCFTaskVertexingHF.cxx
@@ -79,12 +79,19 @@
 #include "AliPIDResponse.h"
 #include "AliAnalysisUtils.h"
 #include "AliGenEventHeader.h"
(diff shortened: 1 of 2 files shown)


xinyepeng:
@laszlogyulai Hi Laszlo, with Susanna, we find that the wagons can't be ran simultaneously in a same train with your commitment. As the time is limited, we suggest to add her new commitment excluded yours, and let you fix it later. We hope it's fine from your side, thanks!

xinyepeng:
+test

xinyepeng:
+test

PR #20290, opened 2022-05-03
Added a variable to track the mother fraction and removed the cut so t…

JColburn-CERN:
…hat cut can be done after analysis. Add fixd a bug with protons being identified.

### DIFF
--- a/PWGJE/EMCALJetTasks/AliAnalysisTaskJetChargeFlavourTemplates.cxx
+++ b/PWGJE/EMCALJetTasks/AliAnalysisTaskJetChargeFlavourTemplates.cxx
@@ -44,7 +44,17 @@ AliAnalysisTaskJetChargeFlavourTemplates::AliAnalysisTaskJetChargeFlavourTemplat
   fCentMax(10),
   fJetRadius(0),
   JetChargeK(0.5),
-
+  Pt(-999),
+  Phi(-999),
+  Eta(-999),
+  JetCharge(-999),
+  ParticlePt(-999),
+  ParticlePhi(-999),
+  ParticleEta(-999),
+  ParticleJetCharge(-999),
+  LeadingTrackPt(-999),
+  PdgCode(0),
+  ProgenetorFraction(-1),
 
   fTreeJets(0)
 {
@@ -62,7 +72,17 @@ AliAnalysisTaskJetChargeFlavourTemplates::AliAnalysisTaskJetChargeFlavourTemplat
   fCentMax(10),
   fJetRadius(0),
   JetChargeK(0.5),
-
+  Pt(-999),
+  Phi(-999),
+  Eta(-999),
+  JetCharge(-999),
+  ParticlePt(-999),
+  ParticlePhi(-999),
+  ParticleEta(-999),
+  ParticleJetCharge(-999),
+  LeadingTrackPt(-999),
+  PdgCode(0),
+  ProgenetorFraction(-1),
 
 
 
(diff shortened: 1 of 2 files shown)


PWGJE/EMCALJetTasks/AliAnalysisTaskJetChargeFlavourTemplates.cxx
```diff
@@ -44,7 +44,6 @@ AliAnalysisTaskJetChargeFlavourTemplates::AliAnalysisTaskJetChargeFlavourTemplat
   fCentMax(10),
   fJetRadius(0),
   JetChargeK(0.5),
-  MotherFraction(0.8),
```
raymondEhlers, line 47: Even if you're not going to set it via your add task, you need to initialize `ProgenetorFraction` in your constructor

PWGJE/EMCALJetTasks/AliAnalysisTaskJetChargeFlavourTemplates.cxx
```diff
@@ -63,7 +62,6 @@ AliAnalysisTaskJetChargeFlavourTemplates::AliAnalysisTaskJetChargeFlavourTemplat
...
```
raymondEhlers, line 66: Same here - even if you're not going to set it via your add task, you need to initialize `ProgenetorFraction` in your constructor

raymondEhlers (changes requested):

PWGJE/EMCALJetTasks/AliAnalysisTaskJetChargeFlavourTemplates.cxx
JColburn-CERN: Hi Raymond, Maybe im being stupid and missing something here, But since The ProgenetorFraction is a varible for the tree output, (ie, Pt, Phi, Jet Charge) then it does not need to be initalized in the same way as MotherFraction was.
Is the confusion here becuase MotherFaction is varible set at runtime and used in the code and progenaator faction is its replacement but is instead an output which i can make a cut on after the train has run? Or am i just incorrect
Thanks for the help!

PWGJE/EMCALJetTasks/AliAnalysisTaskJetChargeFlavourTemplates.h
```diff
@@ -66,13 +63,13 @@ class AliAnalysisTaskJetChargeFlavourTemplates : public AliAnalysisTaskEmcalJet
   Float_t ParticleJetCharge;      //
   Float_t LeadingTrackPt;         //
   Int_t PdgCode;                  //
+  Float_t ProgenetorFraction;     //
```
raymondEhlers, line 66: You've defined `ProgenetorFraction` as a member of your class, so you have to initialize it in your constructor. It has nothing to do with ROOT - this is just standard c++

PWGJE/EMCALJetTasks/AliAnalysisTaskJetChargeFlavourTemplates.cxx
raymondEhlers: Hi Sammy, yeah, you need to initialize it - see my new review below. I wanted to mark it at the constructor, and the `MotherFraction` line was just the easiest place

JColburn-CERN:
Hi Raymond, Thanks  for the advice I have added the initlisation of progenator faction as well as a few other varibles which were missing to the constructor. Ive also made a small change to where some later varibles are initlised since it was causing a bug with meanlessly high pdg codes when jets were empty.
Cheers!

raymondEhlers:
+1

PR #20298, opened 2022-05-04
Restoring version from Jan24

scostanz:
Hi,
I would like to restore the version of AddTastCFVertexing.C from Jan24, since with the latest changes it seems that wagons can't be ran simultaneously in a same train.
Thanks,
Susanna


xinyepeng:
+1

PR #20312, opened 2022-05-06
dowang[PWGCF] not record bad vertex & Nothing changed except AliFemtoSimpleAnalysisOnlyMixP1.cxx

dowangFDU:
Some test outputs were deleted


erogocha:
+1

PR #20319, opened 2022-05-06
[PWG-CF] CMWESE 18 Syst

wenyaCern:
This version aims to do the event-wise syst. estimation and  fill the NUA input files for 18q/r


yzhouCERN:
+1

PR #20323, opened 2022-05-07
To correct the response matrix

PaolaVT:
The filling of the response matrix was modified to exclude neutral particles


aortizve:
+test

PR #20327, opened 2022-05-08
Removed the output  to terminal used for debugging accidently left in.

JColburn-CERN:


raymondEhlers:
+1

For the future, you could try to use something like AliDebugStream and friends to allow you to log (or not) without needing to edit the code

PR #20377, opened 2022-05-17
To correct the response matrix

PaolaVT:
A lock was added so that the response matrix would have contributions of only primary and charged particles


aortizve:
+test

aortizve:
+1

PR #20384, opened 2022-05-17
Modify the Event Selection parts and add a counter for HM[0,100]

JaeYoonCHO:
I modified the Event Selection parts to take the HM event into accounts. 
In previous version of code, at the beginning of the event selection stage, non-MB events are cut off, so I couldn't get HM triggered events much.
And I added one more counter to store the HM [0,100] events.


xinyepeng:
+test

xinyepeng:
+1

PR #20388, opened 2022-05-18
Add optional pt cut on charged and cluster constituents in SD

mfasDa:
Add cut on the pt of the track
constituents and on E of the
cluster constituents (optional)
applied before reclustering the
jet to extract the SD observables.
Cuts can be varied independently.
Affect only the SD params, not
the jet energy.


mfasDa:
+test

mfasDa:
+1

PR #20408, opened 2022-05-20
HFjet - find MC daughters recursively

yatowoo:
Fix issue in the response matrix of Ds-jet.

In `AliHFJetFinder::FindMCJets`, the daughters of HF candidates are read and stored from AliAODMCParticle by GetNDaughters() and GetDaughterLabel(i). But in the source code of `AliAODMCParticle`, it only deals with __2 prongs__ decay case. So for Ds to KKpi, the daughters are phi meson and pion, which means the __Kaon pair is not removed from MC particle array in generator level.__

Comparison with local running on test datasets. Use __relative difference of jet pT between detector reco. and gen. level.
image

### DIFF
--- a/PWGHF/treeHF/AliHFJetFinder.h
+++ b/PWGHF/treeHF/AliHFJetFinder.h
@@ -61,6 +61,7 @@ class AliHFJetFinder : public TObject
   std::vector<AliHFJet> GetJets(TClonesArray *array);
   std::vector<AliHFJet> GetMCJets(TClonesArray *array);
   void FindJets(TClonesArray *array, AliAODRecoDecayHF *cand=nullptr, Double_t invmass=0);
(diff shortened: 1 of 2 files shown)


PWGHF/treeHF/AliHFJetFinder.h
```diff
@@ -61,6 +61,7 @@ class AliHFJetFinder : public TObject
   std::vector<AliHFJet> GetJets(TClonesArray *array);
   std::vector<AliHFJet> GetMCJets(TClonesArray *array);
   void FindJets(TClonesArray *array, AliAODRecoDecayHF *cand=nullptr, Double_t invmass=0);
+  void FindMCdaughters(TClonesArray *array, AliAODMCParticle *mcpart, std::vector<Int_t>& daughter_vec);
```
vkucera, line 64: For consistency, I would suggest calling it `FindMCDaughters`.

yatowoo: Thanks, corrected.

PR #20415, opened 2022-05-20
remove superfluous warning messages to decrease log size

jsaetre:
Removing warning generated for each MC AOD track


iarsene:
+1

PR #20420, opened 2022-05-20
Adds switch for removing MC pileup using AliAnalysisUtils

moliver813:
Adds switch for removing MC pileup using AliAnalysisUtils
This works as a better way of removing pileup than the generator index option.
Confirmed to work with MCs (with pileup) tagged to 15o pass2 and 18qr pass3


raymondEhlers:
+1

PR #20445, opened 2022-05-25
To correct true multiplicity distribution

PaolaVT:
Added a flag to avoid noises in the distribution of true multiplicity


aortizve:
+test

aortizve:
+1

PR #20466, opened 2022-05-31
Mean pT Analysis: Minor change in the class and macro

tutripat:
There is a minor change in some condition in the class. Please consider the newest pull request with the given title.


erogocha:
+1

PR #20479, opened 2022-06-01
Implementation of Spherocity Quantiles

jackal1-66:
First Implementation of spherocity quantiles using Splines for the D mesons vs multiplicity analysis. Check attached macro to obtain the splines and run first the task without isS0Spline enabled in order to have the data to compute them.


xinyepeng:
+test

xinyepeng:
+1

PR #20498, opened 2022-06-03
PWG/EMCAL: Add option to apply Eta dependent cell scale

jokonig:
- Modules behind TRD support structure are more affected by energy loss
and photon conversions in material in front of them
- In energy calibration all cells were treated the same and were
calibrated to one value
- To test the effect of properly calibrating these cells to an expected
value (lower pi0 mass peak as heavily affected by conversions), this new
eta dependent scale is introduced.
- The scale factors are stored in the SMDependent scale factor vector as
the SM dependnt scale factors are not applied in that case
- Highly experimental and should not be used for analysis at this point!


gconesab:
+test

gconesab:
+1

PR #20510, opened 2022-06-07
Update of the AodTrackFilter task to be ROOT6 compatible

ChrisPliatskas:
This task is needed for the application of tracking efficiency on the track collection level rather than the jet collection level.


ChrisPliatskas:
Hi Raymond,
 Thank you for the detailed comment! Indeed I was also reluctant about the hardcoded run periods, however I had the (perhaps false) impression that the filter bit numbers didn't really change for the later periods. Since I want to get at least an estimate on the effect of tracking efficiency on my analysis I would suggest to proceed with the task as it is now so that a run with the LEGO trains would be available. In the meantime, I will look into re-writing the task with your suggestions.

raymondEhlers:
+1

I also think the filter bits are more consistent for later periods, but if I recall correctly, I think there are still some exceptions, so it's still good to check (see [here](https://github.com/alisw/AliPhysics/blob/4ccb32f999f05754fc4d313c3d9c804f23b030ec/PWG/EMCAL/EMCALbase/AliEmcalTrackSelectionAOD.cxx#L212), for example. Taking a quick look, it looks like everything in Run 2 is consistent)

PR #20521, opened 2022-06-08
Fix default OADB container in case of pp 13 TeV

mfasDa:
Condition overwrote custom settings of the OADB
container with "oadb", setting of specific OADB files
was therefore ignored.

In addition: Set default case for the OADB container
also for MC pp 13 TeV - this will have an effect only
in case the user specified loading the FastOR mask
of dead or bad FastORs explicitly. This will not be set
to default to the moment since it requires the trigger
maker and trigger selection task to run in data,
otherwise results will be inconsistent


mfasDa (commented):
+1

PR #20522, opened 2022-06-08
Removed minimal booking of kaons tracks for syst purity

rfacen:
MinimalBooking of Kaons cuts set to false, to get the histograms and compute purity for systematics
TrackPosKaonCuts->SetMinimalBooking(false);
TrackNegKaonCuts->SetMinimalBooking(false);


sartozza:
+1

PR #20549, opened 2022-06-13
Run 3 conversion: Do not store index -1 if no mothers

vkucera:
This PR fixes the list of mothers in the McParticle 001 format introduced in https://github.com/alisw/AliPhysics/pull/19655.

The conversion task now produces an empty array for particles with no mothers instead of an array of length 1 containing index -1.
The change consists of additional lines 1579-1580. The rest are just trailing spaces automatically removed by the editor.

@jgrosseo @pzhristov , please have a look.


jgrosseo:
+1

PR #20557, opened 2022-06-14
[EMCAL-783] Make reduced trigger payload optional

mfasDa:
Reduced trigger payload must be optional for run1 and
run2 data, otherwise it will be impossible masking bad
TRUs offline.


vkucera:
@mfasDa Please fix, it does not compile.
```
AliAnalysisTaskAO2Dconverter.h:56:56: error: use of undeclared identifier 'fEMCALTriggerReducedPayload'
```

PR #20586, opened 2022-06-20
To modify the binning of PbPb collisions

PaolaVT:
A new parameter was added in order to modify the binning of PbPb collisions.


aortizve:
+test

aortizve:
+1

PR #20607, opened 2022-06-23
Add values for Ps analysis systematics

JakeMartinez:
adding values for certain cuts to be able to pass them as parameters on the train


mario-krueger:
+1

PR #20650, opened 2022-06-30
Adding entry for Lambda(1520)

vikasLstar:
This commit is for adding an entry for Lambda(1520) in these macros.


mario-krueger:
+1

PR #20664, opened 2022-07-05
update variable for systematic parameters

JakeMartinez:
make sure input parameter for proton PID doesn't conflict with MC parameter


bdoenigu:
+1

PR #20718, opened 2022-07-13
fixed a bug and testing if the train test would be successful

asasikum:
Fixed a bug related to the initialization of histograms. Testing to see if the train test would be successful.


lhusova:
+1

PR #20735, opened 2022-07-15
Modify the event tree

JaeYoonCHO:
I masked out the event tree to reduce the output file size.
For now, It is somehow hard-corded, but I will fix it soon.


xinyepeng:
+1

PR #20750, opened 2022-07-19
Bug fix on looping over all jets and task clean up

ChrisPliatskas:
In the previous version, there was a bug in the loop over the list of accepted jets where it would only look for a match and calculate energy flow if there were more than one accepted jets in the event. This PR also includes a short cleanup of the task from some old histograms which are no longer used.


ChrisPliatskas:
Hello, it would be very helpful for me to have this fix included in the next AliPhysics tag since I need it to start some new train runs. Could someone have a look for the review check please?

raymondEhlers:
+1

PR #20760, opened 2022-07-20
Automatized rho handling in JetContainer

mfasDa:
- Load rho values automatically if container links are
  specified
- Raise error message in case no jet container is linked
  in the rho task

### DIFF
--- a/PWGJE/EMCALJetTasks/AliAnalysisTaskRho.cxx
+++ b/PWGJE/EMCALJetTasks/AliAnalysisTaskRho.cxx
@@ -42,7 +42,7 @@ Bool_t AliAnalysisTaskRho::Run()
     fOutRhoScaled->SetVal(0);
 
   if (!fJets){
-    AliErrorStream() << "No jet container attached container" << std::endl;
+    AliErrorStream() << "No jet container attached" << std::endl;
     return kFALSE;
   }
 
--- a/PWGJE/EMCALJetTasks/AliAnalysisTaskRho.cxx
+++ b/PWGJE/EMCALJetTasks/AliAnalysisTaskRho.cxx
@@ -41,8 +41,10 @@ Bool_t AliAnalysisTaskRho::Run()
   if (fOutRhoScaled)
     fOutRhoScaled->SetVal(0);
 
-  if (!fJets)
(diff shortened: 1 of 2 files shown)


mfasDa:
@raymondEhlers Please review!

raymondEhlers:
It's been a while since I've loaded rho this way. At first glance, it looks fine. However, it appears that it also shouldn't be needed for standard rho at least - see https://github.com/alisw/AliPhysics/blob/6015b235c21bc8b9d00af9a764be1fb58f7bb32d/PWG/JETFW/AliAnalysisTaskEmcalJet.cxx#L166-L171 ? I suppose you just want to centralize it here?

mfasDa:
@raymondEhlers I see. In principle it should work but the construction is a bit weird (and by the way handles only one rho type) - why not delegating this all to the jet container in set array, it knows what it subscribes to.

mfasDa:
+test

PWGJE/EMCALJetTasks/AliAnalysisTaskRho.cxx
```diff
@@ -41,8 +41,10 @@ Bool_t AliAnalysisTaskRho::Run()
   if (fOutRhoScaled)
     fOutRhoScaled->SetVal(0);
 
-  if (!fJets)
+  if (!fJets){
+    AliErrorStream() << "No jet container attached container" << std::endl;
```
raymondEhlers, line 45: Small typo - "container" here twice

raymondEhlers:
> @raymondEhlers I see. In principle it should work but the construction is a bit weird (and by the way handles only one rho type) - why not delegating this all to the jet container in set array, it knows what it subscribes to.

Yeah, I think it's fine to move this into `SetArray`. Worst case, it's set twice, which I don't think has any side effects

mfasDa:
@raymondEhlers Yes, that's what I thought as well. OK, since I missed the deadline for today I will fix the typo. However the error message must appear here, we have some case where the run function silently exits and we have no clue why and suspect this line.

raymondEhlers:
Yeah, the error message is good - thanks for adding it

mfasDa:
+1

PR #20780, opened 2022-07-22
PWG/EMCAL: Adding MC Rejection for EPOS/other background

aschmier:
PWG/EMCAL: Adding macro for MC Rejection for EPOS/other background

### DIFF
--- a/PWG/EMCAL/EMCALtasks/AliEmcalRejectMCBackground.h
+++ b/PWG/EMCAL/EMCALtasks/AliEmcalRejectMCBackground.h
@@ -1,0 +1,186 @@
+/**************************************************************************************
+ * Copyright (C) 2022, Copyright Holders of the ALICE Collaboration                   *
+ * All rights reserved.                                                               *
+ *                                                                                    *
+ * Redistribution and use in source and binary forms, with or without                 *
+ * modification, are permitted provided that the following conditions are met:        *
+ *     * Redistributions of source code must retain the above copyright               *
+ *       notice, this list of conditions and the following disclaimer.                *
+ *     * Redistributions in binary form must reproduce the above copyright            *
+ *       notice, this list of conditions and the following disclaimer in the          *
+ *       documentation and/or other materials provided with the distribution.         *
+ *     * Neither the name of the <organization> nor the                               *
+ *       names of its contributors may be used to endorse or promote products         *
+ *       derived from this software without specific prior written permission.        *
+ *                                                                                    *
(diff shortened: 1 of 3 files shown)


FriederikeBock:
+test

PWG/EMCAL/EMCALtasks/AliEmcalRejectMCBackground.h
```diff
@@ -0,0 +1,186 @@
...
+  AliNamedArrayI           *fParticlesMap;         //!<! particle index/label
+  AliVEvent                *fEvent;                //!<! event
+  AliMCEvent               *fMC;                   //!<! MC event (ESD)
+  Bool_t                    fIsESD;                //!<! ESD or AOD analysis
+  Bool_t                    fDisabled;             //!<! Disable task if a problem occurs at initialization
+  Int_t                     fDebugLevel;           ///< debug level for interactive debugging
+  Int_t                     fnHeaders;             ///< Number of Headers
+  std::vector<int>          fNotRejectedStart;     //[fnHeaders]
+  std::vector<int>          fNotRejectedEnd;       //[fnHeaders]
+  std::vector<TString>      fGeneratorNames;       //[fnHeaders]
```
mfasDa, line 175: Attention: having [] in comments is actually a ROOT extension for dynamic arrays reserving space (in this case of size fnHeaders) in the streamer. Since you are using a std::vector you don't make use of that, better remove, though without pointer it will probably be ignored by rootcling.

PWG/EMCAL/macros/AddTaskRejectMCBackground.C
```diff
@@ -0,0 +1,55 @@
+AliEmcalRejectMCBackground* AddTaskRejectMCBackground(
```
mfasDa, line 1: This macro should in principle just call the static add function you implemented in the analysis task. Here you basically duplicate the code.

mfasDa (changes requested):
Particularly the constants and the rootcling directives must be removed before merging. The rest is cosmetic stuff.

mfasDa (changes requested):
Code wise it is fine now but you should have squashed the two commits into one. Normally github has a functionality that codeowners (corresponding to approvers in this case) can do it via the webinterface - which is how it is usually done in O2, but the AliPhysics project in github is configured in such an idiotic way that - unlike for O2 for example - all the useful extensions are inaccessible to the reviewers which can only approve or reject commits. So sadly I cannot squash-merge via the webinterface and you have to do it via the command line.

mfasDa:
+1

PR #20810, opened 2022-07-28
To modify the choice of V0

PaolaVT:
The choice of V0 for multiplicity estimation for p-Pb collisions was modified.


aortizve:
+1

PR #20812, opened 2022-07-29
Add tag for D+ -> Kpipi in Ds tree

fgrosa:
@stefanopolitano please have a look


stefanopolitano (approved):
Hi @fgrosa! Thanks for the job, it seems fine to me.

PR #20831, opened 2022-08-03
Fixed the way 100 GeV track cut behaves

osaarima:
Previously: Remove over 100 GeV tracks from event.
Now: Remove jets with over 100 GeV constituents.


dongjokim:
+1

PR #20856, opened 2022-08-08
Modified how the code uses event selection

osaarima:
Now the code can reject detector MC and still run true MC (or vice
versa) if the event selection happens to indicate so, and previously
it was not possible.


dongjokim:
+1

PR #20869, opened 2022-08-10
A minor fix for event selection

osaarima:
If we find jet pt>4*pt_hard we want to discard the whole event,
including the true MC.


dongjokim:
+1

PR #20894, opened 2022-08-16
lower case table names

jgrosseo:


jgrosseo:
+1

jgrosseo:
@pzhristov O2 excepts lower case and max 16 characters.

PR #20922, opened 2022-08-25
test starange train error

11kumaoka:
Now I am facing a strange error on the train run. So I return my task to most simple one


raymondEhlers:
+1

I don't know what "strange error" on a train run means here, but in case you don't already know, odd train issues often  have to do with either memory leaks or streaming issues

PR #20939, opened 2022-08-30
[PAG/Flow] flag to use bootstrap

zhaomr13:
When using bootstrap, the memory consumption is large. We need to reduce the pt binning of trigger particles to 1 (0.2-3.0) to ensure the bootstrap works.


yzhouCERN:
+1

PR #21008, opened 2022-09-14
 PWGGA/GammaConv :  leading particle study

AMechler:
implementing a meson categorization w.r.t the particle with the hight pt (leading) in the event

Added MesonLeadTrackSelection function to test it additionaly to is meson selected.
It was not possible to include it in IsMesonSelected because no event is availible here for data

### DIFF
--- a/PWGGA/GammaConvBase/AliConversionMesonCuts.cxx
+++ b/PWGGA/GammaConvBase/AliConversionMesonCuts.cxx
@@ -44,6 +44,7 @@
 #include "AliAODMCParticle.h"
 #include "AliDalitzAODESDMC.h"
 #include "AliDalitzEventMC.h"
+#include <tuple>
 
 class iostream;
 
@@ -186,7 +187,9 @@ AliConversionMesonCuts::AliConversionMesonCuts(const char *name,const char *titl
   fEnableOmegaAPlikeCut(kFALSE),
   fDoGammaMinEnergyCut(kFALSE),
   fNDaughterEnergyCut(0),
-  fSingleDaughterMinE(0.)
+  fSingleDaughterMinE(0.),
+  fInLeadTrackDir(0),
+  fLeadTrackMinPt(4.)
 {
   for(Int_t jj=0;jj<kNCuts;jj++){fCuts[jj]=0;}
   fCutString=new TObjString((GetCutNumber()).Data());
@@ -312,7 +315,9 @@ AliConversionMesonCuts::AliConversionMesonCuts(const AliConversionMesonCuts &ref
   fEnableOmegaAPlikeCut(ref.fEnableOmegaAPlikeCut),
   fDoGammaMinEnergyCut(kFALSE),
   fNDaughterEnergyCut(0),
-  fSingleDaughterMinE(0.)
+  fSingleDaughterMinE(0.),
+  fInLeadTrackDir(ref.fInLeadTrackDir),
+  fLeadTrackMinPt(ref.fLeadTrackMinPt)
 
 {
   // Copy Constructor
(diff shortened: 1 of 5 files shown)


PWGGA/GammaConvBase/AliConversionMesonCuts.cxx
```diff
@@ -5047,3 +5073,72 @@ Bool_t AliConversionMesonCuts::IsParticleInJet(std::vector<Double_t> vectorJetEt
...
+    return kFALSE;
+  }  
+  if(!Event) {
+    AliError("Event not available\n");
+    return kFALSE;
+  }
+
+  Double_t PtLead = 0;
+  Int_t NTotTrack = 0;
+  Double_t Selection = TMath::Pi()/3.; // 120 grad
```
mhemmer-cern, line 5098: This is 60° not 120°

PWGGA/GammaConvBase/AliConversionMesonCuts.cxx
```diff
@@ -5047,3 +5073,72 @@ Bool_t AliConversionMesonCuts::IsParticleInJet(std::vector<Double_t> vectorJetEt
...
   return particleInJet;
 }
+
+
+    
+
+///_____________________________________________________________________________
+//   Function to find highest pT track in event
+template <class T, class U>
+Bool_t AliConversionMesonCuts::MesonLeadTrackSelection(T Event, U meson)
```
mhemmer-cern, line 5083: If I understand this correctly you will use this function to select Mesons that are in one of the three areas regarding the leading pT, so that one cut setting will only select a single one?

PWGGA/GammaConvBase/AliConversionMesonCuts.cxx
```diff
@@ -5047,3 +5073,72 @@ Bool_t AliConversionMesonCuts::IsParticleInJet(std::vector<Double_t> vectorJetEt
...
+    //  transwerse to lead track direction
+    } else if(fInLeadTrackDir == 2){
+      if ( AngleToLeadTrack > Selection/2 && AngleToLeadTrack < Selection ) return kTRUE;
+    //  opposite direction to lead track (away)
+    } else if(fInLeadTrackDir == 3){
+      if ( AngleToLeadTrack > Selection && AngleToLeadTrack < TMath::Pi() ) return kTRUE;
+    //  in direction and opposite direction
+    } else if(fInLeadTrackDir == 4){
+      if (( AngleToLeadTrack > Selection/2 && AngleToLeadTrack < Selection ) || ( AngleToLeadTrack > Selection && AngleToLeadTrack < TMath::Pi() )) return kTRUE;
+    }
```
mhemmer-cern, line 5140: The angle you get from TVector3 Angle should be the geometric angle between the two 3D vectors, not the phi vector as you state in the comment. I assume you meant the phi vector because afaik in Kristina's thesis she just cut on phi but ignored eta. So you would need to compare T3Vector::Phi() for both particles I think.

PWGGA/GammaConvBase/AliConversionMesonCuts.cxx
AMechler: Thanks I am missing a *2

PWGGA/GammaConvBase/AliConversionMesonCuts.cxx
AMechler: Exactly, thats the idea

PWGGA/GammaConvBase/AliConversionMesonCuts.cxx
AMechler: thats correct, I am changing it to be only phi so we can better relate to that analysis

jokonig:
+1

PR #21014, opened 2022-09-15
Update AliAnalysisTaskAR

ariedel-cern:
- fix GetPointers method to include the list of normalized symmetric cumulants


abilandz:
+1

PR #21022, opened 2022-09-16
[PAG-Flow] Change pt dependent Trigger particle number to Mult dependent

zhaomr13:
To ensure a correct normalization


lhusova:
+1

PR #21029, opened 2022-09-20
QA plots added for AN

rfacen:
Uncommented lines: 

if (suffix == "0") {
       config->SetPtQA(true);
       config->SetMassQA(true);
       config->SetkTBinning(true);
       config->SetMultBinning(true);
       config->SetmTBinning(true);
    }


sartozza:
+1

PR #21031, opened 2022-09-20
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/clean-test.yml
+++ b/.github/workflows/clean-test.yml
@@ -22,7 +22,7 @@ permissions: {}
 jobs:
   cleanup_pr_checks:
     permissions:
-      repository-projects: write # to add to projects (octokit/graphql-action)
+      pull-requests: read # to get last commit for pr (octokit/graphql-action)
 
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/clean-test.yml
+++ b/.github/workflows/clean-test.yml
@@ -18,8 +18,12 @@ on:
         required: true
         default: 'AliPhysics'
 
+permissions: {}
 jobs:
   cleanup_pr_checks:
+    permissions:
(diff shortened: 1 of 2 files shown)


sashashura:
> Hi @sashashura, thanks for your contribution! I think we'll need `statuses: write` in `clean-test.yml` as well, for [`set-github-status` (source linked)](https://github.com/alisw/ali-bot/tree/master/set-github-status) used here:
> 
> https://github.com/alisw/AliPhysics/tree/3bc8b91f615a48bf4c90ec0a318563eb4ff70360/.github/workflows/clean-test.yml#L64
> 
> Have you tested these changes in any way, or should I do that?

You are totally right. I missed that. Fixed.
It is hard to me test workflows as the release.yml. I would appreciate if you could do that.

TimoWilken:
OK, fair enough! It works for me; merging. Thanks again for the PR!

PR #21032, opened 2022-09-20
Set the FromBackgroundEvent flag

pzhristov:


pzhristov:
@jgrosseo The flagging of the background events is now done for ESD, while for AOD I used a "proxy" call since no explicit flag exists. The code works correctly in case of AliEn or local files when the AliAnalysisManager finds the BKG folder. In case of partial local copy when only the "signal" part is copied and the BKG subfolder is missing, we will get again the E-TArrayC messages about the "shifted" indexes.

PR #21037, opened 2022-09-21
Added possibility to filter out EPOS particles.

osaarima:
This feature is designed to be used with AliAnalysisTaskSVtaskMCFilter which will take the default "mcparticle" and "tracks" lists for true MC and detector MC information, and will filter out the particles and tracks with EPOS generator as their origin. These are inserted into new lists "FilteredParticlesTrueMC" and "FilteredTracksDetMC" which are read by the AliJCatalystTask when user gives the flag AliJCatalystTask::FLUC_EXCLUDE_EPOS as an argument.


dongjokim:
+1

PR #21063, opened 2022-09-26
Removed the overcounting of shared daughters of neutral kaons

asasikum:
Added if statements to avoid the overcounting of shared daughters of neutral kaons in K0-K0 correlation functions, Removed a typo in the axis title of histogram fHistK0Phi.


lhusova:
+1

PR #21070, opened 2022-09-27
[PWG-CF]Adding new chi^2 cut

aonnerst:
Adding a new chi^2 cut that differ from the default cut with 30%, this cut will be used as trial for systematic uncertainty evaluation.


dongjokim:
+1

PR #21080, opened 2022-09-30
Use L1 timesums in the calulation of the L0 patch ADCs

mfasDa:
L0 patches also use the timesums for the calculation of
the ADC value. The timesums are calculated at in the
TRU and sent to the STU. Since the timesums at L0 are
not available in the analysis objects the timesums at L1
serve as proxy. The only issue the user has to keep in
mind is that when shipping the timesums from the TRU
to the STU a bit shift from 12 to 10 bits is performed in
order to fit the timesums into the ALTRO format (10 bits).
When working with L1 timesums users need to divide
the L0 threshold by 4.


mfasDa:
+test

PR #21085, opened 2022-10-04
Add tagging effi functions for 2018 HF MC sample

deepathoms:
Adding new functionalities to calculate pi0 and eta weight and tagging efficiency using 2018 Pb-Pb anchored HFe MC sample. 
- The order of MC generators in the MC header is different in the 2018 Pb-Pb anchored MC sample compared to older MC samples.
- Old order of embedded signals: MB, pi0, eta, cele/bele; New order: cele/bele, pi0, eta, MB 
- The ordering information is used to select particles from embedded pi0 and eta sample for tagging efficiency calculation


sashingo:
+1

PR #21096, opened 2022-10-06
Added a seperate analysisTask similar to AliAnalysisTaskJetChargeFlavourTemplates 

JColburn-CERN:
however this one is specialized towards using it with Pb-Pb embedding. As changes have to be made relating to how the particle containers are handled


raymondEhlers:
+1

PR #21117, opened 2022-10-11
Updated jet pt matching

osaarima:
Previously the same det MC jet could match to a true jet multiple times. This is a problem in the very low true jet range, under 5 GeV. Now there is a limit that det mc jet can be matched only once. Added a couple of histograms as well.


dongjokim:
+1

PR #21124, opened 2022-10-12
PWGHF: Adding Lc histos and simplification of the Lc selection code

DelloStritto:
- Adding Lc daughters impact parameters histos.
- Simplification of the Lc selection code. Now it is consistent with the O2 workflow.


lvermunt:
+1

PR #21127, opened 2022-10-13
PWGHF: Apply the event selections before the MC matching

DelloStritto:
Apply the event selections before the MC matching.
Now the event selection is applied before filling the generated candidates histograms (as in O2)
@lvermunt


lvermunt:
+1

PR #21135, opened 2022-10-14
Saving generator info for no Triggered events

amaringarcia:


PWGGA/GammaConv/AliAnalysisTaskMaterialHistos.cxx
```diff
@@ -979,6 +1101,13 @@ void AliAnalysisTaskMaterialHistos::UserExec(Option_t *){
     if(eventNotAccepted){
       // cout << "event rejected due to wrong trigger: " <<eventNotAccepted << endl;
       hNEvents[iCut]->Fill(eventNotAccepted); // Check Centrality, PileUp, SDD and V0AND --> Not Accepted => eventQuality = 1
+      if (eventNotAccepted==3 && fIsMC > 0){
```
fjonasALICE, line 1104: Hi Ana, I am not familiar with your task, but do you normally process no MC gen level particles at all if the event is not accepted? We always even run even the normal ProcessAODMCParticles() loop e.g. in task AliAnlysisTaskGammaCalo.cxx. Otherwise the efficiency does not include the trigger efficiency. I am not sure if your commit aims to fix this problem or is just too bookkeep why a rejection took place

amaringarcia: I want to book keep what was produced at the generator level but that we do not see because we do not look at those events. This is why I fill the info in separate histograms

fjonasALICE: Ok, thanks, this part I understand. My point is that i think you should  run the ProcessAODMCParticles() when the event is not accepted, like we do in the other GA tasks. If you don't look at these events the calculated efficiency will be slightly wrong since the trigger efficiency is not accounted for (probably no problem for MB, but still). If you want I can approve the request as it is, just wanted to make you aware. Let me know

amaringarcia: For GammaConv task we always do it like that. Even now the calo task does it in that way. and the loses are taken into account as signal_loss correction . We discussed this in the 13 TeV analysis. Otherwise the signal loss is corrected blindly and it is generator dependent.  Yes, thanks for pointing this up, but please approve the commit :-)

fjonasALICE: thanks for the clarification :-) just wanted to be sure

fjonasALICE:
+1

PR #21140, opened 2022-10-16
Efficiency tasks for different PID methods

druggian:
The different files contain the efficiency activities in which the exclusive PID method is used and no double counting method. Furthermore, the efficiency activity which uses three different filter bits such as 16, 128, and 768 has been added.


wrzesaCERN:
+1

PR #21160, opened 2022-10-20
PWGHF: Simplify D0 selection to match O2Physics

saganatt:
Analogous change to what Luigi did for Lc in https://github.com/alisw/AliPhysics/pull/21124


xinyepeng:
+1

PR #21171, opened 2022-10-20
Open Fast simulation parameter file correctly in grid environment

MartinSchmidtTuebingen:


raymondEhlers:
I don't go through the rest of your codebase, but just to double check - do you load this one time during your run macro and then stream it in the train.root (or equivalent) to avoid extra load on the grid? (as opposed to loading it n every job separately). Thanks!

PWGJE/EMCALJetTasks/UserTasks/AliPieceWisePoly.cxx
```diff
@@ -149,7 +150,10 @@ double AliPieceWisePoly::Eval (double x, double* p) {
 }
 
 void AliPieceWisePoly::ReadFSParameters(TString parameterFile, TF1** effFunctions) {
-  TFile* f = new TFile(parameterFile.Data(), "READ");
+  if(parameterFile.Contains("alien://") && ! gGrid)
+    TGrid::Connect("alien://");
+  
+  TFile* f = TFile::Open(parameterFile.Data(), "READ");
```
mfasDa, line 156: JAliEn-ROOT I think requires

```
TGrid::Connect("alien");
```
What you have (`TGrid::Connect("alien://")`) I think only worked for legacy AliEn

mfasDa (changes requested):

mfasDa:
In addition to what was commented by @raymondEhlers I would like to add that the train managers strongly advised us to move such runtime parameter into the OADB in order to load them from cvmfs and not from alien, because jet trains are obviously the trains with the largest amount of file handles which slows down the file catalogue. For testing and debugging it would be OK to load files from the grid but once validated they should go to the OADB.

PWGJE/EMCALJetTasks/UserTasks/AliPieceWisePoly.cxx
raymondEhlers: Just as an fyi, it seems to still work in the embedding helper: https://github.com/alisw/AliPhysics/blob/0f0caeafbd1cea78059270550ca5a3feab7f3f24/PWG/EMCAL/EMCALbase/AliAnalysisTaskEmcalEmbeddingHelper.cxx#L68

mfasDa:
+test

PWGJE/EMCALJetTasks/UserTasks/AliPieceWisePoly.cxx
MartinSchmidtTuebingen: I took it from AliTPCPIDResponse where it works with it, but I can change it.

MartinSchmidtTuebingen:
Thanks for your replies. This parameter file is (for now) specific to my analysis, I am not sure how I can push it to OADB. But:
1) The parameters are copied to the analysis task in the AddTask macro
2) As far as I understand it, the resulting task is then copied for every job, the 
3) This means, the file is not loaded in every job but only once - right?

raymondEhlers:
+1

Yeah, sounds like it's only downloaded once during the train generation, so it should be in good shape - thanks!

PR #21182, opened 2022-10-23
Updated the code with QA histograms

asasikum:
Added phi and pt distributions for the neutral kaon particles and also for the pion daughters of each reconstructed kaons. Also, using a wider pT range. (0.2 to 0.5 --> 0.2 to 1.0)


atimmins:
+1

PR #21185, opened 2022-10-23
PWGGA/GammaConv: Added opening-angle dependent omega unsmearing

nstrangm:
- Added a 4th unsmearing option (No Corr, Fixedpz, Subpi0 and now SubLambda)
  - This subtracts the error of the reconstructed pi0 multiplied by a factor lambda from a Toy MC
  - This lambda comes as a function of alpha and has been verified by the 3d histogram from the previous commit
- Made all minv-pT histograms optional with a AddTask argument to reduce memory
- Corrected axis label from previous commit

### DIFF
--- a/PWGGA/GammaConv/macros/AddTask_GammaConvNeutralMesonPiPlPiMiNeutralMeson_CaloMode_pPb.C
+++ b/PWGGA/GammaConv/macros/AddTask_GammaConvNeutralMesonPiPlPiMiNeutralMeson_CaloMode_pPb.C
@@ -45,6 +45,7 @@ void AddTask_GammaConvNeutralMesonPiPlPiMiNeutralMeson_CaloMode_pPb(
 
   //parse additionalTrainConfig flag
   Int_t trackMatcherRunningMode = 0; // CaloTrackMatcher running mode
+  TString unsmearingoutputs = "012"; // 0: No correction, 1: One pi0 mass errer subtracted, 2: pz of pi0 corrected to fix its mass, 3: Lambda(alpha)*DeltaPi0 subtracted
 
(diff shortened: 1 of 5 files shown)


PWGGA/GammaConv/macros/AddTask_GammaConvNeutralMesonPiPlPiMiNeutralMeson_CaloMode_pPb.C
```diff
@@ -40,6 +40,7 @@ void AddTask_GammaConvNeutralMesonPiPlPiMiNeutralMeson_CaloMode_pPb(
     Int_t     runLightOutput              = 0,                        // run light output option 0: no light output 1: most cut histos stiched off 2: unecessary omega hists turned off as well
     Int_t     prefilterRunFlag            = 1500,                     // flag to change the prefiltering of ESD tracks. See SetHybridTrackCutsAODFiltering() in AliPrimaryPionCuts
     Bool_t    enableSortingMCLabels       = kTRUE,                    // enable sorting for MC cluster labels
+    TString   unsmearingoutputs           = "012",                    // 0: No correction, 1: One pi0 mass errer subtracted, 2: pz of pi0 corrected to fix its mass, 3: Lambda(alpha)*DeltaPi0 subtracted
```
fjonasALICE, line 43: i would prefer an implementation where we do not add an additional argument to the add task. Also using a string of numbers like this is not the most intuitive in my opinion. I suggest an implementation via the additionalTrainConfig string. See for example line 93 in CaloMode_pp where the settings for the outlier rejection are passed. This way one could pass an option like UNSMEARING:1 or 0 or 2 etc in the additionalTrainconfig string

fjonasALICE:
+1

PR #21187, opened 2022-10-24
Added on-the-fly check and CosPABachBar parameter to Cascades

JStaa:
Added a check for rejecting on-the-fly V0s/cascades as well as the parameter CosPABachBar for data rejection.


aohlson:
+1

PR #21218, opened 2022-10-26
Efficiency task for pPb and PbPb collisions

druggian:
New efficiency tasks to estimate systematic uncertainties with different PID methods using FB128 by default for pPb and PbPb collisions


wrzesaCERN:
+1

PR #21226, opened 2022-10-27
Restore functionalities from the 5th of April's commit

romainschotter:
- Restore functionalities from the 5th of April's commit
- Remove huge tables


mario-krueger:
+1

PR #21229, opened 2022-10-27
Update of the jet energy flow task to handle the analysis of embedded Pythia events

ChrisPliatskas:
Update of the jet energy flow task to handle the analysis of embedded Pythia events. The main difference is the introduction of an enum object to describe the type of analysis which replaces the previous MC flag. This affects the switch of scaling the output AliEmcalList.


raymondEhlers:
+1

(note that you have a typo "embeded" -> "embedded", but it doesn't matter for functionality, and I'm trying to get you in for the tag, so will just leave it)

PR #21251, opened 2022-10-31
Tuned the jet pt match algo and added pythia info

osaarima:
Now if there is an overlap of several true jets with a single detector jet, the detector jet is not matched multiple times.


dongjokim:
+1

PR #21256, opened 2022-11-01
Propagate updates from O2vertexer in O2 into AliPhysics

lvermunt:
PR to get the O2Vertexer in AliPhysics up to date with the current one in O2. In particular updates regarding the covariance matrix were missing in the AliPhysics version, making it difficult to validate analyses that apply selections based on this (e.g. D meson analyses in HF).

I tried to keep this PR as simple as possible (instead of propagating tens of new header files that O2 now has into AliPhysics). Regarding this I put some considerations below.

@mpuccio, @vkucera, let me know what you think.

Cheers, Luuk

----------------------------------
**Consideration w.r.t. current (01-11-22) O2 version**

General:
 - I didn't split the MathUtils class, since this would require a lot of new header files to be propagated to AliPhysics. Instead I kept the original class (adding two additional functions), and I didn't propagated the changes in O2 regarding the split MathUtils class into `DCAFitterN.h`, `HelixHelper.h`, `Track.h`, and `Primitive2D.h`

In DCAFitterN:
 - I didn't propagate the parts related to `Propagator.h` and `o2::base::Propagator::MatCorrType`. I don't think this is required for Ali-O2Physics comparisons.
 - I didn't propagate the parts related to `mRefitWithMatCorr`. I had some issues with the `restore()` call (which is new), and I don't think this is required for Ali-O2Physics comparisons.
 - For the same reason as above, I commented out `trc.invalidate()` in `getTrackParamAtPCA()`
 - I didn't separate between `TrackPar` and `TrackParCov`, i.e. all `TrackPar` types are changed to `TrackParCov` (as was also done before) and I didn't include the `createParentTrackPar()` function, only the `createParentTrackPar()`

### DIFF
--- a/PWGLF/NUCLEX/Utils/O2vertexer/DCAFitterN.h
+++ b/PWGLF/NUCLEX/Utils/O2vertexer/DCAFitterN.h
@@ -18,6 +18,7 @@
 #include <TMath.h>
 #include <Math/SMatrix.h>
 #include <Math/SVector.h>
+#include "MathUtils.h"
 #include "Track.h"
 #include "HelixHelper.h"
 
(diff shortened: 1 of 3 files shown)


mpuccio:
+1

PR #21258, opened 2022-11-02
Adding Dstar systematics in pp collisions at 13TeV full sample analysis

sjaelani28:
Adding systematics on Dstar analysis in pp at 13 TeV full sample 2016, 2017, and 2018 combined.


xinyepeng:
+1

PR #21260, opened 2022-11-02
Adjustment of the arguments in the AddTask macro and minor typo fixes

ChrisPliatskas:
The main change concerns the AddTask macro, when it comes to the AnalysisType argument. In this version it is explicitly stated that it is part of the Energy Flow task. There are also minor typo fixes in the kEmbedded enum type as it was  pointed out in a previous PR.


raymondEhlers:
+1

PR #21267, opened 2022-11-02
PWGGA/GammaConv Fix for unsafe pointer issue

sstiefel19:
Up to now pointers to the aod mc track array were obtained only once for the first event of an instance of the analysis task. Since it is not guaranteed that the address of this track array stays constant over the lifetime of an analysis task this can produce unexpected behaviour.
With this change the pointer is reset for every event.


fjonasALICE:
discussed PR offline with Stephan. Looks ok to me. He will report back results and then this can potentially be implemented for all GA tasks

fjonasALICE:
+1

PR #21268, opened 2022-11-02
Correctly store and pass fast simulation parameters

MartinSchmidtTuebingen:
Train fails because functions are not copied directly. Store them as variables so we dont have to deal with copying functions


raymondEhlers:
+1

PR #21282, opened 2022-11-04
PWGHF: Lc validation histograms

AnnalenaKa:
- Adding histograms for Lc validation
- Read Lc cuts and pt bins from json file
- Correct configurable names


RUN3/AliAnalysisTaskAO2Dconverter.cxx
```diff
@@ -1327,8 +1327,8 @@ void AliAnalysisTaskAO2Dconverter::FillEventInTF()
 
     collision.fCovXX = AliMathBase::TruncateFloatFraction(covmatrix[0], mCollisionPositionCov);
     collision.fCovXY = AliMathBase::TruncateFloatFraction(covmatrix[1], mCollisionPositionCov);
-    collision.fCovXZ = AliMathBase::TruncateFloatFraction(covmatrix[2], mCollisionPositionCov);
-    collision.fCovYY = AliMathBase::TruncateFloatFraction(covmatrix[3], mCollisionPositionCov);
+    collision.fCovXZ = AliMathBase::TruncateFloatFraction(covmatrix[3], mCollisionPositionCov);
+    collision.fCovYY = AliMathBase::TruncateFloatFraction(covmatrix[2], mCollisionPositionCov);
```
lvermunt, line 1331: Please remove this from this PR. This is an ongoing discussion among the conversion experts how to treat this swap

PWGHF/vertexingHF/AliAnalysisTaskHFSimpleVertices.h
```diff
@@ -306,6 +349,8 @@ class AliAnalysisTaskHFSimpleVertices : public AliAnalysisTaskSE
   Double_t fMaxPtDplus;                                              // D+ max pt
   Double_t fMinPtJpsi;                                               // Jpsi min pt
   Double_t fMaxPtJpsi;                                               // Jpsi max pt
+  Double_t fMinPtLc;                                                 // Lc min pt
```
lvermunt, line 352: You add persistent members, so you should increment the ClassDef

RUN3/AliAnalysisTaskAO2Dconverter.cxx
AnnalenaKa: Thanks. Indeed this should not have entered. I removed it.

lvermunt:
Hi @AnnalenaKa, my other comments:

- [ ] Disable/remove `printf("daughter ptcand_3prong  %g\n", ptcand_3prong);`. 
- [ ] After your fix of `cut3ProngPtCandMin`->`d_pTCandMin`, the `if (ptcand_3prong < 0.) {` should become `if (ptcand_3prong < fMinPt3Prong)` again.
- [ ] I think you shouldn't comment out the line `if (dist12 < fLcCuts[0][4])`. From what I understood from @DelloStritto is that this is a variable that is/can be used in the Lc selector in O2Physics, so for a fair comparison it should also be included in HFSimpleVertices. If how it is currently done is the correct way I let Luigi comment.
- [ ] Increase ClassDef
- [ ] As I commented for your O2Physics PR, would be good to also fill XZ element CovMatrix here
- [ ] As for the O2Physics PR, remove the "Lc" suffix from `d_pTCandMinLc`, `d_pTCandMaxLc`, and `pTBinsLc`. I will try to think of a better way to do the json loading. Will let you know later.

DelloStritto:
Hi @lvermunt. At the beginning the idea was to insert that variable also in O2 but in the end it was not. At the moment the column where this variable is placed corresponds to the chi2PCA variable in O2 (that we don't have in AliPhysics, instead). If the selection cuts are now taken from the json also in AliPhysics, we need to comment out this line. The workaround to ignore the chi2PCA (that is an upper cut on the maximum chi2PCA value allowed) at the moment is to use a super looser cut (putting a large value that does not reject any candidate). This would not work for the dist12 that is a lower cut (on the minimum distance allowed) and requires a value near to 0 to be uneffective.

To summarize, to be consistent, I would comment out the line as @AnnalenaKa did.

lvermunt:
+1

PR #21285, opened 2022-11-04
PWGGA/GammaConv: Fixed bug in heavy meson task for PCMEMC

nstrangm:
- Every NDMcand was deleted in MC leading to seg fault
- Introduced new variable to check, whether NDM is selected or can be deleted to solve the issue


mikesas:
+1

PR #21303, opened 2022-11-08
Fixing missing return

Steffanic:
My Run method had a path that did not return causing undefined behavior. It is now fixed.


raymondEhlers:
+1

PR #21309, opened 2022-11-09
PWGHF: Fix arrays overflow that causes bugs with flags + remove 2-vertex radius cuts absent in O2Physics

saganatt:
Removing the radius cut significantly reduces the discrepancy between AliPhysics and O2Physics secondary vertex histograms. Other radius cuts are already applied inside the DCAFitterN in both frameworks.

Arrays in the HFSimpleVertices class have fixed sizes that are smaller than the default configuration or JSON settings. At the class initialization, the values were written beyond the actual end of the arrays in the memory. This resulted in overwriting the memory of various flags and having spurious flag values. This PR increases the declared array sizes and adjusts the number of cut values.


lvermunt:
+1

lvermunt:
Sorry for that, we just checked ubuntu. We will fix!

saganatt:
Done here: https://github.com/alisw/AliPhysics/pull/21311
Let us know if you will have any further problems.

PR #21315, opened 2022-11-10
Two track cuts using a radial distance of 20cm

druggian:
The track of two particles are calculated throughout the TPC volume every 20 cm from 85 cm to 245 cm


majanik:
+1

PR #21319, opened 2022-11-10
Adding check for positive jet pt values in MC case

ChrisPliatskas:
Adjusted the analysis task to only consider the cases where the jet pt is positive after background subtraction. Previously this existed only for the Dpt histogram in the data case but such a measure is now also foreseen for the MC to handle the embedded Pythia scenario. There is also a binning adjustment for the DR and jet area histograms.


raymondEhlers:
+1 (I don't think you need the ClassDef bump, but it's also fine)

PR #21343, opened 2022-11-14
PWGGA/GammaConv: Modified lambda-unsmearing and added cutstrings

nstrangm:
- Lambda function changed to pol3
- Changed lambda parameters with EMC and EMCPCM option
- New cutstrings for pp PCMEMC+PCMPCM and pPb PCMEMC


jokonig:
+1

PR #21368, opened 2022-11-16
add event cuts

11kumaoka:


raymondEhlers:
+1

Note that AliEventCuts is already implemented in AliAnalysisTaskEmcal, so you're shadowing a ton of functionality that's already there

PR #21404, opened 2022-11-22
adjust cuts to publication 2.0

mwaelde:
adjust cuts to publication 
-Chi2 per ITS cluster 4.5->5.5
-Min ratio TPC cross row over findable 0.8->0.6


rbailhac:
+1

PR #21405, opened 2022-11-22
Update of the jet energy flow analysis task

ChrisPliatskas:
Update the jet energy flow analysis task to handle constituent subtraction as well as analysis of detector to hybrid level for the Pythia embedded case. Some new histograms also added for jet energy response and adjusted binning for some others.


raymondEhlers:
+1

PR #21421, opened 2022-11-24
Small fix one of the jet energy response histograms. 

ChrisPliatskas:
Small fix one of the jet energy response histograms to include the corrected pt for background contributions.


raymondEhlers:
+1

PR #21430, opened 2022-11-26
Updated the analysisTask to try to avoid uncommon seg error

JColburn-CERN:
And floating Point exception to attempt to add stabililty to the task on legotrains


raymondEhlers:
+1

PR #21445, opened 2022-11-29
Bug fix and event selection cuts added

imsurajprasad786:
The task is modified with some bug fixes that are related to more production of antiLambda compared to lambda particles, and required event selection cuts are added.


aortizve:
+test

aortizve:
+1

PR #21448, opened 2022-11-29
memory leak check

11kumaoka:
Now I am facing the error 137 in only train. It happens only in 90% of trains.
So now I want to check the reason one-by-one.

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskRawJetWithEP.h
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskRawJetWithEP.h
@@ -340,7 +340,7 @@ class AliAnalysisTaskRawJetWithEP : public AliAnalysisTaskEmcalJet {
 
 
     /// \cond CLASSIMP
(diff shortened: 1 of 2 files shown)


11kumaoka:
Dear managers, 
could you let me merge my request, please?

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskRawJetWithEP.h
```diff
@@ -168,7 +168,7 @@ class AliAnalysisTaskRawJetWithEP : public AliAnalysisTaskEmcalJet {
   // ### private ###############################################################
   private:
     
-    AliAODEvent* fAOD;                //!<! AOD event
+    AliAODEvent* fAOD;                /// AOD event
```
raymondEhlers, line 171: You're trying to stream a nullptr. If this works, it will only work coincidentally (I guess it will work, but I don't see how this would help)

raymondEhlers (commented):
You changed streamed fields, so you need to bump your ClassDef

raymondEhlers:
+1 since so many of your jobs are fairly, please test this carefully

PR #21471, opened 2022-12-01
PWGGA/GammaConv: Add Cutstrings and one argument to pPb Omega AddTask

nstrangm:
- Added cutstrings for MB, EG1 and EG2
- Standard cuts and cutvariations
- Added pT-dep selection window argument (copied from pp)


jokonig:
+1

PR #21481, opened 2022-12-05
Fill the small Dpt histogram in the case of analysing data.

ChrisPliatskas:
This is a small fix since the histogram was already created in the previous version but not filled.


raymondEhlers:
+1

PR #21488, opened 2022-12-06
PWGCF Add a macro that includes weights for CME calculation in GammaNonIsot…

ShiQiu1:
…ropicCorrUsingWeights, fix a bug in GammaDeltaPID task


rihanphys:
+1

PR #21494, opened 2022-12-08
Critical improvement/fix to qnvector task

moliver813:
Modifies the JEQnVector handler so that the calibration files are only loaded once, at job creation. OADBContainers are stored in task.
This removes an issue where the separate jobs running the task would individually attempt to download the event plane calibration files.


raymondEhlers:
+1

OADBContainers used to leak memory. I think this may have been fixed, but I'm not sure, so something to keep an eye on (I suppose you would have had the same problems before, so it may not change anything just loading them earlier, but just thinking of it now)

PR #21511, opened 2022-12-12
Now calibration files are loaded during add task for TaskJetQnVectors

moliver813:
Now calibration files are loaded during add task for TaskJetQnVectors

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliJEQnVectorHandler.cxx
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliJEQnVectorHandler.cxx
@@ -313,10 +313,10 @@ bool AliJEQnVectorHandler::LoadOADBCalibrations() {
         fOADBFile = nullptr;
     }
 
-    TString pathToFileCMVFNS = AliDataFile::GetFileName(fOADBFileName.Data());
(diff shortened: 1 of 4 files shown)


PWGJE/EMCALJetTasks/UserTasks/AliJEQnVectorHandler.cxx
```diff
@@ -328,11 +328,13 @@ bool AliJEQnVectorHandler::LoadOADBCalibrations() {
     if (pathToFileCMVFNS.IsNull()) {
       cout<<"loading local file "<<pathToFileLocal.Data()<<endl;
       fOADBFile = TFile::Open(pathToFileLocal.Data());
-    }
+    } */
```
raymondEhlers, line 331: Do you want to keep all of this commented out? It seems that it would only support local files now?

moliver813: Yes. The intention is that it only uses local files, which are downloaded during the addtask.

raymondEhlers: Okay - I don't quite see how this works, but I trust that you've implemented it correctly

raymondEhlers:
+1

PR #21521, opened 2022-12-14
Photons tests2

pkurash:
1) *.C  macros updated: added possibility to load PHOS tender, multiplicity and PID selection

2) In lead-lead code histograms for MC are added

3) Extra histograms in pp code

4) Several minor fixes and improvements
 
5) Readme updated

### DIFF
--- a/PWGGA/PHOSTasks/PHOS_gamma/AddTaskGammaPHOSPP.C
+++ b/PWGGA/PHOSTasks/PHOS_gamma/AddTaskGammaPHOSPP.C
@@ -9,10 +9,6 @@
 #endif
 
 AliAnalysisTaskGammaPHOSPP* AddTaskGammaPHOSPP( Bool_t isMC = kFALSE,
-                                                TString tenderOption = "Run2Tune",
-						Int_t tenderPass = 1,
-						TString nonlinearity = "Run2Tune",
-						Int_t recoPass = 1,
                                                 TString name = "GammaAnalysis")
 {
     // get the manager via the static access member. since it's static, you don't need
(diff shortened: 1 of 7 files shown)


PWGGA/PHOSTasks/PHOS_gamma/AddTaskGammaPHOSPP.C
```diff
@@ -22,6 +27,20 @@ AliAnalysisTaskGammaPHOSPP* AddTaskGammaPHOSPP( Bool_t isMC,
...
     }
+
+    //Tender & response
+    
+    const Double_t zsSimulation = isMC ? 0.020 : 0;
+ 
+    AliPHOSTenderTask *tenderPHOS = reinterpret_cast<AliPHOSTenderTask *>
+        (gInterpreter->ExecuteMacro(
+         Form("$ALICE_PHYSICS/PWGGA/PHOSTasks/PHOS_PbPb/AddAODPHOSTender.C(\"%s\", \"%s\", \"%s\", %d, %d, \"%s\", %f)", 
+                           "PHOSTenderTask","PHOStender",tenderOption.Data(), tenderPass, isMC, nonlinearity.Data(), zsSimulation)));
```
peressounko, line 38: Normally tasks like Tender or PIDresponse are called in independent wagons or config macros. Your approach will create another Tender in case if one Tender was already created in special wagon or previous instance of your task and this will result in definitely wrong results. Why do you call these wagons within your analysis task? At least, put check that Tender was not already created, but better move external wagons to confg. macros.

pkurash: Hello Dmitri,

Thank you for your comment.

I will remove these lines in my new commit

PR #21538, opened 2022-12-19
Antiprotons vs RT in p-Pb

m-rasa:
Added new task and addTask macro for antiprotons analysis vs RT in p-Pb collisions, both for data and MC.


bdoenigu:
+1

PR #21544, opened 2022-12-20
Update of the energy flow Response matrix for the embedded Pythia case

ChrisPliatskas:
Fixing the response matrix calculation to use the correct detector level Dpt for the Pythia embedded case.


raymondEhlers:
+1

(in general, if your cast, you should usually check for whether the object is null before using it)

PR #21546, opened 2022-12-21
Add alinormalization for pp Xic mult

Tao-Fang:


xinyepeng:
@Tao-Fang Hi Tao, thanks a lot for the development! However, if I'm not wrong, only one normalization counter is needed, you can find the example for non-prompt fraction vs multi analysis task: https://github.com/alisw/AliPhysics/blob/master/PWGHF/vertexingHF/AliAnalysisTaskSED0MassNonPromptFraction.cxx

Tao-Fang:
Thank you for your feedback. I want to configure the file once and run LGtrain to get the relative yield of Xic under different multiplicity, so I need alinormalization of different mult intervals. If I understand correctly, AliAnalysisTaskSED0MassNonPromptFraction.cxx may require different configurations for different mult intervals. .

> @Tao-Fang Hi Tao, thanks a lot for the development! However, if I'm not wrong, only one normalization counter is needed, you can find the example for non-prompt fraction vs multi analysis task: https://github.com/alisw/AliPhysics/blob/master/PWGHF/vertexingHF/AliAnalysisTaskSED0MassNonPromptFraction.cxx

xinyepeng:
Hi Tao, I see your point, but I would still suggest you follow my suggestion: 1. you can add several wagons to the same train with the same configuration (but a different multiplicity selection), and the result is the same as your request. 2. The main problem with your implementation is the hard code. One may ask you to make different selections in different multiplicity intervals, and the multiplicity intervals are also not configurable in your task.

Tao-Fang:
Hi Xinye
Thanks you for the suggestion, I will close the process and delete the alinormalization part, however I have other changes and I have submitted a new process

PR #21550, opened 2022-12-22
fix K0s MC tree filling w/ only true particles + fix daughter pid res…

maciacco:
…ponse + add hasTOF flag and daughters pT


bdoenigu:
+1

PR #21569, opened 2023-01-03
Add Sigma^0 hyperon analysis code

ssolokhi:
I've added several files that are required for Sigma^0 hyperon analysis. The header file contains documented variables, the implementation file is formatted to ALICE's codestyle, the README file explains the analysis layout. The code has been tested both locally and in GRID with no error occurring. An AddAnalysis... file was added to the macros/ folder.


kharlov:
Hello Sergei,
Thanks for implementing most of the coding conventions in this new merger request. There are still a few data members which names should be adjusted (remember that data members of a class should begin from "f"):

- AliAODEvent *aodEvent (rename to fAODEvent)
- TClonesArray *AODMCTrackArray (rename to fAODMCTrackArray)
- AliMCEvent *mcEvent (rename to fMCEvent)
- Double_t cElectronMass (rename to fElectronMass)
- Double_t cProtonMass  (not used anywhere, remove it)
- Double_t cSigmaMass  (not used anywhere)
- Double_t cPi0Mass  (not used anywhere)
- Double_t c (hard to search this variable in the code. Where is the light speed really needed in analysis? In HEP c=1 always. If you really need it, rename to something meaningful, searchable and starting with "f").
- Double_t primaryVtxPosX, primaryVtxPosY, primaryVtxPosZ (these variables are initialized to 0 in constructors and never redefined. From the usage one can guess that they define the primary vertex, but this vertex should be got from AOD. Correct the usage of these 3 variables. If they are needed as data members, rename to fPrimaryVtxPosX,Y,Z)
- UInt_t EventTriggers (this data member it not set to a real trigger mask which is supposed to be. Correct it of needed, and if it should be really a data member, rename to start from "f")
- Int_t aodEventCounter (this variable is not used, but incremented. Rename to start from "f" and revise its usage in the analysis).

Best regards
Yuri

ssolokhi:
Hello, Yuri!
I've deleted obsolete variables that you've mentioned in your comment and renamed other class members to comply with ALICE's coding style.

kharlov:
+1

PR #21579, opened 2023-01-06
Added Alternative output for jet charge at Higher Track cut 

JColburn-CERN:
to see if seperation from background effects gluon yields


raymondEhlers:
+1

PR #21595, opened 2023-01-10
Methods to retrieve generated vertex position in case of pileup

fprino:
These helper methods allow to get the generated position of the trigger collision in case of simulations with pileup.
By using AliMCEvent::GetPrimaryVertexPosition, a position of (0,0,0) is obtained found for all events simulated with pileup (this was spotted for the general purpose Monte Carlo productions anchored to Pb-Pb). This is because the header of the cocktail generator does not store a primary vertex position, given that different pileup events occur in different position.
A modification in AliMCEvent (inside AliRoot) will also be committed in the next days.


fprino:
+1

PR #21601, opened 2023-01-11
Update of the AOD filter task to include particle containers

ChrisPliatskas:
Updating the filter task to use jet containers for the input and the output. This change is motivated by the intention to use the task for an embedded Pythia run.
 The main changes include the attachment of an input and output container, the latter is either written to the embedded or the Pythia event. The inheritance of the task was changed from AliAnalysisTaskSE to AliAnalysisTaskEmcal. Additionally a lot of legacy code related to the run 1 periods has been commented out.


PWG/EMCAL/EMCALtasks/AliEmcalAodTrackFilterTask.cxx
```diff
@@ -137,15 +121,20 @@ AliEmcalAodTrackFilterTask* AliEmcalAodTrackFilterTask::AddTaskEmcalAodTrackFilt
...
+   if(!OutTrackName.IsNull())aodTask->AddParticleContainer(OutTrackName);
+
     Bool_t includeNoITS  = kFALSE;
     Bool_t doProp        = kFALSE; //force propagation of all tracks to EMCal
     Bool_t doAttemptProp = kTRUE;  //only propagate the tracks which were not propagated during AOD filtering
     Bool_t isMC          = kFALSE;
   
     TString runPeriod(runperiod);
     runPeriod.ToLower();
-  if (runPeriod == "lhc10b" || runPeriod == "lhc10c" || runPeriod == "lhc10d" ||
```
raymondEhlers, line 148: Unless required to make it work for run 2, I wouldn't recommend commenting out Run 1 code. This isn't especially robust code, but there are still folks doing some run 1 work, so I wouldn't recommend throwing out code that still works, even if not relevant for you

PWG/EMCAL/EMCALtasks/AliEmcalAodTrackFilterTask.cxx
```diff
@@ -247,21 +235,32 @@ void AliEmcalAodTrackFilterTask::UserExec(Option_t *)
       return;
     }
   }
-
+*/
+AliParticleContainer* OutputCont= static_cast<AliParticleContainer*>(fParticleCollArray.Last());
```
raymondEhlers, line 239: This is a strong assumption on which particle container is the embedded container. If the embedded is added first (which would be a valid choice), this won't work unexpectedly. You could loop over `fParticleCollArray` and check all containers if any of them are embedded. Or alternatively, does it really needed to be attached to the embedded event? Could it also be fine in the main event? (I didn't review closely enough to answer - just something to consider)

PWG/EMCAL/EMCALtasks/AliEmcalAodTrackFilterTask.cxx
```diff
@@ -247,21 +235,32 @@ void AliEmcalAodTrackFilterTask::UserExec(Option_t *)
...
+        if(!(event->FindListObject(fTracksOutName))){
+        event->AddObject(fTracksOut);
+        OutputCont->SetArray(event);}
+        }
+  else if (!(InputEvent()->FindListObject(fTracksOutName))) {
     InputEvent()->AddObject(fTracksOut);
+        OutputCont->SetArray(InputEvent());
   }
-
+ AliAnalysisTaskEmcal::ExecOnce();
```
raymondEhlers, line 252: This may or may not work, but it's probably a dangerous thing to do. `AliAnalysisTaskEmcal` does a lot of work behind the scenes for the user. I would strongly recommend that you change `UserExec` -> `Run` for this function to avoid being surprised by potentially broken`AliAnalysisTaskEmcal` functionality

raymondEhlers (commented):
I reviewed this a bit more carefully since this is shared code. Please take a look

(if you don't want to take these kind of things into consideration, you could also copy to a new, private task, but I wouldn't generally recommend that since it shifts the maintenance burden to just a few folks. I don't think the requested changes are too substantial)

PWG/EMCAL/EMCALtasks/AliEmcalAodTrackFilterTask.cxx
ChrisPliatskas: As far as I can tell this code does not interfere with running over run 2 so I will reinstate it. Thank you for the comment

PWG/EMCAL/EMCALtasks/AliEmcalAodTrackFilterTask.cxx
ChrisPliatskas: Indeed, there is certainly a better way to pick the correct event from the fParticleCollArray. However, isn't fParticleCollArray a member of only the filter task? Doesn't each analysis task have its own  fParticleCollArray? Because then the order of the containers for this task isn't very user-dependent since first container corresponds to the input to the filter task and second or last to the output of the filter task. This should work for both "normal" and embedded analysis. Did I misunderstand the nature of the object?

I try to avoid adding an extra case for the embedded even though that case is the main motivation behind the changes. The filtered output is assigned to a container which is written either to the main or the embedded event. Given the dependence of the jet finders to the supplied track container name, if one wants to cluster jets at the hybrid level for an embedded event then the filter task will have to be run twice ,once for the embedded and once for the main event. By having the output written in separate event records you exclude cases of overwriting/replacing tracks. Is my reasoning clear or am I missing something?

PWG/EMCAL/EMCALtasks/AliEmcalAodTrackFilterTask.cxx
ChrisPliatskas: This line is certainly needed for the code to work. At the moment I am not sure what the differences between UserExec() and Run() are but will look into it. Do they generally serve the same purpose?

ChrisPliatskas: I think I understand your comment a bit better now so I made replaced the UserExec() here with the combination of Run() and ExecOnce() similar to other modern tasks. Please have a look at commit e494991

PWG/EMCAL/EMCALtasks/AliEmcalAodTrackFilterTask.cxx
```diff
@@ -219,16 +209,140 @@ AliEmcalAodTrackFilterTask* AliEmcalAodTrackFilterTask::AddTaskEmcalAodTrackFilt
...
+    if(OutputCont->GetIsEmbedding()){
+          AliVEvent *event = AliEmcalContainerUtils::GetEvent(InputEvent(),kTRUE);
+          if(!(event->FindListObject(fTracksOutName))){
+          event->AddObject(fTracksOut);
+          OutputCont->SetArray(event);}
+          }
+    else if (!(InputEvent()->FindListObject(fTracksOutName))) {
+      InputEvent()->AddObject(fTracksOut);
+          OutputCont->SetArray(InputEvent());
+    }
```
raymondEhlers, line 227: You can simplify this by using `AliVEvent *event = AliEmcalContainerUtils::GetEvent(InputEvent(), OutputCont->GetIsEmbedding())` and then just using that event. However, it's not really critical, so I'm fine with just approving as is

raymondEhlers:
+1

PR #21602, opened 2023-01-11
PWGGA/GammaConv: Increase number of bins in mass-shift histogram

nstrangm:
- Doubled the number of mass bins for lambda estimation histogram
- Fixed collision system cut digit in cutstring


mikesas:
+1

PR #21603, opened 2023-01-11
add event plane trigger and jet leading track information

jaimenorman:
add histogram to state what event plane bin the trigger belongs to, and leading track info for recoil jets, for IRC checks


raymondEhlers:
+1

PR #21606, opened 2023-01-12
Add an histogram for efficiency in HM analysis

JaeYoonCHO:
The 3D histogram storing the spectrum of generated particle is added to calculate the efficiency for the high multiplicity triggered event analysis.
And the classdef is increased as well.


lvermunt:
+1

PR #21610, opened 2023-01-13
New Xic0 -> eXi codes, revised

Chongk:
Multiple items were updated after a discussion with @lvermunt

### DIFF
--- a/PWGHF/vertexingHF/AliAnaTaskSEXic0SL.C
+++ b/PWGHF/vertexingHF/AliAnaTaskSEXic0SL.C
@@ -1,0 +1,947 @@
+/*
+   Xic0 -> eXi analysis (* newly written, Aug. 2022)
+
+   Chong Kim
+   Pusan National University
+   kimc@cern.ch
+*/
+
+#include "AliAnaTaskSEXic0SL.h"
+ClassImp(AliAnaTaskSEXic0SL);
+
+#include "AliAnalysisManager.h"
+#include "AliAODcascade.h"
+#include "AliAODMCParticle.h"
+#include "AliAODTrack.h"
+#include "AliESDtrack.h"
+#include "AliESDtrackCuts.h"
+#include "AliESDVertex.h"
+
+#include "AliInputEventHandler.h"
+#include "AliMCEvent.h"
+#include "AliMultSelection.h"
+#include "AliNormalizationCounter.h"
+#include "AliPIDResponse.h"
+#include "AliPPVsMultUtils.h"
+#include "AliRDHFCutsXictoeleXifromAODtracks.h" //For fEvtCuts
+#include "AliVEvent.h"
+#include "AliVVertex.h"
+#include "THistManager.h" //For fHisto
+
+#include "TBranch.h"
+#include "TChain.h"
+#include "TDatabasePDG.h"
+#include "TLeaf.h"
+#include "TString.h"
+#include "TTree.h"
+
+#include <cmath>
+#include <iostream>
+#include <vector>
+using namespace std;
+
+AliAnaTaskSEXic0SL::AliAnaTaskSEXic0SL():AliAnalysisTaskSE("AliAnaTaskSEXic0SL")
+{
+	ControlAnaObjects(0);
+	ControlOutputContainers(0);
+	ResetTreeVariables();
+}//Constructor, default
+
+AliAnaTaskSEXic0SL::AliAnaTaskSEXic0SL(const char* name, const char* option):
+	AliAnalysisTaskSE(name), fTaskOpt(option)
+{
+	ControlAnaObjects(0);
+	ControlOutputContainers(0);
+	ResetTreeVariables();
+}//Constructor
+
+AliAnaTaskSEXic0SL::~AliAnaTaskSEXic0SL()
+{
+	ControlAnaObjects(1);
+}//Destructor
+
+void AliAnaTaskSEXic0SL::Terminate(Option_t *)
+{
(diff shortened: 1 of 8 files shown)


lvermunt:
Hi @Chongk, thanks for the new PR.

I had another look. The following points should still be addressed before merging:
 - Add the correct streamers in your .h file. Please have a look at https://alice-doc.github.io/alice-analysis-tutorial/analysis/streamer.html
 - The placing of `StoreEvent()` for the AliNormalizationCounters. It should be placed before the 
 ```if (fEvtVtx->GetNContributors()<cut_vtxNContributors) return; //Return 6d```
 and
 ```else if (t_pileup_MB && t_pileup_HMV0) return;```
 calls.
 - The code below doesn't work to store your 4-body contamination, since it is a chain of 2 decays (i.e. the Xi is not a direct daughter of the Xic). You should have two loops, and use the PDG ID of the Xi* particles.
```
for (int b=fMCPart->GetDaughterFirst(); b<=fMCPart->GetDaughterLast(); b++){
	if (b<0) { cout <<"ERROR: negative index found!\n"; break; }
	const int t_PDG = fMCEvt->GetTrack(b)->PdgCode();
	if (abs(t_PDG) == PDGCode_e) idxEle.push_back(b);
	if (abs(t_PDG) == PDGCode_Xi) idxXi.push_back(b);
}//b
if (idxEle.size()==0 || idxXi.size()==0) continue; //Continue
```

Then a few points for your information/later improvements:
 - You should still add an explicit condition on the number of decay daughters of the Xic decay. Just counting the number of electrons and Xi's is not enough, since then all "e Xi nu + something else" decays will be tagged as signal
 - For the efficiency calculation, I think it is easier to also store the PDG code of the electron-candidate track and Xi-candidate track at reconstructed level when analysing MC. This would be a much easier way to separate signal and background at the reconstructed level.
 - The ```if (fabs(fEvtVtxZ) > cut_vtxZ) return; //Return 9d``` selection is already done in the `AliRDHFCuts::IsEventSelected()` call.
 - I believe there is a difference between `AliMCEvent` (which is the generated event, with MC truth information) and `AliVEvent` (which is the reconstructed event). The event selection in MC should be done on the reconstructed informatin, i.e. the `AliVEvent`. There is one exception: for the effiency calculation you also need a selection on the generated z-vertex position. So your code (see below) is by accident correct :)
 ```
 fMCEvt = fInputHandler->MCEvent();
fEvtVtx = fMCEvt->GetPrimaryVertex();
fEvtVtxZ = fEvtVtx->GetZ();
if (fabs(fEvtVtxZ) > cut_vtxZ) return; //Return 7mc
```

**Comment by Chongk** [comment]
Hello @lvermunt,

Thanks for the check. I'll check those items and send another ping once they're all addressed.

**Comment by Chongk** [comment]
Hello @lvermunt,

For your comments,

- Add the correct streamer to the header: done (hopefully)
- StoreEvent() placing: done
- Address 4-body decay truth: done
- Add # of daughters of an MC truth particle: done
- Add PDG code of e/Xi candidates at reco. level: done
- Repeated "|vtxZ| < 10" cut:
   a. That item is being kept for the legacy setting
   b. Also, please note that two RDHF objects exist for MB or HMV0. IsEventSelected() will be applied offline
- Dealing vtxZ at MC: now I remember... I temporarily wrote that part and planned to update it later but forgot.
  That's a relief it's not a problem, but I'll be careful.

**Comment by lvermunt** [comment]
+test

**Comment by lvermunt** [comment]
Hi @Chongk, the build crashes seemingly because of the streamers. I don't understand the error message (see below). However, after your modifications the streamers still look different wrt other HF tasks. Can you please have a look at other tasks in `vertexingHF` folder and just do as is done there? 

In addition, there are several warnings when compiling your task. Might also be worth to fix those. You can have a look at https://ali-ci.cern.ch/alice-build-logs/alisw/AliPhysics/21610/1780b9fded30e6ffa904278df34c11d6d30062b9/build_AliPhysics_root6/fullLog.txt and look for your task name

------

```
110/116 Test  #78: load_library_PWGHFvertexingHF ....................***Failed   18.63 sec

Processing /sw/slc7_x86-64/AliPhysics/21610-slc7_x86-64-local1/test/load_library/LoadLib.C("libPWGHFvertexingHF")...
cling::DynamicLibraryManager::loadLibrary(): /sw/slc7_x86-64/AliPhysics/21610-slc7_x86-64-local1/lib/libPWGHFvertexingHF.so: undefined symbol: _ZN23AliAnalysisTaskSEXic0SL8StreamerER7TBuffer
(int) -1
```

Chongk:
Hello @lvermunt,

Will do. Hmm, nothing goes easy, I guess...

Chongk:
Hello @lvermunt,

I updated codes to address both warning messages and streamer issues.
Could you check again if it can be merged?

lvermunt:
+test

Building AliPhysics locally shows no issue? Or didn't you try? If not, please try to do so, that's easier debugging than via `+test`

Chongk:
I'm not sure about " Building AliPhysics locally ", but if you meant if it can be built with local AliPhysics I didn't.
To do that, I guess I should have a local build of AliBuild, am I correct? If the test fails continually I'll try.
But preparing that local AliBuild would take some time, I'm afraid.

However, I made sure it runs properly for the local environment, by using a sample AOD, for both data and MC.

Chongk:
Crap, not again... I'll check it.

lvermunt:
+test

lvermunt:
+1

PR #21613, opened 2023-01-13
PWGGA/GammaConv: Add cutstrings for omegas using PCM (EMCal)

nstrangm:
- Add cutstrings for PCM and PCM-EMCal for pp and pPb
- Add parameters for PCM lambda correction
- Remove unneccessary cut variations for EMCal


jokonig:
+1

PR #21629, opened 2023-01-18
changed many lines for train problem

11kumaoka:
I changed many lines to solve train problem

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskRawJetWithEP.h
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskRawJetWithEP.h
@@ -1,88 +1,51 @@
+#ifndef ALIANALYSISTASKRAWJETWITHEP_H
+#define ALIANALYSISTASKRAWJETWITHEP_H
+// ******************************************************************************************
+// \class AliAnalysisTaskRawJetWithEP
+// \brief task used to load the Qn calibrations and get the calibrated Qn vectors for JE analyses
+// \authors:
+// T. Kumaoka, takuya.kumaoka@cern.ch
+// ******************************************************************************************
 
-#ifndef AliAnalysisTaskRawJetWithEP_H
-#define AliAnalysisTaskRawJetWithEP_H
-/**
- * \file AliAnalysisTaskRawJetWithEP.h
- * \brief Declaration of class AliAnalysisTaskRawJetWithEP
- *
- * In this header file the class AliAnalysisTaskRawJetWithEP is declared.
- * This is a sample task that shows how to write a simple user analysis task
- * using the EMCal jet framework. It is also used to do automatic benchmark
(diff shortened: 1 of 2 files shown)


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskRawJetWithEP.h
```diff
@@ -1,88 +1,49 @@
...
- *
- * \author Salvatore Aiola <salvatore.aiola@cern.ch>, Yale University
- * \date Apr 27, 2016
- */
-
-/* Copyright(c) 1998-2016, ALICE Experiment at CERN, All rights reserved. *
- * See cxx source for full Copyright notice                               */
-
+#ifndef ALIANALYSISTASKEMCALJETSAMPLE_H
+#define ALIANALYSISTASKEMCALJETSAMPLE_H
```
raymondEhlers, line 2: This will interfere with the sample task, and may lead to unexpected behavior. Please return these two lines to what you had before:

```c++
#ifndef ALIANALYSISTASKRAWJETWITHEP_H
#define ALIANALYSISTASKRAWJETWITHEP_H
```

(should have been all caps by convention)

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskRawJetWithEP.h
```diff
@@ -1,88 +1,49 @@
...
+// ******************************************************************************************
+// \class AliAnalysisTaskRawJetWithEP
+// \brief task used to load the Qn calibrations and get the calibrated Qn vectors for JE analyses
+// \authors:
+// C. Beattie, caitlin.beattie@cern.ch
+// M. Sas,     mike.sas@cern.ch
```
raymondEhlers, line 8: Not as critical, but all of this is incorrect. It's better to have a correct description and correct author info

raymondEhlers (changes requested):
It's not feasible for me to review such large changes, so I just did a few checks.

It's not clear to me why you've done things like move your AddTask out of compiled code back into an interpreted macro, but if that's what you prefer, it's fine

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskRawJetWithEP.h
11kumaoka: Thank you for telling me that.  I modified these lines.

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskRawJetWithEP.h
11kumaoka: Thank you notice me that. I forgot to fix these lines. Now I modified them.

raymondEhlers:
+1

PR #21635, opened 2023-01-19
Add the possibility to save the TPC track parameters

romainschotter:
Add an option to save the AliExternalTrackParameters of tracks reconstructed using the TPC only


ddobrigk:
+1

PR #21638, opened 2023-01-20
PWGGA/GammaConv: Adjusted PCM-EMCal Lambda correction parameters

nstrangm:
- Introduced parameters from combination of pp & pPb trains/MCs


jokonig:
+1

PR #21664, opened 2023-01-27
update to finer binning of the histograms

mwaelde:
update to finer binning of the histograms
to be able to compare


dsekihat:
+1

PR #21666, opened 2023-01-27
PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx small edit to track container name handling.

aschmier:
PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx small edit to track container name handling.


mfasDa:
This commit allows to use also non-default track containers as input input containers in the jet finder. The behavior for default containers is not affected, but track containers from helper (i.e. selection) tasks can be used as well. The change makes the jet container more flexible and I see no reason not to merge the PR.

mfasDa:
+test

raymondEhlers:
For our future selves, this could cause an issue if the user creates a non-standard track collection name (for example, with filter tracks) that contains "tracks", but is not intended to have the track selection applied (note that apply the track selection twice (ie. once when doing the filtering, and then once later in the jet finder) won't be a problem)

aschmier:
This new commit allows to use also non-default track containers as input containers in the RhoPerpCone and RhoSparse tasks. The behavior for default containers is not affected, but track containers from helper (i.e. selection) tasks can be used as well. The change makes the rho tasks more flexible

PR #21673, opened 2023-01-30
PWGGA/GammaConv add trainconfigs for PbPb

sstiefel19:


PWGGA/GammaConv/macros/AddTask_GammaConvV1_PbPb.C
```diff
@@ -4015,7 +4015,25 @@ void AddTask_GammaConvV1_PbPb(
     cuts.AddCutPCM("11310023", "0d200009ab770c00amd0400000", "0152101500000000"); // 10-30%
     cuts.AddCutPCM("13530023", "0d200009ab770c00amd0400000", "0152101500000000"); // 30-50%
     cuts.AddCutPCM("15910023", "0d200009ab770c00amd0400000", "0152101500000000"); // 50-90%
-  // copy of 2515
+  // 4 more copies of 2515 copy a
+  } else if (trainConfig == 2520){
```
jokonig, line 4019: This will not work as the train config is always 2520 and as such it will only see the first case.
Could you change that?

jokonig (changes requested):

PWGGA/GammaConv/macros/AddTask_GammaConvV1_PbPb.C
sstiefel19: yes of course, oversaw that. thanks :)

jokonig:
+1

PR #21678, opened 2023-01-31
Bugfix and an updated tracking uncertainty

osaarima:
There was a crash when initializing legotrain which was fixed by implementing two functions in the cxx file instead of the header. In addition, recently a pt dependent tracking uncertainty was introduced in PWG-JE which is now implemented in this code for MB collisions.


dongjokim:
+1

PR #21681, opened 2023-01-31
Fix bug in AliNormalizationCounter.cxx and activate the multiplicity in the event counter for Xic0 analysis

zhujh126:
Fix bug with multiplicity evaluation in the function StoreEvent() and activate the multiplicity in the event counter for Xic0 -> Xi pi analysis


lvermunt:
Hi @zhujh126, thanks for spotting, this is indeed a bug. However, for your analysis (which is in V0M multiplicity bins) this will still be incorrect. The `Multiplicity()` function is given you the number of tracklets, why you need call `StoreEvent()` with the V0M percentile.

Please have a look here, you should give the multiplicity from your specific multiplicity estimator as argument: https://github.com/alisw/AliPhysics/blob/2c533a9f90f638f3a8289db119f8e0c41891d2ad/PWGHF/vertexingHF/AliAnalysisTaskSEDvsMultiplicity.cxx#L831-L841 (which is also the reason why we didn't see this bug before)

lvermunt:
+1

PR #21691, opened 2023-02-02
Xic0SL debug for Lego train run

Chongk:
Tested the effect of update in the local environment by using both data and MC sample.
It runs well and created multiple output directories in the output ROOT file, as intended.


lvermunt:
+1

PR #21695, opened 2023-02-02
PWGJE: Update to Track Skimming

bhofman:
PWGJE: To keep track of the charge of a track with track skimming without the need to store a extra int per track we multiply the track pt by -1 for tracks with charge -1. This is only enabled after explicitly setting the correct flag.

### DIFF
--- a/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskTrackSkim.cxx
+++ b/PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskTrackSkim.cxx
@@ -60,6 +60,7 @@ AliAnalysisTaskTrackSkim::AliAnalysisTaskTrackSkim()
  : AliAnalysisTaskEmcalJet("AliAnalysisTaskTrackSkim", kTRUE),
   fYAMLConfig(),
   fConfigurationInitialized{false},
+  fChargeScaling{false},
   fNAcceptedFirstTrackCollection{nullptr},
   fTriggerBitINT7{false},
   fCentralityForTree{-1},
@@ -84,6 +85,7 @@ AliAnalysisTaskTrackSkim::AliAnalysisTaskTrackSkim(const char* name)
  : AliAnalysisTaskEmcalJet(name, kTRUE),
   fYAMLConfig(),
   fConfigurationInitialized{false},
+  fChargeScaling{false},
   fNAcceptedFirstTrackCollection{nullptr},
   fTriggerBitINT7{false},
   fCentralityForTree{-1},
@@ -113,6 +115,7 @@ AliAnalysisTaskTrackSkim::AliAnalysisTaskTrackSkim(
  const AliAnalysisTaskTrackSkim& other)
  : fYAMLConfig(other.fYAMLConfig),
   fConfigurationInitialized(other.fConfigurationInitialized),
+  fChargeScaling{other.fChargeScaling},
   fNAcceptedFirstTrackCollection{nullptr},
   fTriggerBitINT7{other.fTriggerBitINT7},
   fCentralityForTree{other.fCentralityForTree},
@@ -153,6 +156,12 @@ void AliAnalysisTaskTrackSkim::RetrieveAndSetTaskPropertiesFromYAMLConfig()
     fMinCent = centRange.first;
     fMaxCent = centRange.second;
   }
+  bool chargeScaling;
+  res  = fYAMLConfig.GetProperty({ baseName, "chargeScaling" }, chargeScaling, false);
+  if (res) {
+    fChargeScaling = chargeScaling;
+    AliDebugStream(3) << "Charge scaling enabled.\n";
+  }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskTrackSkim.h
```diff
@@ -87,6 +89,7 @@ class AliAnalysisTaskTrackSkim : public AliAnalysisTaskEmcalJet
   // Basic configuration
   PWG::Tools::AliYAMLConfiguration fYAMLConfig; ///<  YAML configuration file.
   bool fConfigurationInitialized;               ///<  True if the task configuration has been successfully initialized.
+  bool fChargeScaling;                        //!<! Flag explicitly whether charge scaling is enabled.
```
raymondEhlers, line 92: Needs to stream: `///<`

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskTrackSkim.h
```diff
@@ -97,6 +100,7 @@ class AliAnalysisTaskTrackSkim : public AliAnalysisTaskEmcalJet
   float fEventPlaneV0MForTree;                //!<! V0M event plane (as a float to reduce storage requirements).
   bool fTriggerBitCentral;                    //!<! Flag explicitly whether trigger bitmap contains kCentral
   bool fTriggerBitSemiCentral;                //!<! Flag explicitly whether trigger bitmap contains kSemiCentral
+
```
raymondEhlers, line 103: Please remove this newline

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskTrackSkim.h
```diff
@@ -61,6 +61,8 @@ class AliAnalysisTaskTrackSkim : public AliAnalysisTaskEmcalJet
   // Configuration is handled via the YAML configuration file
   bool Initialize();
   void AddConfigurationFile(const std::string & configurationPath, const std::string & configName = "") { fYAMLConfig.AddConfiguration(configurationPath, configName); }
+  
+  void SetChargeScaling(bool chargeScaling);
```
raymondEhlers, line 65: Please remove this setter and set the property via `RetrieveAndSetTaskPropertiesFromYAMLConfig()`. I prefer this approach since it gives me a separate file I can used to document settings

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskTrackSkim.cxx
```diff
@@ -411,6 +419,7 @@ Bool_t AliAnalysisTaskTrackSkim::FillHistograms()
     }
 
     for (auto particle : particlesGen->accepted()) {
+      fSecondTrackCollectionKinematics["pt"].push_back(particle->Pt());
```
raymondEhlers, line 422: Seems to be here twice - this is a bug? Did you mean to do the same as for the first track collection? I would think you would want the charge at both det and particle level

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskTrackSkim.cxx
```diff
@@ -386,7 +393,8 @@ Bool_t AliAnalysisTaskTrackSkim::FillHistograms()
   }
   
   for (auto particle : particles->accepted()) {
-    fFirstTrackCollectionKinematics["pt"].push_back(particle->Pt());
+    if(fChargeScaling && particle->Charge() < 0) fFirstTrackCollectionKinematics["pt"].push_back(particle->Pt()*-1);
+    else fFirstTrackCollectionKinematics["pt"].push_back(particle->Pt());
```
raymondEhlers, line 397: Please use `{}` even for one line statements - this is only stylistic, but I strong prefer it since it can avoid bugs in the future

raymondEhlers (changes requested):
The review is nit-picky - sorry, but it comes with the territory of using my task. Once you're done updating, please squash the changes into one commit

One more request:
- Please print out the status of this new flag in `toString` under the "General settings" category

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskTrackSkim.cxx
bhofman: Indeed a bug.. my bad! The idea for particle level is to extract the charge from PDG code

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskTrackSkim.cxx
```diff
@@ -493,6 +504,13 @@ std::string AliAnalysisTaskTrackSkim::toString() const
   else {
     tempSS << "disabled.\n";
   }
+  tempSS << "Charge scaling: ";
```
raymondEhlers, line 507:
```suggestion
  tempSS << "\tCharge scaling: ";
```

raymondEhlers (commented):
Thanks for updating! One last minor change and then it's good to go

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskTrackSkim.cxx
raymondEhlers: Ah, yeah, of course

raymondEhlers:
+1

PR #21696, opened 2023-02-02
Adding TOF nsigma and an AOD filter bit 96, 16, 768

Steffanic:
I'm not sure if this is the right way to apply these filter bits. I had a hard time finding documentation on these particular filter bits. Let me know if I made a mistake.


raymondEhlers:
+1 (I have no particular insight on whether this is the right way to do it or not)

PR #21736, opened 2023-02-09
AliAnalysisTaskRawJetWithEP: Add print out message for each process t…

11kumaoka:
…o clarify the problem that happans on the train


raymondEhlers:
+1 I'm approving this for debugging, but you really should be using a [local test train instead](https://twiki.cern.ch/twiki/bin/view/ALICE/AnalysisTrains#Local_train_test). With these printouts, your task **cannot** be submitted for a full train run in this state. It will heavily pollute the logs, which can cause grid issues

PR #21738, opened 2023-02-10
pT-dependent efficiency fix for cases where the anchor production not…

jaimenorman:
… available

- Haidar reported that for the p-Pb MC productions LHC17h6e2 and LHC17h6f2 the anchored productions are not being found. Investigation found that for whatever reason the anchored productions are not available from AliProdInfo. The MC tag info is still available here.
- The option has thus been added to define the tracking efficiencies for a given MC tag, rather than a data production, in the cases where this info is not available.
- Updated yaml file


jaimenorman:
@raymondEhlers see the following pull request and fix, let me know in case you have any comments/questions on the proposal

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
```diff
@@ -1105,7 +1105,17 @@ void AliEmcalJetTask::SetArtificialTrackingEfficiencyFromYAML() {
...
   AliProdInfo prodInfo(userInfo);
   std::string period = prodInfo.GetAnchorProduction().Data();
-  AliInfoStream() << "anchor production = " << prodInfo.GetAnchorProduction()<< "\n";
+  if(period.size() == 0) {
+    // MC where anchor production is not saved - get the name of the MC production instead
+    period = prodInfo.GetTag(AliProdInfo::kProdTag).Data();
+    if(period.size() == 0) {
+      AliFatal("No information relating to anchored datset or MC tag in this MC production - can't get pT-dependent tracking efficiencies");
+    }
+    AliInfoStream() << "Get MC set tracking efficiency for " << prodInfo.GetTag(AliProdInfo::kProdTag)<< "\n";
```
raymondEhlers, line 1114: A bit pedantic, but I think it would be better to reduce the maintenance burden by using:

```suggestion
    AliInfoStream() << "Get MC set tracking efficiency for " << period << "\n";
```

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
```diff
@@ -1105,7 +1105,17 @@ void AliEmcalJetTask::SetArtificialTrackingEfficiencyFromYAML() {
...
+  }
+  else {
+    AliInfoStream() << "Get anchor production set tracking efficiency for " << prodInfo.GetAnchorProduction()<< "\n";
```
raymondEhlers, line 1117: A bit pedantic, but I think it would be better to reduce the maintenance burden by using:

```suggestion
    AliInfoStream() << "Get anchor production set tracking efficiency for " << period << "\n";
```

raymondEhlers (commented):
Couple of minor suggestions. I guess the p-Pb will be a bit brittle since any new MC would need to be added to the list, but I don't see any obvious way around it, so I think it's fine as is

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
jaimenorman: done

PWGJE/EMCALJetTasks/AliEmcalJetTask.cxx
jaimenorman: done

jaimenorman:
Thanks for the suggestions @raymondEhlers ! I just pushed the suggested updates (merging with the previous commit), you can go ahead and approve if you agree

raymondEhlers:
+1

PR #21749, opened 2023-02-12
Update of the jet energy flow task to calculate mismatches and the effect of the dR criterion

ChrisPliatskas:
This update includes two updates: a) Making the distance for the matching between jets of different size configurable by the user and b) an addition of a histogram to count the fraction of mismatches for the large R in the case of MC analysis.


raymondEhlers:
+1

raymondEhlers:
+1

PR #21774, opened 2023-02-16
Initial commit of jet hadrochemistry task from Sierra

sweyh99:
This is the Initial commit of jet hadrochemistry task by Sierra Weyhmiller and Mesut Arslandok. This analysis extracts pT-spectra of charged kaons, protons, and pions for the inclusive event and in jets. It is based on particle identification via the dE/dx signal of the TPC.

### DIFF
--- a/PWGJE/EMCALJetTasks/CMakeLists.txt
+++ b/PWGJE/EMCALJetTasks/CMakeLists.txt
@@ -36,6 +36,7 @@ include_directories(${ROOT_INCLUDE_DIRS}
                     ${AliPhysics_SOURCE_DIR}/PWG/CaloTrackCorrBase
                     ${AliPhysics_SOURCE_DIR}/PWGHF/jetsHF
                     ${AliPhysics_SOURCE_DIR}/PWGHF/vertexingHF
+                    ${AliPhysics_SOURCE_DIR}/PWGPP
                     ${AliPhysics_SOURCE_DIR}/PWGPP/vertexingHF
                     ${AliPhysics_SOURCE_DIR}/PWGPP/EVCHAR/FlowVectorCorrections/QnCorrections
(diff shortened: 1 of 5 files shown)


PWGJE/EMCALJetTasks/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ include_directories(${ROOT_INCLUDE_DIRS}
                     ${AliPhysics_SOURCE_DIR}/PWG/CaloTrackCorrBase
                     ${AliPhysics_SOURCE_DIR}/PWGHF/jetsHF
                     ${AliPhysics_SOURCE_DIR}/PWGHF/vertexingHF
+                    ${AliPhysics_SOURCE_DIR}/PWGPP
```
raymondEhlers, line 39: Since I'm not familiar with PID analyses are run, what functionality are you getting out of PWGPP? It's generally strongly discouraged to have PWG directories depend on each other since this can become a mess of circular dependencies. Obviously, we have to do this sometimes - and that's fine if so - but it would better if it can be avoided

raymondEhlers (commented):
Hi Sierra,

I don't have time right now to review in depth, but at a quick glance, there is one important question to answer, which I've highlighted in the review. Beyond that, I see that some fairly old code patterns (such as in the AddTask) and no little streaming information, which can make things difficult to maintain, but if it works, I'm inclined to let it go right now

Best,
Raymond

PWGJE/EMCALJetTasks/CMakeLists.txt
sweyh99: Hi Raymond,

Sorry, I wasn't aware that this was discouraged, but it makes sense. I was using this library for a filtered tree task and some ESD tools, but these aren't necessary for what I'm doing now. I've modified my code to not use any PWGPP dependencies. I'll close this pull request and open a new one now with the updated code.

PR #21777, opened 2023-02-17
Fixed a library related bug at SVtaskMCFilter

osaarima:
Reimplemented a function GetGenerator from AliVertexingHFUtils to the SVtaskMCFilter as the run would crash when using the function from HFUtils. I tried fixing the problem without copying it in numerous ways but unsuccesfully. Also added a histogram to the dijet code unrelated to this problem.


dongjokim:
+1

PR #21779, opened 2023-02-17
Optional no-event-selection mode

mfasDa:
Needed for service wagon which should run under
conditions as loose as possible to not create a bias
on consumer wagons which might have a different
(less strict) event selection


mfasDa:
+test

mfasDa:
Feature is needed to exclude that service wagons based on the AliAalysisTaskEmcal (trigger maker, trigger selection, jet finder, jet matcher, ... ) impose a bias on user results if the event selection in their task is different or looser. Since the service wagons have no knowledge about the event selection of consumer tasks to create a minimum overlapping subset of event cuts in such cases the service wagon should simply apply no cut and provide objects no matter whether they will be consumed or not.

raymondEhlers:
ClaasDef bump?

PR #21791, opened 2023-02-20
add print out lines to specify the reason of the train problem.

11kumaoka:
1. I add printout to specify the actual crush line.
2. I add setter to switch active functions to find error function.
3. I modified some histogram setting (not relate this train problem)


raymondEhlers:
+1

PR #21793, opened 2023-02-21
Changed single jet matching behaviour

osaarima:
Now the closest in deltaR is matched instead of highest pt to mimic other similar analysis.


dongjokim:
+1

PR #21795, opened 2023-02-21
Add feature to calculate correlation between Λ-Λ and p-p

ChunzhengLab:
Feature Expansion:

- It is now possible to calculate the correlation between #Lambda-#Lambda and p-p.

Bug Fixes:

- Fixed the issue with incorrect FilterBit settings.
- Optimized the calculation logic of DCA.
- Removed some unnecessary outputs.
- Simplified the parameter settings for AddTask.


shouqiye:
+1

PR #21798, opened 2023-02-21
Add pointer check and printout lines to specify the train problem.

11kumaoka:
I found some reference histograms by previous pull request.
But still remain the crush, so I add more printout lines.


raymondEhlers:
+1

PR #21807, opened 2023-02-22
Modified error of Event plane fanction for train problem

11kumaoka:
I add return lines. I noticed the error function does not end by return line even though it is Bool_t.


raymondEhlers:
+1

PR #21819, opened 2023-02-24
EfficiencyTaskOnfourMult

druggian:
Efficiency task to count the number of events loss after cuts on four multiplicity classes


lgraczykCern:
+1

PR #21822, opened 2023-02-24
PWGGA/GammaConv: Change omega 5TeV standard cut to include TOF

nstrangm:
- TOF (if available) set as standard for pp and pPb and all 3 methods
- pT dep. DCAxy cut included in standard cut as well
- TOF variation cutstrings added


mikesas:
+1

PR #21829, opened 2023-02-24
Added MB multiplicity INEL > 0 to S0 Quantiles

jackal1-66:
This edit allows the user to compute directly the analysis INEL > 0 in the minimum bias multiplicity interval 1 - 199


lvermunt:
+1

PR #21835, opened 2023-02-27
Add flag to store like-sign pairs; remove "weights" and add "pt_gen" in rec. tree; Add more pre-selections and MLmap into Xic0 Pb-Pb analysis task

zhujh126:
Add flag to store like-sign pairs; remove "weights" and add "pt_gen" in rec. tree; Add more pre-selections and MLmap into Xic0 Pb-Pb analysis task


arossi81:
+1

PR #21839, opened 2023-02-27
Added LambdaHadronRatio task (and corresponding AddTask macro)

rhanniga:
Adding my Lambda/Hadron ratio analysis task so I can run trains during these trying times (aka HP2023 preparation)


rhanniga:
Hi everyone! Given that HP is right around the corner I wanted to start running trains ASAP, so I'm "bump"ing this PR to make sure it gets seen! (I know it's ALICE week, I just don't want my code to be left in the dust...)

lhusova:
+1

PR #21841, opened 2023-02-28
Remove event tree in analysis class and add suffix of centrality and pT for output name in AddTask

zhujh126:
Remove event tree in analysis class and add suffix of centrality and pT for output name in AddTask


arossi81:
+1

arossi81:
+1

PR #21855, opened 2023-03-02
PWGGA/GammaConv: fix typo in tree imprementation and change photon cu…

EwaGlimos:
…tnumber to event+photon cutnumber


FriederikeBock:
+1

PR #21889, opened 2023-03-08
Added D0 in spherocity calculation

jackal1-66:
This is a test to reduce the bias obtained in the closure test of D mesons analyses. The D0 is the only neutral particle that we are including in the calculation, while D* and D+ are included already.


lvermunt:
+1

PR #21897, opened 2023-03-09
Add PV infomration in the QA output tree

zhujh126:
Add PV infomration in the QA output tree (mass constraints for Lambda and Xi, topological constraints of Lambda to Xi, Xi to primary, Xi_c to primary).


lvermunt:
+1

PR #21899, opened 2023-03-10
cleaned the code before adding MC part

asasikum:
*Cleaned the code before adding the MC part
*Added functions for filling correlations
*Changed the filter bit from 272 to 96
*Removed kpos-kpos and kneg-kneg correlation functions as the focus is on kpos-kneg and k0-kch correlations as if now.


lhusova:
+1

PR #21903, opened 2023-03-10
Recalculate PV by removing daughters and adding Xic+ in the QA tree

zhujh126:
Recalculate PV by removing daughters and adding Xic+ in the QA tree (mass constraints for Lambda and Xi, topological constraints of Lambda to Xi, Xi to primary, Xi_c to primary).


lvermunt:
+1

PR #21909, opened 2023-03-13
physics selection fix

dushmanta97:
I have tried to fix the issue with physics selection for which the train was not running.


hmurakam:
+1

PR #21928, opened 2023-03-15
Change the number of loop to initialize array

11kumaoka:
Last change make another crush at test train. It successes my mac and lxplus, but local train fail. 
The crush message does not say specific reason.


raymondEhlers:
+1

PR #21931, opened 2023-03-16
Corrected classification for V0M for signal loss and event loss

antpaz:
Signal loss and event loss events where wrongly added in the V0m classification histograms


aortizve:
+1

aortizve:
+1

PR #21994, opened 2023-03-31
Included event mixing for background

asasikum:
Event mixing is implemented for calculating the background for the correlation function. Testing the method on entire dataset as it does not seem to give a flat distribution in deltaPhi while running on a sample of events.


lhusova:
+1

PR #22012, opened 2023-04-03
PWGGA/GammaConv: Change cutstrings for omega 5TeV

nstrangm:
- Reverted the strict DCA cut for primary pions
- Changed DCA variation accordingly


jokonig:
+1

PR #22015, opened 2023-04-03
PWGGA/GammaConv: Add omega background and TOF QA histograms

nstrangm:
- Add a histogram for the two charged pion mass with one coming from an omega
- Change vertical axis on TOF QA from some time to the velocity (beta)
- Add the final histogram for TOF QA
- Evened out previously very asymmetric binning of TPC and TOF histograms


jokonig:
+1

PR #22020, opened 2023-04-04
Fix of memory leak in the calculation of the mismatches fraction

ChrisPliatskas:
Fix of memory leak in the calculation of the mismatches fraction introduced in the last update.


raymondEhlers:
+1

PR #22021, opened 2023-04-04
setting mass bias for event selection for correlation

dsarkaralice:


dsarkaralice:
Setting mass bias for event selection for correlation estimation

EmilGorm:
+1

PR #22032, opened 2023-04-07
Adding a new task for HF-jet tagging

hahassan7:
A new calls for tagging the HF-jet has been implemented. It uses mainly two tagging algorithms, the IP and SV algorithms.


dongjokim:
+1

PR #22047, opened 2023-04-12
PWGGA/GammaConv: Add TOF Kaon and Proton rejection option to omega task

nstrangm:
- Add rejection option, when TOF signal close to Kaon or Proton
- Add Cutstrings for TOF and DCA variations


jokonig:
+1

PR #22055, opened 2023-04-13
Cuts updated. Added signal and Event Loss for Flatenicity

antpaz:
Cuts updated to values indicated by Omar.   Some duplicated cuts now disabled
Added my implementation of Signal and Event Loss calculation for flatenicity classes.  Needed to generate values to implement Closure test.   Will be later compared with implementation that Omar has to provide.


aortizve:
+test

aortizve:
+1

PR #22061, opened 2023-04-14
AliAnalysisTaskNewJetSubstructure - inclusion of EP calibrations

caitiebeattie:
Inclusion of ability to use event-plane angle calibrations in substructure task. (Some spacing alignment was also done.)


raymondEhlers:
+1

PR #22062, opened 2023-04-14
added more options to downsize output, MC capabilities

sweyh99:
I tested building AliPhysics with these changes locally, and it was successful.


raymondEhlers:
+1

PR #22063, opened 2023-04-15
added event mixing background for k+K- correlation

asasikum:
Added event mixing background for K+K- correlation, Centrality 20-30% instead of MinimumBias, Charged kaon tracks selection cut changed


atimmins:
+1

PR #22065, opened 2023-04-15
Updated code with K+K- background

asasikum:
Added event mixing background for K+K- correlation, Track selection cut for charged kaon with TOF info, Centrality changed to 20-30%


atimmins:
+1

lhusova:
+1

PR #22092, opened 2023-04-19
treeHF changes for particle charge

nambrath:
including particle charge information during tree production

@jdmulligan


jdmulligan:
Thank you for these updates @nambrath, it looks good to go from my side. 

Can one of the approvers please have a look at approval? These changes only affect the particletree (which is off by default) by adding some additional information -- and a few general fixes encountered in the sample run macro.

jdmulligan:
Hi @fcolamar @ginnocen, 

Who would be the right person to review/approve this request? We are hoping to produce a couple new HF trees with an additional variable in this PR.

Thank you!

PR #22116, opened 2023-04-25
[PWGEM-13] PWGGA/GammConv: Add new EMCal and PCM-EMCal cuts for PbPb …

mhemmer-cern:
…5TeV

- Add the latest greatest EMCal and PCM-EMCal cuts from the pp 13 TeV Analysis, except track matching which is turned off
- Goal is to rerun all methods for 2015 and 2018 again with the MBW and compare to the previous results


mikesas:
+1

PR #22121, opened 2023-04-26
add bool flag for sys unc cut variation

mwaelde:
add bool flag for sys unc cut variation to create a smaller output


rbailhac:
+1

PR #22123, opened 2023-04-26
Updated prameterisation for pPb cockltail

hscheid:
New set of parameterisations for pPb cocktail. Based on charged pion measurements.


rbailhac:
+1

PR #22134, opened 2023-04-27
[PWGEM-13] PWGGA/GammaConv: PbPb Changed EMCal cuts

mhemmer-cern:
- In the last PR the track matching was set off and the timing cut was changed to pp. This PR should fix these two issues.


mikesas:
+1

PR #22147, opened 2023-04-30
AliAnalysisTaskNewJetSubstructure: rearranged constructor

caitiebeattie:
Rearranged constructor to match order in header file. Done to reduce warnings from task when building AliPhysics.


raymondEhlers:
+1

PR #22148, opened 2023-05-01
PWGGA/GammaConv: Add omega background and DCA histograms

nstrangm:
- Add Armenteros plots for all pi+pi- and those from the HNM
- Add pi+pi- mass histogram for the pair closest to the rho mass, with one pion form the HNM
- Add DCA xy and z histograms for primary pions from rho, kaon and omega
- Implement TOF Kaon-Proton rejection in all 5TeV omega cutstrings


FriederikeBock:
@nstrangm I'll approve this, but we'll have to coordinate with Ewa that we get this under the correct QA flags for her commit (which should come tomorrow), as that one cleans up the handling of the flags in the task and enabling of hists.

FriederikeBock:
+1

PR #22157, opened 2023-05-02
AliAnalysisTaskNewJetSubstructure: addition of EP angle matching 

caitiebeattie:
Addition of event-plane angle matching at truth-level. Some spacing adjustments.


raymondEhlers:
+1

PR #22170, opened 2023-05-04
Update of the jet energy flow task to include the following changes

ChrisPliatskas:
Update of the jet energy flow task to include the following changes:
a) Option to include a leading track pt cut for jets (set up as a class data member) b) Moving from absolute to relative jet energy response histograms c) Updated Dpt binning so that it is centered at zero d) Update of the mismatch histograms to absolute counts instead of ratio and be pt differential e) New RM type histogram for the mismatch cases


raymondEhlers:
+1

PR #22179, opened 2023-05-05
PWGGA/GammaConv: New phi acceptance cuts for PHOS

Steffimro:
- for QA studies the outer PHOS modules are excluded, therefore new phi ranges are defined (4.55 - 5.24)

- added TrainConfigs for new cuts


jokonig:
+1

PR #22288, opened 2023-05-24
Xic0SL: debugged the part causing segfault for pA

Chongk:
Minor debugging to avoid segfault during the pA run


lvermunt:
+1

There is no need to increase the ClassDef though. This only has to be done if you add a persistent member

PR #22305, opened 2023-05-26
[PWGEM-13] PWGGA/GammaConv: Add EMC cut settings for PbPb

mhemmer-cern:
- Add cut setting for rotation background with TM
- Add cut setting for mixed event without TM
- Add cut setting for mixed event with TM

- Without TM is for the usage of the cell track matcher and with track matching for when the cell track matcher is disabled
- Goals:
	- Comaprison of rotation background with template fits and mixed event background
	- Comparison of TM and Cell TM performance and agreement between data and MC


mikesas:
+1

PR #22306, opened 2023-05-26
[PWGEM-13] PWGGA/GammaConv: Fix same config name from PR #22305

mhemmer-cern:
- in the PR #22305 three train configs had the same name


jokonig:
+1

PR #22341, opened 2023-06-04
Implemented MB trigger, updated track cuts and added track cuts' histograms

Archita-Dash:
Hi @raymondEhlers @fkrizek @nzardosh 

In this commit, I have implemented the MB trigger selection, updated the track cuts and some histos for the applied track cuts in the jet validation task. These changes have been made to match with the latest modifications done in the corresponding jet validation task in O2Physics ([AliceO2Group/O2Physics#2541](https://github.com/AliceO2Group/O2Physics/pull/2541)) . Once the PR is accepted, I hope when we run the Jet Validation task locally (which is still under development), this would improve the comparisons between the Run2 and Run3 data than the last time.


raymondEhlers:
+1

PR #22347, opened 2023-06-05
Fixed issue with event selection

alphaflores:
Fixed issue with selecting MB and cent trigger candidates. This should only select cent trigger for now.


sashingo:
+1

PR #22348, opened 2023-06-05
Fixed combinatorial factors for the 3 point correlator, added respons…

anabananana98:
…e matrix hist for  jet pT


raymondEhlers:
+1

Be a bit careful here - you don't specify streamer information for most of your histograms. Will most likely be coincidentally okay, but if it something goes wrong, it will be frustrating to debug. Probably would be worth checking with local folks about how to set them correctly

PR #22351, opened 2023-06-06
Added streamer info for histograms, fixed scaling on EEC and sym E3C

anabananana98:


raymondEhlers:
I don't actually see streamer info here - you should be looking to add `///<` and `//!<!` on the class members in your header

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskJetsEEC.h
```diff
@@ -125,66 +125,66 @@ Bool_t fDoFillEncMC;      ///< to fill the matched mc plane
...
+  TH1D *pt_tru; ///< all tru level
+  TH1D *pt_tru_match; ///< tru level matched
+  TH1D *pt_det; ///< all det level
+  TH1D *pt_det_match; ///< det level matched
   
-  TH1D *test_hist; //test histogram
+  TH1D *test_hist; ///< test histogram
   
-  TH2D *R_matrix;
+  TH2D *R_matrix; ///< response matrix for jet pt
```
raymondEhlers, line 180: Are you sure that you want to stream the hist pointers? Streaming hist pointers is often more trouble than it's worth, especially since they reinitialize just fine. To be concrete, I would suggest `//!<!` for your hist pointers unless you have a particular reason

anabananana98: Ah I see. No, there isn't a particular reason (I was following what I saw in other analyses). I will make the changes and recommit. Thanks for the suggestion!

anabananana98: Sorry, for some reason the changes didn't commit. I'll try again.

anabananana98: Re-committed with the suggested changes.

raymondEhlers:
+1

PR #22359, opened 2023-06-08
Improver file for testing ITS3 with worsen (5%) resolution to mimic i…

arossi81:
…mpact of deadzone (mimics tracks w/o a point only on 1st layer).


arossi81:
+1

PR #22363, opened 2023-06-09
Fixed issue with switching EMCal clusters, tracks, and had info

alphaflores:
Needed to include these functions in AddTask macro to switch on and off EMCal clusters, tracks, and had info for systematic runs


sashingo:
+1

PR #22375, opened 2023-06-13
Changed binning for histograms to match EEC measurement

anabananana98:


raymondEhlers:
+1

just fyi - if you don't change the streamers (which you haven't here), you don't have to bump the ClassDef

PR #22379, opened 2023-06-14
PWG/EMCAL Adding pPb 8.16 TeV production LHC23d7a/b to configuration options.

aschmier:
PWG/EMCAL Adding pPb 8.16 TeV production LHC23d7a/b to configuration options.


FriederikeBock:
+1

PR #22388, opened 2023-06-16
Improved memory issue due to THnSparse

jackal1-66:
There were issues due the size of the THnSparse. This became clear only when running on the full dataset with the latest addition.


fcatalan92:
+1

PR #22390, opened 2023-06-16
Optimized task in hadron cut

alphaflores:
Improved some optimization within the task involving the hadron selection cut.


fcolamar:
+1

PR #22398, opened 2023-06-19
fix reset of flags resulting into no positive-charge particles being …

danielbattistini:
…selected for mc truth reco


otonvd:
+1

PR #22404, opened 2023-06-21
Merged the last multiplicity bins to match the 7TeV analysis

matarzil:
Combined Multiplicity and Sphericity Event Shape analysis; merged the last multiplicity bins to match the 7TeV analysis


mario-krueger:
+1

PR #22414, opened 2023-06-22
Multiplicity Analysis at mid rapidity for pp @ 13TeV using Tracklets

alekhutson:
Uploading 2 analysis for data and for simulation to generate a multiplicity distribution for INEL events in pp collisions


valzacc:
+test

PR #22418, opened 2023-06-23
PWG/EMCAL: Add option for cell scale for run1 w/wo TRD

jokonig:
- For run1, TRD modules with/witout TRD modules in front might need to be treated differently, as they have more/less material in front and as such more/less conversions etc.
- Instead of just having the eta dependent scale activated for run2, a total of 4 options for the cell scale is available now: 1 = default (Scales for Full, 2/3 and 1/3 SM), 2 = diff. scales for regions behind TRD support, 3 = Run1 scales behind SM with TRD in front and without, 4 = Same as 3 but additionally values for with and without TRD support structures
- These options can be set in the yaml file via fAdditionalScaleMode
- In order not to break existing analyses, the variable fUseAdditionalScale still has to be setto true if one wants to use a cell scale (and the variable has to stay a bool as its not evaluated to 1 if one sets it to true)


jokonig:
@FriederikeBock please have a look

FriederikeBock:
+1

PR #22426, opened 2023-06-26
added jet area cut, fixed small output, changed z vertex and pileup cuts

sweyh99:
added jet area cut, fixed small output option, changed z vertex cut and pileup cuts


raymondEhlers (commented):
Only took a quick look, but seems fine (could add comfigurability for z vertex since it's built into the base class but it rarely changes, so probably fine). Only thing is that you need to bump the ClassDef since you added streaming class memebers

sweyh99:
Good catch! I've incremented the ClassDef.

PR #22427, opened 2023-06-26
[PWGJE/EMCALJetTasks/UserTasks/]: Remove EP angle from pp MC running.

hjbossi:
event plane angle is not useful for pp MC and was causing a seg fault. this commit fixes that issue


raymondEhlers:
+1

PR #22440, opened 2023-06-29
Adding some corrections to flatten the event plane angle.

Steffanic:


PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetHdEdxCorrelations.cxx
```diff
@@ -1495,6 +1501,45 @@ namespace PWGJE
...
+  Double_t AliAnalysisTaskEmcalJetHdEdxCorrelations::GetFlattenedEPAngle(Double_t uncorrectedAngle){
+     if (!gGrid)
+      {
+        TGrid::Connect("alien://");
+      }
+        // Open the ROOT file in read mode
+      TFile file("alien:///alice/cern.ch/user/p/psteffan/epCorrections.root", "READ");
+
+      // Read the TTree from the ROOT file
+      TTree *tree = dynamic_cast<TTree *>(file.Get("V0C"));
```
raymondEhlers, line 1513: It appears that this will be opened in every job and read every event. If true, that's incredibly expensive, and will cause huge load on the grid, so it needs to be fixed.

The file doesn't appear to depend on the event, so it should be opened during the AddTask phase and are stored in variables which are streamed. If it for some reason needs to be opened separately job-by-job, then it should be in `UserCreateOutputObjects` or `UserExecOnce`

raymondEhlers (changes requested):

Steffanic:
@raymondEhlers 

Thank you so much for catching that! I really appreciate it. I think it should be ok now as I access the file in the addtask and close it in the destructor.

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetHdEdxCorrelations.h
```diff
@@ -244,6 +251,7 @@ namespace PWGJE
       Double_t fMinSharedMomentumFraction;    ///< Minimum shared momentum with matched jet
       bool fRequireMatchedPartLevelJet;       ///< True if matched jets are required to be matched to a particle level jet
       Double_t fMaxMatchedJetDistance;        ///< Maximum distance between two matched jets
+      TFile *fEPcorrectionFile;               //!<! File containing the EP corrections
```
raymondEhlers, line 254: This needs to be streamed (`///<`) - otherwise, it won't work on the grid. Probably simpler would be to read the tree and stream that instead, but as you like.

In either case, you'll need to bump your claasdef

raymondEhlers (changes requested):

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskEmcalJetHdEdxCorrelations.h
Steffanic: Thanks, I have made the modifications.

raymondEhlers:
+1

PR #22472, opened 2023-07-06
PWGGA/GammaConv: Write all high pt photons to tree

sstiefel19:
For low pt photons only a fraction as it used to be. Boarder between high and low pt can be modified.


jokonig:
+1

PR #22478, opened 2023-07-08
PWGGA/GammaConv: Add omega PbPb cutstrings and bug fixed

nstrangm:
- Cutstring for omega PbPb in EMCal and PCM AddTask added
- Add MBWs in PbPb PCM AddTask
- Fix memory leak in omega task (delete new pointer)


jokonig:
+1

PR #22452, opened 2023-07-03
[PWGEM-13] PWGGA/GammaConv: AddTask_GammaCalo_PbPb - changed traincon…

mhemmer-cern:
…fig

- Changed trainconfigs 751 to 754 to select cent (751) to peripheral collisions (754) only for EMCal+DCal with rotation background method. This can be used to study the effect the mass difference in MC and Data depending on the centrality!

[PWGEM-13] PWGGA/GammaConv: AliAnalysisTaskGammaCal - Add Photon resolution histo

- Add histogram to monitor the energy resolution for photons (E_rec-E_true)/E_true as function of pT_rec. This histogram needs `fDoLightOutput > 0` and `fDoClusterQA > 0` to be created and filled!
Currently no separation between photon and conversion photon clusters is made.

PR #22485, opened 2023-07-10
[PWGEM-13] PWGGA/GammaConv: AddTask_GammaCalo_PbPb - changed trainconfig

mhemmer-cern:
- Changed trainconfigs 755 to 758 to select cent (755) to peripheral collisions (758) only for EMCal+DCal with rotation background method. This can be used to study the effect the mass difference in MC and Data depending on the centrality! Similar to previous PR #22452 only this time with track matching instead of cell track matching and no track matching


jokonig:
+1

PR #22494, opened 2023-07-12
MC closure test (do bkg subtraction in MC)

chengtt0406:
This commit is used for the MC closure test: do bkg subtraction in MC, which is the same as data approach


adubla:
+1

adubla:
+1

PR #22496, opened 2023-07-12
[PWGEM-13] Add secondary cell energy subtraction for cell track matcher

mhemmer-cern:
- In the cell track matcher there is now the option `SubtractMultiple` which can be enabled. If this is enabled and a cell has less energy than the MIP energy of the matched particlethen a cell next to this one will be searched to subtract the remaining energy from this cell. The search direction is only in phi and prefers the direction the track is pointing. If this cell has no energy take the other one, if both have no energy than nothing will be done.

- Changed binning for the resolution histogram added in the PR #22452

### DIFF
--- a/PWGGA/GammaConv/AliAnalysisTaskGammaCalo.cxx
+++ b/PWGGA/GammaConv/AliAnalysisTaskGammaCalo.cxx
@@ -890,6 +890,7 @@ void AliAnalysisTaskGammaCalo::UserCreateOutputObjects(){
   }
 
   // set common binning in pT for mesons and photons
+  double epsilon              = 1.e-6;
   Float_t binWidthPt          = 0.1;
   Int_t nBinsPt               = 250;
   Float_t minPt               = 0;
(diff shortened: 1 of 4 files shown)


PWGGA/GammaConv/AliAnalysisTaskGammaCalo.cxx
```diff
@@ -901,9 +902,12 @@ void AliAnalysisTaskGammaCalo::UserCreateOutputObjects(){
   Float_t maxClusterPt        = 50;
   Int_t nBinsMinv             = 800;
   Float_t maxMinv             = 0.8;
+  float minRes                = -1.f;
+  float maxRes                = +5.f;
```
FriederikeBock, line 906: are the -1.f and +5.f deliberate? I have not seen it done that way yet.

mhemmer-cern: From my understanding any floating point number like 1. would first be assumed to be a double by the compiler, so writing 1.f makes sure it is a float and there is no extra conversion from double to float happening then.

FriederikeBock:
+1

PR #22502, opened 2023-07-14
Add INEL selection before counter

Tao-Fang:


lvermunt:
Hi @Tao-Fang, instead of having the AliNormalizationCounter always fill the events of INEL>0, I would propose you to store two counters. One for minimum bias (so the task can also be used for a MB analysis) and one for the INEL>0 events

Tao-Fang:
Hi @lvermunt 
Thanks for the reminder I'll add another part of the stored the INEL>0 events

PR #22515, opened 2023-07-18
optimizing code for closure

maroozm:
hi, added:
- closure based on the profile plot of shared fraction vs pt, rather than a direct cut on shared fraction and related QA histograms
- bfield systematics
- and some other changes

thanks


abilandz:
+1

PR #22520, opened 2023-07-19
changed to histogram output

sweyh99:
To reduce the output enough to save all the inclusive stats


sweyh99:
Floats were not enough for some of my variables when I was saving them as trees, so I made the histograms doubles to be safe. I will keep an eye out for memory issues and will reduce any that I can to floats if needed.

PR #22528, opened 2023-07-19
NFM for subwagons

maroozm:
- just adding one parameter to the AddTask, so that I will be use subwagons on the legotrain system


abilandz:
+1

PR #22581, opened 2023-07-31
PWGGA/GammaConv: Update AddTask_GammaCalo_PbPb.C

mhemmer-cern:
- Added new trainconfigs `950+` which are similar too the trainconfigs `750+`, only difference is that here the track matching is active , but only clusters that were matched with a track will be used instead of being rejected. This is for a study to find our how many tracks actually match a cluster and what the energy difference between cluster and actual photon is depending on the number of matched tracks.


jokonig:
+1

PR #22588, opened 2023-08-01
PWGGA/GammaConvBase: Update AliCaloPhotonCuts.cxx

mhemmer-cern:
- Add new histogram `fHistClusterENMatchesConv` analog to `fHistClusterENMatchesNeutral` and `fHistClusterENMatchesCharged` for the cluster cases 4 and 5: 
4: Gamma conv cluster R < 180cm 
5: Gamma conv cluster R > 180cm 
This also means that `fHistClusterENMatchesCharged` now excludes conversion cluster! If this should be changed back again please write me!


jokonig:
+1

PR #22595, opened 2023-08-02
Add combination of triggers

mfasDa:
Add option to combine triggers in option
- all: All triggers required to be present in event for
  positive decision
- any: Any trigger required to be present in event for
  positive decision


mfasDa:
Needed for three purposes:

- require L0 for L1
- merge EMCAL and DCAL triggers 
- EMCAL-DCAL correlation triggers

PR #22597, opened 2023-08-03
pT dependent PID added

asasikum:
pT dependent PID combining TPC and TOF nSigma, looser cuts to increase the statistics in reconstructed data.


victor-gonzalez:
+1

PR #22607, opened 2023-08-04
 ML4PhotoConv: Basic tree for photon and meson for ML

an15ms157:
A sub-task is added in AliAnalysisTaskGammaConvV1 which creates trees with properties of photons and mesons to be used in hipe4ml


mikesas:
Hi Abhishek, was the code tested locally? Are the file sizes doable for the grid?

PWGGA/GammaConv/macros/AddTask_GammaConvV1_PbPb.C
```diff
@@ -53,8 +53,18 @@ void AddTask_GammaConvV1_PbPb(
   Bool_t    enablePlotVsCentrality        = kFALSE,
   Bool_t    processAODcheckForV0s         = kFALSE,   // flag for AOD check if V0s contained in AliAODs.root and AliAODGammaConversion.root
   // subwagon config
-  TString   additionalTrainConfig         = "0"       // additional counter for trainconfig + special settings
-)  {
+  TString   additionalTrainConfig         = "0",       // additional counter for trainconfig + special settings
+  
+  // ML parameters
+  Bool_t    enablePhotonTree      =  kTRUE,
```
jokonig, line 59: The additional train config needs to be the last argument in order to properly set up sub-wagons I think. 
Also, you should set the default values to false

PWGGA/GammaConv/AliAnalysisTaskGammaConvV1.cxx
```diff
@@ -564,6 +694,14 @@ AliAnalysisTaskGammaConvV1::AliAnalysisTaskGammaConvV1(const char *name):
...
   DefineOutput(8, TTree::Class());
   DefineOutput(9, TTree::Class());
+  DefineOutput(10, TTree::Class());
+  DefineOutput(11, TTree::Class());
+  DefineOutput(12, TTree::Class());
+  DefineOutput(13, TTree::Class());
+  DefineOutput(14, TTree::Class());
+  DefineOutput(15, TTree::Class());
+  DefineOutput(16, TTree::Class());
+  DefineOutput(17, TTree::Class());
```
jokonig, line 704: Is it possible to define this many output containers?  Why are so many needed?

jokonig (changes requested):
Wouldn't it be better to have your own analysis task for this purpose? (I think we already have some tasks specifically for trees in the ConvCalo folder.)

PWGGA/GammaConv/macros/AddTask_GammaConvV1_PbPb.C
an15ms157: Hello @jokonig 
Yes, setting the value to kFalse is good idea. I'll change the order and set the value

PWGGA/GammaConv/AliAnalysisTaskGammaConvV1.cxx
an15ms157: I will be using https://hipe4ml.github.io/ to analyze the data afterward. The package mentioned here to my knowledge can only access trees from a root file (and not from a directory/subdirectory) and hence the usage of containers to store the trees. The containers seem to be working locally at least for this purpose

an15ms157:
Hello @mikesas 
  Sorry to respond back to you so late as I was on holiday. I have checked it locally. As for using them on the grid, I plan to do them in the next iteration (soon). In this version, I want to have the code available to everyone so that they could run them locally on their own university/institute servers. I have tried to reduce the size of the output tree as much as possible but since I didn't intend to run them on the grid, I'm not sure if they fit the criteria but most possibly they are okay.

PWGGA/GammaConv/macros/AddTask_GammaConvV1_PbPb.C
jokonig: Hi @an15ms157 ,
Sorry for the delay, I am at a conference this week and missed that.
I still think that putting the values after the additional train config will break the sub-wagon configuration on the lego-train page. Because you can only have one additional argument in the subwagon section (at least to my knowlage). And then, one could not change your new parameters anymore

amaringarcia:
> Hello @mikesas
> Sorry to respond back to you so late as I was on holiday. I have checked it locally. As for using them on the grid, I plan to do them in the next iteration (soon). In this version, I want to have the code available to everyone so that they could run them locally on their own university/institute servers. I have tried to reduce the size of the output tree as much as possible but since I didn't intend to run them on the grid, I'm not sure if they fit the criteria but most possibly they are okay.

A little comment here. We should create  representative trees for training on the grid,  do the training locally, and then run over the full statistics "on the fly" using the output of the training.

PWGGA/GammaConv/AliAnalysisTaskGammaConvV1.h
```diff
@@ -352,6 +368,91 @@ class AliAnalysisTaskGammaConvV1 : public AliAnalysisTaskSE {
...
+    Float_t                           fPt_EPos_Gamma1_MesonML;  
+    Float_t                           fPAsymmetry_EPos_Gamma1_MesonML;  
+    Float_t                           fEta_EPos_Gamma1_MesonML;  
+
+    TList**                           fMLFolder;                                  //End ML Variables
+
+
   private:
 
     AliAnalysisTaskGammaConvV1(const AliAnalysisTaskGammaConvV1&); // Prevent copy-construction
```
jokonig, line 458: I think you need to increase the classdef

jokonig (changes requested):

jokonig:
+test

jokonig:
+1

PR #22629, opened 2023-08-15
[PWGEM-13] PWGGA/GammaConv AliConvCalo - Add 3D histo to matched mass…

mhemmer-cern:
…, energy and matched tracks

- PCM-Calo:
	- Added new TH3F `fHistoMotherInvMassECalibNMatchedPrim` which stores the Pi0 candidate mass, the energy of a cluster photon candidate and the number of matched primary tracks to that cluster. The number of matched primary tracks can even be obtained when primary track matching is turned off! This histo could be used to get the difference in pi0 mass depending on the number of matched tracks for a given energy to make some sort of NonLin Calibration.
	- Added new train configs with finer cent
- CaloPhotonCuts:
	- Added new TH1F to track the number of matched tracks per cluster


jokonig:
+1

PR #22633, opened 2023-08-17
Phiresonanceremoval, Correlations without Symmetrization

asasikum:
Phi resonance is removed by rejecting the charged kaon tracks with a mother particle, Symmetrization of correlations is avoided in the filling.


lhusova:
+1

PR #22651, opened 2023-08-22
reject resonances of k0s

asasikum:
cut to remove the resonances of k0s is added. (k0s always have a mother particle)


lhusova:
+1

PR #22673, opened 2023-08-28
[PWGEM-13] PWGGA/GammaConv - Add neutral overlap cluster energy corre…

mhemmer-cern:
…ction

- Added the functionlaty in AliCaloPhotonCut to correct for neutral overlap in the clusters, which is expected in PbPb. Currently this is implemented in the following way: To use it one has to select either TM Cut 't' or 'r'. The first option uses mean number of matched primary charged particles the second option random number. The distribution for these numbers is coming from the Possion fit `fFuncNMatchedTracks`, which was performed to PbPb 2018 and 2015 anchired MCs in different centralities and then the Poisson parameter was obtained as a function of centrality which was fitted and that resulted fit is now in `fFuncPoissonParamCent`. To use this in an AliAnalysisTask one needs to call
```if(((AliCaloPhotonCuts*)fClusterCutArray->At(fiCut))->GetDoEnergyCorrectionForOverlap() > 0){
    ((AliCaloPhotonCuts*)fClusterCutArray->At(fiCut))->SetPoissonParamCentFunction(fIsMC);
    ((AliCaloPhotonCuts*)fClusterCutArray->At(fiCut))->SetNMatchedTracksFunc((centMax + centMin) / 2.);
  }
```
inside the `ProcessClusters()` function to set up the correct TF1 parameters and then call
```
if(((AliCaloPhotonCuts*)fClusterCutArray->At(fiCut))->GetDoEnergyCorrectionForOverlap() > 0){
      clus->SetE(clus->E() - ((AliCaloPhotonCuts*)fClusterCutArray->At(fiCut))->CorrectEnergyForOverlap());
    }
```
To subtract the energy coming from neutral overlap (keep in mind this is just an assumption for the neutral overlap, since we do not know the exact value!). This call needs to be done BEFORE the ClusterIsSelected check, to ensure the change in energy does not lower the cluster energy below the cut value.


mhemmer-cern:
The last commit just added the parameters for the SetPoissonParamCentFunction for data, which was set to the MC parameters before. But after checking with train output from 1131 I saw a difference in the number of matched primary tracks between data and MC; more matched tracks in data then in MC.

mhemmer-cern:
- Add fix when running without track matching. Previously to would simply crash!

jokonig:
+1

PR #22682, opened 2023-08-31
PWGGA: Changed M02 and NCell cut 

Steffimro:
Changed cut to evaluate the influence on nonlinearity discrepancy seen in PHOS Tender and PHOS Tender MC.


jokonig:
+1

PR #22700, opened 2023-09-12
PileUp OOB for PbPb analysis

druggian:
The method SetRejectTPCPileupWithITSTPCnCluCorr(kTRUE) is included in the Reader class for the LHC18 period from PbPb collision


wrzesaCERN:
+1

PR #22703, opened 2023-09-13
OOB pileup method for LHC18 PbPb

druggian:
OOB pile-up method for the LHC18 dataset in the PbPb collision -> fixed


majanik:
+1

PR #22708, opened 2023-09-14
Mezon Phi Eff for custom cuts

AsFastAsSnail:
Separate task to calculate eff of mezon phi production with custom cuts to match femto analysis


majanik:
+1

PR #22711, opened 2023-09-15
Add calculation of opening angle for KF in AliVertexingHFUtils

zhujh126:
Add calculation of opening angle for KF in AliVertexingHFUtils and add opening angle of Xi to pion for Xic0 Pb-Pb analysis


fcatalan92:
+1

PR #22720, opened 2023-09-20
Issue fixed

jpxrk:
There was an error while it's running on train. It was tested on grid and confirmed it runs without error.


sashingo:
+1

PR #22721, opened 2023-09-21
Add new and p_T-dependent preselections

zhujh126:
Add new and p_T dependent, based on 30-50% data and MC, 6% for 2-4 GeV/c and 3% for 4-12 GeV/c


arossi81:
+1

PR #22725, opened 2023-09-21
Eta cut update for MC plots

ChrisPliatskas:
Update on the application of the eta cut for the MC case in the Dpt calculation. In practice QA histograms have the same eta cut as detector level while the Response matrix is unrestricted


raymondEhlers:
+1

PR #22729, opened 2023-09-22
PWGGA/GammaConv: Cutstring to study diff. multiplicities

Steffimro:
Added cutstrings to study different multiplicities.


jokonig:
+1

PR #22736, opened 2023-09-26
Add BaryonBachelorCosPA to FemtoDreamCascade and changes in AddTask f…

otonvd:
…or ML BDT Xi selection

-open cuts in general for ML
-add BachBaryonCosPA word to tree
-fix bug with TPC nsigmas for bachelor in Xi ttree
-allow up to 100 cascades in tree (maybe I can take this back  at some point to reduce tree size, but now is necessary for ML)


otonvd:
+1

PR #22752, opened 2023-10-02
[PWGEM013] PWGGA/GammaConv - Update neutral overlap cluster energy co…

mhemmer-cern:
…rrection

- In the CaloPhotonCut code, changed the energy value that was subtracted from the cluster energy from a static value (300 MeV) to a centrality dependent value. The value comes from mean pT charge particle distributions as functions of centrality that were fitted. A simple simulation for Pi0 decaying into two photons has shown that the mean energy of a decay photon from a Pi0 with a given pT X is X/2. So the energy that is subtracted from the cluster is now 0.5 of the charge particle mean pT at the given centrality. This is under the assumption, that the mean pT distribution of neutral pions follow the charge particle mean pT distribution, which seems resonable.
- Add two cut settings to use this correction for the calo only task, 789 for mean neutral energy and 790 for random energy correction. Mean and random refer to the assumend number of neutral particles per cluster that contribute to the total cluster energy.

### DIFF
--- a/PWGGA/GammaConv/AliAnalysisTaskGammaCalo.cxx
+++ b/PWGGA/GammaConv/AliAnalysisTaskGammaCalo.cxx
@@ -3460,6 +3460,13 @@ void AliAnalysisTaskGammaCalo::ProcessClusters()
     nclus = arrClustersProcess->GetEntries();
   }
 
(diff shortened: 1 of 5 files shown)


jokonig:
+1

PR #22757, opened 2023-10-04
passing non-const pointer

Steffanic:
For some reason passing a non-const vTrack to the AliPIDResponse::CheckPIDStatus method caused a segfault.


raymondEhlers:
+1

PR #22759, opened 2023-10-05
[PWGEM-13] PWGGA/GammaConvBase: AliCaloPhotonCuts - optimise neutral …

mhemmer-cern:
…overlap correction CPU time

- the mean value of the number of matched tracks per cluster is now calculated once per event and stored to recude CPU time
- the fnction to get the mean energy of the neutral overlap is changed from TF1 to a c function with a c-array as parameter to further reduce the CPU time
- fixed wrong factor in `CorrectEnergyForOverlap`, was 0.25 should be 0.5!


jokonig:
+1

PR #22779, opened 2023-10-10
Checking for valid pointer to VTrack

Steffanic:
I think that somewhere a track does not have a valid second() (It's a bit unclear to me what this is, but it is the required input for the PID methods) In any case, I now check that my pointer to the output of this method is valid and skip it if it is not. Sorry for the lengthy debugging session.


raymondEhlers:
+1

How frequently does this happen? I'm a bit concerned this will be quite noisy, since I don't recall if this will get suppressed when running at grid sites (I think ERROR was preserved, but I'm not 100% sure at the moment). I will +1 for expediency here, but if it's quite noisy, please update this before running to something like INFO so it will be suppressed

raymondEhlers:
If it's low frequency, then it should be fine - the logs are de-duplicated, so it's fine if they're not too frequent and aren't interspersed with other messages. I just wasn't sure based on your description

PR #22782, opened 2023-10-11
[Run3JetValidation] modify binning according to tpc sectors

jloemker:
Hi @raymondEhlers and @Archita-Dash,

this PR only modifies the binning of the jet and track phi distributions.
This allows us to properly compare the results from AliPhysics to O2Physics within the JetValidation.

Best,
Johanna


raymondEhlers:
+1

PR #22783, opened 2023-10-11
PWGGA/GammaConv: Add feed-down analysis option to PureMC studies

nstrangm:
- Output mothers of pi0s, etas and omegas
- Output yield of primordial (not from decay) pi0s, etas and omegas


jokonig:
+1

PR #22801, opened 2023-10-18
Fix loop on mothers

fgrosa:


PWGHF/vertexingHF/vHFML/AliAnalysisTaskSEHFResonanceBuilder.cxx
```diff
@@ -1531,7 +1531,7 @@ int AliAnalysisTaskSEHFResonanceBuilder::MatchResoToMC(AliAODMCParticle *partD,
         return 0;
 
     double momSumDaughters[3] = {partD->Px()+partLight->Px(), partD->Py()+partLight->Py(), partD->Pz()+partLight->Pz()};
-    for (auto iMother{commonMothers.size()-1}; iMother>=0; ++iMother) {
+    for (int iMother{commonMothers.size()-1}; iMother>=0; iMother--) {
```
mfasDa, line 1534: clang on mac complains at this line about a narrowing cast from unsigned long to int, for which compilation breaks here (c++11 narrowing)

fgrosa: Thanks for pointing this out @mfasDa, [this PR](https://github.com/alisw/AliPhysics/pull/22829) should fix it.

PR #22834, opened 2023-10-24
[PWGEM-13] AddTask_GammaConv* - Fix single FEPC file option

mhemmer-cern:
- Fixes an issue where giving a single FEPC file would result in an error message instead of just using that one file for all cuts.


jokonig:
+1

PR #22837, opened 2023-10-24
Implementation of AliEventCuts

Archita-Dash:
@raymondEhlers @vkucera @nzardosh @fkrizek @jloemker 

I have implemented AliEventCuts that further improves the jet validation (between AliPhysics and O2Physics).
Note: The task still ensures that the MB Event selection is in use.


raymondEhlers:
+1

PR #22860, opened 2023-10-27
PWGGA/GammaConv: AddTask PbPb (Conv)Calo add train trigger only configs

mhemmer-cern:
- Add trainconfig for inly the triggered 0-10% and 30-50% centralities


jokonig:
+1

PR #22870, opened 2023-10-27
memory error fixed, likely due to non-void function never reaching a …

Steffanic:
…return. Hopefully this is the final fix 🤞


raymondEhlers:
+1

PR #22883, opened 2023-10-31
[PWGEM-13] PWGGA/GammaConv: (Conv)Calo - add traincinfigs with and wi…

mhemmer-cern:
…thout OOB pile up correction

- Added train configs with and without OOB pileup correction to check the effect on EMC-EMC and PCM-EMC


jokonig:
+1

PR #22887, opened 2023-11-01
PWGGA/PHOS: Fine-tune trigger efficiency for pp 13 TeV

jokonig:
- Trigger effiency curves are obtained by fitting ratio of cluster spectra from the L0-trigger flagged INT7 events to all INT7 events per ddl
- In this procedure, statistics is very limited.
- It was observed that the trigger efficiency in 2017 and 2018 in data is lower copared to MC simulations, the shape however fits
- Therefore, a global shift is introduced for the 2017 and 18 dataset to reduce the trigger efficiency to match the data


FriederikeBock:
+1

PR #22888, opened 2023-11-01
[PWGEM] PWGGA: Fix neutral overlap correction:

mhemmer-cern:
- Fix in GammaCalo task: Ppreviously the neutral overlap correction was not corretly applied, because it was only done when we would do the flat energy correction which should not be used together.
- Fix in CaloPhotonCuts: fParamMeanTrackPt values should be set for `fIsMC >= 1` with the MC values and not for `fIsMC > 1` which would only be for special MC like JJ MC. So before this fix normal MC used the same values as data, but we know that in MC the mean pT is lower than in data with HIJING.


jokonig:
+1

PR #22891, opened 2023-11-02
[PWGEM-13] PWGGA/GammaConv: changed neutral overlap cut to not use TM

mhemmer-cern:
- Neutral Overlap Correction should be run with cell track matching and hence should not be used with normal primary track matching, so the cut settings `t` and `u` where changed to only use the neutral overlap correction but no primary track matching


jokonig:
+1

PR #22893, opened 2023-11-02
Xi-pi/K femto: Allow tune mixing for FemtoDream Pairing checks and MC resolution matrix

otonvd:
… resolution matrix


otonvd:
+1

PR #22899, opened 2023-11-03
[PWGEM-13] PWGGA/GammaConv: Add Trainconfig for PbPb 2018qr

mhemmer-cern:
- Add new traincofnigs for ConvCalo for PbPb 2018qr with no TM


jokonig:
+1

PR #22903, opened 2023-11-04
[PWGEM-13] PWGGA/GammaConv: AddTaskConvCalo PbPb - Add light output f…

mhemmer-cern:
…lag settings

- Added the `SetLightOutput` and `SetECalibOutput` function calls for the Task (not the cuts!) similar to what Adrian introduced into the pp AddTask, so we have excess to all the QA plots again in PbPb that were blocked off by not having `SetECalibOutput`, e.g. the InvMass vs E photon plots.


jokonig:
+1

PR #22922, opened 2023-11-08
Renewed update of filter task

ChrisPliatskas:
The track filter task has been updated to include a pt- dependent tracking efficiency according to a yaml configuration. The relevant methods were replicated from the AliEmcalJetTask.


raymondEhlers:
+1

Note that I didn't trace all of the logic through - I'm trusting that you ported it over correctly - but at a quick glance, it seems about right

raymondEhlers:
+1

PR #22939, opened 2023-11-10
Update of AOD track filter

ChrisPliatskas:
This update contains:
a) An option for pt-dependent tracking efficiency reduction based on a yaml configuration. This update was ported based on the AliEmcalJet task 
b)Minor bugfixes of previous versions


raymondEhlers:
+1

PR #22946, opened 2023-11-13
[JetValidationFramework] Fix pi mass and reduce track cuts

jloemker:
Hi @Archita-Dash ,

these are the small changes that helped increasing the similarity between AliPhysics and O2Physics.
- the track cut SetMaxFractionSharedTPCClusters does not exist in O2Physics and removing it increases the number of tracks in AliPhysics slightly (we still have less tracks in AliPhysics wrt. O2Physics).
- the pi mass in O2Physics is rounded to 0.139 now we do this in AliPhysics as well and only miss one jet in AliPhyisics wrt. O2Physics.

Cheerio,
Johanna


jloemker:
Sorry @raymondEhlers !
I forgot to tag you for approval of these mini changes that benefit the jetfindercomparison.

raymondEhlers:
+1

Thanks for making sure that I see it. Just fyi, you don't need to tag me unless you've been waiting more than 24 hours - the automatic post requesting approval takes care of it

Best, Raymond

PR #22951, opened 2023-11-13
[PWGEM-13] PWGGA/GammaConv: AddTask ConvCaloPbPb - Add train configs

mhemmer-cern:
- Add train configs for PbPb 5TeV with fine cent binning 10% seteps except 0-5 and 5-10 and with out TM, to check the Number of cells per cluster


jokonig:
+1

PR #22994, opened 2023-11-24
[PWGEM-13] PWGGA/GammaConv: Add EMCal neutral overlap correction base…

mhemmer-cern:
…d on mean Nch/cell

- Add new option for neutral overlap correction which is based on the mean number of charged particles per cell instead of mean matched tracks per cluster as before. This new option is track matching option 'v'


jokonig:
+1

jokonig:
+1

PR #22995, opened 2023-11-24
fixed trackQA binning+ removed redundant comments

Archita-Dash:
- the track QA histos bininngs are modified to match with those of in O2Physics validation task
- VtxZ histogram renamed to a more standard name
- redundant lines of code and comments are now removed


raymondEhlers:
+1

PR #23001, opened 2023-11-29
PWGGA: GammaConvFlow - adapated V0Reader usage from PCM task

mhemmer-cern:
- Adapted the V0ReaderTask to be similar to the PCM  tasks we use, where it gets the photonCutNumberV0Reader as input from the AddTask
- In AnalysisTask changes are just blanks for better overvie when trying to collapse code blocks ind IDEs


jokonig:
+1

PR #23002, opened 2023-11-29
PWGGA/PHOS: parameter for subwagons is added

vlad-kuskov:
Parameter char *sub_name is added to AddTaskPHOSPi0EtaToGammaGamma_PbPb_5TeV in order to run subwagons on the analysis-train


kharlov:
+1

PR #23015, opened 2023-12-04
Adding protections against inf and nan to make Mixed Events finite.

Steffanic:
As well as passing the AliVTrack object to the AliEventPool so I can get the TOF PID info in mixed events.


raymondEhlers:
+1

PR #23017, opened 2023-12-04
Fixing macOS RPATH for libRooUnfold

mfasDa:
link_direcotries of the RooUnfold libdir
needed in order to trigger adding it
to RPATH


pzhristov:
+1

PR #23020, opened 2023-12-05
Setting macOS RPATH also for TreeLite libraries

mfasDa:


mfasDa:
Either `link_directories` in `find...` or `target_link_directories` in CMakeLists.txt of the library needed to trigger the CMake policy to set the RPATH to the library - `-L...` in `target_link_directories` not sufficient as it has no impact on the CMake policy.

pzhristov (approved):

pzhristov:
+1

PR #23030, opened 2023-12-08
[PWGEM-13] PWGGA/GammaConv: Addition of NonLin like Neutral Overlap C…

mhemmer-cern:
…orrection

- Add case `w` for TrackMatching as implementation of first iteration of NonLin like neutral overlap correction
- Add specific cases for this to the GammaConvCalo and GammaCalo Analysis tasks
- Add and updated GCoCa and GCa trainconfigs (most notably added GCa version without OOB pileup correction for MC!)


jokonig:
+1

PR #23034, opened 2023-12-09
[PWGEM-13] PWGGA/GammaConv: Updated Neutral Overlap Correction via tr…

mhemmer-cern:
…ack density

- After a brief discussion with Joshua, the number of cells that will go into the correction for number of tracks per cell was changed to `12.f` from the number of cells of the current cluster. The idea behind this is that a normal photon cluster has on average 2 cells with 10 neighbouring cells, so in total there are on average 12 cells where neutral overlap could happen.
- Updated train configs for GammaConvCalo


jokonig:
+1

PR #23037, opened 2023-12-11
Got rid of redundant histograms, added track matching abilities, adde…

anabananana98:
…d plots needed for fast simulation checks


raymondEhlers:
+1

PR #23043, opened 2023-12-13
Fix to previous fix to eff

AsFastAsSnail:
Something was overlooked, and resulted in one significant histogram to be empty


majanik:
+1

PR #23046, opened 2023-12-14
Femto: Add MC Truth selection and ancestor analysis.

makor:
I needed to code some additional getter fct. in the DreamBase class. This will *NOT* impact default behaviour.


rdelgran:
+1

PR #23056, opened 2023-12-19
PWGGA/GammaConv: Add Jet output to isolated gamma tree

nstrangm:
- Add Jet option in AddTask and AnalysisTask for gamma-jet correlation studies
- Add true and reconstructed jet momenta leafs to the tree


jokonig:
+1

PR #23062, opened 2023-12-20
changing the inheritance of AliFemtoPairCutMInv

asasikum:
Changing the inheritance of the class AliFemtoPairCutMInv from AliFemtoPairCut to AliFemtoPairCutAntiGamma in-order to use more than one paircuts in the code at the same time.


majanik:
+1

PR #23063, opened 2023-12-20
New code to analyse mean pt and pt fluctuations with respect to N charged particles

chlorine1777:
Added AliAnalysisPtN to calculate mean pt and higher order pt fluctuations with respect to Nch. This task will be further on contributed to the research in speed of sound in QGP.


yzhouCERN:
+1

PR #23072, opened 2023-12-28
R2 & P2 with different pt ranges

subhadee-ali:
These new set of macros are added to study R2 and P2 correlation functions in different pt ranges


victor-gonzalez:
+

victor-gonzalez:
+

victor-gonzalez:
+1

PR #23085, opened 2024-01-11
Modifying electron rejection cut

asasikum:
Modified the electron rejection cut to make it stricter. The default value is set to zero which rejects no electrons. Higher values of this parameter `fMaxNsigmaERejection` would reject more electrons.


atimmins:
+1

PR #23087, opened 2024-01-11
Update AliAnalysisTaskOmegaDielectron_AccEff.cxx

mwaelde:
Add rec_omega histograms with only TrackCuts to the output


rbailhac:
+1

PR #23089, opened 2024-01-11
[PWGEM-13] PWGGA/GammaConv: AliAnalysisTaskGammaConvCalo.cxx - Add mi…

mhemmer-cern:
…ssing if(!fDoLightOutput)

- In the UserExec there was a missing `if(!fDoLightOutput)` when trying to use the `fHistoMCHeaders`


jokonig:
+1

PR #23096, opened 2024-01-12
Added cut on Lifetime of K0s Reconstruction

asasikum:
Added cut on Lifetime of K0s Reconstruction and a switch for turning on/off the armenteros-podolonski cut in the class `AliFemtoV0TrackCut`


majanik:
+1

PR #23106, opened 2024-01-17
[PWGEM-13] PWGGA/GammaConv: (Conv)Calo AddTask - Add pT weighting

mhemmer-cern:
- Add functionality to use pT weighting for added signal MC for LHC18q anchored MC


jokonig:
+1

PR #23119, opened 2024-01-20
PWGGA/GammaConv: AliCaloPhotonCuts - Increment ClassDef

mhemmer-cern:
- Previous commit did not increment ClassDef, thus should fix it
- Also added spaces to some comments for AliAnalysisTaskConvCaloCalibration memebers which produced warnings before

PWGGA/GammaConv/AliAnalysisTaskConvCaloCalibration.cxx fix QAMesonFlag

- Fix a bug with the QAMesonFlag where a histogram is only initilized if its set, but was always accessed later.


jokonig:
+1

PR #23123, opened 2024-01-22
PWGGA/GammaConvBase/AliCaloPhotonCuts: Fix PbPb Nonlin like tempE

mhemmer-cern:
- Set `tempE` variable for 2nd iteration to E inside `CorrectEnergyForOverlap` if the energy is not at the threshold where it will be constant. Since the first iteration threshhold was lower than the energy threshold now, before this fix all values between the old threshhold and the new one would be set to the old treshold value


jokonig:
+1

PR #23138, opened 2024-01-24
PWGGA/GammaConv: Prepare (PCM-)EMC for LHC24a1 MC

mhemmer-cern:
- Added new LHC24a1 added signal MC to the GCoCa and GCa AddTasks to the injector header list
- Fixed `FindEnumForMCSet` implementation for `LHC24a1` since there are `LHC24a1a` and `LHC24a1b` it was changed to `namePeriod.BeginsWith("LHC24a1")` which was previously `namePeriod.ComparesTo("LHC24a1") == 0`.


jokonig:
+1

PR #23140, opened 2024-01-24
PWG/EMCAL: Add option for cell energy calibration per column

jokonig:
- The absolute cell energy calibration is done in order to account for material in front of the detector, and hence not all cells should be calibrated to the nominal Pi0 mass. Effects due to the material are taken from MC simulations
- Previously, the Full, 2/3 and 1/3 SM types were calibrated seperatly, or the part where TRD support structures are visible had a different absolute calibration than the rest of the detector
- This new commit enables a column by column calibration, to refine the idea o calibrating the TRD support regions differntly
- This is only an esperimental setting! And is used to test calibration improvments for Run3


FriederikeBock:
+1

PR #23158, opened 2024-01-29
PWGGA/GammaConvBase/AliCaloPhotonCuts: PbPb enable TM for neutral ove…

mhemmer-cern:
…rlap again

- enable track matching for the neutral overlap track matching cuts. Previously the CellTM setting was set to also match secondaries. Now we only want the CellTM to match primaries and the normal TM to match secondaries, for this we need to reanable the track matching in the track matching cuts. Keep in mind do set the `doPrimaryTrackMatching` flag in the AddTask to false, when running with CellTM and Neutral Overlap Correction!


jokonig:
+1

PR #23166, opened 2024-01-31
PWGGA/GammaConv: AddTask_GammaCalo_PbPb - Add rotation background tra…

mhemmer-cern:
…in config

- Add NonLin like neutral overlap train configs that use rotation background


jokonig:
+1

PR #23171, opened 2024-02-01
PWGGA/GammaConv: - Changed CutSetting from no TM to standard PbPb TM

mhemmer-cern:
- we need this change since we use the TM Cut also for secondaries

IMPORTANT: This will remove MIPSubtraction from the CaloPhotonCuts where it should not be used from anymore. MIPSubtraction should be used via the yaml config file for the EMCal Correction task.


jokonig:
+1

PR #23187, opened 2024-02-03
Changing inheritance of AliFemtoPairCutMInv class

asasikum:
The inheritance of AliFemtoPairCutMInv class is changed inorder to perform the rejection of phi meson resonance candidates using invariance mass cut.


wrzesaCERN:
+1

PR #23191, opened 2024-02-06
Fixes for c++20

mfasDa:
Obviously forward declarations are no longer allowed
if the object is used in templated class members as
template argument.


mfasDa (commented):
Let's see if that fixes the compilation issues ...

mfasDa:
+test

mfasDa:
+1 

Works, AliPhysics compiles with the fix under c++20, can be merged immediately.

mfasDa:
+1

PR #23195, opened 2024-02-07
PWGGA/GammaConv: EMC MC timing cut + new NOC

mhemmer-cern:
- Add more options for injection header selection to `AddTask_GammaCalo_PbPb.C` and `AddTask_GammaConvCalo_PbPb.C`
- Add cluster timing cut for all MCs that are `kPbPb5T18HIJING`. This timing cut uses the width as in data, but also uses an offset, since we have seen an offset of 61.6 ns in all MCs anchored to LHC18qr
- Add a new NOC (neutral overlap correction) which was determined by fitting the mass^2 ratio of Pi0s in pp 13 TeV and PbPb 5.02 TeV with PCM-EMC as function of the cluster energy. The old NonLin like correction was fitted to EMC only data where the energy range was much smaller and the uncertainties made the fit very unstable. This should be a much better correction now.

- Add histogram `fHistoClusOverlapMBHeaderGammaPt` to AliAnalysisTaskGammaCalo, to track if cluster has overlap with MB header if MB header is not selected


jokonig:
+1

PR #23200, opened 2024-02-08
PWGGA/GammaConvV1: Add mc pt histo

sstiefel19:
add two histos that allow for better comparison of meson true efficiencies between MC productions with differing pt shapes of generated particles


jokonig:
+test

jokonig:
+1

PR #23203, opened 2024-02-11
PWGGA/PHOSTasks/PHOS_Run2 added kCentral/kSemiCentral trigger

vlad-kuskov:
- Added options for analyses with kCentral/kSemiCentral/kMB triggers in AddTask macro
- The same is done for the corresponding AliAnalysis code


peressounko (commented):
What is the reason for this chager? It will make a mess of events with different triggers and will not be useful for analysis. Rather you should run 3 subwagons with 3 separate triggers.

vlad-kuskov:
These changes are aimed exactly on the trigger selection. One could use kCentral OR kSemiCentral OR kMB (in addition to INT7 and PHI7 triggers) trigger. Previously, only INT7 OR PHI7 triggers were possible.

peressounko (approved):

peressounko:
+1

PR #23225, opened 2024-02-17
PWGGA/GammaConv: Fix `fHistoMotherBackInvMassECalib` for ConvCalo

mhemmer-cern:
- Swapped `currentEventGoodV0` and `previousGoodV0` when filling `fHistoMotherBackInvMassECalib` and `fHistoMotherBackInvMassECalibPCM` to match the expected type. This was wrong for mixing current clusters with previous conversion photons.


jokonig:
+1

PR #23237, opened 2024-02-21
[PAG-Flow] changed place to fill Nch-centrality histogram and add two histograms to test Nch distribution

chlorine1777:
I found that we should consider the difference it makes when filtering the 0-denominator of higer order pt. So I changed place to fill Nch-centrality histogram to make its entries consistent with pt, and add two histograms to test how Nch distribution change after filtering the 0-denominator.


yzhouCERN:
+1

PR #23239, opened 2024-02-21
Removed pT rebinning from wt_res_eec and wt_res_e3c, moved histograms…

anabananana98:
… around in UserCreateOutputObjects


raymondEhlers:
+1

PR #23245, opened 2024-02-22
PWGGA/GammaConv: add two histos for measuring the occupancy of the TPC

sstiefel19:


sstiefel19:
picked two commits by accident

PR #23251, opened 2024-02-23
PWGGA/GammaConv: Reduce Memory for PbPb

mhemmer-cern:
- changed minv and pT binning for PbPb 5.02 TeV for GCoCa and GCa
- excluded EventCutList, ClusterCutList and MesonCutList from Output if enableLightOutput > 1


jokonig:
+1

PR #23253, opened 2024-02-24
Added resolution and match histograms as a function of truth level je…

anabananana98:
…t pT, changed bad binning of the wtnojet_e3c histograms, added if statements for flags


raymondEhlers:
+1

PR #23261, opened 2024-02-26
[PAG-Flow] using Ntracks from MC Truth or MC Reco 

zhaomr13:
Investigate the difference between using Ntracks from MC Truth or MC Reco for Monte Carlo


yzhouCERN:
+1

PR #23269, opened 2024-02-27
PWGGA/GammaConv: Gamma(Conv)Calo Add PbPb 2015 configs

mhemmer-cern:
- Added trainconfigs for LHC15o PbPb similar to the LHC18qr traincnofigs except no trigger for 0-10 and 30-50% centralities


jokonig:
+1

PR #23280, opened 2024-03-02
- Correct projection Cfs

makor:
- Add QA for 2B Cfs
- Allow pT selection of rhos for optimisation
- Add option for much finer pT binning of M(V0) vs pT
- Update Addtask


rdelgran:
+1

rdelgran:
+1

PR #23282, opened 2024-03-04
Add histograms to check the vertices in MC

JaeYoonCHO:
Several histograms were added for checking the primary vertex in MC, and use MC vertex for creating 3Prong object


fcatalan92:
+1

PR #23294, opened 2024-03-07
PWGGA/GammaConv: Gamma(Conv)Calo - fixed `fHistoTrueClusGammaEResE` y…

mhemmer-cern:
…-axis

- fHistoTrueClus(Photon)GammaEResE y-axis and hence filling was changed from (E_rec-E_true)/E_true to (E_rec-E_true)/E_rec, since 1 - (E_rec-E_true)/E_rec = E_true/E_rec which is what we want as a factor to correct the cluster energy.
- Also made a change to ensure the minv binning in PbPb is not made even smaller in GammaConvCalo when using the lightout option 2


fjonasALICE:
+1

PR #23310, opened 2024-03-16
update to k0sKch efficiency task

asasikum:
Added a few setters in the header to make the systematics check easier, using V0M instead of RefMult as centrality estimator, changed some binning


majanik:
+1

PR #23328, opened 2024-03-21
Fix errors when building in debug mode

mhemmer-cern:
This PR aims to fix all error messages that appear when building in debug mode and would hence prevent building in debug mode.

**PWG/EMCAL/EMCALbase/AliEMCALRecoUtils.cxx:**
- remove unsused variable `bool behindSupport = false;` PWG/EMCAL/EMCALbase/AliTrackContainer.cxx:
- comment out unsused varaible `// int nmatch(0);`
---
**PWG/EMCAL/EMCALtasks/AliAnalysisTaskEmcalTriggerPatchClusterMatch.cxx:**
- comment out unused variables `Int_t nG1 = 0, nG2 = 0, nJ1 = 0, nJ2 = 0, nL0 = 0;` 
---
**PWG/EMCAL/EMCALtasks/AliEmcalAodTrackFilterTask.cxx:**
- changed order of variables in constructor to match order in header file
- Fixed Warning message to get message via `"%s"` to be secure (Error: format string is not a string literal) 
 ---
**PWG/EMCAL/EMCALtasks/AliEMCALClusterParams.cxx:**
- remove unused variables `ncell` and `ncell1` 
---
**PWG/EMCAL/EMCALtasks/AliEmcalPatchFromCellMaker.cxx**:
- remove unsused variable `sum` and a double nested loop which was not needed since it only added values to `sum`, but as stated `sum` is unused.
---
**PWG/EMCAL/EMCALtrigger/AliEmcalTriggerDecisionContainer.cxx:**
- change for loop from access by value, to access by reference `for(const auto& trgclass : selectclasses) {` 
---
**PWG/Tools/THistManager.h:**
- change `iterator` because std::iterator is deprecated
---
**PWG/Tools/yaml-cpp/include/yaml-cpp/node/detail/iterator.h:**
- change `iterator_base` because std::iterator is deprecated 
---
**PWG/Tools/yaml-cpp/include/yaml-cpp/node/detail/node_iterator.h:**
- change `node_iterator_base` because std::iterator is deprecated

### DIFF
--- a/PWG/EMCAL/EMCALbase/AliEMCALRecoUtils.cxx
+++ b/PWG/EMCAL/EMCALbase/AliEMCALRecoUtils.cxx
@@ -490,7 +490,6 @@ Bool_t AliEMCALRecoUtils::AcceptCalibrateCell(Int_t absID, Int_t bc,
     //____________________________________
     } else if (fUseAdditionalScale == 4) { // Run1 special values: SM with TRD in front (+ with/without Support structure) and SM without TRD in front (+ with/without Support structure)
       Int_t iCol = ieta;
-      bool behindSupport = false;
       // select columns with TRD support in front
       if( (imod > 11 && imod < 18) && imod%2) iCol+= 65;
       else if (imod%2) iCol+=49;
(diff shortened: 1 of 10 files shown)


mfasDa:
+1

PR #23331, opened 2024-03-27
fixed initialization of the variable binning vectors

sweyh99:
The previous initialization caused run-time errors


raymondEhlers:
+1

PR #23335, opened 2024-03-30
PWGCF: removing pidresponse as dep

maroozm:
related to recent memory issues in CF lego trains, where PIDResponse is a dependency.


abilandz:
+1

PR #23339, opened 2024-04-03
add information on generated out-of-bunch pile up in McParticle::fFlag

maciacco:


jgrosseo:
This can be added and then also needs to be added here: https://github.com/jgrosseo/AliceO2/blob/dev/Framework/Core/include/Framework/DataTypes.h#L80 
Then this bit is also reserved for Run 3 of course...

jgrosseo:
+1

PR #23341, opened 2024-04-03
PWGGA/GammaConv: TaskGamma(Conv)Calo: Addition of fHistoTruePi0(Eta)I…

mhemmer-cern:
…nvMassECalib

- Add `fHistoTrueEtaInvMassECalib` and `fHistoTruePi0InvMassECalib` histograms to the Calo task, in ConvCalo only added `fHistoTrueEtaInvMassECalib` since Pi0 one was already there.
- Added comment to filling of one of the ECalib histograms in the Calo task to explain why we use the energy of the meson candidate and not the energy of one of the photons directly, since this is not straight forward and might cause some confusion.
- Tweaked the flag so that these histograms only require the `fDoECalibOutput` flag and nothing else. Previously they sometimes could be active for a different set of flags which results in a very uneasy to follow structure.


jokonig:
+1

PR #23347, opened 2024-04-08
Fixes for JE trains

sweyh99:
Fixed pointer initialization to work with JE trains. The passed variable histogram binning wasn't working either, so changed to hard-coded binning.


raymondEhlers:
+1

PR #23348, opened 2024-04-08
On-The-Fly-Code for PYTHIA8 with rope 

wenyaCern:
Copy the "getMCEvent()" function from the marco which uses "the PYTHIA8 with rope" data for the on-the-fly system
Added some switches to call the function


yzhouCERN:
+1

PR #23356, opened 2024-04-16
PWGGA/GammaConv: AddTaskGammaConvCalo_PbPb.C fix config 2566

mhemmer-cern:
- train config 2565 was there two times, this is now fixed by properly changing one to 2566


jokonig:
+1

PR #23366, opened 2024-04-17
Create a new class for corr flow with FMD for on-the-fly trains 

wenyaCern:
We create a new class for corr flow measurement with FMD for on-the-fly trains. In order to avoid the conflict between data and on-the-fly setup.


yzhouCERN:
+1

PR #23380, opened 2024-04-22
PWGGA/PHOSTasks/PHOS_Run2: bugfix + added nonlin and weights for MC

vlad-kuskov:
Minor changes:
- added the reaction plane selection (for mixed events)
- added option for min bias PHOS clusters selection at low energies
- fixed PID criteria for PIDCutEff histos
- Mgg in PHOS M1 is lower than in other modules -> added option for energy correction in M1


peressounko:
+1

PR #23389, opened 2024-04-24
fixed bug in the task

sashingo:


raymondEhlers:
+1

Since I think this is for your student, just note that this change prevents the use of subwagons on the LEGO train. To preserve that functionality, the suffix needs to be the last argument. If you don't need this (which I'm assuming), then it's as fine as is

PR #23393, opened 2024-04-24
PWGGA/GammaConv/TaskGamma(Conv)Calo: adjust flags for ECalib histos

mhemmer-cern:
- Cluster QA flag > 0 now also enables the Cluster energy calib histograms instead of only the fDoECalib flag

- After changing the normalization for the cluster energy resolution from E_true to E_rec this commit changes the ranges for the resolution from (-1,+5) to (-5,+1)


jokonig:
+1

PR #23410, opened 2024-04-29
Memory fixes for JetHadro AOD and ESD tasks

sweyh99:
Deleting a pointer after its use in a loop and updating the ESD task with previous AOD changes


raymondEhlers:
+1

PR #23440, opened 2024-05-13
Add jet angular correlations task

ljoergen:


ljoergen:
Third time's the charm. Hopefully fastjet/PseudoJet.hh will be found now that it's in the source file, not the header.

bdoenigu:
+1

bdoenigu:
+1

ljoergen:
alibuild keeps complaining various FastJet files can't be found while checking but the task runs both on the Grid and locally, and the error isn't even in my own code but in FJ_includes.h, which _I_ didn't write. How can that be?

PR #23442, opened 2024-05-14
[Fix] Solve error due to QA pair settings and fix pt selection of pions

danielbattistini:
Revert "Clean the code and use the correct systematic variations"

This reverts commit e8e4fc4ccaa29dee3ed032af1848de2082fe1f53.

Remove unused parameters

Update systematics and fix pt selection of pions


sartozza:
+1

PR #23448, opened 2024-05-15
PWGGA/GammaConv/macros: AddTask PbPb - change base train config 750

mhemmer-cern:
- Change the base train config `750` for PbPb 5.02 TeV for EMC and PCM-EMC to use the new cluster and meson cut string. The config is still 00-90% without using PileUp Removal, only using the MB header in MC and standard |Vz| < 10cm


jokonig:
+1

PR #23449, opened 2024-05-15
PWGGA/GammaConv: Modifications to the GammaIsoTree

nstrangm:
- Add two leafs that contain the mass when pairing the cluster closest to the pi0 and eta (Tagging)
- Add the central event selection when analyzing jets
- Add trainconfigs for 13TeV analyses
- Seperate the gen and rec pT cuts for jets from those for gammas


jokonig:
+1

jokonig:
+1

PR #23450, opened 2024-05-16
PWGGA/GammaConv/macros: AddTask_Gamma(Conv)Calo - Adjust base config …

mhemmer-cern:
…750 to be V0AND

- the trainconfig 750 was using KINT7 | kCent | kSemiCent, but for QA purposes this is changed to V0AND now, so we only have MBs events


jokonig:
+1

PR #23453, opened 2024-05-17
PWGGA/GammaConv: Change IsoGammaTree container name

nstrangm:
- Add possible additionalFileEnding to the container names to enable running of multiple Jet settings in the same Train


jokonig:
+1

PR #23454, opened 2024-05-17
PWGGA/GammaConv: AddTask G(Co)Ca - Add HeaderList for PbPb MCs

mhemmer-cern:
- To be able to only select MB header from the PbPb MCs we need the HeaderList which is given in the AddTask. This commit adds the HeaderList containing only Hijing (MB header) for the MC productions starting with `LHC18e1`, `LHC19h2` and `LHC18l8`. For the productions `LHC20j6`, `LHC20e3` and `LHC22b` Hijing and Pileup are added as HeaderList, since those productions contain simulated Pile Up.
- Some minor code style changes, especially changing the indentation of some comments so IDEs can properly collapse {} again...


jokonig (approved):

jokonig:
+1

PR #23470, opened 2024-05-24
PWGGA/GammaConv/: add functionality for calculating pT-weights in variant (=!invariant) mode

sstiefel19:
As is, this does not change any functionality. As is it only adds one member function 'Float_t AliConvEventCuts::GetWeightForMesonNew...)' without ever calling it. 
The function allows to compute the pt weights from variant spectra in contrast to invariant spectra that were used up to now. 
 
 There will be another commit in short, that will add functionality for calling the function.


sstiefel19:
close because unwanted formatting changes.

PR #23471, opened 2024-05-24
PWGGA/GammaConv: add functionality for calculating pT-weights in variant (=!invariant) mode

sstiefel19:
As is, this change does not alter any functionality. It only adds one member function 'Float_t AliConvEventCuts::GetWeightForMesonNew...)' without ever calling it. The function allows to compute the pt weights from variant spectra in contrast to invariant spectra that were used up to now.  There will be another commit in short, that will add functionality for calling the function.

### DIFF
--- a/PWGGA/GammaConv/macros/AddTask_GammaConvV1_PbPb.C
+++ b/PWGGA/GammaConv/macros/AddTask_GammaConvV1_PbPb.C
@@ -54,7 +54,7 @@ void AddTask_GammaConvV1_PbPb(
   Bool_t    enableChargedPrimary          = kFALSE,
   Bool_t    enablePlotVsCentrality        = kFALSE,
   Bool_t    processAODcheckForV0s         = kFALSE,   // flag for AOD check if V0s contained in AliAODs.root and AliAODGammaConversion.root
-  
+  bool      theUseGetMesonWeightNew       = false,
   // subwagon config
   TString   additionalTrainConfig         = "0")       // additional counter for trainconfig + special settings
   {
@@ -103,8 +103,8 @@ void AddTask_GammaConvV1_PbPb(
     return ;
   }
 
-  if (doWeightingPart<0) {
-    Error(Form("%s_%i", addTaskName.Data(),  trainConfig), "No negative values allowed for doWeightingPart");
+  if (acceptedAddedParticles<0) {
+    Error(Form("%s_%i", addTaskName.Data(),  trainConfig), "No negative values allowed for acceptedAddedParticles");
     return;
   }
 
@@ -3795,6 +3795,8 @@ void AddTask_GammaConvV1_PbPb(
       cuts.AddCutPCM("11310023", "0d200009ab770c00amd0404000", "0152101500000000"); // 10-30%
       cuts.AddCutPCM("13530023", "0d200009ab770c00amd0404000", "0152101500000000"); // 30-50%
       cuts.AddCutPCM("15910023", "0d200009ab770c00amd0404000", "0152101500000000"); // 50-90%
+  } else if (trainConfig == 998){ //____________________-___
+      cuts.AddCutPCM("10130023", "0d200009ab770c00amd0404000", "0152101500000000"); // 0-10%
 //****************************************************************************************************
 
   } else if (trainConfig == 1001){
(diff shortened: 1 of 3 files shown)


PWGGA/GammaConv/macros/AddTask_GammaConvV1_PbPb.C
```diff
@@ -4310,7 +4314,7 @@ void AddTask_GammaConvV1_PbPb(
     TString mcInputMultHisto    = "";
     if (enableMultiplicityWeighting){
       cout << "INFO enableling mult weighting" << endl;
-      if(periodNameAnchor.CompareTo("LHC15o")==0 || periodNameAnchor.CompareTo("LHC18q")==0){
+      if(periodNameAnchor.BeginsWith("LHC15o")==0 || periodNameAnchor.CompareTo("LHC18q")==0){
```
sstiefel19, line 4317: remove ==0

sstiefel19 (commented):
looks good

PWGGA/GammaConv/macros/AddTask_GammaConvV1_PbPb.C
sstiefel19: done

sstiefel19:
Get...WeightOld and Get..WeightNew run without substantial differences.
some bins deviate at the order of % which will be rare corner cases. I still need to investigate which of the two methods is correct in those.
Ready to be merged and run on the grid. Tested locally extensively.

PWGGA/GammaConv/macros/AddTask_GammaConvV1_PbPb.C
```diff
@@ -4337,106 +4341,120 @@ void AddTask_GammaConvV1_PbPb(
...
+          periodNameAnchor.BeginsWith("LHC18q"))
+      {
+        TString eventCutString = cuts.GetEventCut(i);
+        TString eventCutShort = eventCutString(0, 6);                                                // first six digits
+        histoNameMCPi0PT = Form("Pi0_%s_5TeV_%s", periodNameV0Reader.Data(), eventCutString.Data()); // MC
+        fitNamePi0PT = Form("Pi0_Data_5TeV_%s", eventCutShort.Data());                               // fit to data
+        histoNameMCEtaPT = Form("Eta_%s_5TeV_%s", periodNameV0Reader.Data(), eventCutString.Data());
+        fitNameEtaPT = Form("Eta_Data_5TeV_%s", eventCutShort.Data());
       }
-      analysisEventCuts[i]->SetUseReweightingWithHistogramFromFile(weightPi0, weightEta, weightK0s, fileNamePtWeights, histoNameMCPi0PT, histoNameMCEtaPT, histoNameMCK0sPT, fitNamePi0PT, fitNameEtaPT, fitNameK0sPT);
```
jokonig, line 4355: Why were the different weights, like "weightPi0" changed to "intPtWeightsCalculationMethod" and the K0s was disabled completely?

sstiefel19: 1) the semantics of  weightPi0 and alike do not change. Unless I oversee sth. Which change do you see?
2) I renamed the macro argument parameter 'enablePtWeighting' -> 'intPtWeightsCalculationMethod' because I also changed its type from bool to int, because Im using it steer if  invariantly or variantly calculated weights shall be used. I should have used a separate commit just for the renaming and type change of this variable to make this more clear, but forgot about it in the beginning. Sorry for that.
3) Note: 
 -  that the lines here are in the scope of 'if(periodNameAnchor.BeginsWith("LHC15o") || periodNameAnchor.BeginsWith("LHC18q"))'. Meike and I are not using any weights for K0s, thats why I don't switch them on.
 - that also in the current version of the code all the weights are initialised as off and then afterwards, depending on anchored data and mc tags switched on.

PWGGA/GammaConvBase/AliConvEventCuts.cxx
```diff
@@ -7808,10 +7897,153 @@ Float_t AliConvEventCuts::GetWeightForMultiplicity(Int_t mult){
...
+        mesonPt = aodMCParticle->Pt();
+        PDGCode = aodMCParticle->GetPdgCode();
+      }
+      else
+      {
+        return 0;
+      }
+    }
+    return 1;
+  };
```
mhemmer-cern, line 7990: Do we need this lambda function if it is only called once directly after its definition?

PWGGA/GammaConvBase/AliConvEventCuts.cxx
```diff
@@ -7808,10 +7897,153 @@ Float_t AliConvEventCuts::GetWeightForMultiplicity(Int_t mult){
...
+                                 : 0 // to signal problem
+                           : -1.;    // to signal problem      
+    // will reset to 1 and throw a warning if weight is not >=0 and finite
+    return checkSanitizeAndReturnWeight(lWeight);
+  };
+
+  bool lCaseEtaPi0 = (PDGCode == 111) || (PDGCode == 221);
+  double lResult = lCaseEtaPi0
+             ? calcWeight()
+             : checkSanitizeAndReturnWeight(lDenomMC);
```
mhemmer-cern, line 8037: So if the PDGCode is neither Pi0 nor eta, why do we need to run `checkSanitizeAndReturnWeight(lDenomMC)`? why not directly return 0 or something?

mhemmer-cern (commented):
Just a few comments from my side to the code

PWGGA/GammaConvBase/AliConvEventCuts.cxx
```diff
@@ -1041,6 +1064,71 @@ void AliConvEventCuts::LoadGammaPtReweightingHistosMCFromFile() {
...
+    }
+    return true;
+  };
+
+  // execution starts here
+  AliInfo("AliConvEventCuts::InitializeMapPtWeightsAccessObjects(): start.\n");
+  bool lSuccess = true;
+  lSuccess &= calculateVariantSpectraAndInsert(111, fDoReweightHistoMCPi0, *fFitDataPi0_inv, *hReweightMCHistPi0_inv);
+  lSuccess &= calculateVariantSpectraAndInsert(211, fDoReweightHistoMCEta, *fFitDataEta_inv, *hReweightMCHistEta_inv);
+  lSuccess &= calculateVariantSpectraAndInsert(310, fDoReweightHistoMCEta, *fFitDataEta_inv, *hReweightMCHistEta_inv);
```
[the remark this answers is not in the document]
sstiefel19: yes, thanks! That was a provisoric leftover. I fixed it now.

PWGGA/GammaConvBase/AliConvEventCuts.cxx
sstiefel19: its not strictly needed (as any function) but it's still good to have in my opinion for readability (it labels this part of the code, and allows to see directly which variables are needed in this section of code). Admittedly, the last point is   abolished by me lazily capturing everything (see the comment above). Its on my list to correct but does not do any harm the way it is.
Set aside, these lines we see many times all over our GammaConv Code base, so in priniciple the right thing would be to move into a private or static member function. Which is also on my list (with a low priority)

PWGGA/GammaConvBase/AliConvEventCuts.cxx
sstiefel19: good catch. This was a remnant from before the early exit in case particle not found in map. Back then, I used it to issue a warning and reset to 1 now indeed it was completely superflous. I fixed it, thanks :)

PWGGA/GammaConv/macros/AddTask_GammaConvV1_PbPb.C
```diff
@@ -54,7 +54,7 @@ void AddTask_GammaConvV1_PbPb(
   Bool_t    enableChargedPrimary          = kFALSE,
   Bool_t    enablePlotVsCentrality        = kFALSE,
   Bool_t    processAODcheckForV0s         = kFALSE,   // flag for AOD check if V0s contained in AliAODs.root and AliAODGammaConversion.root
-  
+  bool      theUseGetMesonWeightNew       = false,
```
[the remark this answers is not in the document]
sstiefel19: We don't strictly need it you're right. It serves as an additional master switch on top of the int variable you mention. The idea was to preserve complete same code execution as before this PR for everthing but if you explicitly switch it on (which is only done from this addtask, where it default to false)

PR #23489, opened 2024-06-03
Femto:

makor:
-Set-up required to estimate the true MC background for the rho0 -Use clang formatting for the all the new commits
-Include new options in the AddTask for the MC closure test and rho0 comb. background in data -Include functionality to reconstruct a fake V0 using same charges in order to estimate comb. background in data


rdelgran:
+1

PR #23492, opened 2024-06-04
MC PDG study

Marcellocosti:
Implemented TH3F to store kstar, PDGpart1, PDGpart2 for MC SE distributions Common and NonCommon ancestor and MC ME distribution. Other histograms are not filled with this configuration to reduce the memory consumption and the runtime.


sartozza:
+1

PR #23493, opened 2024-06-04
PWGGA/GammaConv: add weights for AOD for eta prime jj MC, add skimmin…

EwaGlimos:
…g to reduce running time


FriederikeBock:
Skimmed output might not be the right terminology for what are doing... maybe "optimize runtime"?

FriederikeBock:
+1

PR #23494, opened 2024-06-04
PWGJE/EMCALJetTasks/AliAnalysisTaskScale Adding 3xEMC and manual set of EMCal max phi

aschmier:
PWGJE/EMCALJetTasks/AliAnalysisTaskScale Adding 3xEMC and manual set of EMCal max phi


mfasDa:
+1

PR #23501, opened 2024-06-07
PWGGA/GammaConv: Updated MC Header handling

mhemmer-cern:
- New bool variable `fHasMBNotFirst` to check if the minimum bias head is NOT first in the list of headers. If that is the case it will be set in `SetPeriodEnum` to true for the corresponding MC. This is currently only impemented for the few known Hijing MCs which are from 2020 or newer. Everyone whos working with MCs that include pile up need to check this and possibly set the variable to true for their used MC.
- Fixed a bug which made it necessary to have a headerlist in the AddTask when running only min bias header or exclude pile up. (See changes to code where `fHeaderMap` is used)
- Thanks to that bug fix some header lists in the AddTask could be removed.


jokonig:
+1

PR #23542, opened 2024-06-25
PWGGA/GammaConv: TaskGamma(Conv)Calo - Increase PbPb V0 Mult max x

mhemmer-cern:
- Change x axis for V0 Mult histogram from (3000, 0, 30000) to (2000, 0, 40000) which is identical to the ConvV1 task. This change was necessary since we see mults larger than 30000.


FriederikeBock:
+1

PR #23554, opened 2024-07-01
New task for Xi and Omega analysis in jets

eeremenk:
The new task for cascades, in principle, is the same as the StrangenessInJets task but analyzes cascades.


raymondEhlers:
+1

PR #23557, opened 2024-07-02
PWGGA/GammaConv: Fix particleFromBGEvent bug

mhemmer-cern:
Currently particles which are coming from the MB header where marked as 2 from `IsParticleFromBGEvent`. For some reason particles that have 2 returned from that function which is only the case if they are MB AND ACCEPTED, they are marked as `fIsFromDesiredHeader = false`, so not from desired Header. This seems outdated, so I removed the case where particles from MB header are marked as `fIsFromDesiredHeader = false`


jokonig:
+1

PR #23582, opened 2024-07-18
PWGGA/GammaConv: Add QA trainconfigs for PbPb cent and semicent

mhemmer-cern:
- Add new trainconfigs for cent and semi cent (triggered for 18qr and not triggered for 15o)


jokonig:
+1

PR #23584, opened 2024-07-19
PWGGA/GammaConv: Add more PbPb QA train configs

mhemmer-cern:
- Add trainconfigs using predefined MC track mult for cents.
- Add trainconfig for MC from 0-15% to see the influence of the 5% and if this could help with staistics


jokonig:
+1

PR #23586, opened 2024-07-21
Adding MC id to tree_Particle and tree_Particle_gen in HF trees

nambrath:
Including MC particle ID information during tree production (for unfolding)


nambrath:
This is a pretty minor but somewhat urgent change -- we need to regenerate ntuples to finalize some analyses for Hard Probes. Would appreciate a quick approval!

jezwilkinson:
+1

PR #23594, opened 2024-07-24
ML4PhotoConv: PCM XGB on grid

an15ms157:


PWGGA/GammaConv/AliAnalysisTaskGammaConvV1.h
```diff
@@ -498,7 +501,12 @@ class AliAnalysisTaskGammaConvV1 : public AliAnalysisTaskSE {
     Float_t                           fEta_EPos_Gamma1_MesonML;  
     Float_t                           fPAsymmetry_EPos_Gamma1_MesonML;  
 
-    TList**                           fMLFolder;                                  //End ML Variables
+    TList**                           fMLFolder;                                 
+    
+    AliGammaMLResponseGammatoEPlusEMinus* fMLResponse = nullptr;  //!<! object to handle ML response
+    TString fConfigPath   = "";                     /// path to ML config file
```
jokonig, line 507: I think you need to increase the classdef by one.
Because you add a new persistent object to the class, so the streamer has to know that it needs to stream the value of this variable. This can work in local tests but then fails on the grid

jokonig:
+1

PR #23602, opened 2024-07-26
PWGGA/GammaConv: remove external pT dependent flag for NDM task, inco…

EwaGlimos:
…rporate it into cutstrings


FriederikeBock:
+1

PR #23623, opened 2024-08-08
PWGGA/GammaConv: fix a pt weights related bug that I introduced with my latest commit.

sstiefel19:
The effect of the bug was, that for the MBMC productions for LHCh15o and LHC18qr pt weighting could not be enabled because for these MCs we don't explicitly name the accepted particle headers (since we take all particles)


jokonig:
+1

PR #23626, opened 2024-08-11
PWGJE: Improve getting underlying event, 

MartinSchmidtTuebingen:
check perpendicular cone for collision with reconstructed jet


raymondEhlers:
+1

PR #23630, opened 2024-08-15
Added Mgg histos + minor changes

vlad-kuskov:
- added histos for Mgg distributions in real and mixed events (both for the general analysis task and embedding)
 - added PHOS M4Exclude option for embedding


vlad-kuskov:
Hello?

kharlov:
+1

PR #23632, opened 2024-08-16
add track extras for run 2 + store SPD tracklet flag for free vs. shared clusters

maciacco:


RUN3/AliAnalysisTaskAO2Dconverter.cxx
```diff
@@ -163,7 +164,8 @@ const TString AliAnalysisTaskAO2Dconverter::TreeTitle[kTrees] = {
   "HF D* candidates",
   "O2 HepMc Cross Sections",
   "O2 HepMc Pdf Info",
-  "O2 HepMc Heavy Ion"
+  "O2 HepMc Heavy Ion",
+  "Barrel trackes Extra Run 2"
```
jgrosseo, line 168:
```suggestion
  "Barrel tracks Extra Run2"
```

RUN3/AliAnalysisTaskAO2Dconverter.cxx
```diff
@@ -2031,17 +2045,19 @@ void AliAnalysisTaskAO2Dconverter::FillEventInTF()
         tracks.fRho1PtSnp = 0;
         tracks.fRho1PtTgl = 0;
         tracks.fTPCinnerP = NAN;
-        tracks.fFlags = 0;
+        tracks.fFlags = mlt->FreeClustersTracklet(itr, 0) && mlt->FreeClustersTracklet(itr, 1);
```
jgrosseo, line 2048: The content of tracks.fFlags follows the enum TrackFlagsRun2Enum in DataTypes.h (O2). Please prepare also a comment change there, or even simply a second name with the same enum value (0x1).

jgrosseo (changes requested):
Thanks! Generally fine. Please see my minor comments.

RUN3/AliAnalysisTaskAO2Dconverter.cxx
maciacco: @jgrosseo thanks for the review. As you suggested, I added a new name specifically for tracklets in O2 (also in AliPhysics for completeness), here is the link to the PR: https://github.com/AliceO2Group/AliceO2/pull/13393

jgrosseo:
+1

jgrosseo (approved):

jgrosseo:
I approved to early as the change in the O2 PR may need changes here. I close it to avoid the automatic merging (which I cannot stop now anymore). We can reopen in case we decide no changes are needed.

PR #23641, opened 2024-08-22
Added code to use embedding framework

sweyh99:
Added capabilities for hybrid, reco, and truth jets
Adding jet matching
Saved all information needed as histograms to calculate UE bias scaling factor
Saved all JES and JER info
Saved distributions of pi,k,p as fxn of jet pT


raymondEhlers:
+1

There's a lot of helper functionality that would have simplified this code, but since it's already done, no need to rewrite

PR #23643, opened 2024-08-23
Add option to select jets with reduced EMCAL fiducial accpetance

mfasDa:
Study of the impact of edge effects in the EMCAL


mfasDa (commented):
+1

raymondEhlers:
+1

(I think you unintentionally reviewed rather than commented, which as I recall, the system doesn't pick up)

mfasDa:
Thanks @raymondEhlers ! This happens from time to time when I comment from within VSCode

For the record: Looking at the constituent distribution we see lower counts at the detector edges, particularly in eta but also in phi, and we suspect this to come from edge effects at the detector boundaries, consequently we want to cross check with a reduced acceptance where we are sure the cluster reconstruction is not biased.

PR #23650, opened 2024-08-26
PWGJE/EMCALJetTasks/Tracks/AliAnalysisTaskEmcalJetEnergySpectrum Fixing rho scale type bool -> double

aschmier:
PWGJE/EMCALJetTasks/Tracks/AliAnalysisTaskEmcalJetEnergySpectrum Fixing rho scale type bool -> double


mfasDa:
+1

PR #23659, opened 2024-09-03
Added K0/h analysis task (Run 2)

rhanniga:
This PR contains my (somewhat) new analysis task for investigating the K0/h ratio in jets and UE using azimuthal correlations.


lhusova:
+1

PR #23684, opened 2024-09-23
PWGGA/PHOSTasks: Changed mixed events selection + added Mgg histos

vlad-kuskov:
- Changed binning of mixed events container: added bins on the centrality and reaction plane
 - Added Mgg histos for embedding and MC


kharlov:
+1

PR #23733, opened 2024-10-17
PWGGA/GammaConv:  AddTask_GammaConvV1_PbPb.C systematics

sstiefel19:
add trainconfigs for systematics


jokonig:
Hi Stephan, 
just out of curiosity, would it be better to arrange the cuts according to the variation and not having different centralities in one trainconfig? Because if I remember correctly, the PbPb MCs are different for 0-10 and 30-50%, so it would not be good having them in the same trainconfig.

sstiefel19:
Hi Joshua,
yes, that indeed would be better :) Thanks a lot for your comment! I'll change it.

PR #23737, opened 2024-10-22
Improve AliFJWrapper/Add Functionality To SoftDropUtility/Add to analysis task

MartinSchmidtTuebingen:
Replace SoftDrop in AliFJWrapper with RecursiveSoftDrop, standard value for recursive depth is N=1 which corresponds to standard soft drop. Add settings for recursive depth, zCut and beta in the soft drop functionality

Recursive Soft Drop with N=1 corresponds to standard soft drop: https://phab.hepforge.org/source/fastjetsvn/browse/contrib/contribs/RecursiveTools/trunk/README

### DIFF
--- a/PWGJE/EMCALJetTasks/AliEmcalJetUtilitySoftDrop.h
+++ b/PWGJE/EMCALJetTasks/AliEmcalJetUtilitySoftDrop.h
@@ -34,6 +34,9 @@ class AliEmcalJetUtilitySoftDrop : public AliEmcalJetUtility
   void Prepare(AliFJWrapper& fjw);
   void ProcessJet(AliEmcalJet* jet, Int_t ij, AliFJWrapper& fjw);
   void Terminate(AliFJWrapper& fjw);
+  void SetZCut(Double_t zCut) {fZCut = zCut;}
+  void SetBeta(Double_t beta) {fBeta = beta;}
+  void SetRecursiveDepth(Int_t n) {fRecursiveDepth = n;}
 
  protected:
 
@@ -44,6 +47,10 @@ class AliEmcalJetUtilitySoftDrop : public AliEmcalJetUtility
   TString                fRhomName;                           ///< name of rhom
   Double_t               fRho;                                ///< pT background density
   Double_t               fRhom;                               ///< mT background density
+    // condition to stop the grooming (rejection of soft splitting) z > fZCut theta^fBeta
+  Double_t               fZCut;                              //< fZCut = 0.1                
+  Double_t               fBeta;                              //< fBeta = 0
+  Int_t                  fRecursiveDepth;                    //< 0: No Soft Drop; 1: Apply once; -1: Apply until the condition stops the grooming
 
   TClonesArray          *fGroomedJets;                        //!<! groomed jet collection
   TClonesArray          *fGroomedJetParticles;                       //!<! groomed particle collection
--- a/PWGJE/EMCALJetTasks/AliEmcalJetUtilitySoftDrop.h
+++ b/PWGJE/EMCALJetTasks/AliEmcalJetUtilitySoftDrop.h
@@ -57,6 +57,6 @@ class AliEmcalJetUtilitySoftDrop : public AliEmcalJetUtility
   AliRhoParameter       *fRhoParam;                           //!<! event rho
   AliRhoParameter       *fRhomParam;                          //!<! event rhom
 
-  ClassDef(AliEmcalJetUtilitySoftDrop, 1) // Emcal jet utility that implements the constituent subtractor form the fastjet contrib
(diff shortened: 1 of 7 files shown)


raymondEhlers:
Please gimme a couple of days to check on this (and ask for someone else to look it over). It looks fine at first glance, but changing something so deep in the jet framework needs careful checks

If you don't hear anything in a few days, please ping me

MartinSchmidtTuebingen:
Sure. If you feel safer, I can also introduce the recursive soft drop functionality as its own function. Just felt a bit redundant.

PWGJE/EMCALJetTasks/AliFJWrapper.h
```diff
@@ -143,7 +143,9 @@ class AliFJWrapper
   void SetEventSub(Bool_t b) {fEventSub = b;}
   void SetMaxDelR(Double_t r)  {fMaxDelR = r;}
   void SetAlpha(Double_t a)  {fAlpha = a;}
-
+  void SetRecursiveDepth(Int_t n) {fRecursiveDepth = n;}
+  void SetZCut(Double_t zCut) {fZcut = zCut;}
+  void SetBeta(Double_t beta) {fBeta = beta;}
```
raymondEhlers, line 148: I dug a little but can't find the answer yet, so in case you found out - we've e.g. varied the zcut in other measurements, so how did we use this in the past?

raymondEhlers (commented):
Agreed that implementing it separately seems like a waste - I'm just worried given the late stage of AliPhysics that there may e.g. be a bug or difference in the RecursiveSoftDrop code vs the standard SoftDrop object. Let's see if something comes up when Florian reviews

PWGJE/EMCALJetTasks/AliFJWrapper.h
MartinSchmidtTuebingen: I have no clue, I also try to find it. I guess everyone did it with using manual code. Also, the AliEmcalJetUtilitySoftDrop class is, in my opinion, not really useful for the SoftDrop analyses I found (at least it is not useful for my analysis, since it does not update the jet content - a second MR will follow after this one adding this functionality)

raymondEhlers: Indeed, I implement it by hand in my analysis. Thanks for the info - it sounds like for your analysis your interested in grooming vs the usual tagging mode. Okay, let's see what comes of the other review. Perhaps the existing use of this class is actually pretty limited

MartinSchmidtTuebingen:
I have added the functionality for the SoftDropUtility class directly in this MR

raymondEhlers:
LGTM, thanks. @fjonasALICE , when you get a moment, can you please take a look?

MartinSchmidtTuebingen:
I have also tested it locally, it works (and produces different results depending on the recursive depth, with it corresponding to the old code for n=0 and n=1)

MartinSchmidtTuebingen:
Also added changes to my analysis task to the MR

PWGJE/EMCALJetTasks/AliEmcalJetUtilitySoftDrop.h
fjonasALICE, line 1: missing increase of   ClassDef(AliEmcalJetUtilitySoftDrop, 1) to  ClassDef(AliEmcalJetUtilitySoftDrop, 2)

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskIDFragmentationFunction.h
fjonasALICE, line 1: missing increase of ClassDef

fjonasALICE (changes requested):
Hey, in my opinion, while the code is deep in the framework, I do not see it really used anywhere in AliPhysics. If you ensured that RecursiveDepth=1 is identical to the previous version, the proposed changes are fine by me.

However, I commented on a few missing ClassDef increases, which need to ba addressed. Otherwise, when running on the grid you will encounter problems in the streaming of the newly introduced vairables.

After these changes, good to go from my side

PWGJE/EMCALJetTasks/AliEmcalJetUtilitySoftDrop.h
MartinSchmidtTuebingen: Done

PWGJE/EMCALJetTasks/UserTasks/AliAnalysisTaskIDFragmentationFunction.h
MartinSchmidtTuebingen: Done

raymondEhlers:
+1

Thanks @fjonasALICE for the review and @MartinSchmidtTuebingen for the PR, careful testing, and hanging on with the long review. It will only get trickier to get these changes lower level into AliPhysics as it begins to phase out

PR #23749, opened 2024-10-29
PWGGA/GammaConv: fix for var. ptW in AddTask_GammaConvV1_PbPb.C

sstiefel19:
small fix. The issue was that even when the new pt weights method (= pt variantly calculated pt weights) was specified, the old was method was used.


jokonig:
+1

PR #23751, opened 2024-10-30
PWGJE/EMCAL: Fix cluster efficiency application in jet task

jokonig:
- Dcerease cluster efficiency by rejecting clusters according to a predefined function
- The random value was compared to the wrong variable. This is now fixed


raymondEhlers:
+1

PR #23752, opened 2024-10-31
New histograms for leading V0s

eeremenk:


raymondEhlers:
+1

Perhaps I already mentioned this (I don't recall), but since you've made a number of commits in AliPhysics recently - please remember that Run 2 AliPhysics preliminaries won't be possible going forward. You'll have to either go directly to a paper or use run 2 converted data in o2physics. As far as I can tell, you're still in active development of your analysis, so you should consider carefully what path forward you'd like to take going forward. If I've already mentioned this or you already know your plan, feel free to disregard

PR #23753, opened 2024-11-01
[RUN3, PWGGA] Fix circle dependency

mhemmer-cern:
The `AliAnalysisTaskAO2DConverter` needed to know of `AliV0ReaderV1` and vice versa. This commit adds the class `AliAODRelabelInterface` which is a new base class that inherits itself from the `AliAnalysisTaskSE` class and that `AliV0ReaderV1` and `AliAnalysisTaskAO2DConverter` inherit from. This way both classes only need the one new class and do not need to know from each other. Inside PWGGA/GammaConvBase/AliConversionPhotonCuts both classes were also needed because there in the function `GetTrack` one needs to check if the V0s are already relabeled.

### DIFF
--- a/PWGCF/FEMTOSCOPY/FemtoDream/CMakeLists.txt
+++ b/PWGCF/FEMTOSCOPY/FemtoDream/CMakeLists.txt
@@ -20,6 +20,7 @@ add_definitions(-D_MODULE_="${MODULE}")
 # Module include folder
 include_directories(${AliPhysics_SOURCE_DIR}/PWGCF/FEMTOSCOPY/FemtoDream
                     ${AliPhysics_SOURCE_DIR}/PWGGA/GammaConvBase
+                    ${AliPhysics_SOURCE_DIR}/PWGGA/Common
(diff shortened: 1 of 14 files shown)


mhemmer-cern:
@ddobrigk there was a problem in my PR for the converter that I didn't get on my local computer probably because it didn't rebuild everything from scratch, which lets all PWGGA jobs crash currently. This PR should fix it. @jokonig tested it separately.

mhemmer-cern:
@pzhristov could you please check this. Without this the PWGGA can not run any trains on the latest Tags

jokonig:
looks good from PWGGA side! Can be merged

jokonig:
+1

FriederikeBock:
+1

jokonig:
+test

mhemmer-cern:
@singiamtel that's really weird, @jokonig had that problem, but after a fresh rebuild it disappeared. In the CMakeLists.txt we included the PWGGA/Common folder where `AliAODRelabelInterface.h` is, so it should be found by everything inside the RUN3 and PWGGA/GammaConvBase folder.

PR #23756, opened 2024-11-06
RUN 3: add number of findable minus PID TPC clusters in converter

maciacco:
@mpuccio @jgrosseo @ddobrigk @pzhristov here is the addition of the TPC pid clusters info in the track extra table, similarly to what is planned for run 3


ddobrigk:
@pzhristov shall we carry on? When you have a moment, could you please look and approve? Thanks!

PR #23754, opened 2024-11-06
[PWG, PWGDQ, PWGGA, PWGHF] Fix CMakeLists.txt missing include directory

mhemmer-cern:
- Added `${AliPhysics_SOURCE_DIR}/PWGGA/Common` to the `include_directories` everywhere, where the `AliV0ReaderV1` is included, since the `AliV0ReaderV1` includes the new `AliAODRelabelInterface.h` which is in `PWGGA/Common`. This should fix the compile errors we saw after the merge of PR #23753

PR #23757, opened 2024-11-06
[RUN3] Try to fix build crashing with undefined symbol: _ZTI22AliAODC…

mhemmer-cern:
…onversionPhoton

- I assume the problem now is, that since we include the AliAODConversionPhoton class in RUN3 we need the PWGGA/GammaConvBase lib in RUN3 since AliAODConversionPhoton is defined there. Why multiple other builds worked without that fix, I do not know.
Should fix remaining issue seen in PR #23754


pzhristov:
+test

mhemmer-cern:
I saw it earlier. I am just so glad that it's finally building. 
Thanks for checking the builds.

PR #23784, opened 2024-11-27
Run 3 converter: add PMD information

ddobrigk:
@abmodak @sustripathy


ddobrigk:
This adds PMD information according to our first discussion. The primary desired variables should already be there in full; there are no filters on the PMD tracks and all entries originally contained in the ESDs are currently saved. 

Note: the development of the 'receiving' side in O2Physics is still pending and it would be desirable to have that for QA as soon as possible.

PR #23794, opened 2024-12-04
Change in MCGenerated function

eeremenk:
solution for fastjet bg estimator fail in GeneratedMCParticles because of an empty bg subtractor


raymondEhlers:
+1

PR #23797, opened 2024-12-05
[RUN3] Rename V0otfs to Run2V0otfs

mhemmer-cern:
- Renaming of the V0otfs table to Run2V0otfs to go hand in hand with the table name we want in the O2 framework.


jgrosseo:
+1

PR #23804, opened 2024-12-11
Edited SVtaskMCFilter to be able to filter HIJING part of MC tracks

lmhuhta:
EPOS is still the default filtered gen in the task, added the option to instead filter HIJING if one sets the last setting to true.


raymondEhlers:
LGTM, excpe that you need to bump the ClassDef

raymondEhlers:
+1

PR #23811, opened 2024-12-15
PWGGA/GammaConv: Modify weighting for jet particle abundance

jokonig:
- Add option for different weightings:
- 1) Add weight from every particle
- 2) Only add weights from particles with a weight factor different from 1


jokonig:
+1

PR #23817, opened 2024-12-17
Run 3 conv: add ev sel extras

ddobrigk:
@victor-gonzalez


victor-gonzalez (commented):
@ddobrigk Thanks!
Just one thing I spotted, probably not significant
- the way the total number of `SDD+SSD` is taken in a different way in the `AliEventCuts` object
```
  AliVMultiplicity* mult = ev->GetMultiplicity();
...
  int nCluSDDSSD=0;
  for(Int_t iLay=2; iLay<6; iLay++) nCluSDDSSD+=mult->GetNumberOfITSClusters(iLay);
```
so, it is taken from the multiplicity object, but probably they are equivalent

PR #23852, opened 2025-01-22
Check the fraction with and without matching to the decay of Xi->Lambda+pi and Lambda->p+pi

zhujh126:
Check the fraction with and without matching to the decay of Xi->Lambda+pi and Lambda->p+pi


fgrosa:
+1

PR #23860, opened 2025-02-01
[PWGJE] changes for pp MC jet hadrochem

sweyh99:
Made rho=0 for 2017 data and added event histos for MC rec and truth jets


raymondEhlers:
+1

PR #23864, opened 2025-02-07
removed PWGHFvertexingHF library dependency

lmhuhta:
AliPhysics builds normally without this library in the CMakeFiles, and removing it allows to run the AliAnalysisTaskSVtaskMCFilter which needs lib PWGHFvertexingHF


dongjokim:
+1

PR #23866, opened 2025-02-10
[PWGGA] AddTask_GammaCalo_PbPb - Add MC train config for 10% steps

mhemmer-cern:
- 7000 for 0-30 and 70-90 with TM for secondaries
- 7001 for 30-70% with TM for secondaries
- 7100 for 0-30 and 70-90 without TM
- 7101 for 30-70% without TM


jokonig:
+1

PR #23872, opened 2025-02-18
PWGJE: Updates to HFE tagging jet analysis

leo-barreto:
Multiple update and fixes to the analysis: removed old observables, revert DCal changes, more options for MC tests, etc.


raymondEhlers:
+1

Just in case you're not already aware, AliPhysics will be deprecated very soon. Given that these commits are modifying a task which hasn't been updated in quite some time, moving to O2Physics would likely be more appropriate

PR #23877, opened 2025-02-23
Embedding code for EEC

anabananana98:


raymondEhlers:
+1

I didn't review in any detail, but I'm approving immediately given the short time available

PR #23881, opened 2025-02-25
[RUN3] AOD conversion: save hybrid tracks, event bits

tch285:
- Save hybrid tracks in AOD -> AO2D conversions
- Save event bit information in AOD -> AO2D conversions
- Implement flags to turn hybrid track and event bit saving on (the saving is off by default)

tagging also @ddobrigk


ddobrigk:
Thanks a lot! Shall we merge and give this a go? Right now the grid is very packed but we can start with a single run as a test or so, especially for the AOD-only MC. How does that sound?

tch285:
sounds good to me!

PR #23897, opened 2025-03-11
[RUN3] AOD conversion: save MultSelection event bits

tch285:
- Save event bits stored in MultSelection during AOD -> AO2D conversions
tagging also @ddobrigk

### DIFF
--- a/RUN3/AliAnalysisTaskAO2Dconverter.cxx
+++ b/RUN3/AliAnalysisTaskAO2Dconverter.cxx
@@ -1816,28 +1816,28 @@ void AliAnalysisTaskAO2Dconverter::FillEventInTF()
   }
   else if (fAOD && fSaveEventBitsInAODConversion) {
     // // Get multiplicity selection
-    // AliMultSelection *multSelection = (AliMultSelection*) fAOD->FindListObject("MultSelection");
+    AliMultSelection *multSelection = (AliMultSelection*) fAOD->FindListObject("MultSelection");
     // AliMultSelection *multSelection = (AliMultSelection*) fVEvent->FindListObject("MultSelection");
-    // if (!multSelection)
-    //   AliFatal("MultSelection not found in input event");
+    if (!multSelection)
+      AliFatal("MultSelection not found in input event");
 
-    // if( multSelection->GetThisEventINELgtZERO() )
-    //   SETBIT (run2bcinfo.fEventCuts, kINELgtZERO);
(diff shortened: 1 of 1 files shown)


ddobrigk:
Hi Tucker, thanks for looking into the hybrid tracks and the converter! I am still wondering, though: was there any specific motivation behind storing these values (in the sense that they were not already there in the conversions)? It seems to me a number of them are redundant (and in fact we already store event selection bits from AliEventCuts in the Run2BCInfo table if memory serves me correctly)...

tch285:
Hi David, yes, I checked initially with storing only those not from MultSelection (of which the kAliEventCutsAccepted is one) but if one does the sel7 selection inside O2Physics, no events will pass, I think because the event selection framework there checks each bit.

ddobrigk:
Yeah, I think that's a bit unexpected - it would be best to check with @ekryshen about the event selection, since it's supposed to be done there, and identify any issues that we may still have ... @ekryshen can you confirm that without this change one can already perform e.g. sel7 selections? Thanks!

ekryshen:
Hi @ddobrigk, @tch285 A few years ago we were working on ESD->AO2D conversion trying to reuse available info and replay event selection bits. So sel7 is expected to work for this kind of conversions. We've never tried to validate event selection for AOD->AO2D conversions, so indeed sel7 might fail in this case.

tch285:
hi @ekryshen , thank you for the clarification! To confirm, in the case that not all event bits are saved, one should not expect the event selection to work, since it replays the selections rather than just checking the one AliEventCutsAccepted bit? Should we then consider not saving these bits to save space and make a new event selection in O2Physics that checks AliEventCutsAccepted rather than replay the selection (since they are equivalent)? I suppose there is not many datasets that are AOD-only, and in my local testing the increase in disk space is not large...

ekryshen:
@tch285 As far as I can see, you are just setting bits in the already available run2bcinfo.fEventCuts mask, so I don't see a problem with disk space here. Moreover, setting some bits (kIncompleteDAQ etc) could be a step towards working sel7 in AOD->AO2D conversions.

ekryshen:
@tch285 to confirm, yes event seleciton in O2 replays the selections one-by-one and is not relying on AliEventCutsAccepted bit.

tch285:
@ekryshen thanks for the clarification! @ddobrigk could we then get this merged and attempt a test with LHC23a3 like before?

PR #23900, opened 2025-03-13
add more nogap correlators

EmilGorm:


ktf:
This PR breaks the builds.

ddobrigk:
Hi @ktf - I'm terribly sorry for force-merging, and thanks a lot for fixing the resulting problem! 

We should follow up on the operation of the AliPhysics CI in some cases - the issue in this case was that the security approval was never given for the CI to run, so the error was not caught, and I haphazardly approved (again, sorry). In the past, I had discussed with @pzhristov and neither of us are quite sure exactly how to recover the legacy behaviour of i.e. issuing a `+test` to trigger / approve the CI or so for this repository. When there is a similar case of a PR needing a security approval, we should absolutely make sure this gets sorted out...

PR #23967, opened 2025-05-12
PWGGA/GammaConv: Small fix for pt-weights

sstiefel19:
Initialize pt-weights only for centrality classes that are contained in the used MC production.


FriederikeBock:
+1

PR #24005, opened 2025-06-16
 PWG/EMCAL: Fix CrossTalk same cell check

mhemmer-cern:
When inserting a new cell at the end of the cross talk the check to ensure that the same cell is not inducnig to itslef was wrong:
```c++
if ( iphii==0 && ietai == 0) continue;
```
It should be
```c++
if ( iphii == iphi && ietai == ieta) continue;
```

The impact of this bug is unknown. I guess it basically just means that before this fix the cell (0,0) was not allowed to induce energy into a new cell, which is a very specific edge case.


gconesab:
Nice catch!
+1

PR #24027, opened 2025-07-09
PWGJE: Extra histograms for electron PID and fixes

leo-barreto:
Final steps for electron identification using EMCal and MC studies, alongside code clean-up


raymondEhlers:
+1

A friendly reminder: AliPhysics is near end of life. The exact date isn't yet set, but it won't be supported for long

leo-barreto:
> +1
> 
> A friendly reminder: AliPhysics is near end of life. The exact date isn't yet set, but it won't be supported for long

Thanks @raymondEhlers, I know haha. This should be the last update :pray:

PR #24030, opened 2025-07-14
Flatten EP with shift correction in JetHdEdxCorrelations task

aohlson:


raymondEhlers:
+1

Okay, perfect -  thanks for the info! It's great to hear that you're pushing this forward!

PR #24167, opened 2026-06-02
PWGGA/GammaConv: add option for photon weights based on meson weight

sstiefel19:
only on for  theUseGetMesonWeightNew=true  (AddTask)


jokonig:
+1

PR #24169, opened 2026-06-04
PWGGA/GammaConv: Add pT-eta photon weights for added signal gammas

sstiefel19:
Apply photon pT-eta weights for flat injected added-signal productions while keeping inclusive true-gamma histograms free of photon reweighting.


jokonig:
+1

PR #24184, opened 2026-09-09
PCM4ML: data bdt distribtion hist + config with Pion cuts

an15ms157:
Added configs 
Added one histogram for data bdt distribution


jokonig (approved):

jokonig:
+1

